Issue #14, opened 2017-10-31
debops.php5 and debops.mysql roles required but missing


pniederlag:
Hi,

today I executed 

    pip install --upgrade debops && debops-upgrade

afterwards running debops I ran into missing the two roles 'debops.php5' and 'debops.mysql'. After I added the missing roles into my local debops repository via ansible galaxy everything went smooth again.

drybjed:
I did not include these roles since they were superseded by `debops.php` and `debops.mariadb{_server}` respectively. The roles that still rely on them will need to be updated to use the new ones.

You can read more about the merge and what roles were not included [on the mailing list](https://lists.debops.org/pipermail/debops-users/2017-October/000102.html)

ypid:
@drybjed The issue is that these roles are still referenced by playbooks which are included in the site playbook. I would propose to include the two roles until they are not referenced by any role anymore. You have those roles probably still in your role path.

drybjed:
@ypid You are probably right about having the roles at the time in `roles_path`. I removed them and now running `ansible-playbook site.yml --syntax-check` properly fails on the playbook level and role `meta/main.yml` level. This test will need to be added to the Travis-CI tests to avoid cases like this in the future.

Do you think that you could try updating the small number of roles that currently use the `debops.php5` and `debops.mysql` as dependencies as well? That's most likely `debops.phpipam`, not sure what else. Otherwise I'll merge the PR when you give it a go and then we can enable the `--syntax-check` test on Travis.

ypid:
> Do you think that you could try updating the small number of roles that currently use the debops.php5 and debops.mysql as dependencies as well?

I don’t use the other roles which use the old roles. That task would be good for getting more into DebOps development. Any volunteers? The new roles like `debops.php` and `debops.mariadb{_server}` are really well done and nicely documented. Also, there are a lot of example roles which already use the new roles.

cc: @ombah, @carriercomm, @GR360RY, @adeelahmad84

drybjed:
I meant that without updating the role dependencies in `meta/main.yml` for the `debops.phpipam` and `debops.phpmyadmin` roles the `ansible-playbook --syntax-check` would still fail (which it did). I updated these roles with as small changes as possible in https://github.com/debops/debops/pull/19, so it's all good now. The `--syntax-check` test is now enabled on Travis as well.

Of course the roles could still be refreshed to DebOps coding standards. Currently they seems to work fine, so it's not a pressing issue.

drybjed:
[...]

PR #15, opened 2017-10-31
Remove mysql and php5 playbooks. Roles are already removed.

ypid:
Playbooks which refer to missing roles causes Ansible to abort execution.

Potentailly wait with merging this. Ref: https://github.com/debops/debops/issues/14


wrgeorge1983:
the role `debops.php5` is also referenced by role `debops.phpipam\meta\main.yml`, which stops ansible from running on a fresh install.  

Not sure what real work if any needs to be done to convert that role to `debops.php`, but I was able to proceed by simply commenting that out (and doing the other changes you have listed here)

ypid:
@wrgeorge1983 https://github.com/debops/debops/issues/14 is mainly about what you talk here. Btw, you used a path `debops.phpipam\meta\main.yml` which is invalid under any sane operating system. I only know legacy systems which use this 😉 Are you using DebOps on Windows by any chance? That is unsupported ;)

wrgeorge1983:
@ypid I must have just typed it weird.

It *is* windows, but it's 'bash on ubuntu on windows 10 linux subsystem blah blah blah' which presents paths in the unix fasion.  For extra insulation I'm running it in a pipenv virtualenv, and that also works fine as far as I can tell.  I'm still chasing some issues, but none of them smell like that kind of compatibility issue so far.  I use ansible for other things without issue, generally.

PR #17, opened 2017-10-31
debops-contrib.firejail might get vars passed using debops.secret

ypid:


drybjed:
In cases like this it might be better idea to include `debops.secret` role as a hard dependency; that way you don't need to remember to add it in the custom playbooks that might use `debops-contrib.firejail`. Can you move it there, into `meta/main.yml`?

ypid:
Could be done but I think this is to DebOps specific to include it in the role itself. The variable I had in mind is [`firejail__fix_for_users`](https://debops-contrib.readthedocs.io/en/latest/ansible/roles/ansible-firejail/docs/defaults.html#envvar-firejail__fix_for_users). When you use DebOps, it might come in handy to use variables of `debops.users` but if you are not using DebOps, listing the users there by hand is also fine. I think having it at the playbook level in DebOps is fine.

`debops.secret` is completely optional here. I know other roles that are more closely tight with `debops.secret` where it makes more sense to declare it as hard dependency.

drybjed:
I see. Very well then.

PR #33, opened 2017-11-11
Symlink node to fix npm error on etherpad plugins

scosist:
I ran into a number of npm errors (namely `/usr/bin/env: ‘node’: No such file or directory`) when running the `debops.etherpad` "Manage Etherpad plugins" task. Setting up this symlink seemed to do the trick for me.

### DIFF
--- a/ansible/roles/debops.nodejs/tasks/main.yml
+++ b/ansible/roles/debops.nodejs/tasks/main.yml
@@ -40,6 +40,16 @@
     - '{{ nodejs__host_packages }}'
     - '{{ nodejs__dependent_packages }}'
 
+# need this to resolve error(s) on debops.etherpad: Manage Etherpad plugins
+# /usr/bin/env: ‘node’: No such file or directory
+- name: Create symlink for node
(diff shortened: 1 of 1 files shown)


drybjed:
This is still a problem then... Debian has a [`nodejs-legacy`](https://packages.debian.org/stretch/nodejs-legacy) APT package which fixes this issue with a symlink. Unfortunately the package is not compatible with upstream NodeJS package installed by the role when enabled.

Furthermore, the upstream package has `/usr/bin/node` as a real binary, not a symlink, and the `/usr/bin/nodejs` is a symlink to it, via the alternatives system. I suspect that when you enable the upstrean NodeJS package, the task you added will error out because the link you are trying to create is an existing case.

Because `/usr/bin/node` is still needed, the role will need to be updated to handle these cases. The installation of `nodejs-legacy` package [was removed in a PR adding support for Nodesource packages](https://github.com/debops/ansible-nodejs/pull/4). Now it will need to be a conditional. I'll try to look into it soon.

PR #34, opened 2017-11-11
Add postgresql database support to debops.etherpad

scosist:
Note, I found the need to exclude the database port for postgres in the `settings.json` template when using a socket connection. I didn't get a chance to see how mysql handles this so assuming it was good as it was I maintained previous behavior.

### DIFF
--- a/ansible/roles/debops.etherpad/defaults/main.yml
+++ b/ansible/roles/debops.etherpad/defaults/main.yml
@@ -142,36 +142,48 @@ etherpad_welcome_text: |
 
 # .. envvar:: etherpad_database_server [[[
 #
-# FQDN of the MariaDB database host. It will be configured by
-# the debops.mariadb_ role.
+# FQDN of the database host. It will be configured by
+# the debops.mariadb_ or debops.postgresql_ role.
 etherpad_database_server: '{{ ansible_local.mariadb.server
                               if (ansible_local|d() and ansible_local.mariadb|d() and
                                   ansible_local.mariadb.server|d())
-                              else "" }}'
+                              else (ansible_local.postgresql.server
+                                    if (ansible_local|d() and ansible_local.postgresql|d() and
+                                        ansible_local.postgresql.server|d())
+                                    else "") }}'
 
                                                                    # ]]]
 # .. envvar:: etherpad_database_user [[[
 #
-# MariaDB database user to use for Etherpad.
+# Database user to use for Etherpad.
 etherpad_database_user: '{{ etherpad_system_name }}'
 
                                                                    # ]]]
 # .. envvar:: etherpad_database_name [[[
 #
-# Name of the MariaDB database to use for Etherpad.
+# Name of the database to use for Etherpad.
 etherpad_database_name: '{{ etherpad_system_name }}'
 
                                                                    # ]]]
 # .. envvar:: etherpad_database_password [[[
 #
-# MariaDB database password for the Etherpad database user.
-etherpad_database_password: "{{ lookup('password', secret + '/mariadb/'
+# Database password for the Etherpad database user.
+etherpad_database_password: "{{ lookup('password', secret + '/'
+                                + ('mariadb'
+                                   if etherpad__database == 'mysql'
+                                   else ('postgresql'
+                                         if etherpad__database == 'postgres'
+                                         else etherpad__database)) + '/'
                                 + (ansible_local.mariadb.delegate_to + '/'
                                    if ansible_local|d() and ansible_local.mariadb|d() and ansible_local.mariadb.delegate_to|d()
-                                   else '')
+                                   else (ansible_local.postgresql.delegate_to + '/'
+                                         if ansible_local|d() and ansible_local.postgresql|d() and ansible_local.postgresql.delegate_to|d()
+                                         else ''))
                                 + (ansible_local.mariadb.port + '/'
                                    if ansible_local|d() and ansible_local.mariadb|d() and ansible_local.mariadb.port|d()
-                                   else '')
+                                   else (ansible_local.postgresql.port + '/'
+                                         if ansible_local|d() and ansible_local.postgresql|d() and ansible_local.postgresql.port|d()
+                                         else ''))
                                 + '/credentials/' + etherpad_database_user + '/password length=48') }}"
 
                                                                    # ]]]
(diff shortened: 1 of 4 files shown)


ansible/roles/debops.etherpad/defaults/main.yml
```diff
@@ -465,6 +483,41 @@ etherpad__mariadb__dependent_databases:
...
+    password: '{{ etherpad_database_password }}'
+    flags: [ 'NOSUPERUSER', 'NOCREATEDB', 'LOGIN' ]
+
+                                                                   # ]]]
+# .. envvar:: etherpad__postgresql__dependent_groups [[[
+#
+# Group configuration for the debops.postgresql_ Ansible role.
+etherpad__postgresql__dependent_groups:
+
+                                                                   # ]]]
```
drybjed: The common way of setting up PostgreSQL databases and users in DebOps is to create two roles:

- role for the database that cannot directly login, for example `etherpad_lite_production`
- role for the application user that can login, has a password and has access to the database role, for example `etherpad_lite`
That way, permissions in the database are somewhat manageable. you can check the `debops.gitlab` or `debops.netbox` roles for examples.

You used only one role in your PR, and I guess that's fine as well. However if you go that route, you shouldn't leave empty variables like this one and `etherpad__postgresql__dependent_pgpass` below. Either remove them, and their corresponding entries in the playbooks, or switch to the separated PostgreSQL roles.

ansible/roles/debops.etherpad/defaults/main.yml
```diff
@@ -465,6 +483,41 @@ etherpad__mariadb__dependent_databases:
...
+# .. envvar:: etherpad__postgresql__dependent_databases [[[
+#
+# Database configuration for the debops.postgresql_ Ansible role.
+etherpad__postgresql__dependent_databases:
+
+  - name: '{{ etherpad_database_name }}'
+    port: '5432'
```
drybjed: I suppose this forces the usage in the `main` PostgreSQL cluster, but I guess it's fine for a initial support. We can update this later, when the PR is merged.

drybjed:
The code is stashed in the main repository as https://github.com/debops/debops/tree/pending/etherpad-postgresql-support

ansible/roles/debops.etherpad/defaults/main.yml
scosist: Right now, the database name and user are based off of `etherpad_system_name` as are many other variables. The default value of `etherpad_system_name` is `etherpad-lite` but you suggest `etherpad_lite_production` for the database role name--looking at `debops.gitlab` and `debops.netbox` the database name and role name are ideally the same. How should this be handled here? Should the `etherpad_database_name` change or the `etherpad_system_name` change? Either way, I believe the system user and database user name need to be the same for a unix socket connection. I just don't know which way to go as this could change a number of other things.

ansible/roles/debops.etherpad/defaults/main.yml
scosist: I have a quick fix for this I was going to commit to my PR branch but I'm unsure what the desired process here is. The DebOps development process doc states `don't rewrite history in branches that are already pending as pull requests` but soon after that recommends a rebase during development as the upstream branch may have new commits. In this context I assume the rebase occurs before the initial feature branch push. I'm used to doing a rebase on PR branches as upstream changes because sometimes it seems necessary to allow it to be automatically merged and pass the checks. However, that seems to be against the desired process here. Have I understood this correctly? Am I to avoid a rebase once the PR has been created?

drybjed: My reasoning is this: having a large commit, unless it's something completely new like a new Ansible role, is bad for debugging and bisecting. With the PRs, if you create one, try and make small and sensible commits with any changes needed and push them to that PR's branch. You don't need to create new PRs (happened in the past), and you shouldn't try and `--force` new changes to existing branches, that messes with the reviewer's head when he sees the same commit names but there's a new code.

If there are no merge conflicts with the base branch, even if it moved on, you don't need to worry about rebasing. Merging code in a different place in the history is fine, as long as there are no conflicts.

if the PR cannot be merged cleanly, or the project moved so much forward that the changes don't make sense, then there's no other way around than rebasing the entire PR on the new commits in the base branch. That's acceptable as well, as long as you keep the commit history in that PR.

Since I added this PR on a separate branch to the DebOps monorepo, I think that the best case would be to close this PR, and create a new one against the `proposed/` branch. Then, when the time comes, that branch can be merged with the `master` branch without much issues, hopefully.

Keep in mind that I'm still new at this 'monorepo' thing and using multiple branches instead of separate repositories, we are developing best practices as we go. Any input is greatly appreciated. :-)

scosist: Thanks, that clears up my confusion and it sounds reasonable so I don't have any recommendations at the moment.

ansible/roles/debops.etherpad/defaults/main.yml
scosist: @drybjed any thoughts on this?

drybjed: Ah, sorry, must have missed it. The way `debops.postgresql_server` configures `pg_hba.conf`, you can create two roles:

- one role as a "database holder", `etherpad_lite_production`

- another role as an "application client", `etherpad_lite`

Now, when you add the `etherpad_lite` role to the `etherpad_lite_production` role inside PostgreSQL, the `etherpad_lite` will be able to login and access the `etherpad_lite_production` database. For this to work, application needs to explicitly request that database, otherwise the connection will be refused because there's no `etherpad_lite` database.

I'm not sure if that's the case with Etherpad. If it is, ie. the role name and database name needs to be the same, then I guess that we can stay with the current configuration; just remove the empty default variables.

scosist: That is the case with `debops.etherpad`. The role currently has the postgres user and database with the same name: `'etherpad-lite'` (with a hyphen as opposed to an underscore). These are both based off of `etherpad_system_name` (set to `'etherpad-lite'`) as are a number of other variables in the role.

I had intended on changing the database name to append `'_production'` and changing both the db name and db user name to use underscores (`'etherpad_lite'` as opposed to `'etherpad-lite'` to be clear) to be more like the `netbox` and `gitlab` roles but then I realized I should probably also change `etherpad_system_name` (to `'etherpad_lite'`) but that would change a lot as so many other things build off of this variable in the role. I believe such a change however would be disruptive to anyone who may already be using this role.

I'm happy to make the changes but what do you prefer in this case?

drybjed: In that case let's stay with current setup, `etherpad-lite`. I'm trying to preserve current DebOps roles as-is, fixing any bugs so that the existing setup can be released as `stable-1.0` branch. More disruptive changes can be done after that, and if somebody will want to stick to the current setup, they will be able to by selecting a specific branch.

PR #42, opened 2017-11-14
debops-contrib.etckeeper: Fix custom Git ignore list entries not applied

gajdusek:
This role adds custom entries to /etc/.gitignore after invoking
'etckeeper init'.

However, at that moment all files are already added to Git index, so we
need to remove them again before calling 'etckeeper commit'. Otherwise
custom ignore list has no effect.

See /etc/etckeeper/init.d/70vcs-add

### DIFF
--- a/ansible/roles/debops-contrib.etckeeper/tasks/main.yml
+++ b/ansible/roles/debops-contrib.etckeeper/tasks/main.yml
@@ -104,6 +104,7 @@
     mode: '{{ etckeeper__repository_permissions|d("0700") }}'
     owner: 'root'
     group: '{{ etckeeper__repository_group|d("root") }}'
+  when: (etckeeper__vcs == 'git')
 
 ## Set VCS user, email used for auto committing changes [[[
 - name: Set user, email for the git repository


ansible/roles/debops-contrib.etckeeper/tasks/main.yml
```diff
@@ -111,6 +111,14 @@
   tags: [ 'role::etckeeper:vcs_config' ]
 ## ]]]
 
+- name: Unstage ignored files from Git index
+  shell: etckeeper vcs ls-files -i --exclude-standard -z | xargs -0 etckeeper vcs reset --mixed --
+  register: git_reset
```
drybjed: You should namespace this variable, something like: `etckeeper__register_git_reset`.

ansible/roles/debops-contrib.etckeeper/tasks/main.yml
```diff
@@ -111,6 +111,14 @@
...
+  when:
+    - etckeeper__vcs == "git"
+    - etckeeper__register_init|changed
```
drybjed: Hmm, I wasn't aware that syntax could work... I need to check it out.

gajdusek: I think the filter `changed` just checks `etckeeper__register_init.changed` property. I am new to Ansible. I took it from task "Create root commit" (right after this one):

```yaml
- name: Create root commit
  command: etckeeper commit 'Initial commit created by Ansible.'
  when: etckeeper__register_init|changed
```

drybjed:
@gajdusek Thanks for updating, works like a charm. I had a quick look at what was unstaged (`.gitignore`? weird), and I noticed this warning from `git`:
```
warning: --mixed with paths is deprecated; use 'git reset -- <paths>' instead
```
Can you update the command that's used in the task? Just future-proofing.

ansible/roles/debops-contrib.etckeeper/tasks/main.yml
drybjed: No, I meant use of YAML list syntax in a `when:` statement. Apparently it works. Maybe I'll start using it more then.

PR #48, opened 2017-11-16
Etherpad postgresql get port from facts

scosist:
and other minor changes as suggested in #34.


drybjed:
This PR was merged in https://github.com/debops/debops/pull/221, therefore I'm closing it.

scosist:
Thanks! Any reason why the default port definitions for mysql and postgres remain [static](https://github.com/debops/debops/pull/221/files#diff-f2ce153869efbff0fd49cd29c4a6ec39L206) when they exist in local facts?

drybjed:
I exposed them through local facts in case the service ports ever changed, but I don't remember why I didn't hook them up here. I guess there's no reason to not switch to assigment through facts.

PR #49, opened 2017-11-16
Fix rsyslog network modules without requiring TLS

gaudenz:
The UDP and TCP modules are also useful if you don't require everything
to use TLS.


drybjed:
Trying to re-run Travis-CI check...

drybjed:
Fixed via https://github.com/debops/debops/pull/51

PR #50, opened 2017-11-16
Jessie debian-installer compatible GnuPG configuration

gaudenz:
The stripped down version of gpgv used in the jessie installer does not
support SHA512 and SHA384 as digest algorithms. Neither for singing
release files nor key certifications (ie. self signatures).

This adds a new variable reprepro_installer_jessie_compatibility to
disable the unsupported algorithms. This defaults to False. If you
change this variables to True, you have to regenerate your archive
signing key. Otherwise your key won't work as it contains unsupported
self-signatures.


drybjed:
@gaudenz Travis-CI flagged your PRs as 'Abuse detected' and refused to process them. I'm seeing this for the first time, and that's only your PRs so far. I created new PR with your changes incorporated and it seems to be processed fine.

What did you do that angered the CI gods? ;-)

drybjed:
Fixed by https://github.com/debops/debops/pull/52

PR #57, opened 2017-11-20
Fix defaults

cultcom:
debops-defaults complained about missing playbooks due to an old defaults path.


drybjed (approved):
Good catch, thanks. I'll try to release a new version of the scripts soon so this fix is available on normal installs. Unfortunately the release process from the monorepo is not yet ironed out...

drybjed:
@cultcom You should also fix the test cases. :-)

drybjed (approved):

PR #71, opened 2017-11-28
debops.dovecot: Add support for iterate_query

### LINKED ISSUE #70: debops.dovecot: Add support for "iterate_query"
Very simple PR incoming supporting this:
https://wiki2.dovecot.org/AuthDatabase/SQL#User_iteration

I just read about the GPG signed commits which is a very good idea, will set it up for future PRs.

evilham:
Added the `dovecot_sql_iterate_query` variable and modified the
`dovecot-sql.conf.ext.j2` accordingly.

Fixes #70


drybjed (approved):
Sounds good, thanks.

I don't think that we have any concrete set of SQL tables to work off from at the moment, but that would probably depend on the used e-mail solution.

PR #94, opened 2017-12-19
Some Python3 compatibility fixes.

cultcom:
I tried debops within a Python 3 virtual environment and encountered some exceptions. Maybe there is more to be fixed but this could be a starting point.


drybjed:
Thanks! Looks like there are some issues with the test suite, any idea what's the fix?

cultcom:
I guess for an issue with unicode vs. bytestring... Maybe I find time to have a look tomorrow.

cultcom:
the last commit fixes "debops-update" so far. More to come ...

Merry Christmas!!

drybjed (changes requested):
Awesome, thanks. While you're at it, can you update the `debops-defaults` script to look in the correct location for the roles? It should look in the monorepo directory, but it expects `debops-playbooks`.

cultcom:
I cannot proof that it looks under "debops-playbooks" here. Maybe you have a wrong path configured in your local projects ".debops.cfg"?

"debops-defaults" utilizes debops.config to read the defaults which results in ~/.local/share/debops/debops/ansible/playbooks for me.

drybjed:
@cultcom I just checked and indeed, `debops-defaults` seems to be working as expected. It might have been an old configuration file.

So, do you think this PR is ready, or do you want to mess some more with it?

cultcom:
let me sleep over it ;-)
I'm currently stuck making "debops" itself PEP8 compliant ...
```
debops-tools/bin/debops|84 col 80 C| E501 line too long (93 > 79 characters) [pep8]                                                                   
debops-tools/bin/debops|85 col 80 C| E501 line too long (80 > 79 characters) [pep8]
debops-tools/bin/debops|88 col 80 C| E501 line too long (84 > 79 characters) [pep8]
debops-tools/bin/debops|99 col 80 C| E501 line too long (80 > 79 characters) [pep8]
debops-tools/bin/debops|129 col 1 C| C901 'main' is too complex (14) [mccabe]
debops-tools/bin/debops|147 col 80 C| E501 line too long (84 > 79 characters) [pep8]
```

ansible/roles/debops.nsswitch/templates/etc/ansible/facts.d/nsswitch.fact.j2
```diff
@@ -32,7 +32,7 @@ try:
                         service_actions.update({ words[i]: words[i+1]})
                 service_list = []
                 for service in words[1:]:
-                    if service in service_actions.keys():
+                    if list(service in service_actions.keys():
```
ypid: Here be dragons (missing closing parentheses). Also, I don’t think `list` is required here.

lib/debops-tools/debops/config.py
```diff
@@ -39,7 +42,7 @@
 
 DEBOPS_CONFIG = ".debops.cfg"
 
-DEFAULTS = """
+DEFAULTS = u"""
```
ypid: In Python strings support Unicode by default which I think is a good default. I guess you could leave it at:

```Python
DEFAULTS = """
```

? This is more common in Python 3 projects I would say.

ypid (changes requested):
Thanks very much for working on this! Two minor details attached.

ansible/roles/debops.nsswitch/templates/etc/ansible/facts.d/nsswitch.fact.j2
drybjed: Good catch! The GitLab CI test environment didn't catch this, perhaps it's a good idea to add an Ansible facts check after the playbook completes.

ansible/roles/debops.nsswitch/templates/etc/ansible/facts.d/nsswitch.fact.j2
```diff
@@ -9,7 +9,7 @@ import os
 
 nsswitch_conf = '/etc/nsswitch.conf'
 
-nss_databases = loads('''{{ nsswitch__combined_database_map.keys() | to_nice_json }}''')
+nss_databases = loads('''{{ nsswitch__combined_database_map | list | to_nice_json }}''')
```
bfabio: Despite being correct, I feel `nsswitch__combined_database_map | list` takes out a little bit of readability.

How do you feel about using `nsswitch__combined_database_map.keys() | list`, even if it's not as efficient?

bfabio (dismissed):

ansible/roles/debops.nsswitch/templates/etc/ansible/facts.d/nsswitch.fact.j2
cultcom: oh, this is not only the missing bracket but the "keys()" part should be removed as well.

lib/debops-tools/debops/config.py
cultcom: But it breaks the IOString stuff.
In py2 the default string type is a bytestring. Using u""" is more precise and should work on both versions.

ansible/roles/debops.nsswitch/templates/etc/ansible/facts.d/nsswitch.fact.j2
cultcom: Well, but that makes the lines even loger. I am not sure if that supports readability.

cultcom:
BTW: here is a very nice overview of py2/p3 compatibility examples.

drybjed:
@cultcom Do you know if `pycodestyle` can be configured to check for Python2/3 compatibility? Alternatively, is there any other linter that could do this?

cultcom:
I only use Pylint within vim python-mode so far. Checking for Py2/Py3 compatibility is more a logic check than a syntax check, so I do not believe there are checking tools for that.

ansible/roles/debops.nsswitch/templates/etc/ansible/facts.d/nsswitch.fact.j2
bfabio: It's a little bit unexpected to the casual eye that `list` on a map returns its keys, having `keys` is more explicit and reassuring. It's just nitpicking though, I don't have a strong opinion about it.

lib/debops-tools/debops/config.py
ypid: Works for me too then :) I remember, I also worked with IOString already.

ypid:
> @cultcom Do you know if pycodestyle can be configured to check for Python2/3 compatibility? 

I don’t know `pycodestyle` but for checking Python2 and Python3 compatibility a CI matrix could be used where all tests and so on are run with the corresponding Python version. I think I did this for one of my Python projects in the past but seems I dropped Python2 support in all of them already because they are too irrelevant that I would care to support legacy stuff. It is basically the same as testing different Python3 versions. Example: https://github.com/ypid/fdeunlock

But even without all this complexity, the simplest test would be to just call the scripts with both Python2 and 3.

drybjed:
@cultcom Thanks again for your work! I'm currently testing this PR in GitLab CI to see if there are any issues, tests will be performed using old `debops` version and Python 2. If there are no issues, I think we can merge this.

The release process of `debops-tools` from the monorepo is still a bit wonky... I guess that I'll release DebOps `v0.7.0`, publish the tarball with the scripts on the monorepo release page and point the PyPI configuration there. I'll wait for @leegarrett's https://github.com/debops/debops/pull/111 PR for that as well. The new release will probably happen sometime in the next week.

drybjed (approved):

PR #97, opened 2017-12-22
[debops.gitlab] Use the --retry switch with bundle install.

bfabio:


drybjed (approved):
Sounds good, I didn't know `bundle` supported that option. Still, `until` method was a proof of concept, and it seems like it does the job, will be handy in other cases like this.

I'll merge it with the rest of the PRs.

PR #102, opened 2018-01-02
[debops.gitlab] Run gitlab-pages' bin/install as root.

bfabio:
Run bin/install as root because it needs to install libraries system-wide.


drybjed (approved):
Sounds good. I'll merge it with the other PRs when my test environment is back up.

PR #103, opened 2018-01-02
[debops.gitlab] Add support for GitLab 10.3.

bfabio:


drybjed (approved):
Looks good, I'll merge it with the rest when my test environment is back up.

NB, @bfabio, in the main `gitlab` repository, there are files like VERSION, GITALY_VERSION, etc. I wonder if the order of tasks would allow for something like:

- grab the selected `gitlab` version
- get versions of related repositories
- check out the rest of the repositories

This probably would drop the need to add or update version numbers and just stick to selecting the main branch.

bfabio:
> NB, @bfabio, in the main gitlab repository, there are files like VERSION, GITALY_VERSION

Yes, I thought about that but never had the time follow through. I'll give it I go next week.

PR #121, opened 2018-01-19
Add python-setuptools to mosquitto packages

Cadair:
I had a faliure when running the mosquitto role on a Debian 9 box because setuptools was not present. I am assuming this is the correct place to add it, and it won't have unintended concequences.

### DIFF
--- a/ansible/roles/debops.mosquitto/defaults/main.yml
+++ b/ansible/roles/debops.mosquitto/defaults/main.yml
@@ -39,7 +39,7 @@ mosquitto__upstream_repository:
 # .. envvar:: mosquitto__base_packages [[[
 #
 # List of base APT packages to install for Mosquitto support.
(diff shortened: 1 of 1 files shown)


drybjed:
Yes, that's correct. Any more info about the failure? What caused it? Let me check why there's Python stuff in the role... I'm asking because [there are no issues during testing](https://gitlab.com/debops/debops/-/jobs/47058767), was that something during Mosquitto usage?

Cadair:
It was pip installing the paho-mqtt module that caused the issue. I was in a massive rush to get the thing back up, so didn't have time to debug it properly, I just sshed in and installed this.

I don't really understand this because the version on pypy hasn't been updated in over 2 years: https://pypi.python.org/pypi/paho-mqtt/1.1

PR #139, opened 2018-02-02
Merge debops-contrib.roundcube into monorepo

ganto:
As I was asked in IRC to merge the roundcube role into the monorepo here my PR. Documentation and playbook updates will follow after merge.


drybjed:
And I just started writing my own Roundcube role, which among other things will support both Debian version and upstream... @ganto, would you agree to wait with merging this PR to see my version? I'll try and release it in the next week or so.

ganto:
How you like. Is there anything missing in my role that you are looking for? I only see now that there are official Debian packages again. They weren't there yet, when I updated the role last year, that why I removed support for that.

Other than that I think this role is pretty capable as I put a lot of effort in it and as you can see there are also some users other than me.

drybjed:
After a quick review in the original repository, I think that most of the role could be adapted for the upstream support in mine, mostly for deployment. What I would change however is redesign the Roundcube configuration to use the `parse_kv_config` filter used for example in the Postfix-related roles.

I think that we could merge your role in to keep the `git` history in the DebOps monorepo, and then I would redesign it with my modifcations in mind, drop the reliance on the `task_src`/`template_src` lookups, move the `composer` installation to `debops.php` role, etc. The role is not yet integrated with the main playbook, documentation and testing in this PR, but this can be done later; I still need to document these steps so don't worry about not doing it.

How does that sound to you, @ganto?

ganto:
Sounds great, thanks. :-)

PR #140, opened 2018-02-05
prosody role

TuxCoder:
This role is not tested yet. Updates comming soon

TODO:
 * see @TODO
 * split and document main config `prosody__config_global`
 * doc refer in `ansible/roles/debops.prosody/README.md` 

Open for improvements and ideas


ansible/playbooks/srv/prosody.yml
```diff
@@ -0,0 +1,20 @@
+---
```
drybjed: This file should be in `ansible/playbooks/service/prosody.yml` and then it can be symlinked to `../srv/prosody.yml`.

ansible/roles/debops.prosody/CHANGES.rst
```diff
@@ -0,0 +1,23 @@
+.. _apt__ref_changelog:
```
drybjed: The separate Changelogs for each role are hard to manage. You should focus instead on writing [more verbose `git` commit messages](https://chris.beams.io/posts/git-commit/), and document more visible things about the changes in the main project Changelog. For example with initial release of a new role, a note about adding new role in the main changelog should be enough.

ansible/roles/debops.prosody/README.md
```diff
@@ -0,0 +1,24 @@
+## [![DebOps project](http://debops.org/images/debops-small.png)](http://debops.org) prosody
```
drybjed: The `README` files in individual roles aren't currently needed since the roles are not published separately on GitHub anymore. Role documentation should be the go-to place for information about the roles themselves.

ansible/roles/debops.prosody/defaults/main.yml
```diff
@@ -0,0 +1,162 @@
+---
+
```
drybjed: The `defaults/main.yml` files should have a header that enables vim folding, for usability. There should also be a header that creates a separate section in the autogenerated documentation. See other roles for examples.

ansible/roles/debops.prosody/defaults/main.yml
```diff
@@ -0,0 +1,162 @@
...
+# -----------------------------
+
+# .. envvar:: apache__base_packages [[[
+#
+# List of base packages to install.
+prosody__base_packages:
+  - prosody
+  - lua-zlib
+  - prosody-modules # for ldap auth
+#  - lua-event # for use_libevent = true
```
drybjed: Since the role files are supposed to be read-only, having commented out parameters like this is counter-productive. Instead, I would create a condition that enables a package installation if specific parameter is enabled, and skips the package otherwise. For example:
```yaml
prosody__base_packages:
  - '{{ "lua-event" if prosody__use_libevent|bool else [] }}'
```
This requires the use of the `with_flattened` parameter lookup in the APT installation task to work properly.

ansible/roles/debops.prosody/defaults/main.yml
```diff
@@ -0,0 +1,162 @@
...
+                                                                   # ]]]
+                                                                   # ]]]
+
+
```
drybjed: Whitespace between variables like this should be avoided to make the vim folded view more readable.

ansible/roles/debops.prosody/defaults/main.yml
```diff
@@ -0,0 +1,162 @@
...
+# PKI [[[
+# ~~~~~~~
+
+# .. envvar:: prosody_pki [[[
+#
+# Enable or disable support for PKI/SSL/TLS in prosody.
+# Defaults to ``True`` if debops.pki_ is enabled on the remote host.
+prosody_pki: '{{ ansible_local|d() and ansible_local.pki|d() and
```
drybjed: Ensure that all role variables are properly namespaced with a `prosody__` prefix.

ansible/roles/debops.prosody/defaults/main.yml
```diff
@@ -0,0 +1,162 @@
...
+                                                                   # ]]]
+# .. envvar:: prosody__admins [[[
+#
+# list of admins jabber accounts of this prosody instance
+prosody__admins: []
+                                                                   # ]]]
+# .. envvar:: prosody__config_global: [[[
+#
+# main prosody configuration
+# @see prosody.im
```
drybjed: Since this will be included in the role documentation on https://docs.debops.org/, you should include at least a full URL for the documentation here, https://prosody.im/doc/configure

ansible/roles/debops.prosody/defaults/main.yml
```diff
@@ -0,0 +1,162 @@
...
+  allow_registration: False
+  daemonize: True
+  pidfile: "/var/run/prosody/prosody.pid"
+  #ssl:
+  #  key: '..'
+  #  certificate: '..'
+  c2s_require_encryption: True
+  s2s_secure_auth: True
+  s2s_insecure_domains:
+    - "gmail.com"
```
drybjed: This is an arbitrary configuration choice and might not be obvious for the first time user, therefore it should be avoided as a default. Instead you can add it as an example in the Getting Started documentation of the role, to inform the user that this might be needed for the role to work correctly.

DebOps is aimed for production use, and we should avoid an "insecure by default" approach whenever possible.

ansible/roles/debops.prosody/defaults/main.yml
```diff
@@ -0,0 +1,162 @@
...
+    - "mam"
+    - "blocking"
+    - "smacks"
+    #- "s2s_auth_dane"
+  allow_registration: False
+  daemonize: True
+  pidfile: "/var/run/prosody/prosody.pid"
+  #ssl:
+  #  key: '..'
+  #  certificate: '..'
```
drybjed: This should be hooked up by default to the DebOps PKI using the `prosody__pki_*` variables defined earlier.

ansible/roles/debops.prosody/defaults/main.yml
```diff
@@ -0,0 +1,162 @@
...
+  c2s_require_encryption: True
+  s2s_secure_auth: True
+  s2s_insecure_domains:
+    - "gmail.com"
+  authentication: ..
```
drybjed: I don't think that this is a legal value in the Prosody configuration. Are there any good alternatives for a default value here? Maybe an empty dictionary like `{}`?

ansible/roles/debops.prosody/defaults/main.yml
```diff
@@ -0,0 +1,162 @@
...
+  log:
+    info: "/var/log/prosody/prosody.log"
+    error: "/var/log/prosody/prosody.err"
+    #{ levels = { "error" }; to = "syslog";  }
+                                                                   # ]]]
+# .. envvar:: prosody__config_virtual_hosts: [[
+prosody__config_virtual_hosts:
```
drybjed: The `[[` marker will break vim folding.

ansible/roles/debops.prosody/defaults/main.yml
```diff
@@ -0,0 +1,162 @@
...
+# .. envvar:: prosody__ferm__dependent_rules: [[[
+#
+# Configuration for :command:`iptables` firewall manged by :program:`ferm`.
+prosody__ferm__dependent_rules:
+  - type: 'accept'
+    dport: [ '5222', '5269', '5280', '5281' ]
+    accept_any: True
+    weight: '40'
+    by_role: 'prosody'
+    name: 'http'
```
drybjed: This should probably be a different name, like `prosody` to avoid conflict with other potential configuration files generated by other roles.

ansible/roles/debops.prosody/defaults/main.yml
```diff
@@ -0,0 +1,162 @@
...
+    multiport: True
+    delete: '{{ prosody__deploy_state != "present" }}'
```
drybjed: You should use the `state` parameter instead of `delete` to manage the state of the configuration file properly.

ansible/roles/debops.prosody/handlers/main.yml
```diff
@@ -0,0 +1,6 @@
+---
+
+
+
+- name: 'Restart prosody'
+  service: name=prosody state=restarted
```
drybjed: A whitespace above the task is unnecessary. You should also write the task in a YAML format instead of inline.

ansible/roles/debops.prosody/meta/ansigenome.yml
```diff
@@ -0,0 +1,24 @@
+---
```
drybjed: The `meta/ansigenome.yml` files are not needed anymore, since roles don't need their own `README` files.

ansible/roles/debops.prosody/tasks/main.yml
```diff
@@ -0,0 +1,47 @@
+---
+
+
+- name: Install required packages
+  package:
+    name: '{{ item }}'
+    state: 'present'
+  with_flattened:
+    - '{{ prosody__base_packages }}'
+
```
drybjed: You could add here an additional `prosody__packages` variable to allow installation of additional APT packages with Prosody, if needed.

ansible/roles/debops.prosody/tasks/main.yml
```diff
@@ -0,0 +1,47 @@
...
+- name: copy config
```
drybjed: I would rephrase it as 'Generate Prosody configuration' to be more descriptive about what the task does during the playbook execution.

ansible/roles/debops.prosody/defaults/main.yml
```diff
@@ -0,0 +1,162 @@
...
+#
+# list of admins jabber accounts of this prosody instance
+prosody__admins: []
+                                                                   # ]]]
+# .. envvar:: prosody__config_global: [[[
+#
+# main prosody configuration
+# @see prosody.im
+prosody__config_global:
+  admins: '{{ prosody__admins }}'
```
drybjed: I wonder if the Prosody configuration could be rewritten to use the list of YAML dictionaries format used by Postfix-related roles. This would allow users to override the specific parameters without needing to copy entire configuration into the Ansible inventory. You can read more about how it works [in the `debops.postfix` documentation](https://docs.debops.org/en/latest/ansible/roles/debops.postfix/defaults-maincf.html).

drybjed (changes requested):
First of all, thanks for your contribution! :-)

There are few things I would like to see changed. It would be nice if you could try and integrate the role with the DebOps infrastructure right away, such as:
- enable a role test in the `.gitlab-ci.yml` file
- add the role documentation, remember to add an entry in the `role-index` file in the apropriate sections
- add a short Changelog entry mentioning a new role

ansible/playbooks/srv/prosody.yml
TuxCoder: thanks done.

ansible/roles/debops.prosody/CHANGES.rst
TuxCoder: thanks for pointing that out, will try to do good git messages.

ansible/roles/debops.prosody/README.md
TuxCoder: k, deleted

ansible/roles/debops.prosody/defaults/main.yml
TuxCoder: k thanks

ansible/roles/debops.prosody/defaults/main.yml
TuxCoder: thanks, done

ansible/roles/debops.prosody/defaults/main.yml
TuxCoder: thanks, should be fixed

ansible/roles/debops.prosody/defaults/main.yml
TuxCoder: I'm not sure what is the better choice. Agree with the aim to be as secure as possible.
But as gmail.com does not support secure connection and this is the only way to use,
a client can so choose to add a *@gmail.com contact or not.

I have deactivate it for now an offer a new parameter `prosody__insecure_domains`

drybjed: Separate variable for insecure domains is good. You can explain the issue with `gmail.com` contacts in the "Getting started" section of the role documentation where hopefully users will look when they want to enable this role and encounter issues. Then they can decide if they want to enable insecure connections for `gmail.com` domain or not. This should at least ensure that they are aware of the risk - otherwise they would just enable the role, which works out of the box, but users don't know that connections to `gmail.com` are insecure.

drybjed:
@TuxCoder You will need to resolve the conflicts in the Changelog, then the PR will most likely be tested again on Travis. I will also push a public test on GitLab to see if Prosody installs correctly when the conflicts are resolved.

.gitlab-ci.yml
```diff
@@ -947,6 +947,13 @@ stages:
     JANE_DIFF_PATTERN: '.*/debops.preseed/.*'
     JANE_LOG_PATTERN: '\[debops\.preseed\]'
 
+'prosody role':
+  <<: *test_role_1nd_deps
```
drybjed: This should be `*test_role_1st_deps`, then the syntax will be correct.

drybjed:
Here's a [GitLab CI test for `debops.prosody` role with latest changes](https://gitlab.com/drybjed/debops/-/jobs/51425124).

Edit: Test passed. The `debops.pki` hook was not installed because `debops.pki` role was not enabled for the test VM. However, I noticed that the role configured the expected certificate/key paths; I guess Prosody didn't flinch while starting up, but that might just be `systemd` not reporting an error.

If you want, you can enable `debops.pki` in the test to have more code paths (namely, the hook) tested, like this:
```yaml
'prosody/pki role':
  <<: *test_role_2nd_deps
  variables:
    JANE_TEST_PLAY: '${DEBOPS_PLAYBOOKS}/service/pki.yml ${DEBOPS_PLAYBOOKS}/service/prosody.yml'
    JANE_INVENTORY_GROUPS: 'debops_service_prosody'
    JANE_DIFF_PATTERN: '.*/debops.prosody/.*'
    JANE_LOG_PATTERN: '\[debops\.prosody\]'
```
Add this section in the `.gitlab-ci.yml`, after the `prosody role` one. That way, you can have both cases (with and without PKI) covered by the tests.

TuxCoder:
Thank you very much for your nice review/feedback/help, It really simplified my work.

Yeah prosody doesn't log so much, also default log is /var/log/prosody/prosody.{log,err}.
Probably change this to syslog/journald, not sure what is possible.
What I remember, it also doesn't crash with invalid configuration it only logs the error.

Thank for the last comment, added the second test to gitlab ci. I have to take a look into the taks/template file to also enable the service to work without debops.pki. That should be also done fast.

Getting started documentation is not complete yet, but plan also to be done tomorrow.
The rest should be more or less complete.

docs/ansible/roles/debops.prosody/getting-started.rst
```diff
@@ -8,6 +8,14 @@ Getting started
...
 ---------------------
 
+The configuration is split into 3 basic parameters,
+this is because of limitation of YAML and easier representation.
+
+- prosody__*_config_global
+- prosody__*_config_components
+- prosody__*_config_virtual_hosts
+
+By default there are two components active :ref:`prosody__http_upload` :ref:`prosody__muv`
```
drybjed: You can reference variables using the `:envvar:` instead of `:ref:`, for example:
```
:envvar:`prosody__http_upload`
```

drybjed:
The [test with PKI enabled](https://gitlab.com/drybjed/debops/-/jobs/51575349) completed, but I forgot that the `debops.pki` playbook needs a corresponding inventory group as well, and it wasn't activated. To do it, you need to change the test in `.gitlab-ci.yml` like this:
```yaml
'prosody/pki role':
  <<: *test_role_2nd_deps
  variables:
    JANE_TEST_PLAY: '${DEBOPS_PLAYBOOKS}/service/pki.yml ${DEBOPS_PLAYBOOKS}/service/prosody.yml'
    JANE_INVENTORY_GROUPS: 'debops_service_pki,debops_service_prosody'
    JANE_DIFF_PATTERN: '.*/debops.prosody/.*'
    JANE_LOG_PATTERN: '\[debops\.prosody\]'
```
(add the `debops_service_pki` group).

TuxCoder:
Known bugs:
 - ldap config is not complete
 - debops.pki is required, otherwise invalid cert/key paths are used
 - logs goes currently into /var/log/prometheus/* can be changed to syslog

I'm not sure how to fix the first two issues nicely 
Logs have to be testes, will try today.

The rest should be fine, probably will get a bit improved in the future.

drybjed:
I think that LDAP support in DebOps is in a half-backed state. Since I would like to preserve as much of the current state of roles as possible (moving existing features to separate roles, but not changing much otherwise) for the `v1.0.0` release, so that existing environments could switch to it reasonably easily, work on better LDAP support throughout the project will happen after `v1.0.0` is released.

For the PKI requirement, perhaps [this thread](https://groups.google.com/forum/#!topic/prosody-users/jfksbtvVKxE) could help?

For the logs, I hope you mean `/var/log/prosody/`? That's fine, as long as there's `logrotate` configuration to remove old logs.

Do you want to work on this PR some more, or do you think it's ready to go? Did you check that XMPP connections to the resulting server work? I'll test the PR in GitLab shortly.

drybjed:
Both tests, [normal](https://gitlab.com/drybjed/debops/-/jobs/51998616) and [with PKI](https://gitlab.com/drybjed/debops/-/jobs/51998656), passed fine.

TuxCoder:
sorry for the late response,
>For the logs, I hope you mean /var/log/prosody/? That's fine, as long as there's logrotate >configuration to remove old logs.

Oh of curse prosody, to much similar names :P

Think the rolle works for now, have tested it with the LenticularCloud system with some ldap configs.
Haven't tested it with default configs, but should work fine.

Also I think the role isn't working proper without PKI, but it doesn't make much sense to support prosody without TLS today. (It is possible that prosody only log the errors and start without tls support, so the tests run fine)

For me this is ok for now, a few parts are not perfect but will probably fixed later.
Start with a prometheus pull request :D

drybjed:
OK, in that case remove the GitLab CI test without PKI, since it's false, and I'll merge this after proper review.

ansible/roles/debops.prosody/handlers/main.yml
```diff
@@ -0,0 +1,8 @@
+---
+
+
+
```
drybjed: Can you leave just one empty line here? otherwise it looks kinda weird. :-)

ansible/roles/debops.prosody/meta/main.yml
```diff
@@ -0,0 +1,29 @@
+---
+
+dependencies:
+  - role: debops.ferm
```
drybjed: This dependency is defined in the `prosody.yml` playbook and shouldn't be defined here, since the `debops.ferm` role will be executed twice in this case. Please remove it and set `dependencies: []`.

docs/ansible/roles/debops.prosody/getting-started.rst
```diff
@@ -0,0 +1,70 @@
...
+- `5280` http
+- `5281` https
+
+
+
+Example inventory
+-----------------
+
+``debops.prosody`` is included by default in the :file:`common.yml` DebOps playbook;
+To enable it, add a host to the group `[debops_service_prosody]`.
```
drybjed: This is false, role is not included in the common playbook. You can replace this text with something similar to [NFS server inventory](https://docs.debops.org/en/latest/ansible/roles/debops.nfs_server/getting-started.html#example-inventory), for example.

drybjed (changes requested):
Looks good, thanks. There are a few small changes left, and it looks like this role can make it to `v0.7.0` release after all.

drybjed (approved):

PR #144, opened 2018-02-06
[debops.roundcube] Remove 'contrib' references, move docs

ganto:
This is an initial cleanup after moving the role to the monorepo. It doesn't contain Gitlab-CI and playbook integration yet as I first have to study the integration a bit more carefully.


drybjed:
Nice work, thanks. What do you think about merging the copyright and introduction documentation pages with the index, like in the [`debops.machine` documentation](https://docs.debops.org/en/latest/ansible/roles/debops.machine/index.html)? This should make browsing the docs a little easier, IMO. I want to convert the rest of the role docs eventually to this format.

drybjed:
BTW, you can test the documentation locally with `make docs`, you need `python-sphinx`, `python-sphinx-rtd-theme` and `graphviz` packages for it to work.

drybjed:
@ganto, you can refer to other roles in the documentation directly now, like this:
```
:ref:`debops.nginx`
```

ganto:
Thanks for your feedback.

> What do you think about merging the copyright and introduction documentation pages with the index

Ok, looks good. Will do that.

> BTW, you can test the documentation locally with make docs

Ya, I underrated the documentation update. First I thought that it can't be so difficult, but I became a bit rusty in the meantime. Will definitely test locally before sending the next commit.

> you can refer to other roles in the documentation directly now

Nice! Have to check if/where I'm using this.

ganto:
The doc updates should be fine now. 

So I had a look at the Gitlab-CI definition. Intuitively I would add a section like:
```
'roundcube role':
  <<: *test_role_2rd_deps
  variables:
    JANE_TEST_PLAY: '${DEBOPS_PLAYBOOKS}/service/roundcube.yml'
    JANE_INVENTORY_GROUPS: 'debops_service_roundcube'
    JANE_DIFF_PATTERN: '.*/debops.roundcube/.*'
    JANE_LOG_PATTERN: '\[debops\.roundcube\]'
```
Anything wrong with that? Or something missing? The playbook should be available from this PR. You want me to add another PR, or should I add it here?

drybjed:
You probably would want to have a MariaDB server on the test host, to setup the database. And the name of the template is wrong, it should be `test_role_2nd_deps`, as in `1st`, `2nd`, `3rd`, `4th`, etc. So, something like this:
```yaml
'roundcube role':
  <<: *test_role_2nd_deps
  variables:
    JANE_TEST_PLAY: '${DEBOPS_PLAYBOOKS}/service/mariadb_server.yml ${DEBOPS_PLAYBOOKS}/service/roundcube.yml'
    JANE_INVENTORY_GROUPS: 'debops_service_roundcube'
    JANE_DIFF_PATTERN: '.*/debops.roundcube/.*'
    JANE_LOG_PATTERN: '\[debops\.roundcube\]'
```
You can see [in an example GitLab CI pipeline](https://gitlab.com/debops/debops/pipelines/17220790) how the tests are separated into stages. The idea is to test the dependent roles in previous stage, so that if there are issues with them, your role won't be needlessly tested. There can be also multiple tests of a role with different parameters, different databases, webservers, etc. - for example ownCloud role is tested multiple times.

Yes, you can add the test in this PR. I will then push it to my public GitLab instance so you will be able to see how it performs. I'm doing this manually, so I'm not sure if it will be done today, but definitely before the PR is merged.

ganto:
I added the GitLab-CI definition but without the MariaDB dependency. By default Roundcube will use SQlite which should be fine for most people. If GitLab supports multiple configurations another run with MariaDB can be added later.

drybjed:
I've [tested the role](https://gitlab.com/drybjed/debops/-/jobs/51575354); the test failed but I know why. For some time now, Ansible uses the user account shell instead of `/bin/sh` to run commands using `become` keyword. This breaks the `command` and `shell` modules on accounts that use non-viable shells like `/bin/false` or `/usr/sbin/nologin`.

To avoid that, you can add the `executable` argument to the task arguments, this works only with the `shell` module. Basically something like this:
```yaml
- name: Run a command
  shell: ls -l
  args:
    executable: '/bin/sh'
  become: True
  become_user: 'some-user'
```

ganto:
Hmpf, I couldn't reproduce the issue on my Fedora with the 'command' module, therefore I thought it's safe to use. But only now I find that also the 'shell' module isn't failing. Will investigate...

drybjed:
I think that [this was introduced in Ansible 2.4](https://github.com/ansible/ansible/issues/26741), what Ansible version are you using?

ganto:
Something with the `become` is unexpected I guess:
```
$ ansible --version
ansible 2.5.0 (devel ea2b89c7ae) last updated 2018/01/26 18:55:53 (GMT +200)
  config file = /home/ganto/.ansible.cfg
  configured module search path = [u'/home/ganto/.ansible/plugins/modules', u'/usr/share/ansible/plugins/modules']
  ansible python module location = /home/ganto/repositories/upstream/ansible/ansible/lib/ansible
  executable location = /home/ganto/repositories/upstream/ansible/ansible/bin/ansible
  python version = 2.7.14 (default, Dec 11 2017, 16:08:01) [GCC 7.2.1 20170915 (Red Hat 7.2.1-2)]
```
The playbook:
```yaml
---
- hosts: localhost
  gather_facts: False

  tasks:
    - name: Save current checkout hash for update
      become: True
      become_user: gofed
      shell: git rev-parse HEAD
      args:
        chdir: '/home/ganto/repositories/upstream/ansible/ansible'
      register: register_head
      changed_when: False
    - debug: var=register_head
```
The user:
```
gofed:x:967:961:Gofed user:/:/sbin/nologin
```
The output:
```
$ ansible-playbook -K test.yml                                                                                                         
SUDO password:                                                                       
 [WARNING]: provided hosts list is empty, only localhost is available. Note that the implicit localhost does not match 'all'
                                                                                     
                                                                                     
PLAY [localhost] *********************************************************************************************************************************************************
                                                                                                                                                                          
TASK [Save current checkout hash for update] *****************************************************************************************************************************
ok: [localhost]                                                                      
                                                                                                                                                                          
TASK [debug] *************************************************************************************************************************************************************
ok: [localhost] => {                      
    "failed": false,                                                                 
    "register_head": {                    
        "changed": false,                                                            
        "cmd": "git rev-parse HEAD", 
        "delta": "0:00:00.004321", 
        "end": "2018-02-08 09:42:19.485267", 
        "failed": false, 
        "rc": 0, 
        "start": "2018-02-08 09:42:19.480946", 
        "stderr": "", 
        "stderr_lines": [], 
        "stdout": "ea2b89c7ae69424055652db30f49182d4a987bf7", 
        "stdout_lines": [
            "ea2b89c7ae69424055652db30f49182d4a987bf7"
        ]
    }
}

PLAY RECAP ***************************************************************************************************************************************************************
localhost                  : ok=2    changed=0    unreachable=0    failed=0
```

ganto:
Reading through the link you provided, it seems to me that the culprit is the shell of the user executing the playbook, not the target user (`become_user`) running the module. Can you confirm this? Does the test user on Gitlab not have a valid shell?

I just like to understand the issue. Arriving at the office now, but will fix it tonight.

drybjed:
I tested this locally with Ansible 2.4.3 and I couldn't reproduce it as well. A [new test with your latest changes passes OK](https://gitlab.com/drybjed/debops/-/jobs/51642526) so I suppose you're in the clear.

The test environment is a `debian/stretch64` Vagrant VM as a base, with installed Ansible from `stretch-backports`... Hmm, maybe that's a factor, not sure.

.gitlab-ci.yml
```diff
@@ -1009,6 +1009,15 @@ stages:
     JANE_DIFF_PATTERN: '.*/debops.roundcube/.*'
     JANE_LOG_PATTERN: '\[debops\.roundcube\]'
 
+'roundcube/mariadb role':
+  <<: *test_role_2rd_deps
```
drybjed: That would need to be `test_role_2nd_deps`. Maybe I need to change that to a simplier version...

.gitlab-ci.yml
```diff
@@ -1009,6 +1009,15 @@ stages:
...
+  variables:
+    JANE_TEST_PLAY: '${DEBOPS_PLAYBOOKS}/service/mariadb_server.yml ${DEBOPS_PLAYBOOKS}/service/roundcube.yml'
+    JANE_INVENTORY_GROUPS: 'debops_service_mariadb_server,debops_service_roundcube'
+    JANE_INVENTORY_HOSTVARS: 'roundcube__database=mysql-default'
```
drybjed: The role has no database autodetection implemented? That's fine, we can do it later. The goal should be that extra variables for tests aren't needed, that way the roles can automatically adapt to the existing environment.

ansible/playbooks/srv/roundcube.yml
```diff
@@ -0,0 +1 @@
+../service/roundcube.yml
```
drybjed: RoundCube seems to me as more of an end user application instead of a service. How about moving it to the `app.yml` playbook instead? That one is meant for user-facing services, basically.

docs/ansible/roles/debops.roundcube/upgrade.rst
```diff
@@ -0,0 +1 @@
+.. include:: ../../../../ansible/roles/debops.roundcube/UPGRADE.rst
```
drybjed: The upgrade notes from individual roles will need to be moved to the central location in `docs/news/upgrades.rst` file. In this particular case, I guess the upgrade notes don't need to be included right now, since the role was just introduced in the project. If somebody with an older installation comes over, they will be able to look up the notes in the old location via `git log`.

After the `v0.7.0` release the role directories will be cleaned up from old files (Travis, readmes, etc.).

drybjed (changes requested):
Very nice work, @ganto! There are just a few small issues with `.gitlab-ci.yml` and placement in the playbook, when that's resolved I think that this PR can be merged. You might make it into `v0.7.0` release after all. :-)

.gitlab-ci.yml
ganto: The role still uses the ancient database configuration table lookup pattern (originally copied from the wordpress role, or the like). IMHO, SQLite is a much better choice for most installations as the database has only infrequent updates and a file is much easier to backup. A real DBMS is only needed if you have multiple Roundcube instances and then you would want a separate database server anyway.

If you feel MariaDB should be auto-configurable I can send a PR for this after v0.7.0. It likely would break existing installations that's another reason why I hesitated to change this so far. But it might be a good occasion now, as there is some compatibility code which could be cleaned up in the role anyway.

docs/ansible/roles/debops.roundcube/getting-started.rst
```diff
@@ -41,10 +41,10 @@ Example playbook
 The following playbook can be used with DebOps. If you are using these role
 without DebOps you might need to adapt them to make them work in your setup.
 
-.. literalinclude:: ../../../../ansible/playbooks/service/roundcube.yml
+.. literalinclude:: ../../../../ansible/playbooks/app/roundcube.yml
```
drybjed: Argh, no... All playbook files are in the `service/` subdirectory, and then they can be symlinked to `app/`, `srv/` etc. And you can always call them like, `debops service/roundcube` this way.

.gitlab-ci.yml
drybjed: Sure, doing that after `v0.7.0` is released is fine. I want to add PostgreSQL support as well, so we will se how that goes. Another thing to consider is `memcached` autodetection to switch session management from database to memcache.

drybjed:
I think the PR is ready now. I'm just gonna test it in GitLab CI to see if everything passes.

drybjed:
Looks like the MariaDB test machine [needs more RAM](https://gitlab.com/drybjed/debops/-/jobs/51946035). @ganto, can you add a `mem-6GB` tag to this test, similarly how the `gitlab role` test is done? I'll test it most likely in the morning.

.gitlab-ci.yml
```diff
@@ -1017,6 +1017,7 @@ stages:
     JANE_INVENTORY_HOSTVARS: 'roundcube__database=mysql-default'
     JANE_DIFF_PATTERN: '.*/debops.roundcube/.*'
     JANE_LOG_PATTERN: '\[debops\.roundcube\]'
+  tags: [ 'mem-6GB' ]
```
drybjed: Great. I just forgot about [setting the Vagrant box memory](https://github.com/debops/debops/blob/master/.gitlab-ci.yml#L450) sorry about that. Can you add the
```yaml
  VAGRANT_MASTER_MEMORY: '4096'
```
variable as well?

drybjed:
Both tests, [with Sqlite3](https://gitlab.com/drybjed/debops/-/jobs/51986048) and [with MariaDB](https://gitlab.com/drybjed/debops/-/jobs/51986049) tests passed now. Nice work! Merging then.

drybjed (approved):

ganto:
Thanks for merging 👍

PR #149, opened 2018-02-08
[debops.php] Add a version condition for the mcrypt package

gabor-udvari:
In PHP 7.2 the MCrypt extension was moved out from core to PECL.
A condition was added to only include the mcrypt package if the PHP
version is lower than 7.2.

The [issue was originally raised](https://github.com/debops/ansible-php/issues/43) in the ansible-php repo.

### DIFF
--- a/ansible/roles/debops.php/defaults/main.yml
+++ b/ansible/roles/debops.php/defaults/main.yml
@@ -83,7 +83,7 @@ php__server_api_packages: [ 'cli', 'fpm' ]
 # .. envvar:: php__base_packages [[[
 #
 # Install set of standard PHP packages.
-php__base_packages: [ '{{ "php" + php__version }}', 'curl', 'gd', 'mcrypt' ]
+php__base_packages: [ '{{ "php" + php__version }}', 'curl', 'gd', '{{ "mcrypt" if php__version | version_compare("7.2","<") else None }}' ]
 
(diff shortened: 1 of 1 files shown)


ansible/roles/debops.php/defaults/main.yml
```diff
@@ -83,7 +83,7 @@ php__server_api_packages: [ 'cli', 'fpm' ]
 # .. envvar:: php__base_packages [[[
 #
 # Install set of standard PHP packages.
-php__base_packages: [ '{{ "php" + php__version }}', 'curl', 'gd', 'mcrypt' ]
+php__base_packages: [ '{{ "php" + php__version }}', 'curl', 'gd', '{{ "mcrypt" if php__version | version_compare("7.2","<") else None }}' ]
```
drybjed: For better readability, I would convert this to a vertical list. Also, using `None` like this might create issues if/when APT will try to install a `None` package. I would change this to `[]`, the package list is passed through a `flattened` Ansible lookup which will remove an empty list. In effect, something like this:
```yaml
php__base_packages:
  - '{{ "php" + php__version }}'
  - 'curl'
  - 'gd'
  - '{{ "mcrypt"
        if (php__version | version_compare("7.2","<"))
        else [] }}'
```

drybjed (approved):

PR #152, opened 2018-02-08
Allow disabling the nginx maintenance page

gaudenz:
While the maintenance page is generally a nice thing under some
circumstance it's not needed or desired. Eg. if another mechanism is
used to show a maintenance page.


drybjed (approved):
Sounds good, thanks.

PR #153, opened 2018-02-08
Fix variable name and call pattern.

cultcom:
and an indention complained by flake8


drybjed (approved):
Thanks for catching that omission! :+1:

PR #160, opened 2018-02-11
[WIP] debops.prometheus - add new roles

TuxCoder:
This is an early state but will be updated soon.

Goal of the PR is to add Prometheus client/server with Graphana to monitor one or more server(s).

TODO:
 - ~~upgrade to debian backport of prometheus to fix RAM issues~~
 - see known issues
 - `<add comment>`
- test tcpwrapper configuration

# Prometheus Server

prometheus_server__dependent_jobs is used to add the ability to use this role in another playbook.
(Not sure I done it right needs testing)

### known bugs/missing
 * nginx site is still enable after disable it by prosody_server__nginx
 * ~~nginx basic_auth is configurated but don't know how to create a htpasswd file~~
 * ~~prometheus_server__dependent_jobs can only manipulate  the list of target but not settings like protocol, metric-path,..~~
 * ~~docu~~

# Prometheus exporter

### known bugs/missing
 * ~~docu is only partial partial~~
 * same nginx problems as prometheus server
 * postfix exporter not tested
 *~~use debops.go role~~

# Prometheus Alertmanager

### known bugs/missing
 * ~~default varibales~~
 * ~~config~~
 * ~~split main config into variables~~
 * ~~docu~~
 * same nginx problems as prometheus server


ansible/playbooks/service/grafana.yml
```diff
@@ -0,0 +1,18 @@
...
+  roles:
+    - role: debops.apt
+      tags: [ 'role::apt', 'role::grafana' ]
+      apt__repositories: '{{ grafana__apt__repositories }}'
+      apt__keys: '{{ grafana__apt__keys }}'
+    - role: debops.ferm
+      tags: [ 'role::ferm', 'role::grafana' ]
+      ferm__dependent_rules:
+        - '{{ grafana__ferm__dependent_rules }}'
+    - role: debops.nginx
```
evilham: This is not doing anything and having nginx should probably be optional.

ansible/playbooks/service/grafana.yml
```diff
@@ -0,0 +1,18 @@
+---
+
+
+- hosts: [ 'debops_service_grafana' ]
+  roles:
+    - role: debops.apt
+      tags: [ 'role::apt', 'role::grafana' ]
+      apt__repositories: '{{ grafana__apt__repositories }}'
+      apt__keys: '{{ grafana__apt__keys }}'
+    - role: debops.ferm
```
evilham: Just setting up `ferm` is not enough, you need `tcpwrappers` as well.

ansible/playbooks/service/prometheus_server.yml
```diff
@@ -0,0 +1,16 @@
+---
+
+
+- hosts: [ 'debops_service_prometheus_server' ]
+  roles:
+    - role: debops.apt_preferences
+      tags: [ 'role::apt_preferences', 'role::prometheus' ]
+      apt_preferences__dependent_list:
+        - '{{ prometheus_server__apt_preferences__dependent_list }}'
+    - role: debops.ferm
```
evilham, line 22: Just setting up ferm is not enough, you need tcpwrappers as well.

ansible/playbooks/service/prometheus.yml
```diff
@@ -0,0 +1,18 @@
...
+
+
+- hosts: [ 'debops_service_prometheus' ]
+  environment: '{{ inventory__environment | d({})
+                   | combine(inventory__group_environment | d({}))
+                   | combine(inventory__host_environment  | d({})) }}'
+  roles:
+    - role: debops.prometheus
+      tags: [ 'role::prometheus' ]
+    - role: debops.ferm
```
evilham: Just setting up ferm is not enough, you need tcpwrappers as well.

ansible/playbooks/service/prometheus.yml
```diff
@@ -0,0 +1,18 @@
...
+      tags: [ 'role::ferm', 'role::prometheus' ]
+      ferm__dependent_rules:
+    - role: debops.mariadb
```
evilham: This probably does not belong in this role. See proposal in #160

ansible/roles/debops.grafana/defaults/main.yml
```diff
@@ -0,0 +1,28 @@
...
+
+grafana__ferm__dependent_rules:
+  - type: 'accept'
+    dport: [ '3000' ]
+    accept_any: True
+    weight: '40'
+    by_role: 'grafana'
+    name: 'http'
+    multiport: True
+    rule_state: '{{ "present" if (grafana__deploy_state|bool and grafana__public) else "absent" }}'
```
evilham: This rule can be generalised a bit, instead of having `grafana__public`, you can use something more like what `nfs` uses.
https://docs.debops.org/en/latest/ansible/roles/debops.nfs_server/defaults.html?highlight=allow_any#envvar-nfs_server__allow
See `nfs_server__allow` and `nfs_server__accept_any`.

ansible/roles/debops.prometheus/defaults/main.yml
```diff
@@ -0,0 +1,58 @@
...
+prometheus__ferm__dependent_rules_postfix_exporter:
+  - type: 'accept'
+    dport: [ '9154' ]
+    saddr: '{{ prometheus_servers }}'
+    accept_any: True
+    weight: '40'
+    by_role: 'prometheus'
+    name: 'prometheus_postfix_exporter'
+    multiport: True
+    delete: '{{ prometheus__deploy_state != "present" }}'
```
evilham: Same comments as for grafana's rules apply here: beter to have `__allow` and `__allow_any` variables, with sane defaults (e.g. `__allow` being all prometheus servers and `__accept_any` being `False`).

ansible/roles/debops.prometheus_server/defaults/main.yml
```diff
@@ -0,0 +1,44 @@
...
+  - prometheus
+  - grafana
+
+prometheus_server__packages: []
+
+prometheus_server__apt_preferences__dependent_list:
+  - packages:
+    - prometheus
+    backports: [ 'stretch' ]
+    reson: 'Stable verson has some ram leaking problems'
```
evilham: Should read `reason`.

evilham:
Ahà! Glad to see this @TuxCoder 
I was, as a matter of fact, just starting this myself :).

I know you said you'd ping when it's ready for review, but since it was also my intention to have my roles land on debops and I had given the thing some thought, I have some comments, so we can discuss and come up with a role that works out for the debops user base. I am of course, willing to help make this happen :) if it's two of us working on it and with some debops-father (@drybjed) guidance, it can be a very good thing.

First things first: it looks like a good starting point, particularly the decision of separating into 3 roles (`grafana`, `prometheus` and `prometheus_server`).

I made a couple comments in a review, you may already be aware of some things.

To have this land on debops, probably the easiest is if you split this pull requests in a couple, just like the roles:
- `grafana` role can be perfectly usable on its own with data sources other than prometheus and therefore its acceptance should not depend on this as a whole.
- `prometheus_server` role: to setup the actual prometheus server, it shouldn't assume grafana runs on the same host, it also shouldn't have anything to do with exporters.
- `prometheus_exporters` role: to provision nodes with a small set of well-supported exporters like the `node-exporter`, `mysql`, `postfix`, `dovecot`, ...
        Having the role being called `prometheus` only may end up being confusing.
- `prometheus_alertmanager` role: to setup an alertmanager.

In order to allow for redundancy, high availability and multiple prometheus instances, it should be possible to specify on the exporters / alertmanagers which prometheus_servers they should be added to. Similar to the way `mariadb` and `postgresql` roles work.
At some point it should be possible to have alertmanagers communicate and be part of a mesh.

For a full-instance deployement that runs grafana + prometheus_server + prometheus_alertmanager + nginx together, I think we can create a special playbook that runs on a special group; @drybjed will know better how to handle that. But I do think that keeping it flexible here is a very good thing :).

Firewall
========
You setup `ferm` but don't offer a way to setup `tcpwrappers`, the way you are setting up `ferm` is also quite limited.

In general, I'd think that the way `nfs` does it is pretty much it. See `nfs_server__allow` and `nfs_server__accept_any`:
https://docs.debops.org/en/latest/ansible/roles/debops.nfs_server/defaults.html?highlight=allow_any#envvar-nfs_server__allow

Re:`graphana`, I would think that the default should be having it *not* accessible over TCP, but require admins to either override the config or setup nginx in front (with TLS and all the awesome debops goodies).

Re:`prometheus_server`, I would also think that the default should be having it *not* accessible over TCP, but offer ways to change that. E.g. it should be possible to whitelist grafana's host.

`grafana`
=========

Authentication
--------------
There should be variables that manage the default user and leave the setup with a sane config (*not* admin/admin).

Dashboard and datasource management
-----------------------------------
Ideally, but this may be too much work for a first version, dashboards and datasources could be managed by debops, maybe one way of doing that would be to expose `grafana__dependent{dashboards,datasources}` and have other roles use that.

This way, the `prometheus_server` role could just add itself as a datasource to graphana and setup its firewall to allow the connection.

`prometheus_server`
===================

This role should not install grafana by default; they could be running in different machines.

It should also setup file_discovery, that way we can add exporters and rules from other roles without having to fiddle with the config file.
See: https://prometheus.io/docs/prometheus/latest/configuration/configuration/#file_sd_config

drybjed:
@evilham Thanks for the review, very nice ideas, especially about role namespacing. I haven't really played with Prometheus myself, so I'm trusting you guys to make this right. :-)

As for the "full stack" playbook, I'm playing with an idea of stacks like LEMP/LAMP playbooks myself, most likely in the `ansible/playbooks/stack/` subdirectory, so that they could be executed like:
```
debops stack/lemp -l hostgroup
```
The playbooks wouldn't be included in the "main" site.yml (hmm, maybe... not sure yet), and would use `[debops_stack_*]` group namespace to distinguish them from standalone services. It's at least an idea we can play around with and see what we get.

evilham:
I had forgotten to post the rest of the review :-D.

@drybjed: that sounds quite sane!

I'll wait a couple days on @TuxCoder's reply before resuming work on this myself, it'd be nice to finish the work with some consensus :).

ansible/playbooks/service/prometheus_server.yml
TuxCoder: changed my minde about exporting prometheus/exporter to the network

TuxCoder: I would make them only available through a nginx proxy to enable TLS by default

TuxCoder:
@evilham  Thanks for your review.
will try to go today your comments

My idea to get a bit more security by default is to proxy all prosody traffic throw nginx to enable TLS by default. This can be enabled by `prometheus_*_public: True` 
Should solve all ferm/tcpwrapper issues.

I also like the naming of the your roles, will use them.

TuxCoder:
For the exporters I going to use subdirs like

 * https://monitor.example.com/exporter/postfix
 * https://monitor.example.com/exporter/node
 * https://monitor.example.com/exporter...

nginx forwards this then to local ports.

scibi:
Hi Guys!

Long time ago I've made role for Grafana which I intended to add to the project. Unfortunately I didn't had time to add documentation and polish it. Feel free to take a look at it and use it if you like it: https://github.com/scibi/ansible-grafana

evilham:
@scibi that looks pretty good :). I'd think it's a better starting point regarding graphana, so it totally is worth it.

As for the exporters @TuxCoder:
I was also considering doing something like that (nginx proxy in front of exporters) and decided against it. A few reasons:
* You effectively introduce a single point of failure for your metric gathering. If nginx is down for whatever reason (or its TLS cert is not working or....) then prometheus will see EVERYTHING as being down.
* With debops it is fairly trivial to limit TCP access to the exporter's port only from the prometheus instances that are supposed to be pulling.
* It actually simplifies deployment of exporters by not having to fiddle around with nginx config.
* Some machines that may need monitoring, may also not be running nginx at all. It is not a nice thing to mix with the concept of exporters.

TuxCoder:
@scibi Thanks for offer take a look an probably copy some suff

@evilham 
Good point, change it back to port based for the exporters and prometheus server.
Think for grafana it is ok to let the nginx proxy


Grafana
======
I have looked into the config
It is posible to set a default admin password and the database backend
But I didn't find a way to setup dashboard or datasource excepct editing the database.
If you have a good idea you are welcome

drybjed:
Guys, keep in mind that TCP ports are a limited resource, and thus scarce (you can only have so much). But I guess going with ports is inevitable to not rely on too much services in the "stack". TLS will still be supported, I hope?

If you go with ports, please ensure that you register each port in the `/etc/services` database via `debops.etc_services`. The `debops.postgresql_server` role has an example how to do this with dynamic list of services.

evilham:
@drybjed: re: TCP ports, it took me a while to accept that as well, but it's just the way prometheus exporters are designed and in that use-case there's no way around it.
As for TLS support/authentication, it is explicitly left out from the design of prometheus itself and of any exporters.
I found this ML thread on the topic, where the developers explain a bit why/how: https://groups.google.com/forum/#!topic/prometheus-developers/hICTGceHUGQ
That being said, it totally is possible and in fact fairly easy to put nginx in front with debops :).
I would just think it's a matter better left to the administrator, with a nice example in the documentation on how to do it and a quick explanation with links to why one would want one solution or the other.
Making too many assumptions here would end up biting us later because it will be a matter of taste: for some the metrics won't be something super-secret and having TCP access whitelisted is OK, for others that may be unacceptable and they'll want to go with Nginx with TLS + HTTP Auth in between; it'd just be a matter of the admin making sane use of the rest of debops :).

@TuxCoder: re: dashboards and datasources in grafana: AFAIK there's no way around editing the database,. Which is doable but may be an overkill for a first version of the grafana role :). Probably something we'll want to add at some point though.

drybjed:
@evilham In that case I think that sane policy would be to block TCP access to exporters in the firewall by default, and allow easy unblocking for specific subnets. The nginx setup doesn't need to be included in the role/playbook, but it would be very good to have detailed explanation in the documentation about how to publish exporters via `debops.nginx`, maybe even with an alternate playbook like ownCloud and Nexcloud, for example, to make it easier.

drybjed:
NB: YAML tests are now enabled, you can use `test yaml` to check the syntax via `yamllint`.

scibi:
@TuxCoder What do you think about removing Grafana role from this PR and making new PR just for Grafana? IMHO this way it could be finished and merged faster.

I don't think it's a good ideat to manage dashboards in Grafana through Ansible. Dashborads configuration is just application data and it is edited by users. In the future it may be nice to be able to provide initial configuration but IMHO you should focus on in in the first version.

TuxCoder:
Made a fork of the current master and added (hopfully) all grafana files.
Is available here https://github.com/TuxCoder/debops/tree/grafana
Thanks to @scibi have merged the best partes.
Not sure what way to add the repose are better,
try to use the debops way with the roles.


In that case will concentrate me on the prometheus roles first.

TuxCoder:
@evilham 

### Prometheus
Thanks for the explanation. Have totally forgot to add also base auth -.-
My plan is currently to ether add a example inventory
or a boolen switch to enable nginx role (proxy).
But tending to have a boolean like `prosody_exporters__nginx_proxy` to enable it.

### Grafana
Agree move fancy default dashboard to a later release.
Would add a link/description how to get started and that it has to be get changed.
Also were you can find the generated default username/password

drybjed:
@TuxCoder For adding APT repositories for Grafana, you can just use the Ansible `apt_key` and `apt_repository` modules, no need to delegate this to `debops.apt` role at all.

Small suggestion, create `grafana__upstream` variable, set to `True` by default and install the repos with it enabled, this way users can decide to use either the upstream repo, or provide the packages themselves via local APT repository if they wish.

TuxCoder:
@drybjed Thanks for the tips

Prometheus Server
---------------------
Most stuff expect from docu should be done,
`prometheus_server__dependent_jobs` is used to add the ability to use this role in another playbook.
(Not sure I done it right needs testing)

### known bugs
 * nginx site is still enable after disable it by `prosody_server__nginx`
 * nginx basic_auth is configurated but don't know how to create a htpasswd file

TuxCoder:
Have updatend the first post with a status of the roles.

Has anyone a idea how to solve the problem that a few settings from a job are in the main config file `promehteus.yml` and  not in the included file `jobs/<name>.json`.
So my solution with `prometheus_server__dependent_jobs` doesn't work that way,
because on the next run i don't have this `prometheus_server__dependent_jobs` variable information in the promehteus_server playbook.

drybjed:
For the dependent variable, you should ensure that its state is stored somewhere; keep in mind that multiple different roles can keep their own state. As for how to do this, there are different solutions - if an application supports `conf.d/` style configuration, writing a separate config file for each dependent role is ideal. Less cooperative applications, like `/etc/hosts.allow`, can be handled by doing `conf.d/` directory "manually" and assembling the finished result, but this is of course a bit brittle. Alternatively you could move the dependency handling entirely to Ansible side and do something like variable store and merging on Ansible Controller, a'la `debops.postfix`. That gives you powerful control over the entire config, but you push application state back to Ansible Controller. Pros and cons.

If the data you need to track is small, keeping them in Ansible facts and making sure that dependent config is preserved is also an option, a few different DebOps roles use that.

scibi:
Hi Guys!

I was just reading Grafana changelog. In Grafana 5.0 (it's beta currently) you can configure data sources and dashboards with config file. So it will be easy to configure with ansible :)

http://docs.grafana.org/administration/provisioning/#datasources

TuxCoder:
@scibi
Thanks for the info.
Will take a look, but wait until it is released as stable to add it to the role.

drybjed:
@TuxCoder Thanks for including GitLab CI configuration. I [pushed your PR for testing](https://gitlab.com/drybjed/debops/pipelines/17925973), let's see how it will go.

Edit: The tests failed because playbooks are executed on an unprivileged user account. You need to add `become: True` to the playbooks for them to work as expected.

TuxCoder:
@drybjed 
Cool.
Yeah forgot that, thanks for pointing out.

@evilham , @drybjed ,...

Can you give me some feedback if this is fine for you.

## Prometheus Server

Because of I can not save all infos in the `file_sd_config` included file from prometheus I changed the format, that it saves one `scrape_config` entry from prometheus in such a file in `/etc/prometheus/jobs/*.yml`

When building the main config `prometheus.yml` it puts all files together, what should work fine.
It needs more testing haven't tried yet to use the `prometheus_server__dependent_jobs` variable

Documentation is found here https://github.com/TuxCoder/debops/blob/8af6819337e2771af94dae05f3ebc4e46c2dfa58/docs/ansible/roles/debops.prometheus_server/defaults-detailed.rst

Is it also needed to be able to change the global/rule_files config?
Think it would be fine, will try to do this later.

More infos about the prometheus config https://prometheus.io/docs/prometheus/latest/configuration/configuration/

drybjed:
@TuxCoder I think that for a first implementation, assembling YAML from file parts should be OK. Prometheus configuration seems to be very extensive, but I suspect that only some sections of it would be valid to configure via role dependent variables and track them, the rest seems to be strictly related to the service itself and probably shouldn't be exposed to other roles.

In the future it might be interesting to remodel the configuration, so that parts of it can be enabled or disabled conditionally. This would enable some useful default configuration in the role, since right now you have hardcoded configuration like use of the `example.com` domains, which would require initial modification by the user. For now, I would move these to documentation as examples and set the default variables as empty, unless some initial configuration is required by the service to even start properly. In that case, try to use `ansible_domain` and other such facts; `debops.core` role provides private and public admin mail addresses as local facts, it might be interesting to hook into these.

docs/ansible/role-index.rst
```diff
@@ -160,9 +160,6 @@ Logging
 - :ref:`debops.kibana`
 - :ref:`debops.logrotate`
 - :ref:`debops.rsyslog`
-- :ref:`debops.prometheus_server`
-- :ref:`debops.prometheus_exporters`
-- :ref:`debops.prometheus_alertmanager`
```
drybjed: You can definitely put roles in multiple index sections, it will be easier to find them that way.

drybjed:
Another GitLab CI test after changes: https://gitlab.com/drybjed/debops/pipelines/17940810

docs/ansible/role-index.rst
TuxCoder: I don't think it matches logging, it was a misstage to put it there.
It matches much more monitoring.

SuperQ:
There is no need to "namespace" exporters under `/srv/prometheus`.  They are stand-alone software with no explicit dependency on Prometheus itself.  For example, it should be simply `/srv/node_exporter` or `/srv/postfix_exporter`

TuxCoder:
@SuperQ thx for review, change the namespace, testing it an will be pushed

drybjed:
New test with GitLab CI for current changes: https://gitlab.com/drybjed/debops/pipelines/17962204

drybjed:
I [tested latest changes in the PR](https://gitlab.com/drybjed/debops/pipelines/18457972). I'm not sure about the [alertmanager error](https://gitlab.com/drybjed/debops/-/jobs/55916233), it might be something to do with the default example configuration.

The [prometheus server error](https://gitlab.com/drybjed/debops/-/jobs/55916234) is due to UUID-based hostnames used during testing. I'll increase the `nginx` hash bucket size in the default inventory used for tests to mitigate that.

drybjed:
[Test of the current changes in the PR](https://gitlab.com/drybjed/debops/pipelines/18884383) via GitLab CI. The `nginx` issue with long server names should be resolved now.

ansible/roles/debops.prometheus_exporters/defaults/main.yml
```diff
@@ -14,9 +14,9 @@
 #
 # List of base packages to install.
 prometheus_exporters__base_packages:
-  "{{ 'prometheus-node-exporter'     if 'node' in prometheus_exporters__exporters else []
-    + 'prometheus-mysqld-exporter'   if 'mysqld' in prometheus_exporters__exporters else []
-    + [ 'golang', 'libsystemd-dev' ] if 'postfix' in prometheus_exporters__exporters else [] }}"
+  "{{ ([ 'prometheus-node-exporter'   ] if 'node' in prometheus_exporters__exporters else [])
+    + ([ 'prometheus-mysqld-exporter' ] if 'mysqld' in prometheus_exporters__exporters else [])
+    + ([ 'golang', 'libsystemd-dev', 'build-essential' ] if 'postfix' in prometheus_exporters__exporters else []) }}"
```
drybjed: Perhaps relying on `debops.golang` to pull Go dependencies conditionally could be better, the role can setup APT preferences to install Go support from backports. Remember to include its config for `debops.apt_preferences` in the playbook.

TuxCoder:
Idea to provide a easy to setup TLS connection by default:

Facts about prometheus config.
You need a seperated entry for different `path`, `protocoll`, `credentials`,...
see https://prometheus.io/docs/prometheus/latest/configuration/configuration/#<scrape_config>

Add dynamic settings are not so easy to be added to `debops.nginx`, for a global service like `https://.*/.well-known/promehteus/...`


So my solution.
additional to the default metric port of an exporter (normal 9xxx) add a secound secure tls encryped one (probably 19xxx), with nginx as proxy.
Firewall blocks traffic to 9xxx by default.

Benefits:
* no extra domain/cert is needed,
* no extra path for different exporters, only different ports

Cons:
* more ports are needed

ansible/playbooks/srv/all.yml
```diff
@@ -71,3 +71,10 @@
 - import_playbook: fail2ban.yml
 
 - import_playbook: prosody.yml
+
+- import_playbook: prometheus_exporters.yml
+
+- import_playbook: prometheus_server.yml
+
+- import_playbook: prometheus_alertmanager.yml
+
```
gasuketsu, line 88: Seems this blank line causes test fail in travis.
https://travis-ci.org/debops/debops/jobs/359868266

alinalexandru:
I was looking on how to install prometheus using debops.
I can't wait to see it merged into upstream.

drybjed:
@alinalexandru You could make this happen faster by checking if there are any issues with deployment. If you want to know how to do this, let me know.

TuxCoder:
So a few last question.

[common]
This role has the ability to secure the the blank http servers from `prometheus_*` by adding an nginx proxy.
It listen only on https traffic and needs base auth login.
Currently it is configured that the nginx server runs on the domain `'{{ ansible_fqdn }}'` with the port 19xxx instant of 9xxx.
Nginx proxy is currently enabled by default only for the exporters, but can be done by one variable `prometheus_*__nginx: {True,False}`
I don't add currently the secured ports to `debops.etc_services`, only the http ports, should I?
What kind of service name? ("nginx_prometheus_*" or only "https")

[debops.prometheus_alertmanager]
smtp from address is currently `'alertmanager@{{ ansible_domain }}'`, should this be changed to ansible_fqdn?
host is localhost:25
ref: https://github.com/debops/debops/pull/160/files#diff-7b67d77ded8e7fff46620e0ca386d8e3R90


currently can not confirm that all is working, have some problems with the postfix exporter...

[edit]
try to clarify that I add a '1' in front to the insecure server

TuxCoder:
[debops.prometheus_server]
The rules for generating alerts can get kind of complicated and have to be the same for multiple hosts.
Is there a good way to have some "templates" stored?
Probably there is also a good collection out in the Internet.
The rules are yaml code.
example:
```yml
        - name: 'disk_usage.rules'
          rules:
            - alert: DiskUsage
              expr: node_filesystem_avail/node_filesystem_size < .05
              for: 5m
              labels:
                severity: error
              annotations:
                summary: !unsafe 'Instance {{ $labels.instance }} Disk usage'
                description: !unsafe "{{ $labels.instance }} of job {{ $labels.job }} has less space on partition {{ $labels.device }} mounted at {{ $labels.mountpoint }}, {{ $value }}% left"
```

drybjed:
TCP port names should be fairly short, otherwise the output of various commands might not be readable. Perhaps something like:
```
prom-https-*
```
would suffice. Don't use underscores for TCP port names.

Is alertmanager cluster-wide, or only host-based? If cluster-wide, I would stick to `ansible_domain` by default.

As for the rules, that depends on how you configure them and what can be used as "input". By default inventory variables are probably a must, since role code is read-only. You can go with flat list of YAML dictionaries; if you use `parse_kv_items` filter you could even allow for modification of specific parameters without the need to copy entire entries from role defaults to the inventory.

Is that example you pasted a way a given template is represented in the role defaults, or is that entire snippet a "template" which needs to be configured as-is in Prometheus? In latter case having Jinja in it will probably mess with Ansible and including the templates in the inventory variables directly. In that case you could go with either telling users to use `debops.resources` role to copy the template files over, or include 2-3 tasks that create required directories and copy files from specific directory on Ansible Controller. Including the templates in the role could also be an option, but in that case you probably would want to provide a way to disable specific or all templates configured that way, for example with a list of templates enabled/disabled.

TuxCoder:
ok will change port names and add both http and https.

the alertmanager is normaly used for the monitored cluster,
I would say from is `prometheus@{{fqdn}}` and to is `root@{{ansible_domain}}`.

ah nice will take a look at `parse_kv_items`, thanks for the tip.

**template:**
currently I use the config from the inventory directly as prometheus config, because both is yaml.
So you have the max flexibility and less effort for support new features.

I thought about a few predefined variables, probably in default/main.yml, if thats the rights place.
The structure should allow concat the required rules in groups and add them to the `prometheus__*_config`

```yml
#in default/main.yml
prometheus__rules_node_disk_full_: 
 name: '...'

prometheus__rules_node:
 - {{ prometheus__rules_node_disk_full }}

# in inventory
prometheus__hosts_config:
 - {{ prometheus__rules_node }}
 - {{ prometheus__rules__smtp }}

**Comment by drybjed** [comment]
If you want to go with `parse_kv_items`, the proposed variables are already obsolete. :-) Think of this solution as a pipeline, where the role takes a list of YAML dictionaries, combines them and merges into one YAML dictionary of elements that define the configuration, and expands back to a YAML list of dictionaries to be parsed by the final template tasks. What you can do with this is define the rules in one default variable:
```yaml
prometheus__default_config:
 - name: 'disk_usage.rules'
    state: 'present'
    options:
      - name: 'Alert_Disk_Usage'
        alert: DiskUsage
        expr: node_filesystem_avail/node_filesystem_size < .05
        for: 5m
        labels:
          severity: error
        annotations:
          summary: !unsafe 'Instance {{ $labels.instance }} Disk usage'
          description: !unsafe "{{ $labels.instance }} of job {{ $labels.job }} has less space on partition {{ $labels.device }} mounted at {{ $labels.mountpoint }}, {{ $value }}% left"
```
Then,  in the inventory, users can modify some parameters indirectly, for example to disable a rule:
```yaml
prometheus__host_config:
  - name: 'disk_usage.rules'
    state: 'absent'
```

Or, disable a specific alert in the rules file:
```yaml
prometheus__host_config:
  - name: 'disk_usage.rules'
    options:
      - name: 'Alert_Disk_Usage'
        state: 'absent'
```
The filter is pretty flexible, however it's limited in some aspects, for example there's no support for recursive configuration. There are now many examples of how the filter can be used, you can look at newest DebOps roles for details (check in the Changelog).

mikenowak:
@TuxCoder - thanks for you work on this.

A few things that came to mind while looking at the code.

1) Installing golang and all dependencies on each and every host where we want to run exporter (in this case postfix_exporter) isn't probaby the best idea, we could perhaps delegate compilation to localhost (ansible controller?) and then just copy the binary across?
2) A more flexible format for exporter definition would be welcome - such as:

```
- name: 'node'
  repo: 'https://github.com/kumina/postfix_exporter.git'
  version: 'master'
  port: '9100'
  options: '--collector.systemd'
```
Also some repositories provide releases, so as an alternative to compiling we could just fetch a binary instead. This gist might be of assitance https://gist.github.com/lukechilds/a83e1d7127b78fef38c2914c4ececc3c

This should be enough to create firewall rules, services, nginx vhosts for each of these and will definitely allow for more flexibility going forward.

Let me know what you think.

TuxCoder:
@mikenowak Thanks for the tips, will test a bit more and clean up code
Also in the future go builds will be done on host, probably to that optional, not sure yet.
I will deploy it on a bit more machines, and test it again.

drybjed:
About the builds on the Ansible Controller - it's tricky to do that correctly. You don't know what OS and environment is on the Ansible Controller (Linux? macOS? WSL?), you don't know if any required development packages are available and there's no way to handle that from the playbook "from the inside" - you would have to treat the Ansible Controller as a remote host to do something like that on it.

Therefore compiling it on the remote hosts is an only option. But there might be a better way - ask the upstream authors of the plugins you want to use to publish binary releases, signed with their own GPG keys. This is usually done on GitHub it's a good alternative to building them from source. Of course you would have to verify the code you download that way, but using a GPG key for that is easy. An example of something like this in DebOps is `debops.hashicorp` role that lets you download Go binaries of many HashiCorp applications.

Another alternative would be to push the exporters to Debian, but of course that requires finding maintainers, and the results will be readily available on Stable+1 release (at this point Debian Bullseye). But that can bring long-term benefits, not only to DebOps environment and its users, but to the wider FLOSS community.

TuxCoder:
@drybjed thanks for your feedback, will look for binary releases, makes all simpler. I'm still not sure if this postfix exporter works, have to debug some stuff

At leased fixed another bug with binary upload :)

drybjed:
@TuxCoder I just found [ITP request for `prometheus-postfix-exporter`](https://lists.debian.org/debian-devel/2019/01/msg00298.html), it seems that it will be available in Debian Archive at some point. Not in Debian Buster though, it's too late for that. Cross your fingers for a backport in the future. :-)

mikenowak:
@TuxCoder, @drybjed,

> About the builds on the Ansible Controller - it's tricky to do that correctly. You don't know what OS and environment is on the Ansible Controller (Linux? macOS? WSL?), you don't know if any required development packages are available and there's no way to handle that from the playbook "from the inside" - you would have to treat the Ansible Controller as a remote host to do something like that on it.

We actually don't need any development packages other than golang itself on the machine where the code will be compiled (again, my preference would be that it was the ansible controller). Golang programs generally don't use any external dependences (unlike c programs), and the required libs are usually fetched from github with `go get ...`. 

Also, architecture wise i.e. MacOS controller vs Linux endpoints - Go supports cross compiling for other architectures, so not a problem either.
```
env GOOS=target-OS GOARCH=target-architecture go build ...

i.e. 

env GOOS=windows GOARCH=amd64  go build ...
```

> Another alternative would be to push the exporters to Debian, but of course that requires finding maintainers, and the results will be readily available on Stable+1 release (at this point Debian Bullseye). But that can bring long-term benefits, not only to DebOps environment and its users, but to the wider FLOSS community.

While backporting is generally a good idea, its more of a mission impossible for over 100 exporters that are currently available for prometheus. So lets park that idea for the moment.

ganto:
Thanks for this cool role. I'm very interested in prometheus support and will try it out later to give you some feedback.

> Also in the future go builds will be done on host, probably to that optional, not sure yet.

Maybe you can temporarily drop support for the postfix exporter for now to get the role merged and then add it back later once the Debian package is ready? It's a pity that such a great role is hanging around as PR for so long.

ganto:
I created TuxCoder/debops#1 with some cleanup issues which hit my eye.

Here some feedback after using the server and exporter role on a few hosts:
* I locally removed the dependency to `debops.golang` as I don't need/want any go packages to be installed on my hosts
* I saw that there are already a few additional exporter packages available in stretch-backports (apache, postgres) would be nice to have support for them too
* I used both roles without Nginx proxy which was working fine
* I was using the Stretch backports packages (v0.17) of `node_exporter` (to be consistent with my Gentoo, Fedora and CentOS hosts. To do so, I set in my DebOps inventory (`group_vars/debops_service_prometheus_exporters/apt_preferences.yml`):
  ```
  apt_preferences__group_list:
    - packages:
        - prometheus-node-exporter
      backports: [ 'stretch' ]
      reason: 'Use Prometheus 0.17 from backports'
  ```
* It would be nice if one could set the command line arguments for both services (`/etc/default/prometheus-node-exporter`, `/etc/default/prometheus`)

It's not a lot of magic in the roles yet (e.g. still have to manually define the scraping jobs even we would have a list of exporters managed by debops) but the roles are useful, functional and I also would be willing to do some improvements to them.

@TuxCoder: Do you think you can get it in a shape that it can be merged? From my pov get rid of the golang stuff for now.

inthehack:
Hi, what is the status of this PR?

ypid:
There seems to be a a different PR #1310 which also adds Prometheus support. @TuxCoder did you look at it yet?

I am starting to review #1310 to make up my mind.

TuxCoder:
sorry lost a bit this  PR, will close as not working on it any more.
Fell free to copy paste as much code from it as you want.

PR #172, opened 2018-02-20
Fix issues with 'debops-padlock' on Python 3

drybjed:


drybjed:
@cultcom I think that this should make the script work on Python 3. Unfortunately, I don't have a Debian Jessie Py3 environment handy to test this, and [`encfs` in Debian Stretch is broken for our usage](https://github.com/debops/debops-tools/issues/184) - I checked and with these changes script errors out the same on Python 2 and Python 3 so I guess it's "fixed"?

If you can check if this works on Python 3 that would be great, let me know if it does what it's supposed to.

drybjed:
I tested these changes on Debian Stretch with older `encfs` package from Jessie installed, on Python 2.7 and Python 3.5; everything seems to work as it's supposed to (filesystem is opened and closed correctly), so I'm going to merge it in.

PR #189, opened 2018-02-25
[debops.apt_proxy] Drop unneeded dependency `python3-future`

### LINKED ISSUE #181: [debops.apt_proxy] The 'python3-future' package requires 'jessie-backports' Debian repository
The `debops.apt_proxy` role contains a Python script which allows easy handling of unreachable APT caches, implented in https://github.com/debops/debops/commit/5bdb2c148c4. This script apparently relies on code provided by the `python3-future` APT package, which is available on Debian Jessie via the `jessie-backports` repository.

In certain situations, particularly initial host deployment, where the APT proxy support is enabled, playbook run can be aborted due to the `python3-future` package not being available when `jessie-backports` repository hasn't been configured yet. A solution would be to drop the `python3-future` from the script requirements.

ypid:
Closes #181

### DIFF
--- a/ansible/roles/debops.apt_proxy/defaults/main.yml
+++ b/ansible/roles/debops.apt_proxy/defaults/main.yml
@@ -15,7 +15,7 @@
 #
 # List of base packages to install.
 apt_proxy__base_packages:
-  - '{{ [ "python3-future", "python3-apt"  ]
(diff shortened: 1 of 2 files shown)


drybjed (approved):
Thanks for checking it!

PR #193, opened 2018-02-26
replace codecs with io: https://mail.python.org/pipermail/python-list…

umeboshi2:
…/2015-March/687124.html

Github chopped the commit msg.

https://mail.python.org/pipermail/python-list/2015-March/687124.html

The debops-defaults command wasn't working in py3 venv.


drybjed (approved):
Looks good, thanks!

PR #195, opened 2018-02-27
[debops.nginx] Fix order in nginx.conf.j2 template in order to allow …

pedroluislopez:
Example for use a custom log_format

```
nginx_http_options: |
  log_format timed_combined '$remote_addr - $remote_user [$time_local] '
                    '"$request" $status $body_bytes_sent '
                    '"$http_referer" "$http_user_agent" '
                    '$request_time $upstream_response_time $pipe';

nginx_access_log_format: 'timed_combined'
```
You need to define a log format before you can reference it, so `nginx_http_options` must to be written before `access_log` definitions.

### DIFF
--- a/ansible/roles/debops.nginx/templates/etc/nginx/nginx.conf.j2
+++ b/ansible/roles/debops.nginx/templates/etc/nginx/nginx.conf.j2
@@ -39,6 +39,10 @@ http {
         server_names_hash_bucket_size {{ nginx_http_server_names_hash_bucket_size | default('64') }};
         server_names_hash_max_size {{ nginx_http_server_names_hash_max_size | default('512') }};
 
+{% if nginx_http_options|d() and nginx_http_options %}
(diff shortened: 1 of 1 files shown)


ansible/roles/debops.nginx/templates/etc/nginx/nginx.conf.j2
```diff
@@ -51,9 +54,6 @@ http {
 
         add_header X-Clacks-Overhead "GNU Terry Pratchett";
 
-{% if nginx_http_options|d() and nginx_http_options %}
-{{ nginx_http_options | indent(8, true) | regex_replace("(?m)^\s*$", "") }}
-
 {% endif %}
```
drybjed: You should move the corresponding `endif` as well, otherwise the condition scope is incorrect.

drybjed (approved):
Looks good, thanks!

PR #199, opened 2018-03-01
[debops.owncloud] Fix shellcheck issues

drybjed:


ypid (approved):
LGTM. I tested that your changes behave as intended. About the script itself. I don’t know if anyone uses that currently.

drybjed:
@ypid Thanks for testing. It's apparently only used with ownCloud, not Nextcloud, but `shellcheck` still does the linting so it should be fixed.

PR #209, opened 2018-03-08
Improve dmz rule type

gaudenz:
The first commit fixes some issues with setting dport on a dmz rule and the second commit adds an optional snat_ip parameter.


drybjed (approved):
@gaudenz First of all - awesome commit messages, I want more of this! :-)

Do you know if there are any changes needed in existing configuration (inventory), or will they work as expected?

gaudenz:
AFAICS no changes to existing configurations should be needed. Everything else would be an unintended bug.

ganto:
Thanks a lot for this PR. I guess this would finally fix debops/ansible-ferm#55 :+1:

PR #232, opened 2018-03-17
[debops.cryptsetup] Support an macOS Ansible controller. Fixes #31.

### LINKED ISSUE #31: debops.cryptsetup uses `dd` option that is unsupported on macos
I'm using debops.cryptsetup from a macos controller and have found that the [command to generate the encryption key](https://github.com/debops/ansible-cryptsetup/blob/master/tasks/manage_devices.yml#L37) on the ansible controller makes use of the `dd` option `iflag=fullblock`.  This option isn't available to the default macos version of `dd`.

I can think of three ways to address this:

1. Remove the `iflag=fullblock` option.  From the man page this option will `accumulate full blocks of input`.  It seems to me this is a "nice to have" option rather than a requirement, but I'm not certain.
2. Use something other than `dd`.  I'm thinking `openssl` might be a suitable alternative.
3. Determine whether the controller is `darwin` or `linux` in a prior task, then conditionally include or exclude the option.

It seems to me option 1 may acceptable, though I'm happy to explore the other options too and raise a PR.

ypid:
It has been checked that `head --byte` behaves like `iflag=fullblock` and the role has been locally tested with this change.

Fixes: #31 

I don’t think this needs to be merged right away. I would rather have some times and ideally multiple reviewers. Mandatory is one DebOps Developer and one guy using macOS.

Note that support for macOS in DebOps is provided on best effort bases. No DebOps Developer uses macOS.

@jinnko Can you try that this works on macOS?

### DIFF
--- a/ansible/roles/debops.cryptsetup/tasks/manage_devices.yml
+++ b/ansible/roles/debops.cryptsetup/tasks/manage_devices.yml
@@ -34,7 +34,7 @@
 ## Generate keyfile and copy it to the remote system [[[
 
 - name: Generate binary keyfile on the Ansible controller
-  command: dd if={{ cryptsetup__keyfile_source_dev | quote }} of={{ (item.keyfile | d(cryptsetup__secret_path + "/" + item.name + "/keyfile.raw")) | quote }} bs={{ ((512/8) if (item.key_size|d(cryptsetup__key_size) == "default") else ((item.key_size|d(cryptsetup__key_size))/8)) | int }} count=1 iflag=fullblock
+  shell: head --byte {{ ((512/8) if (item.key_size|d(cryptsetup__key_size) == "default") else ((item.key_size|d(cryptsetup__key_size))/8)) | int }} {{ cryptsetup__keyfile_source_dev | quote }} > {{ (item.keyfile | d(cryptsetup__secret_path + "/" + item.name + "/keyfile.raw")) | quote }}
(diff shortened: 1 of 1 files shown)


drybjed (dismissed):
Looks good to me, but I'll wait with merging for other reviewers who use macOS to confirm the change works as expected.

stek29:
There's no `--byte` option on BSD/Apple `head`.
`head -c X` is the right command for it -- see [manpage](http://www.manpagez.com/man/1/head/)

Also, from what I see even GNU head doesn't have "--byte" option advised in man/help (however, it works, so I guess it's just left for backward compatibility)

Both GNU and BSD head support `-c` option, and it is in fact equivalent to `--byte` afaiu:

```
❯ ghead --help|grep -- --byte
  -c, --bytes=[-]NUM       print the first NUM bytes of each file;
```

And I assume that `--byte` is same as `-c`:

```
❯ ghead --byte
ghead: option '--bytes' requires an argument
Try 'ghead --help' for more information.
```

ypid:
@stek29 Thanks. `--byte` seems to work because it is unambiguous. This works with most (all?) GNU coreutils. Regardless, `--bytes` should have been used because `--byte` is not documented. I changed it to `-c`.

drybjed (approved):
OK guys, let me know if the proposed change works on macOS. :-)

PR #236, opened 2018-03-18
Correct the gmail example

jankatins:
Without the `=` between the option name and the value, I get this error:

```
nullmailer-send[25293]: /usr/lib/nullmailer/smtp: option --port requires a value.
nullmailer-send[25293]: Sending failed: Unspecified temporary error
```

Adding an `=` between option name and value does not have this problem.


drybjed (approved):
Sounds good, thanks. Usually the space separator works, I guess `nullmailer` doesn't support that syntax in the config file.

PR #245, opened 2018-03-20
[debops.postfix] Allow TLSv1 connections for now

drybjed:
There are many third-party MTAs with older software that don't speak
newer protocols. This is an issue outside of DebOps control, therefore
to make the life of system administrators easier, TLSv1 protocol for
SMTP connections will be allowed for now.


drybjed:
I'm not sure if this move is a good one. This needs further discussion, therefore I'm moving to v1.0.0 release, it could be merged earlier though.

PR #248, opened 2018-03-21
Add apache__*packages lists for group and host

### LINKED ISSUE #246: debops.apache2 Add apache__*packages lists
I think it might be useful (at least for me…) and allow this role to have the same behaviour/variables like almost other DebOps's roles.

In my case, i already use **apache__packages** in debops_service_php group_vars to automatically add _libapache2-mod-php_. And on another host i also need _libapache2-mod-perl2_.

**apache__group_packages** and **apache__host_packages** will be perfect.

gardouille:
Try to fix #246.

### DIFF
--- a/ansible/roles/debops.apache/defaults/main.yml
+++ b/ansible/roles/debops.apache/defaults/main.yml
@@ -28,6 +28,20 @@ apache__base_packages:
 apache__packages: []
 
                                                                    # ]]]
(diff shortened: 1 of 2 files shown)


gardouille:
With this config a package previously set can't be unset… But i guess it's ok like this ^^

drybjed (approved):
Looks good, thanks.

PR #249, opened 2018-03-21
Redesign of the 'debops.sysctl' Ansible role

drybjed:


ypid (approved):
LGTM. Tested against one Debops Stretch test system. Nice work as always :+1:

Issue #238, opened 2018-03-18
sshd fails in MAC config line


jankatins:
I installed sshd via debops on my local laptop. I didn't set any sshd related config option in my vars. `debops`now fails with this message:

```
RUNNING HANDLER [debops.sshd : Test sshd configuration and restart] **********************************************************************************************************************************
fatal: [littlebottom2]: FAILED! => {"changed": true, "cmd": ["sshd", "-t"], "delta": "0:00:00.004924", "end": "2018-03-19 00:28:30.666923", "msg": "non-zero return code", "rc": 255, "start": "2018-03-19 00:28:30.661999", "stderr": "/etc/ssh/sshd_config line 18: Bad SSH2 mac spec 'hmac-sha2-512-etm@openssh.com,hmac-sha2-256-etm@openssh.com,umac-128-etm@openssh.com,hmac-sha2-512,hmac-sha2-256,umac-128@openssh.com,hmac-ripemd160'.", "stderr_lines": ["/etc/ssh/sshd_config line 18: Bad SSH2 mac spec 'hmac-sha2-512-etm@openssh.com,hmac-sha2-256-etm@openssh.com,umac-128-etm@openssh.com,hmac-sha2-512,hmac-sha2-256,umac-128@openssh.com,hmac-ripemd160'."], "stdout": "", "stdout_lines": []}
```

I'm running debian unstable with this ssh packages:

```
λ dpkg -l |grep ssh
[...]  
ii  openssh-client                          1:7.6p1-4                           amd64        secure shell (SSH) client, for secure access to remote machines
ii  openssh-server                          1:7.6p1-4                           amd64        secure shell (SSH) server, for secure access from remote machines
```

As far as I understand the problem, my sshd version needs a new versioned list of MACs, similar to this: https://github.com/dev-sec/ansible-ssh-hardening/pull/139

jankatins:
Adding this in a vars file let it run throught, but now fails on a different line:

```
sshd__macs_map:
  '7.6': [ 'hmac-sha2-512-etm@openssh.com', 'hmac-sha2-256-etm@openssh.com',
           'umac-128-etm@openssh.com', 'hmac-sha2-512', 'hmac-sha2-256' ]
```

```
RUNNING HANDLER [debops.sshd : Test sshd configuration and restart] **********************************************************************************************************************************
[...]

PR #259, opened 2018-03-24
[debops.sshd] Fix sshd config file

jankatins:
There were two problems:
* the dict was not ordered as inserted
* the flag wasn't overwritten inside the loop, resulting in all mac getting added

This also reverts ecfe618d0ff68474ff3423fd56a2ebb9ceef92c8

Closes: #238

### DIFF
--- a/ansible/roles/debops.sshd/templates/etc/ssh/sshd_config.j2
+++ b/ansible/roles/debops.sshd/templates/etc/ssh/sshd_config.j2
@@ -67,12 +67,13 @@ KexAlgorithms {{ sshd__tpl_kex_algorithms | join(",") }}
 
 {% endif %}
 {% set sshd__tpl_macs = [] %}
-{% set sshd__tpl_macs_match = False %}
-{% for key, value in sshd__macs_map.items() %}
+{# workaround for jinja not beeing able to overwrite a variable inside the loop #}
+{% set sshd__tpl_macs_match = [] %}
+{% for key, value in sshd__macs_map|dictsort|reverse %}
 {%   if not sshd__tpl_macs_match and
         sshd__register_version.stdout|d() and
         sshd__register_version.stdout | version_compare(key, '>=') %}
-{%     set sshd__tpl_macs_match = True %}
+{%     set _ = sshd__tpl_macs_match.append(True) %}
 {%     for element in value %}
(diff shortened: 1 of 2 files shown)


ansible/roles/debops.sshd/templates/etc/ssh/sshd_config.j2
```diff
@@ -67,12 +67,13 @@ KexAlgorithms {{ sshd__tpl_kex_algorithms | join(",") }}
...
-{% set sshd__tpl_macs_match = False %}
-{% for key, value in sshd__macs_map.items() %}
+{# workaround for jinja not beeing able to overwrite a variable inside the loop #}
+{% set sshd__tpl_macs_match = [] %}
+{% for key, value in sshd__macs_map|dictsort(reverse=true) %}
 {%   if not sshd__tpl_macs_match and
         sshd__register_version.stdout|d() and
         sshd__register_version.stdout | version_compare(key, '>=') %}
-{%     set sshd__tpl_macs_match = True %}
+{%     if sshd__tpl_macs_match.append(True) %}{% endif %} {# more workaround... #}
```
drybjed: This can be written as:
```
{% set _ = sshd__tpl_macs_match.append(True) %}
```

jankatins: updated

drybjed (approved):
Excellent, thank you. :-)

drybjed (changes requested):
@jankatins Unfortunately, bad news. After testing this on Debian Jessie with Jinja 2.8, I got an error during templating, from Jinja:
```
do_dictsort() got an unexpected keyword argument 'reverse'
```
It looks like support for `docsort(reverse=true)` [wasn't added until Jinja 2.10](https://groups.google.com/forum/#!searchin/ansible-project/do_dictsort()$20got$20an$20unexpected$20keyword$20argument$20$27reverse$27%7Csort:date/ansible-project/-ZADTSEXEbk/f7wvGUQIAAAJ). That breaks the role on older installations.

Unfortunately, removing `reverse=true` does not end well, the wrong entry is picked. I'll look for a workaround.

jankatins:
Try this...

drybjed (approved):
Latest changes seem to work as expected. Thanks, merging.

PR #261, opened 2018-03-24
[debops.kmod] Fix undefined variable abortion when tasks are skipped; [debops.etckeeper] Drop orphan reference to old debops-contrib.etckeeper

ypid:
Fixing my own mistakes :)

### DIFF
--- a/ansible/roles/debops.kmod/tasks/main.yml
+++ b/ansible/roles/debops.kmod/tasks/main.yml
@@ -38,6 +38,7 @@
   include_tasks: 'modprobe.yml'
   loop_control:
     loop_var: 'module'
+  register: kmod__register_modprobe_tasks
   with_items: '{{ kmod__combined_modules | parse_kv_items }}'
   when: kmod__enabled|bool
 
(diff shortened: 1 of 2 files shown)


ansible/roles/debops.kmod/tasks/main.yml
```diff
@@ -60,5 +61,4 @@
 
 - name: Update Ansible facts if the list of load kernel modules might have changed
   action: setup
-  when: (kmod__register_module_config_delete is changed or
-         kmod__register_module_config_create is changed)
+  when: kmod__register_modprobe_tasks is changed
```
drybjed, line 64: Just to be sure I understand this correctly - the outer loop task will record the results of inner loops... Will it have a `changed` state for all of the results, even if some of them later on are not changed? Just want to make sure it works as expected, I don't think I ever registered an outer loop like this.

drybjed (approved):

ansible/roles/debops.kmod/tasks/main.yml
ypid: Damn, I was wrong again. I played with it a bit more now and it seems there is no easy way to run a task when one or more tasks in included tasks have the "changed" state. I am changing it back in #263.

drybjed: As I suspected. Thanks for checking. :-)

PR #263, opened 2018-03-25
[debops.kmod] Fix missed Ansible facts update

ypid:
Introduced in: 70d2dba1ca0a465d9bced3edfa168c6c831e3706
Updated but not solved in: 5c3e8fe22bf5f601c3d4fffa756699c05f06877a

Updates: #261

### DIFF
--- a/ansible/roles/debops.kmod/tasks/main.yml
+++ b/ansible/roles/debops.kmod/tasks/main.yml
@@ -38,7 +38,6 @@
   include_tasks: 'modprobe.yml'
   loop_control:
     loop_var: 'module'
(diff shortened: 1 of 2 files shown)


drybjed:
So, what's the change between the original version, you are just moving the `action: setup` task to the list of tasks in `modprobe.yml`?

ypid:
Refer to https://github.com/debops/debops/pull/261#discussion_r176944508. It did not work as I thought it would. The second "setup" was never executed. This PR should finally fix it.

drybjed (approved):

PR #264, opened 2018-03-25
Fix various deprecation warnings from Ansible v2.5

drybjed:
This patch fixes deprecation warnings reported by Ansible v2.5:

- warning about usage of 'include' instead of 'import_playbook' to
  include additional playbooks;

- warning about usage of various tests as filters in Jinja expressions;


ypid (approved):
LGTM. Tested against a workstation (no server roles).

PR #277, opened 2018-04-09
 [debops.owncloud] Support Nextcloud 13 and partially ownCloud 10

ypid:
Status: WIP. Commits might be rebased later. I just wanted to open the PR here to avoid duplicate work in case someone else noticed that the ownCloud and NextCloud releases deployed by the role are EOL. Anyone :) ?

### DIFF
--- a/ansible/roles/debops.owncloud/tasks/ldap.yml
+++ b/ansible/roles/debops.owncloud/tasks/ldap.yml
@@ -3,8 +3,6 @@
 
 - name: Check default ldap config exists
   command: php --file "{{ owncloud__app_home }}/occ" ldap:show-config ""
-  args:
-    executable: '/bin/sh'
   changed_when: False
   when: (not ansible_check_mode)
   register: owncloud__register_ldap_default_config_exits
@@ -13,16 +11,12 @@
 
 - name: Create empty LDAP config
   command: php --file "{{ owncloud__app_home }}/occ" ldap:create-empty-config
-  args:
-    executable: '/bin/sh'
   become: True
   become_user: '{{ owncloud__app_user }}'
   when: ((not ansible_check_mode) and 'Invalid configID' in owncloud__register_ldap_default_config_exits.stdout|d(""))
 
 - name: Set LDAP password
   command: php --file "{{ owncloud__app_home }}/occ" ldap:set-config "" ldapAgentPassword "{{ owncloud__ldap_password }}"
-  args:
-    executable: '/bin/sh'
   changed_when: False
   become: True
   become_user: '{{ owncloud__app_user }}'
@@ -31,8 +25,6 @@
 
 - name: Set LDAP parameters
   command: php --file "{{ owncloud__app_home }}/occ" ldap:set-config "" "{{ item.key }}" "{{ item.value }}"
-  args:
-    executable: '/bin/sh'
   changed_when: False
   become: True
   become_user: '{{ owncloud__app_user }}'
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -63,6 +63,11 @@ Changed
   reflects its intended purpose. Variable names and their default values
   changed; see the :ref:`upgrade_notes` for more details.
 
+- [debops.owncloud] Support Nextcloud 13 and partially ownCloud 10. Nextcloud
+  11 and ownCloud 9.1 are EOL, you should update. The role can help you with
+  the update to ensure that everything works smoothly with the new versions.
+  Currently, the role can not do the update for you.
+
 - [debops.sshd] The role will now check the :ref:`debops.system_groups` Ansible
   local facts to define what UNIX groups are allowed to connect to the host via
   the SSH service.
--- a/ansible/roles/debops.owncloud/COPYRIGHT
+++ b/ansible/roles/debops.owncloud/COPYRIGHT
@@ -2,8 +2,8 @@ debops.owncloud - Install and manage ownCloud instances
 
 Copyright (C) 2015-2016 Maciej Delmanowski <drybjed@gmail.com>
 Copyright (C) 2015 Hartmut Goebel <h.goebel@crazy-compilers.com>
-Copyright (C) 2015-2017 Robin Schneider <ypid@riseup.net>
-Copyright (C) 2015-2017 DebOps https://debops.org/
+Copyright (C) 2015-2018 Robin Schneider <ypid@riseup.net>
+Copyright (C) 2015-2018 DebOps https://debops.org/
 
 This Ansible role is part of DebOps.
 
--- a/ansible/roles/debops.owncloud/defaults/main.yml
+++ b/ansible/roles/debops.owncloud/defaults/main.yml
@@ -1687,7 +1687,7 @@ owncloud__nginx_access_log_assets: True
 # uses to a path with more space available.
 # This directory is used to cache uploaded files when using Apache.
 # See also :envvar:`owncloud__temp_path`.
-
+#
 # Empty string will not change the temp directory of PHP.
 owncloud__php_temp_path: ''
 
@@ -1845,7 +1845,7 @@ owncloud__apache__dependent_vhosts:
       SetEnv HTTP_HOME {{ owncloud__app_home }}
 
       {# Does not work.
-      ## Tested with while uploading with:
+      ## Tested while uploading with:
       ## while true; do df -h /tmp|tail -n 1; sleep 0.1; done
       ## Currently configured in PHP Apache scope: owncloud__php__dependent_configuration
       {% if owncloud__php_temp_path != "" %}
--- a/ansible/roles/debops.owncloud/defaults/main.yml
+++ b/ansible/roles/debops.owncloud/defaults/main.yml
@@ -372,7 +372,10 @@ owncloud__post_upgrade_hook_host_list: []
 # * ``owncloud`` (legacy variant, not recommended for new deployments, not well tested)
 # * ``nextcloud`` (Main supported variant in the future. Currently supported on Debian stretch, refer to `issue 45 <https://github.com/debops/ansible-owncloud/issues/45>`_ for details)
 #
-owncloud__variant: 'nextcloud'
+owncloud__variant: '{{ ansible_local.owncloud.variant
+                       if (ansible_local|d() and ansible_local.owncloud|d() and
+                           ansible_local.owncloud.variant|d())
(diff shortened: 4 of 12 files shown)


ansible/roles/debops.owncloud/tasks/ldap.yml
```diff
@@ -3,8 +3,6 @@
 
 - name: Check default ldap config exists
   command: php --file "{{ owncloud__app_home }}/occ" ldap:show-config ""
-  args:
-    executable: '/bin/sh'
```
drybjed, line 7: I'm not sure about the removal of the `executable` keywords... Isn't ownCloud user acount using `/usr/sbin/nologin` as a default shell? That would fail with recent Ansible which uses the user account shell by default, that's why these were even added.

drybjed (commented):
Looks good so far.

One thing I noticed recently is that people report that they cannot get ACME certificates when installing ownCloud/Nextcloud with this role. I suspect that somewhere the `/srv/www/` path is created incorrectly with `0750` permissions, where it should be at least `0751`, so that the webserver user account can traverse down to `/srv/www/sites/acme/` directory. I'll try to find this issue and fix it this release cycle, or you could check it out if you have the time.

ansible/roles/debops.owncloud/tasks/ldap.yml
ypid: Its the `command` module ;-) There is no shell. This was reported by Ansible 2.4 as warning. I guess nobody noticed this before that it does not make any sense.

drybjed: I see. Sounds good then. I need to update the Ansible used in GitLab CI tests and check all of them for any new deprecation warnings...

ypid:
I am done for now. I have not looked into the new CI tests, will be done later.

drybjed:
I tested your changes using a fresh install of the default variant on a Debian Jessie (ownCloud) and Debian Stretch (Nextcloud) hosts. Here are my findings.

The `debops.owncloud` after an upgrade in both variants is idempotent, that's good. Nextcloud was not updated from 11.x to the newer version automatically in the existing instance. I used the web updater to perform the upgrade to the 12.0.6 version (is that the current stable? Channel "production"). After downloading the update Nextcloud performed an upgrade and reported that it failed, but still after a reload it seems to work. Perhaps some of the applications returned and error, I haven't investigated further yet. After an upgrade, Ansible is idempotent.

The existing ownCloud instance was replaced with Nextcloud 12.0.6.The user files seem to be intact, but Nextcloud started with a new database and it seems that everything needs to be set up from scratch.

I'm not sure if this is a good move - it is certainly unexpected. Perhaps the `owncloud__variant` variable  could check the existing Ansible facts and use the variant that is present on the host, and if there are no facts yet, default to Nextcloud. This would certainly not catch existing ownCloud installations by surprise and let the administrators convert them to Nextcloud at their own pace.

If you want, I can update the variable before merging the PR. Let me know.

ypid:
Thanks very much for testing.

> That is expected. I used the web updater to perform the upgrade to the 12.0.6 version (is that the current stable? Channel "production")

12.0.6 is the latest patch release of Nextcloud 12.0. I read in the Nextcloud 13 release post that they are doing staged updates so I guess the missing update to 13.0 can be explained with that.
https://nextcloud.com/changelog/#latest12

I had the same issue updating for all my instances (all where on freshly setup with Nextcloud 11 once). Now that you are also having issues with upgrading, I guess this is a Nextcloud 11 issue. What I did was a manual upgrade following: https://docs.nextcloud.com/server/12/admin_manual/maintenance/manual_upgrade.html

The role can still do most of the manual upgrade for you. Basically I did:

* Run the playbook `service/owncloud.yml --tags role::owncloud:tarball` to download and verify the latest version
* Full Backup
* Stop php-fpm, nginx
* `mv /srv/www/nextcloud /srv/www/nextcloud-old`
* Run the playbook `service/owncloud.yml --tags role::owncloud:tarball` again
* Move `config` and `data` from `/srv/www/nextcloud-old` to `/srv/www/nextcloud`
* Start php-fpm, nginx
* Run `occ upgrade --no-app-disable` and if needed `occ maintenance:mode --off`
* Quick test of the website
* Run the full `service/owncloud.yml` playbook to optimize php and so on so that all warnings on the admin page in Nextcloud change to a nice "All checks passed."
* Enjoy.

I realize that this upgrade help should go into the docs but I hope to rather automate it in the role or a script in the near future directly.

About `owncloud__variant` you are absolutely right, my bad. I did as you suggested and should have done that in the first place.

drybjed:
Does that mean that a clean install will result in Nextcloud 13.0? Interesting, I'll test it out tommorow then. Thanks for updating the role.

ypid:
> Does that mean that a clean install will result in Nextcloud 13.0?

No, the role still defaults to Nextcloud 12, but I tested 13 with the role. Not sure how Nextcloud staged updates work. I thought everyone should get 13 by now. Thanks.

Refer to: "Get started now" in https://nextcloud.com/blog/nextcloud-13-brings-secure-file-sync-and-collaboration-to-the-next-level/

drybjed:
> `mv /srv/www/nextcloud /srv/www/nextcloud-old`

This one line really shows the advantages of a real package manager over distributing software via tarballs... I realize that this step is done so that there are no files from the previous release that were removed in the current one.

This is a problem because we are upgrading a live server. Is there a way in Nextcloud to even see if any users are connected, or should I just pick say, 3AM to do the upgrade then? Also, with Nextcloud it seems that when the web updater pulls an upgrade, the information about an upgrade shows up for all users - or was that because I was logged in as an admin? I will need to check that - I don't want random users seeing an information about an upgrade and clicking that to perform it, come on... Should I firewall the entire installation behing HTTP Basic Auth before an upgrade?

What do you think about moving Nextcloud to its own role, built from scratch? Seems that ownCloud is on its way out, I wonder if a fresh start would benefit us.

ypid:
> Should I firewall the entire installation

That is what I did in the past. Now I simply update. It is only for a few minutes when users will see the upgrade. But you are right, users could potentially hit the "upgrade" button which I also don’t like.

> What do you think about moving Nextcloud to its own role, built from scratch? Seems that ownCloud is on its way out, I wonder if a fresh start would benefit us.

Sounds good. But this is not a priority for me anytime soon. This would be polishing, so far the role works as it should be in the current state but it does not look very nice, lots of fixme notes.

It would only benefit us if we drop ownCloud completely which I would approve. But I am not in a hurry for that.

drybjed:
In that case, perhaps the maintenance mode in `nginx` needs an upgrade, where you can specify the IP addresses/subnets that are allowed to access the website normally, and the rest gets redirected elsewhere or shown a maintenance page. I wonder if that can be accomplished somewhat dynamically by creating a specific file, so that `nginx` doesn't have to be reconfigured...

OK then, I'll test the changes again to confirm and merge the PR. Thanks for your work on this. :-)

drybjed:
Well, I tried it again. The playbook run on ownCloud stopped due to some `occ` commands being required, but after running the upgrade via the web interface the application runs fine. NB: ownCloud suggested to use Redis-based file locking, is that also possible with Nextcloud? Maybe we should add Redis support for better performance.

Nextcloud... The playbook on the Nextcloud host ran fine, no errors. But now the application stays on the 11.0.8 version and does not detect any upgrades. W. T. F. Did Nextcloud team change something on their side? I guess that I'll try reinstalling Nextcloud with the PR on a clean host, see what version will be installed by default...

Edit: After a while Nextcloud detected a newer version and let me perform an upgrade which again apparently failed for some reason. Now Nextcloud that 12.0.6 is the current version, last checked... Jan 1st 1970. Some epoch counter must have been reset along the way. I'll try a clean install next.

drybjed:
@ypid I checked your fix against the `/srv/www/` issue not being readable, and on a clean install, the result looks like this:
```
# namei -om /srv/www/nextcloud/
f: /srv/www/nextcloud/
 drwxr-xr-x root     root     /
 drwxr-xr-x root     root     srv
 drwxrwx--- www-data www-data www
 drwxr-x--- www-data www-data nextcloud
```
That's not good, especially that `www-data` has `rw` access to `/srv/www/` now. I'll add a task that explicitly ensures that `/srv/www/` exists with `root:root` ownership and `0755` permissions so that the parent directory is correctly created.

Edit: Ah, no, sorry - my bad. I checked this against the host installed by the old version of the role, let me check that against the newer one. Stay tuned.

Edit2: OK, with the PR the result is:
```
# namei -om /srv/www/nextcloud 
f: /srv/www/nextcloud
 drwxr-xr-x root     root     /
 drwxr-xr-x root     root     srv
 drwxr-xr-x root     root     www
 drwxr-x--- www-data www-data nextcloud
```
So it's fixed.

drybjed:
Oh, an Redis is supported :laughing: I think that I need a break...

drybjed:
@ypid I just noticed that you set `owncloud__release` variable to `12` in this PR. Is that intentional, or was that supposed to be `13`? I suppose that's why Nextcloud sticks to `12.0.6` version. I'll check if setting `13` will install the latest release without issues.

Clean install with `owncloud__release: '13.0'` went fine. I'll test an upgrade from an old installation next.

Edit: An upgrade from Nextcloud 11.0 went fine, finished with 12.0.6. So I think that switching the `owncloud__release` variable to `13.0` would be OK - existing installations would end up with `12.0.6`, and new installations would start with `13.1`. @ypid, what do you think?

ypid:
Sounds like a busy day. My day was busy as well. Even more now, thanks very much for testing this! 

Maintenance page with Nginx sounds very nice. Should be done.

About the Nextcloud version, Nextcloud 13 is supported as well as I wrote in the docs. I tested it (13) this time thoughtfully, and also optimized using the NC 13 docs. I just like to run solid and major software versions which is why I still run 12 and so this is also the version which this role has been most tested with. I see no issue with 13 by default but I chose and would go with the more conservative version. It is up to you. See also: [Any thoughts with starting with 12 instead of 11?](https://github.com/debops/ansible-owncloud/pull/83#issuecomment-303960385)

drybjed:
Very well, sounds good to me - we can go with Nextcloud 12 for now. The PR is ready for merge then?

ypid:
Ready I would say. My updated Nextcloud 12 instances are in production for a while now and the PR has been tested and discussed by us quite a bit. It is not up to latest DebOps standards, but that needs to be addressed laster.

drybjed (approved):

PR #289, opened 2018-04-12
Add secret/salt encode/decode for Python3 compatibility

anzil:
The original version seems not to work with Python3 (throws error: "Unicode-objects must be encoded before hashing"). The proposed changes cure the problem and work with Python2 and Python3.

The salt gets now base64-encoded before being processed further. This might require some checking regarding potential security issues... I cannot judge on the implications.


drybjed (approved):
Looks good to me, thanks. I'll wait a bit to see if @scibi has any comments before merging.

PR #290, opened 2018-04-12
[debops.sshd] Fix sshd macs ,ciphers and kex_algorithms generation

prahal:
The for loop contains a test to avoid more than one iteration.
This has no effect as the "iterations" are evaluated at once.

This issue is documented in http://jinja.pocoo.org/docs/2.10/templates/#for
"Please note that assignments in loops will be cleared at the end of the
iteration and cannot outlive the loop scope. Older versions of Jinja2
had a bug where in some circumstances it appeared that assignments would
work. This is not supported. See Assignments for more information about
how to deal with this.

More details and alternatives constructs at:
http://jinja.pocoo.org/docs/2.10/templates/#assignments

---
Two items are RFCs:

- if sshd version is too low we ends up with a runtime error: map[max_version] will not match.
I am undecided: fixing the errors means set blank values to those configuration items which is no better.

- I defaults sshd tpl version to 0 ... not that nice either.

### DIFF
--- a/ansible/roles/debops.sshd/templates/etc/ssh/sshd_config.j2
+++ b/ansible/roles/debops.sshd/templates/etc/ssh/sshd_config.j2
@@ -26,18 +26,8 @@ HostKey /etc/ssh/ssh_host_{{ hostkey }}_key
 {%   endif %}
 {% endfor %}
 
-{% set sshd__tpl_ciphers = [] %}
-{% set sshd__tpl_ciphers_match = False %}
-{% for key, value in sshd__ciphers_map.items() %}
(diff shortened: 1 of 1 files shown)


drybjed:
The package version should be available at runtime via `sshd__register_version.stdout`, why not default to that in the template?

Looks like the Travis-CI test is broken, I'm going to fix that ASAP.

prahal:
Indeed. Thanks for the review and travis fix !

drybjed (approved):
It works as expected on Wheezy, Jessie and Stretch hosts. Thanks!

PR #292, opened 2018-04-13
[debops.debops] Allow defining git repositories like user@host:repo

pedroluislopez:
If you define a repository like user@host:repo, you get this error in when conditions:

`TASK [debops.debops : Clone project repository] ***************************************************************************************************************************************************************************************
fatal: [debops]: FAILED! => {"msg": "The conditional check 'debops__project_git_repo' failed. The error was: template error while templating string: unexpected char u'@' at ...`

With this fix it works.


drybjed (approved):
Good idea, especially that the project repositories are usually accessed via private repositories. Thanks!

PR #293, opened 2018-04-13
debops-padlock script OS X compatibility

pedroluislopez:
Implemented compatibility of debops-padlock script for OS X systems


drybjed (approved):
Nice work, thanks! :+1:

PR #294, opened 2018-04-13
[debops.apt]: reverse the boolean for autoremove configurations

prahal:
APT::Autoremove::SuggestsImportant and APT::Autoremove::RecommendsImportant
value tells - when true - that we prevent removal of suggested or recommended
packages.

---
Speedup apt as it stop telling to autoremove recommanded packages while apt__autoremove_recommends: False.


drybjed:
This is very confusing, but the problem is not the value, but the meaning of the variables themselves. Thanks for bringing this up. I'll update the variables so that they have the same meaning that's described [in the documentation](https://aptitude.alioth.debian.org/doc/en/ch02s05s05.html).

PR #297, opened 2018-04-14
[debops.php] Set track_errors only for PHP<7.2

amette:
track_errors has been deprecated in PHP 7.2.
    See: https://secure.php.net/manual/en/migration72.deprecated.php


drybjed (approved):
Nicely done, thanks!

PR #298, opened 2018-04-15
Refresh configuration of UNIX system groups used by DebOps

### LINKED ISSUE #252: [debops.debops_legacy] New role which helps with DebOps upgrades/removes legacy files
Often, redesigns of existing roles result in new filenames on managed hosts. Old files currently don’t get removed by DebOps and just lay on managed hosts. This can become an issue because the old files are still in effect. Currently, the admin is expected to review all changes to DebOps and remove those files manually. I propose we create a role which deletes common legacy files to centralize this effort.

The role does not need to do much. I would keep the scope minimal for now. Just delete legacy files. I maintain a list for this already using the `debops.resources` role. Example:

```YAML
resources__files:

  - path: '/etc/sysctl.d/80-local.conf'
    state: 'absent'

  - path: '/etc/apt/apt.conf.d/000apt-cacher-ng-proxy'
    state: 'absent'

  - path: '/etc/apt/apt.conf.d/00apt-proxy'
    state: 'absent'

  - path: '/etc/modprobe.d/blacklist-ansible-kernel_module-role.conf'
    state: 'absent'
  - path: '/etc/modprobe.d/options-ansible-kernel_module-role.conf'
    state: 'absent'
  - path: '/etc/modules-load.d/ansible-kernel_module-role.conf'
    state: 'absent'

  - path: '/etc/sysctl.d/10-debops_sysctl.conf'
    state: 'absent'
```

So we could just provide a role similar to `debops.resources` which only cares about deleting files and ship it with a preset of `absent` definitions. This does not need to be defined by default.

Edit: Alternative name: `debops.debops_legacy`. I guess we should prefer this.

Related to: https://github.com/debops/debops/pull/229, https://github.com/debops/debops/pull/249

drybjed:
This set of patches cleans up and reorganizes the custom UNIX system groups used by various DebOps roles for access control of different services. The patch includes enchancements to the 'debops.sudo' role, 'bootstrap.yml' playbook, clean up of the 'debops.auth' role and introduces the 'debops.system_groups' and 'debops.debops_legacy' Ansible roles.

Fixes #252

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -1223,6 +1223,14 @@ stages:
     JANE_DIFF_PATTERN: '.*/debops.sudo/.*'
     JANE_LOG_PATTERN: '\[debops\.sudo\]'
 
+'system_groups role':
+  <<: *test_role_1st_deps
+  variables:
+    JANE_TEST_PLAY: '${DEBOPS_PLAYBOOKS}/service/system_groups.yml'
+    JANE_INVENTORY_GROUPS: 'debops_service_system_groups'
+    JANE_DIFF_PATTERN: '.*/debops.system_groups/.*'
+    JANE_LOG_PATTERN: '\[debops\.system_groups\]'
+
 'swapfile role':
   <<: *test_role_no_deps
   variables:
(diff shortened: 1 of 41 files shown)


ypid (commented):
LGTM. Also thanks for working on the debops.debops_legacy role, looks good :)

drybjed:
@ypid Thanks. Looking forward for your PRs against it. :-)

ypid:
Thats what I figured :) But lets have a clean start. Otherwise I would need to go though git history to figure out a good check if the file can safely be removed as you did for the old sudoers/admin file.

PR #302, opened 2018-04-16
Update Python support, add 'debops.python' Ansible role

drybjed:


ansible/roles/debops.python/templates/etc/ansible/facts.d/python.fact.py2.j2
```diff
@@ -0,0 +1,27 @@
+#!/usr/bin/python
```
ypid, line 1: Why have two files where the only difference is the shebang. Couldn‘t you just template that?

ypid (commented):
LGTM, one question attached.

ansible/roles/debops.python/templates/etc/ansible/facts.d/python.fact.py2.j2
drybjed: Due to [PEP-394](https://www.python.org/dev/peps/pep-0394/) which Debian Python Policy adheres to, the `#!/usr/bin/python` interpreter in Debian will always be Python 2.7. So, if you want to switch to Python 3-only environment, you need to use `#!/usr/bin/python3` explicitly. But we want to have DebOps backwards compatible with older OS releases at least for a little while longer, so doing this explicitly isn't an option everywhere.

I thought that having the shebang templated will break `pycodestyle` and `file` commands, but surprisingly, they still work as before. I'll prepare a new PR which contains only templated shebangs in the fact scripts, stay tuned. There still will be some scripts used by roles internally that cannot be templated like this and will need to be duplicated, but that is much less than doing this with all fact scripts.

PR #303, opened 2018-04-16
[debops.sudo]: fix bashism in profile

prahal:
With user shell as dash I get:
$ sudo -i -u git
-sh: 16: [: Illegal number:


drybjed (approved):
Ah, so that's what that was... Thanks for finding it out and fixing. :-)

PR #304, opened 2018-04-16
[debops.librenms] Fix resources loading for sublevel pages

prahal:
Set base_url to upstream default '/'.

Resources are relative to base_url. With an empty one
they end up relative to the current subfolder we are
in which fails for pages below home, ie /devices/ or
such.


drybjed (approved):
Good catch, thanks!

PR #305, opened 2018-04-16
[debops.postgresql]: Fix pgpass escaping - password field

prahal:
pgpass fields colon and backslash should be escaped by mean of a backslash
https://www.postgresql.org/docs/current/static/libpq-pgpass.html

before 9.2 the fields where not unescaped properly though:
https://www.postgresql.org/message-id/20111217082754.GB30887%40rice.edu

Note: I only apply it to password as other fields are unlikely to
have colon or backspace in them.


drybjed:
Sounds good. Can you add similar escaping for the `root` password in `debops.postgresql_server`?

prahal:
pgpass file only requires the escaping. Its syntax delimits with colons.
`hostname:port:database:username:password`
The other fields might not allow for colon or slash so no need to escape them.

drybjed (approved):
Thanks!

PR #309, opened 2018-04-17
Allow Python 3-only environment in the 'common' playbook

drybjed:


ypid (changes requested):
I only checked parts of the PR yet. I am hesitant about adding duplicate files, ref: https://github.com/debops/debops/pull/302. Is this really needed?

drybjed:
It turns out that both `file` and `pycodestyle` can handle templated Python shebangs correctly. I'm closing this PR in favor of a new one which will have much less duplicated content. Some scripts that are used internally by the roles will still need to be duplicated, but that much less than duplicating everything.

PR #311, opened 2018-04-18
Fix gitlab

prahal:


ansible/roles/debops.gitlab/templates/etc/default/gitlab.j2
```diff
@@ -43,3 +43,7 @@ gitlab_pages_dir="{{ gitlab_pages_checkout }}"
 gitlab_pages_pid_path="$pid_path/gitlab-pages.pid"
 gitlab_pages_log="$app_root/log/gitlab-pages.log"
 gitlab_pages_options="-pages-domain {{ gitlab_pages_domain }} -pages-root {{ gitlab_pages_path }} -listen-proxy 127.0.0.1:{{ gitlab_pages_port }}"
+
+gitaly_dir=$(cd $app_root/../gitaly && pwd)
```
drybjed, line 47: Hmm, that's unusual... But it looks like [the upstream does it this way as well](https://gitlab.com/gitlab-org/gitlab-ce/blob/master/lib/support/init.d/gitlab.default.example#L90) so I guess we can use it. And this file is not touched by `systemd` at all.

I think that you should add
```
gitaly_enabled=true
```
for completeness and parity with upstream.

bfabio: @drybjed Uhm, I never noticed we support SystemV init scripts. Do we even want that?

drybjed: I think that @prahal uses the SysVinit version, otherwise he wouldn't notice that the required variables are missing.

I know that some users use DebOps with Devuan, but I'm not sure if they use GitLab specifically.

I don't see any reasons to drop SysVinit support at this time. In the future, perhaps. Since, as you said, you hadn't even noticed it, seems to do good enough job so far. :-)

prahal:
I believe I let gitaly_enabled off as it is already handled by a systemd service ... but added the required bit to let gitlab init script detect gitaly was running. I somehow managed to forget.

Indeed I have gitlab.service a systemd automatic LSB init wrapper.

```
systemctl  show gitlab
ExecStart={ path=/etc/init.d/gitlab ; argv[]=/etc/init.d/gitlab start ; ignore_errors=no ; start_time=[n/a] ; stop_time=[n/a] ; pid=0 ; code=(null) ; status=0/0 }

```

Is the proper set of SystemD services: gitlab-gitaly + gitlab-workhorse + gitlab-mailroom ?
If so we might be better off masking the SysVinit script ins systemd.
My fix might restore sanity to the SysVinit systemd wrapper and it will mess with expected systemd services set behavior.

drybjed:
@prahal Your `debops.gitlab` installation must be very old. Currently there's no `gitlab.service` but only `gitlab.slice` which does not have any `ExecStart=` configuration.

Your fix will not affect the current `systemd` configuration managed by the role, but it will probably affect your particular installation of it. The `systemd` units currently present and managed by `debops.gitlab` don't read the `/etc/default/gitlab` file so any configuration set there will not affect them.

prahal:
No gitlab.service file but `systemctl show gitlab.service` is defined. That is systemd generates runtime definition when it finds a SysVinit init files.
 gitlab.slice is up and running.
Mind the "restart gitaly" handler restart a (systemd) service that does not exists. I seems to remind it aborts the debops run when it triggers and fail so.

drybjed:
@prahal, do you have other `gitlab-*.service` systemd units? In that case you can remove `/etc/init.d/gitlab`, is not needed anymore. It's even installed [only when running without systemd](https://github.com/debops/debops/blob/master/ansible/roles/debops.gitlab/tasks/configure_gitlab_ce.yml#L168-L172).

Running `systemctl status gitlab.slice` should let you see a bunch of GitLab services under that `.slice`, `gitlab-gitaly.service` should be one of them.

The `gitaly` handler needs to be renamed, agreed.

prahal:
Thanks:
```
cat /etc/ansible/facts.d/gitlab.fact 
{
"database": "postgresql",
"installed": true,
"systemd_services": false,
"version": "10.3"
}
```
All in all once we ran debops.gitlab while it was SysVinit , gitlab_use_systemd stays at no until one remove /etc/ansible/facts.d/gitlab.fact and run anew debops.gitlab.

 Closing this PR. I will open a new one about the gitaly restart handler.

PR #312, opened 2018-04-19
Improve Python 3 support in DebOps roles

drybjed:


ansible/roles/debops.apt_mark/files/script/apt-mark-status3
```diff
@@ -0,0 +1,26 @@
...
+from json import dumps
+from subprocess import Popen, PIPE
+import os
+import signal
+
+
+def run_command(command_string):
+    """Gather information about APT packages"""
+    process = Popen([command_string], stdout=PIPE, stderr=PIPE,
+                    shell=True, preexec_fn=os.setsid)
```
ypid: Please use this opportunity to remove the `shell=True` and pass a list of arguments.

ypid (changes requested):
LGTM, one comment attached.

ansible/roles/debops.apt_mark/files/script/apt-mark-status3
drybjed: You mean [this](https://medium.com/python-pandemonium/a-trap-of-shell-true-in-the-subprocess-module-6db7fc66cdfd)? I'll check it out, thanks.

ypid: I was more referring to the fix I did in `etckeeper.fact.j2` but that article was interesting as well.

drybjed: @ypid Check if the update works when you have some time.

ypid (commented):
LGTM, thanks :)

ansible/roles/debops.ansible/templates/etc/ansible/facts.d/ansible.fact.j2
```diff
@@ -1,4 +1,4 @@
-#!/usr/bin/env python
```
ypid, line 1: I only thought about this now after seeing the diff in etckeeper. Why drop the use of `/usr/bin/env`. This is still useful I would say for the sysadmin to install different versions of Python in the `$PATH` and have it used. In practice you would probably not do that but still, I use `/usr/bin/env` in shebangs all the time.

drybjed: The Debian Python Policy says this about [Interpreter Location](https://www.debian.org/doc/packaging-manuals/python-policy/python.html#interpreter_loc):

> Python scripts should specify the Debian Python interpreter, to ensure that the Debian Python installation is used and all dependencies on additional Python modules are met.

>Maintainers should not override the Debian Python interpreter using /usr/bin/env name. This is not advisable as it bypasses Debian's dependency checking and makes the package vulnerable to incomplete local installations of Python.

However that could be interpreted as valid only for official `.deb` packages with Python software that are included in Debian. If we aim to have DebOps officially included in Debian, I think that sticking to the system Python in case of facts, and perhaps other scripts that are executed on remote host could be benefical... Although that might be valid only for the '`debops*` set of scripts.

Local Ansible facts are installed in `/etc/ansible/facts.d/` directory and are executed most of the time under `root` account via the `setup` module. This is usually done by the system administrators during DebOps management runs, and can be done by unprivileged users if they use Ansible to manage their accounts without `root` access but still want to use some of the DebOps facts and don't override the `setup` default path.

What would change if we use `#!/usr/bin/env python` to run the local facts? Let's see...

Unprivileged users can have their own Python virtualenv and the Ansible fact scripts could be executed with Python from that virtual environment by specifying custom `ansible_python_interpreter`. Since none of the Ansible local facts provided by DebOps use setuid, I think that using them should be fairly safe - users won't have access to files and other data protected by UNIX permissions. Of course all needed Python modules would need to be installed in the `virtualenv`, and since this is most of the time via PyPI, untrusted code could complicate things by attacking the user account. Using system Python to run local facts that operate on the system itself seems to be a safer option.

Privileged execution via `root` using custom virtual environment could also be possible. Again, the virtual environment would have Python packages installed from PyPI which would be executed with full `root` privileges this time. On one hand, this is an environment prepared by the DebOps users themselves, on the other - can we trust PyPI modules enough to run the facts with it? Perhaps sticking to the Python packages included in the OS would be a safer bet.

I think that official DebOps Python policy, and an additional local fact policy could be defined as a DEP, then this could be easier to discover, and we can get better community input.

ypid:
To me this was more a theoretical need. Using `/usr/bin/python` directly is totally fine with me, also, now that I am aware of the Debian Python Policy. I don’t like installing Python modules using `pip` either when there are Debian packages for it. So lets just go with that `/usr/bin/python`.

ansible/roles/debops.apt_mark/files/script/apt-mark-status
```diff
@@ -13,7 +13,7 @@ def run_command(command_string):
                     shell=True, preexec_fn=os.setsid)
     stdout, stderr = process.communicate()
     if stdout:
-        return stdout.strip().split('\n')
+        return stdout.decode('utf-8').strip().split('\n')
```
bfabio: This will fail when the user's default locale is not UTF-8. I don't think there is a good reason for the user to change it to something else, but still.

drybjed: @bfabio This seems to me as a go-to way to handle strings in code compatible with both Python 2.7 and Python 3.x, do you know a better way? Python 3 is supposed to use Unicode (not sure if specifically UTF-8) by default.

bfabio: @drybjed We can use `Popen([command_string], ..., env=dict(os.environ, LANG="C")`) and `stdout.decode('ascii')`.

PR #313, opened 2018-04-19
[docs] Redesign DebOps Policies and Guidelines

drybjed:
This patch proposes a new design for DebOps Policies and Guidelines,
inspired by Python Enchancement Proposals. Some of the existing DebOps
policies are rewritten as DEPs.


drybjed:
Now I'm thinking that the `DEP-0001` should be something more akin to [PEP-1](https://www.python.org/dev/peps/pep-0001/) but I'm not sure if copying the whole thing and replacing/modifying parts relevant to DebOps is a good course of action... Opinions? Propably defining a framework for DEPs in the first one would be better than a list of keywords there.

docs/dep/dep-0001.rst
```diff
@@ -0,0 +1,599 @@
...
+- You, the DEP author, fork the `DebOps monorepo`_, and create a file named
+  :file:`docs/dep/dep-9999.rst` that contains your new DEP. Use "9999" as your
+  draft DEP number.
+- Push this to your GitHub fork and submit a pull request.
+- The DEP editors review your PR for structure, formatting, and other errors.
+- Once approved, they will assign your DEP a number, and label it as Standards
+  Track, Informational, or Process, and give it the status "Draft"
+
+Once the review process is complete, and the DEP editors approve it (note that
+this is *not* the same as accepting your DEP!), they will squash commit your
```
ypid: Why "squash commit", we don’t otherwise do this to my knowledge? I think we discussed this before, "squash committing" breaks signatures.

docs/dep/dep-0001.rst
```diff
@@ -0,0 +1,599 @@
...
+DEP will occur.
+
+For a DEP to be accepted it must meet certain minimum criteria. It must be
+a clear and complete description of the proposed enhancement. The enhancement
+must represent a net improvement. The proposed implementation, if applicable,
+must be solid and must not complicate the codebase unduly. Finally, a proposed
+enhancement must be "debopsy" in order to be accepted by the DebOps
+maintainers. (However, "debopsy" is an imprecise term; it may be defined as
+whatever is acceptable to the DebOps maintainers. This logic is intentionally
+circular.).
```
ypid, line 247: debopsy :smile: Throwing in a joke here and there is a good idea with such long and otherwise welcoming precise specifications.

docs/dep/dep-0001.rst
```diff
@@ -0,0 +1,599 @@
...
+3. Copyright/public domain -- Each DEP must either be explicitly labeled as
+   placed in the public domain or licensed under the
+   `GNU General Public License v3`_.
+
+4. Specification -- The technical specification should describe the
+   syntax and semantics of any new project feature. The importance of various
+   specification details can be emphasised by usage of the capitalized key
+   words, as described in the `RFC 2119`__.
+
+   .. __: https://tools.ietf.org/html/rfc2119
```
ypid: "must" (lower case) is used a lot in the document and the meaning is defined by RFC 2119. Although the RFC seems to be not 100 % precise about upper/lower case. I would say it is better to always write them in upper case (as done in the keyring docs and "older"/not newly written DEPs) so that it is clear that the author was aware of the meaning. That is also what the RFC implies and what it uses in itself.

Also because you write "details can be emphasised", if we don’t use upper case then, it is unclear.

docs/dep/dep-0001.rst
```diff
@@ -0,0 +1,599 @@
...
+
+DEP Editors
+~~~~~~~~~~~
+
+The DEP editors are individuals responsible for managing the administrative
+and editorial aspects of the DEP workflow (e.g. assigning DEP numbers and
+changing their status).  See `DEP Editor Responsibilities & Workflow`_ for
+details. The current editors are:
+
+- Maciej Delmanowski
```
ypid, line 91: I might add myself when I have more time, but not right now, at least not officially. But I will try to keep up and should meet the DEP Editor requirements. But maybe don’t request a review from me where you think I am not involved in the discussion or which is unrelated to me, for now.

ypid (changes requested):
Nice work! Pretty detailed and lets say, ready for the future :) Please find my, minor, fixes in: https://github.com/drybjed/debops/pull/6 and as comments.

ypid:
Defining DEPs is interesting and should help DebOps when it further grows. I approve the idea.

docs/dep/dep-0001.rst
drybjed: Good point. In Python as I understand it, this was done to keep the `git` repository with the PEPs clean. Original author of the PEP could have done multiple commits against PEP 9999, and after passing it off to the PEP editors they were free to renumber the PEP to the correct new number and squash all of the author commits together.

But we are tracking the commit origins via GPG signatures which breaks this system... I suppose that some of the current PEP workflow would need to be adapted for DEPs to handle this. I'll add a quick explanation on how to squash commits before submitting the work so that DEP authors could still have their work sighed with GPG keys, and there would need to be a mechanism to request new DEP numbers. I guess a reservation list in DEP 0 would suffice and DEP authors could be given the numbers after discussing an idea over the mailing list or GitHub issues.

docs/dep/dep-0001.rst
drybjed: I cannot take the credit here, it's in the original PEP 1 as well. :-)

docs/dep/dep-0001.rst
drybjed: [RFC 8174](https://tools.ietf.org/html/rfc8174) makes this more precise - only UPPERCASE words have the intended meaning defined by the RFC 2119.

In RFC 2119 there's a section about usage of the terms which states:

> Imperatives of the type defined in this memo must be used with care
> and sparingly.  In particular, they MUST only be used where it is
> actually required for interoperation or to limit behavior which has
> potential for causing harm

In my understanding it means that if you are writing a specification for an obvious thing that cannot be interpreted in multiple ways, using lowercase version is sufficient, as in speaking in normal voice. Only ambigious things require an uppercase version of the words which then carry **very specific meaning** defined by the RFC and that causes the given specification to be interpreted in a certain way, intended by the author.

docs/dep/dep-0001.rst
drybjed: I'm aware of this and I did not intentionally add you or anybody else just yet, community is still small.

DEPs (and PEPs which they were based on) are designed for long-term discussions and mesh perfectly well with `git` history. This allows for interested parties to voice their concens at their own pace. But since this PR only introduces them, I wanted to have a few other people in the community review them beforehand to polish all the issues we might have.

docs/dep/dep-0001.rst
ypid: Works for me. To be explicit, we should the use:


```
      The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL
      NOT", "SHOULD", "SHOULD NOT", "RECOMMENDED", "NOT RECOMMENDED",
      "MAY", and "OPTIONAL" in this document are to be interpreted as
      described in BCP 14 [RFC2119] [RFC8174] when, and only when, they
      appear in all capitals, as shown here.
```

ypid (changes requested):
Thanks for merging. A few comments are open, otherwise I approve it.

ypid (approved):
LGTM.

PR #316, opened 2018-04-21
debops-optimize: Mechanical edit: Cleanup

ypid:
I would suggest we include `../../../includes/global.rst` in all `defaults/main.yml` files by default. This can be done by yaml4rst if we agree on it. Conditionally including is not supported by yaml4rst yet and I would like to avoid it :)

@drybjed I would suggest you give it a try maybe in a time when no PR are open to avoid rebasing. I can also do it when I notice a low PR count.


drybjed (approved):
Thanks, I'll merge this while there are no conflicts.

As for the `global.rst`, I'm more inclined into using anonymous links these days, but I suppose that having global static list of links doesn't sound that bad. :-)

ypid:
> anonymous links

Seen that :) I guess they are handy but once you define the same URL a second time, I would strongly recommend to define the URL in `global.rst`.

PR #321, opened 2018-04-24
[debops.ifupdown] Add 'debops.kmod' dependency

drybjed:
Some of the network interface types (bridges, bonding, VLANs) require
additional kernel modules to be loaded at boot time or during
configuration. This can be handled by the operating system on demand,
however this results in warnings about missing modules and can cause
issues at boot time.

This patch adds the 'debops.kmod' role as a dependency of the
'debops.ifupdown' role. The 'debops.ifupdown' role will check what
network interfaces are to be configured on a host, and will dynamically
generate configuration for the 'debops.kmod' role which will then load
the missing kernel modules and mark them to be loaded at boot time.

### DIFF
--- a/ansible/playbooks/service/ifupdown.yml
+++ b/ansible/playbooks/service/ifupdown.yml
@@ -13,6 +13,11 @@
     - role: debops.ifupdown/env
       tags: [ 'role::ifupdown', 'role::ferm' ]
 
+    - role: debops.kmod
+      tags: [ 'role::kmod' ]
+      kmod__dependent_load:
+        - '{{ ifupdown__env_kmod__dependent_load }}'
+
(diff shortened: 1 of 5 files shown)


ansible/playbooks/service/ifupdown.yml
```diff
@@ -13,6 +13,11 @@
     - role: debops.ifupdown/env
       tags: [ 'role::ifupdown', 'role::ferm' ]
```
scibi, line 14: @drybjed shouldn't `'role::kmod'` be added here too?

drybjed: @scibi Yes, you're right. Good catch! Will do that shortly.

PR #328, opened 2018-05-01
[debops-contrib.tor] Automatic firewall config, Fix auto update of tor

ypid:
* Add automatic firewall config based on tor_ports
* Fix auto update of tor

This is still work in progress but it should be usable.

PS: I know `debops-contrib.tor` is not anywhere near the DebOps Standards and this role is unfinished but it is already usable and polishing will need to be done later. Note that it is technically still a DebOps Contrib role.

Tested on: Debian Stretch
Uses: https://github.com/nusenu/ansible-relayor/pull/163

### DIFF
--- a/ansible/debops-contrib-playbooks/service/tor.yml
+++ b/ansible/debops-contrib-playbooks/service/tor.yml
@@ -1,6 +1,6 @@
 ---
 
-- name: FIXME
+- name: Manage Tor relay
   hosts: [ 'debops_service_tor' ]
   become: True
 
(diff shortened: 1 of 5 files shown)


drybjed (approved):
Sounds good. Your suggestion about symlinking files to inventory, etc. is kinda weird, but I haven't played with the relayor role so I'm not sure how this role works with it. I guess this is what happens when you try to combine stuff with roles outside of DebOps; normally this would be done via Ansible local facts. But since this is still a Contrib role, not enabled by default, I think its fine, for now.

ypid:
Thanks for merging. To be improved.

PR #329, opened 2018-05-02
[debops.mariadb_server] Fix failure with non-default datadir

gasuketsu:
Fix debops.mariadb_server role failure when `mariadb_server__datadir` have non-default value.

The database service gets started when packages are installed even if environment variable `RUNLEVEL=1` is set and files under default datadir (`/var/lib/mysql`) should be moved to new datadir,
so this fix performs following:
* Stop database server on first install
* Move files under default datadir to given datadir


drybjed (approved):
Sounds good. The database server is restarted on first install to apply new configuration, so stopping it and starting should work similarly. I haven't tested this PR, but any issues will quickly show up in GitLab CI on the next full test.

PR #330, opened 2018-05-03
Add mode parameter for docker sources.list file

gardouille:
The default value from the module (420) might block some users from
using apt* commands (such as monitoring tools, standard users,…).


gardouille:
The default mode is already overwrite for some other roles (apt_install, mariadb,…), i choose the same value (0644).

I set it in a variable to allow users to overwrite it if they really want to block this access.

In my case, i wasn't able to use apt-cache policy without sudo/su permissions ^^

drybjed (changes requested):
On my hosts the Docker `sources.list` file already has `0644` permissions. Maybe its due to your OS choice or Ansible version?

Anyway, I guess we can set the mode in the task, there's no issue with using `0644` there. Since APT sources are "public", I don't see any need to have it as a default variable though, you could just set it directly. I don't think that anybody would want to change the default `0644` to anything else.

gardouille:
Close this and create a new clean PR… :)

PR #347, opened 2018-05-23
Update the arch linux package for Python 3

Cadair:
It seems that the main arch ansible package migrated to Python 3. So I had to update this to make debops run on my machine again.

Do you have any issues with me uploading this to the AUR?


drybjed (approved):

drybjed:
@Cadair you can upload it to AUR, no issue there. I don't use Arch personally, but I trust that you know what needs to be done. Thanks for the support! :-)

PR #355, opened 2018-05-28
Fail2ban: group and host management for ignoreip

gardouille:
In order to add IP to whitelist from a group or a host.


drybjed (approved):
Sounds good, thanks.

PR #358, opened 2018-05-31
Chrony support

gaudenz:
Three fixes for chrony support. The first is primarily relevant to support Debian Jessie, the two others are relevant if you want to run chrony as an NTP server. If you prefer individual pull requests I can also create two or even three.


drybjed (approved):
Looks good, thanks. Small commits are more important rather than multiple PRs, useful for bisecting. :+1:

PR #368, opened 2018-06-06
[debops.gitlab] Fix syntax error in Unicorn service file.

bfabio:


drybjed (approved):
That probably happened when MariaDB support was removed. Thanks for the fix!

drybjed:
Looks like the Travis-CI API is experiencing slowdown at the moment. The tests passed, therefore I'm merging the PR.

PR #377, opened 2018-06-25
[debops.nginx] php*.conf.j2 : use item.try_files instead of hardcoded values

alo-is:
In `php.conf.j2` and `php5.conf.j2`, hardcoded values of `try_files` were used, instead of the user definable `item.try_files` variable. 

This was no suitable for all cases : for example, can't use index.php for all URL for custom handling with many frameworks.

### DIFF
--- a/ansible/roles/debops.nginx/templates/etc/nginx/sites-available/php.conf.j2
+++ b/ansible/roles/debops.nginx/templates/etc/nginx/sites-available/php.conf.j2
@@ -11,7 +11,7 @@
 
 {% endblock %}
 {% block nginx_tpl_block_location_root %}
-                try_files $uri $uri/ $uri.php;
+                try_files {{ item.try_files|default("$uri $uri/ $uri.php") | join(' ') }};
 {% endblock %}
 {% block nginx_tpl_block_custom_status_locations %}
 {%   if item.php_status|d()|bool %}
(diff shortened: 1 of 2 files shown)


ansible/roles/debops.nginx/templates/etc/nginx/sites-available/php.conf.j2
```diff
@@ -11,7 +11,7 @@
 
 {% endblock %}
 {% block nginx_tpl_block_location_root %}
-                try_files $uri $uri/ $uri.php;
+                try_files {{ item.try_files|default("$uri $uri/ $uri.php") | join(' ') }};
```
drybjed, line 14: The idea is good, however I'm not sure if `join()` is needed here. Just expect a string, otherwise the generated line could have space between every letter since Python treats a string like a list of characters in this case.

alo-is: I followed the same pattern the current implementation used in `default.conf.j2` : https://github.com/debops/debops/blob/480482048a96e6758176ed38dd5fdbef318d659f/ansible/roles/debops.nginx/templates/etc/nginx/sites-available/default.conf.j2#L397

Changing this might lead in other inconsistencies, which can impact many other playbooks and users.

drybjed: Sounds good, thanks for clarification. :+1:

drybjed (approved):

Issue #376, opened 2018-06-24
Error using debops.postfix


computerlyrik:
I got an error using the postfix-role within debops.mono:

```
TASK [debops.postfix : Generate Postfix 'main.cf' configuration] ***************************************************************************
[...]

PR #379, opened 2018-06-26
Fix for python3 compatibility on list keys

computerlyrik:
Fixes an issue on templating with python3 and jinja2.

Came up on this with my archlinux machine.

Please have a look into https://github.com/debops/debops/issues/376 for more details


drybjed (approved):
Seems to be working OK on Python 2.7 as well, thanks for the fix!

PR #381, opened 2018-06-28
Fix exception thrown by debops-default for Py2/3.

cultcom:
Without the encoding debops-defaults throws exceptions with Py2 and Py3.


drybjed (approved):
Thanks for the fix!

PR #382, opened 2018-06-29
Py3

cultcom:
This appears more elegant to me.


drybjed:
The `past` Python module is not part of standard Python install, at least on Debian. That probably means that the list of Python requirements of the PyPI package would need to be updated as well, and Arch packaging too. Did you check if the filter works on a clean Ansible Controller install?

cultcom:
no I have no "clean controller" around... in the end all this is only necessary to remain Py2 compatibility. So in my opinion everything should migrate to Py3 only to get clean code without compromises.

drybjed:
In this specific case this is an Ansible plugin. Ansible is not yet fully Python3 compatible, so the internal code used by it should work reliably on Python2 as well. On top of that, [Ansible package in Debian uses `#!/usr/bin/env python` shebang](https://sources.debian.org/src/ansible/2.5.5+dfsg-1/bin/ansible/) which results in usage of Python 2 to run it in Debian. So I'm afraid that Py2 compatibility will need to be here for a while longer.

i'll test this change on a clean host and see what needs to be added to dependencies to make it work properly.

cultcom:
You need these:
python-future
python3-future

PR #386, opened 2018-07-09
fix bad link for getting started guide

fermulator:
fix debops-init inventory/hosts file link to the getting started guide
(previous reference was invalid likely due to doc reorg)


drybjed (approved):
Thanks for the fix!

PR #389, opened 2018-07-09
Fix nameConstraints to allow subdomains.

cultcom:
According to https://tools.ietf.org/html/rfc5280#section-4.2.1.10 the
DNS name should be prefixed with a leading "." to allow hosts and
subdomains based on that name.


drybjed (approved):
Sounds good, thanks.

PR #394, opened 2018-07-13
fix some `docs.debops.org` URLs in .rst files

myrdd:
some of the changed URLs didn't work anymore. the others I have changed for consistency.


drybjed (approved):
Thanks for fixing the docs!

PR #397, opened 2018-07-18
Change ifupdown forward configuration per documentation

jstruebel:
Discovered that the Ferm forward configuration from the ifupdown role didn't use the forward_interface_ferm_rule_enabled option like the documentation specified so I fixed it.


drybjed (approved):
Nice catch, thanks.

PR #412, opened 2018-07-26
[debops.ferm] Handle IPv4/IPv6 addresses in 'dmz'

drybjed:
The 'ferm' language does not support iterating over lists or converting
lists to strings, this caused an issue where a list of IP addresses is
concatenated with a string representing a port when the 'dmz' rule type
is used. The result was an error:

    variable 'PRIVATE_IP' must be a string, but it is an array

This patch changes the template for the 'dmz' rule so that the IPv4 and
IPv6 addresses are separated on Ansible level instead of directly in
ferm. Next, only 1 public and private IP address is selected for the
'dmz' rule so that the string concatenation with a port can be performed
correctly. This should solve the issue with the 'dmz' rule type.

Fixes https://github.com/debops/ansible-ferm/issues/55
Fixes https://github.com/debops/ansible-ifupdown/issues/73


drybjed:
This PR is related to https://github.com/debops/debops/pull/209/, the other one fixed the `dmz` rule type only for the "old" `debops.ferm` rules, but not the current ones stored in `/etc/ferm/rules.d/`. I'll add the rest of the changes from the linked PR as well, for completion.

PR #413, opened 2018-07-28
[debops.php] Added support for newer PHP versions

carlalexander:
Added newer PHP versions `php__version_preference` since `php7.0` isn't actively supported anymore and isn't available on newer Linux distributions like bionic.

Side note: It would be useful to maybe throw an error if `php__version` and `php__long_version` are empty. You get a weird python error otherwise.


drybjed (approved):
Sounds good, thanks.

PR #419, opened 2018-08-17
[debops.nullmailer] Accept integers in port.

bfabio:
Cast the port in nullmailer__remotes to string, so the user can set
that variable as integer (as he may be tempted to do).


drybjed (approved):
Good idea, thanks.

PR #425, opened 2018-08-26
Added option to specify interface in ferm dmz rules

jstruebel:
Also made public_ip optional.
This change allows to specify only the interface used for the dmz DNAT rules which is necessary when the public interface has a dynamic address e.g. for a home router.


drybjed (approved):
Nicely done, thanks.

PR #427, opened 2018-08-28
[minor] remove `*__nginx__dependent_servers.*.default`

myrdd:
I'm using `debops.owncloud` as a template to create a new role based on it. In the `owncloud__nginx__dependent_servers` default variable, I wondered what the `default` variable is for.

* It's not documented (https://docs.debops.org/en/latest/ansible/roles/debops.nginx/defaults-detailed.html#nginx-ref-servers)
* it seems unused. I've checked these files:
  * `ansible/roles/debops.nginx/tasks/nginx_servers.yml`
  * `ansible/roles/debops.nginx/templates/etc/nginx/sites-available/*.conf.j2`

If I'm wrong, simply close this PR.

### DIFF
--- a/ansible/roles/debops.owncloud/defaults/main.yml
+++ b/ansible/roles/debops.owncloud/defaults/main.yml
@@ -1882,7 +1882,6 @@ owncloud__nginx__dependent_servers:
   ## https://docs.nextcloud.com/server/13/admin_manual/installation/nginx.html
   - type: 'default'
     enabled: True
-    default: False
     by_role: 'debops.owncloud'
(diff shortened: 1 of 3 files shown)


drybjed (approved):
Good idea, thanks.

The `default` parameter was used in the `debops.nginx` role to designate the "default server"; its configuration would include the `default_server` and `ipv6only=off` parameters, since these need to be unique per port `nginx` listens on.

At some point the way that the default server was chosen was changed so that `debops.nginx` would pick one by itself. After that, this parameter is not used anymore.

PR #431, opened 2018-08-29
fix missing unicode in python 3

computerlyrik:
Fixes a issue for unicode class in python 3.

Found it elsewhere in the code of debops and adapted for this file.


drybjed (approved):
Sounds good, thanks!

PR #449, opened 2018-09-14
elastic_co: add curator repository

scibi:


ansible/roles/debops.elastic_co/defaults/main.yml
```diff
@@ -31,10 +31,16 @@ elastic_co__version: '{{ ansible_local.elastic_co.version
...
 #
 # APT repository entry to configure for Elastic repository.
-elastic_co__repository: 'deb https://artifacts.elastic.co/packages/{{ elastic_co__version }}/apt stable main'
+elastic_co__repositories:
+  - repo: 'deb https://artifacts.elastic.co/packages/{{ elastic_co__version }}/apt stable main'
+  - enabled: "{{ ansible_distribution_release in [ 'wheezy', 'jessie' ] }}"
+    repo: 'deb [arch=amd64] https://packages.elastic.co/curator/5/debian stable main'
+  - enabled: "{{ ansible_distribution_release in [ 'stretch' ] }}"
+    repo: 'deb [arch=amd64] https://packages.elastic.co/curator/5/debian9 stable main'
+
```
drybjed: This is kind of a bit hard to read, I would separate each entry with a line, and reverse the order of `repo` and `enabled` keys.

I don't think that specifying `[arch=amd64]` is needed, APT should handle this automatically. This might cause issues on other architectures.

I suppose that the Curator version number to install could exposed later on, when new release happens.

drybjed (changes requested):
Nice work, I'd just like to see the list of repositories a bit more readable. :-)

drybjed (approved):

PR #463, opened 2018-09-28
Fix wrong indention of service/mailman.yml. Fixes #462.

ypid:
Fixes #462.


drybjed (approved):
Thanks. No idea why `yamllint` on Travis didn't catch that, local test returns with an error.

drybjed:
No, wait [it did catch the error](https://travis-ci.org/debops/debops/jobs/43477). Seems to be a bug in test suite, will investigate.

PR #470, opened 2018-10-04
[debops.owncloud] Disabled deny_hidden in nginx conf, fixes #469

### LINKED ISSUE #469: [debops.owncloud] Nginx deny_hidden breaking chunked upload in Nextcloud
Hi,

It seems that this block : 

https://github.com/debops/debops/blob/7cbc3df22718e4d379cab0814659454faf6f40ed/ansible/roles/debops.owncloud/defaults/main.yml#L1880-L1897

is missing `deny_hidden: False`, since Nextcloud is using URLs ending with `.file` with the desktop client & the webapp for large files (dav & chunked upload) :

```
MOVE /remote.php/dav/uploads/xxxxxxxx-xxxx-xxxx-xxxx-xxxxxxxx/1234567890/.file HTTP/1.1" 404 4194 "-" "Mozilla/5.0 (Windows) mirall/2.3.3 (build 1) (Nextcloud)
```

Removing the following block from my Nginx config fixed the issue :

```
        # Disallow access to hidden files and directories
        location ~ /\. {
                return 404;
        }
```

alo-is:
Fixes #469.

### DIFF
--- a/ansible/roles/debops.owncloud/defaults/main.yml
+++ b/ansible/roles/debops.owncloud/defaults/main.yml
@@ -1886,6 +1886,7 @@ owncloud__nginx__dependent_servers:
     filename: 'debops.owncloud'
     name: '{{ owncloud__fqdn }}'
     root: '{{ owncloud__deploy_path }}'
+    deny_hidden: False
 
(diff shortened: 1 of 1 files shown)


drybjed (approved):
Sounds good, thanks for the fix.

PR #471, opened 2018-10-04
[debops.grub] Re-introduce grub__timeout_* variables removed in #466

ypid:
I found/find them quite useful to make precise changes in the inventory.

https://github.com/ypid/ypid-ansible-inventory/search?q=grub&unscoped_q=grub

Updates: #466

### DIFF
--- a/ansible/roles/debops.grub/defaults/main.yml
+++ b/ansible/roles/debops.grub/defaults/main.yml
@@ -129,10 +129,10 @@ grub__default_configuration:
   # Set the default GRUB timeout depending on the host type (hardware or
   # virtual machine) to allow for faster booting of VMs
   - name: 'timeout'
-    value: '{{ "5"
+    value: '{{ grub__timeout_hardware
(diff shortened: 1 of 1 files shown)


drybjed:
They way to handle that now would be something like this in the inventory:
```
grub__configuration:
  - timeout: 10
```
But sure, I guess they can be exposed in separate variables.

drybjed (approved):

ypid:
@drybjed How would I make changes only to hardware or virtual in the group inventory with this? I only want to change either hardware or virtual. I do this quite a bit in my inventory.

drybjed:
@ypid Have separate groups for hardware and virtual hosts, and put `grub__group_configuration` in each with desired value. It can be a bit tricky if you want to change something more though at the group level, you would need to do some list contatenation tricks in the inventory.

ypid:
Thanks for merging. I guess with those variables, it will be the simplest.

PR #472, opened 2018-10-04
Various small fixes

ypid:
Various small fixes I had laying around.


drybjed (approved):
Looking good, thanks!

PR #473, opened 2018-10-04
[debops.avahi] Add avahi__enabled configuration option

ypid:
Some packages like Mandos hard depend on Avahi but can be configured to work without it. Thus disabling the service is the only option if I don’t want to have Avahi running.


drybjed (approved):
Interesting way to handle this. I guess a boolean should fine here, instead of `state` parameter.

PR #474, opened 2018-10-04
[debops.grub] grub__configuration: value does not support list of lists

ypid:


drybjed (approved):
Right, my bad. Good catch.

PR #476, opened 2018-10-05
Check for docker in ansible_virtualization_type

nickjj:
In `sysctl__writable` you are checking against `lxc` and `openvz` as potential container runtimes. Ansible is able to identify `docker` as well. You are able to write `net` parameters inside of a Docker container.


drybjed (approved):
Sounds good, thanks.

PR #480, opened 2018-10-09
[debops.rsyslog] Ensure that the rsyslog__user can access the ssl certs

mikenowak:
I hit an issue today with the following setup

```
rsyslog__capabilities: [ 'tls' ]
rsyslog__forward: [ '*.* @@loghost1.{{ ansible_domain }}:6514' ]
```

It turns out that rsyslog refuses to start as it can not access the certificates.

### DIFF
--- a/ansible/roles/debops.rsyslog/defaults/main.yml
+++ b/ansible/roles/debops.rsyslog/defaults/main.yml
@@ -81,6 +81,14 @@ rsyslog__user: '{{ "syslog" if rsyslog__unprivileged|bool else "root" }}'
 rsyslog__group: '{{ "syslog" if rsyslog__unprivileged|bool else "root" }}'
 
                                                                    # ]]]
+# .. envvar:: rsyslog__append_groups [[[
+#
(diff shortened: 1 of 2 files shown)


drybjed (changes requested):
Good idea, however this issue is probably apparent only on Ubuntu. On Debian, `rsyslog` service is running under the `root` account; you can see this reflected in the `rsyslog__user` and `rsyslog__group` variable values.

I would propose adding a `rsyslog__additional_groups` default list, and in it add the `ssl-cert` group if the `rsyslog__unprvileged` variable is True, make the list empty otherwise. Don't forget to add `append: True` to the task, so that the list of additional groups won't get reset by Ansible when something else added the `syslog` account to a different UNIX group.

This should avoid adding `root` to the `syslog` group - this account shouldn't be in any other groups.

mikenowak:
@drybjed, this hopefully should do it

drybjed (approved):
Nicely done, thanks. :+1:

PR #482, opened 2018-10-09
[debops.snmpd] Bionic uses Debian-snmp user/group

mikenowak:
The default user/group for the snmpd daemon on **bionic** is `Debian-snmp`.


drybjed (approved):
I guess the changes from Debian have been merged into Ubuntu, thanks for the fix. :+1:

PR #485, opened 2018-10-12
[debops.librenms] Adding the ability to modify the config file mode

mikenowak:
The use case is that I forward syslog to librenms and in **syslog.php** (executed as syslog user) currently fails if it can not read the config file.

### DIFF
--- a/ansible/roles/debops.librenms/defaults/main.yml
+++ b/ansible/roles/debops.librenms/defaults/main.yml
@@ -156,6 +156,14 @@ librenms__log_dir: '{{ (ansible_local.root.log
                             ansible_local.root.log|d())
                         else "/var/log") + "/librenms" }}'
                                                                    # ]]]
+
+# .. envvar:: librenms__config_mode [[[
+#
+# LibreNMS config mode. This allows to override the default config mode
+# so that the other system users (i.e. syslog) could execute scripts
+# such as syslog.php.
+librenms__config_mode: '0600'
+                                                                   # ]]]
                                                                    # ]]]
 # LibreNMS sources [[[
 # --------------------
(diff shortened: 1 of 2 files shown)


ansible/roles/debops.librenms/defaults/main.yml
```diff
@@ -156,6 +156,10 @@ librenms__log_dir: '{{ (ansible_local.root.log
                             ansible_local.root.log|d())
                         else "/var/log") + "/librenms" }}'
                                                                    # ]]]
+# .. envvar:: librenms__config_mode [[[
+#
+# LibreNMS config mode.
+librenms__config_mode: '0600'
```
drybjed, line 165: You should add another `# ]]]` here so that the `vim` folds will be handled correctly.

ansible/roles/debops.librenms/defaults/main.yml
```diff
@@ -156,6 +156,10 @@ librenms__log_dir: '{{ (ansible_local.root.log
                             ansible_local.root.log|d())
                         else "/var/log") + "/librenms" }}'
                                                                    # ]]]
```
drybjed, line 158: You should add empty line @158, so that the variables are easier to notice.

drybjed (changes requested):
Sounds good, although you should add small cosmetic fixes so that `vim` folding works fine.

BTW, how about that syslog config? Perhaps you could add an example in the role documentation? :-)

mikenowak:
@drybjed, is this acceptable?

drybjed (approved):
Empty line is in the wrong place, but I'll take it. :-) I can fix it while merging the PR. Thanks.

PR #493, opened 2018-10-22
[debops-contrib.btrfs] Include missing module

drybjed:
The custom 'btrfs_subvolume' Ansible module is required by the
'debops-contrib.btrfs' role to correctly manage BtrFS subvolumes.


ypid (approved):
LGTM, thanks very much. I had it included outside of the role: https://github.com/ypid/ypid-ansible-common/tree/master/ansible/library

PR #494, opened 2018-10-22
Use secrets with the nullmailer role

rei-ifesca:
Just like the postfix role.

1. Use aliases to send notification to the correct email address/alias
2. Sometimes is necessary to use a SMTP transport server from which nullmailer needs credentials, thus use the secret role to store and retrieve this credentials

### DIFF
--- a/ansible/playbooks/service/nullmailer.yml
+++ b/ansible/playbooks/service/nullmailer.yml
@@ -10,9 +10,6 @@
 
   roles:
 
-    - role: debops.etc_aliases/env
-      tags: [ 'role::etc_aliases', 'role::secret', 'role::nullmailer' ]
-
     - role: debops.nullmailer/env
       tags: [ 'role::nullmailer', 'role::ferm', 'role::tcpwrappers' ]
 
@@ -31,9 +28,6 @@
       tags: [ 'role::tcpwrappers', 'skip::tcpwrappers' ]
       tcpwrappers__dependent_allow:
         - '{{ nullmailer__tcpwrappers__dependent_allow }}'
-
-    - role: debops.etc_aliases
-      tags: [ 'role::etc_aliases', 'skip::etc_aliases' ]
 
     - role: debops.nullmailer
       tags: [ 'role::nullmailer', 'skip::nullmailer' ]
--- a/ansible/playbooks/service/nullmailer.yml
+++ b/ansible/playbooks/service/nullmailer.yml
@@ -10,8 +10,17 @@
 
   roles:
 
+    - role: debops.etc_aliases/env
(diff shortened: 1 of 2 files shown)


drybjed:
Good idea, however [`nullmailer` does not support `/etc/aliases` lookups](https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=95624). Even if you configure `/etc/aliases` database, `nullmailer` will not use it. If you want local forwarding support, currently you should enable Postfix support on a host, for example via `debops.postfix` role, which will automatically replace `nullmailer`.

An alternative would be to switch the default SMTP server in DebOps from `nullmailer` to [`dma`](https://wiki.debian.org/Debate/DefaultMTA/DMA) which supports `/etc/aliases` database. Of course the role for it should be similar to `debops.nullmailer` and support automatic replacement by Postfix. DMA looks interesting and I plan to try it out in the future and if it works, use it as default in DebOps; it's just that I currently have other things to work on. If you want to take a crack at a role for it, go ahead.

As for the support for `debops.secret` in the `debops.nullmailer` role - good idea, however `debops.secret` should be added in the role dependencies, ie. `ansible/roles/debops.nullmailer/meta/main.yml`. That way it will be always available and you don't need to specifically include it in the playbook.

rei-ifesca:
@drybjed you got a point.
I will remove the aliases code and leave the role::secret stuff. And of course update the meta information.

Regarding the discussion about a new mail daemon I will rather use this Isssue #495 .

ansible/playbooks/service/nullmailer.yml
```diff
@@ -13,6 +13,12 @@
     - role: debops.nullmailer/env
       tags: [ 'role::nullmailer', 'role::ferm', 'role::tcpwrappers' ]
 
+    - role: debops.secret
+      tags: [ 'role::secret', 'role::nullmailer' ]
+      secret__directories:
+        - '{{ etc_aliases__secret__directories }}'
+        - '{{ postfix__secret__directories }}'
+
```
drybjed, line 21: This is not needed in the playbook, having the dependency in `meta/main.yml` of a role makes it a "hard dependency" which means that it will always be included in the run.

Also, you referenced a variable `postfix__secret__directories` which is not in any of the included roles and will result in Ansible erroring out.

rei-ifesca: Apparently I have not drank enough coffee this morning :coffee:

rei-ifesca: @drybjed So I just need to edit the meta information without touching the playbook?

drybjed: @rei-ifesca Yes, exactly. Just modify `meta/main.yml` and you're good.

PR #505, opened 2018-10-29
Call the debops.users role before the debops.resources role

rei-ifesca:
otherwise the resources role may crash if a path/file is created with an owner, who does not exists yet, because the users role has not been executed. Placing the path owner (user) in the bootstrap configuration, if is not an sysadmin, is not always a good idea.

### DIFF
--- a/ansible/playbooks/common.yml
+++ b/ansible/playbooks/common.yml
@@ -110,6 +110,9 @@
     - role: debops.machine
       tags: [ 'role::machine', 'skip::machine' ]
 
+    - role: debops.users
+      tags: [ 'role::users', 'skip::users' ]
+
     - role: debops.resources
       tags: [ 'role::resources', 'skip::resources' ]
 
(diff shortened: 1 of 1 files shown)


drybjed:
The `debops.resources` role was executed after [`debops.users` role](https://github.com/debops/debops-playbooks/pull/360) but it was changed to solve issues with `debops.ferm`. I'd like to check any issues with other `common.yml` roles before applying this change, I can do this properly in about a week (currently on holidays).

Do you know that `debops.users` has separate set of tasks to handle per-user resources? Perhaps these would be sufficient for your use case? Check the role documentation.

rei-ifesca:
I will give it a try with the users resources

rei-ifesca:
Placing the debops.users role before the debops.resources role is meaningful. Specially if I need to clone several git repositories, for with an owner been an user to be created with the debops.users role.
Then somethimes is handy to give access to other users and groups to paths or the git repos, however this users and groups need to be created before debops.resources runs.

Sadly using the user__resources is not powerful enough for my needs

drybjed (approved):
I checked the `common.yml` playbook with this change on a fresh host and I don't see any issues with playbook execution with the changed order. Let's do this then. Thanks for the PR!

PR #506, opened 2018-10-29
Install "gunicorn3" package for Py3 web applications.

cultcom:
Add the "gunicorn3" packages to the base_packages.

### DIFF
--- a/ansible/roles/debops.gunicorn/templates/etc/systemd/system/gunicorn@.service.j2
+++ b/ansible/roles/debops.gunicorn/templates/etc/systemd/system/gunicorn@.service.j2
@@ -13,7 +13,7 @@ Wants=network-online.target
 PermissionsStartOnly=true
 EnvironmentFile=-/etc/default/gunicorn-%i
 PIDFile=/run/gunicorn-%i/%i.pid
-ExecStart=/usr/bin/{{ item.binary if (item.binary is defined) else gunicorn__binary }} --config=/etc/gunicorn/%i.conf.py ${GUNICORN_OPTS}
+ExecStart=/usr/bin/{{ gunicorn__binary }} --config=/etc/gunicorn/%i.conf.py ${GUNICORN_OPTS}
 ExecReload=/bin/kill -s HUP $MAINPID
 ExecStop=/bin/kill -s TERM $MAINPID
 Restart=always
--- a/ansible/roles/debops.gunicorn/templates/etc/systemd/system/gunicorn@.service.j2
+++ b/ansible/roles/debops.gunicorn/templates/etc/systemd/system/gunicorn@.service.j2
@@ -13,7 +13,7 @@ Wants=network-online.target
 PermissionsStartOnly=true
 EnvironmentFile=-/etc/default/gunicorn-%i
 PIDFile=/run/gunicorn-%i/%i.pid
-ExecStart=/usr/bin/gunicorn --config=/etc/gunicorn/%i.conf.py ${GUNICORN_OPTS}
+ExecStart=/usr/bin/{{ item.binary if (item.binary is defined) else gunicorn__binary }} --config=/etc/gunicorn/%i.conf.py ${GUNICORN_OPTS}
 ExecReload=/bin/kill -s HUP $MAINPID
 ExecStop=/bin/kill -s TERM $MAINPID
 Restart=always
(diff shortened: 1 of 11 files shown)


drybjed:
@cultcom. Sounds good, however what would you say about converting this role to use `debops.python` to install the Py2 and Py3 packages, ie. `gunicorn` and `python-setproctitle` for Py2, `gunicorn3` and `python3-setproctitle` for Py3. You could look at the `debops.redis_server` role/playbook for example usage, and the `debops.python` documentation.

I can look at this in a week or so (currently on holidays) if you are not sure how to make the conversion.

cultcom:
happy holiday then.

There is still another issue with the generated config file for gunicorn. With gunicorn3 it complains about malformed JSON data.

cultcom:
Ok, have a look at my last changes then. I hope I understood you correctly. In case I actually did I would suggest to remove the base_packages handling from this role completely as well.

I also added a fix for some JSON errors leading to decoder exceptions.

drybjed:
@cultcom Nicely done! I agree that `gunicorn__base_packages` variable can be dropped; remove it and I think this PR will be ready for merging.

As for the gunicorn config file, it will have to be ignored by pyCodeStyle and the Jinja wrapping can then be dropped completely, this should solve several issues during servie execution. It can be done in a separate step.

cultcom:
@drybjed I removed the package installation tasks as requested but I don not completely understand your point regarding the config file. With my last changes the service starts running at least.

Another point might be the systemd service files. They currently call "gunicorn" which is no problem when you prepend a Py3 virtualenv path. But without venv a Py3 app should be called with "gunicorn3" instead...

drybjed:
@cultcom Perhaps this could be solved by adding `item.gunicorn_binary` parameter that would specify `/usr/bin/gunicorn` by default and could be overriden on a per-app basis. Do you want to add that as well before I merge this PR? If to, don't forget to update the documentation.

cultcom:
I added a item.binary option which configures the service per app and a global gunicorn__binary as a default fallback.

ansible/roles/debops.gunicorn/templates/etc/systemd/system/gunicorn@application.service.d/instance.conf.j2
```diff
@@ -28,11 +28,11 @@ TimeoutStopSec={{ item.systemd_timeout | d(gunicorn__systemd_timeout) }}
 {% if item.app_module|d() %}
 
 ExecStart=
-ExecStart={{ (item.python + ' ') if item.python|d() else '' }}/usr/bin/gunicorn --config=/etc/gunicorn/%i.conf.py ${GUNICORN_OPTS} {{ item.app_module }}
+ExecStart={{ (item.python + ' ') if item.python|d() else '' }}/usr/bin/{{ item.binary if (item.binary is defined) else gunicorn__binary }} --config=/etc/gunicorn/%i.conf.py ${GUNICORN_OPTS} {{ item.app_module }}
```
drybjed: I think that this can be shortened with:
```
{{ item.binary | d(gunicorn__binary) }}
```

ansible/roles/debops.gunicorn/templates/etc/systemd/system/gunicorn@.service.j2
```diff
@@ -13,7 +13,7 @@ Wants=network-online.target
 PermissionsStartOnly=true
 EnvironmentFile=-/etc/default/gunicorn-%i
 PIDFile=/run/gunicorn-%i/%i.pid
-ExecStart=/usr/bin/gunicorn --config=/etc/gunicorn/%i.conf.py ${GUNICORN_OPTS}
+ExecStart=/usr/bin/{{ item.binary if (item.binary is defined) else gunicorn__binary }} --config=/etc/gunicorn/%i.conf.py ${GUNICORN_OPTS}
```
drybjed: This change doesn't make sense - per-item variables are available in the override file, not in the main service file. You should only specify `gunicorn__binary` here.

drybjed (changes requested):
Nicely done, although there a few minor changes I'd like to see.

The `debops.gunicorn` role is used in the `ansible/playbooks/service/netbox.yml` playbook. Because you changed the dependencies of `debops.gunicorn`, you should also update all of the other playbooks it is included in - the soft dependencies will not be pulled in automatically. I think that this also warrants a mention in the Changelog "Changed" section, in case that other `debops.gunicorn` users will have to do the same in their own playbooks.

You should also add the explanation of the `binary` parameter to the documentation section that describes the `gunicorn__applications` configuration. Putting it in the example without short explanation of its use might be confusing.

Something to think for the future: the `debops.python` role provides facts that indicate if support for Python 2 is enabled or not. This could be used to automatically enable/disable `gunicorn3` default binary. DebOps is not fully Py3-only compatible yet, so this can be implemented later.

drybjed (approved):
Nicely done, thanks!

PR #509, opened 2018-11-02
Support for iPXE on EFI systems

gaudenz:
EFI systems configured to do EFI PXE booting can't use the iPXE binary
compiled for legacy BIOS PXE. They need the ipxe.efi variant compiled
for EFI.


drybjed (approved):
Sounds good, thanks!

PR #511, opened 2018-11-05
Fix issues in 'lxc_ssh.py' now python3 compatible

TuxCoder:
cleanup and a few fixes.
change to newer import names and be compatible to the old one.
tested on python2 and python3 with ansible 2.7.1

Probably remove the old ansible support in the future release.


drybjed (approved):
Thanks for the update!

PR #512, opened 2018-11-05
[debops.prosody] fix typo

TuxCoder:
Fix a few typos.
Ferm rules aren't working because of false usage of the variable `prosody__deploy_state` as bool.
There was also a copy paste issue with the ports


drybjed (approved):
Thanks for the fixes. :-)

PR #522, opened 2018-11-10
[debops.owncloud] Fix PHP packages for OwnCloud 10

sread:
mbstring, intl and zip are required by OwnCloud 10, but were only installed for the NextCloud variant.


drybjed (approved):
Sounds good, thanks!

PR #530, opened 2018-11-14
Add debops.secret dependency to debops.gitlab_runner

rei-ifesca:
Allow to store the Runner Token and Runner API Token in the secret folder.

Example:

```
gitlab_runner__token: '{{ lookup("password", secret
                           + "/credentials/" + ansible_fqdn
                           + "/gitlab/runner/token chars=ascii,numbers") }}'
gitlab_runner__api_token: '{{ lookup("password", secret
                           + "/credentials/" + ansible_fqdn
                           + "/gitlab/runner/api_access_token chars=ascii,numbers") }}'
```

### DIFF
--- a/ansible/roles/debops.gitlab_runner/meta/main.yml
+++ b/ansible/roles/debops.gitlab_runner/meta/main.yml
@@ -1,6 +1,8 @@
 ---
 
-dependencies: []
+dependencies:
+  - role: debops.secret
+    tags: [ 'role::secret' ]
 
 galaxy_info:
 
--- a/ansible/roles/debops.gitlab_runner/meta/main.yml
+++ b/ansible/roles/debops.gitlab_runner/meta/main.yml
@@ -2,7 +2,6 @@
 
 dependencies:
   - role: debops.secret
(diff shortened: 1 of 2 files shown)


rei-ifesca:
@drybjed maybe do you have an idea on how to ask the user (prompt?) during installation for the API tokens and store the tokens in the secret folder.

drybjed (changes requested):
Sounds good, although I would remove the `tags` keyword, it's not needed in this case.

Could you add an example about this usage in the Token section of the documentation? It will be more discoverable this way.

drybjed:
@rei-ifesca As for your question, I don't want to add any prompts to DebOps roles, the inventory is a good enough, and idempotent interface. Good documentation should be sufficient.

rei-ifesca:
> @rei-ifesca As for your question, I don't want to add any prompts to DebOps roles, the inventory is a good enough, and idempotent interface. Good documentation should be sufficient.

I understnad.

Updated the `GitLab CI registration token` documentation.

drybjed (approved):

Issue #483, opened 2018-10-11
[debops.auth] pam_mkhomedir not enabled


mikenowak:
@drybjed,

I am trying to make this to work on Ubuntu 18.04 bionic.

The template is generated correctly

```
$ cat  /usr/share/pam-configs/mkhomedir
# This file is managed remotely, all changes will be lost

# Enable support for libpam-mkhomedir in PAM

Name: activate mkhomedir
Default: yes
Priority: 900
Session-Type: Additional
Session:
        required                        pam_mkhomedir.so umask=0077 skel=/etc/skel
```

And the following handler is called `pam-auth-update --package libpam-modules`, however it doesn't enable the pam_mkhomedir.

If I execute `pam-auth-update` manually **activate mkhomedir** is disabled.

```
   │ PAM profiles to enable:                                                                                                                                                        │   
   │                                                                                                                                                                                │   
   │    [*] Cracklib password strength checking                                                                                                                                     │   
   │    [ ] activate mkhomedir                                                                                                                                                      │   
   │    [*] Unix password history                                                                                                                                                   │   
   │    [*] Unix authentication                                                                                                                                                     │   
   │    [*] LDAP Authentication                                                                                                                                                     │   
   │    [*] Register user sessions in the systemd control group hierarchy                                                                                                           │   
   │    [*] Inheritable Capabilities Management                                
```

It might be something very bionic specific - I am still looking, but meanwhile can you confirm if this is working on any other recent distribution?

drybjed:
I tested this on Debian Stretch, and seems to be working as expected, `mkhomedir` in the dialog options is enabled, and I can see the correct entry in `/etc/pam.d/common-session` configuration file.

mikenowak:
Thanks for checking @drybjed , and I believe the behaviour has been altered slightly in bionic - i.e. I can see an addition of the following switch to the **pam-auth-update**:

```
       --enable profile [profile...]
              Enable the specified profiles in system configuration. This is used to enable profiles that are not on by default.
```

I've just checked an older ubuntu 16.04 machine and that switch does not exist there.

Now, testing this:

```
[root@host1:~]# grep -ri mkhomedir /etc/pam.d
[root@host1:~]# pam-auth-update  --enable mkhomedir
[root@host1:~]# grep -ri mkhomedir /etc/pam.d
/etc/pam.d/common-session-noninteractive:session	required                        pam_mkhomedir.so umask=0077 skel=/etc/skel
[...]

PR #531, opened 2018-11-14
Adding support for enabling pam_mkhomdir on bionic/buster

mikenowak:
Relates to #483.

@drybjed, would this work for you?

Let me know if anything needs changing.


drybjed (approved):
Sounds good, thanks. This will probably be refined later when it's moved out from `debops.auth` to a separate role, but should work fine for now.

PR #533, opened 2018-11-14
 Fix docker.service-j2 systemd template

rei-ifesca:
Fix docker.service-j2 systemd template. Between docker docker version 18.06.1-ce and 18.09.0 the systemd unit is not more dependent on 'docer.sock' unit. Therefore update the template with the latest example systemd unit from docker-ce debian deb pkg.

Use the unix socket and not the file descriptor. Otherwise the systemd unit won't work...

### DIFF
--- a/ansible/roles/debops.docker/templates/etc/systemd/system/docker.service.j2
+++ b/ansible/roles/debops.docker/templates/etc/systemd/system/docker.service.j2
@@ -3,9 +3,9 @@
 [Unit]
 Description=Docker Application Container Engine
 Documentation=https://docs.docker.com
-After=network-online.target docker.socket firewalld.service
+BindsTo=containerd.service
+After=network-online.target firewalld.service
 Wants=network-online.target
-Requires=docker.socket
 
 [Service]
 Type=notify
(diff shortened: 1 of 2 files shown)


drybjed:
@rei-ifesca Will this work with older versions of Docker (reasonably, of course), or should that `fd:///` presence/absence be conditional?

rei-ifesca:
AFAIK it will work, however it would be useful if other debops users could test if it works.

How would you check which Host value should be set? With the docker-ce version or the systemd version?

rei-ifesca:
Fyi this problem occurred with the latest stable release of docker-ce using the upstream channel on a Debian stretch (9.5). Fresh system install on monday.

gardouille:
> Fyi this problem occurred with the latest stable release of docker-ce using the upstream channel on a Debian stretch (9.5). Fresh system install on monday.

Got the same issue with an upgrade on a Debian Stretch 9.6.

ansible/roles/debops.docker/templates/etc/systemd/system/docker.service.j2
```diff
@@ -3,41 +3,56 @@
...
-# restart the docker process if it exits prematurely
-Restart=on-failure
-StartLimitBurst=3
-StartLimitInterval=60s
 
 [Install]
 WantedBy=multi-user.target
+
+
+
```
drybjed: Can you remove these empty lines?

drybjed (changes requested):
I tested this in a local VM and without these changes `docker.service` doesn't even want to start, but starts with this PR applied. Good job! Although the lines at the end could go away... :-)

ansible/roles/debops.docker/templates/etc/systemd/system/docker.service.j2
rei-ifesca: wilco

drybjed (approved):

PR #535, opened 2018-11-14
[fix] don't create sudoers file if `debops.sudo` is not used

### LINKED ISSUE #516: debops.nginx seems to require sudo
I'm using the debops roles without DebOps, as described in the ”example playbook“ documentation sections.

I've been executing `debops.gitlab`, and ansible failed at this `debops.nginx` task:

https://github.com/debops/debops/blob/2806b5517dfa69ba8eb08ea1c49b5726c75209ca/ansible/roles/debops.nginx/tasks/main.yml#L126-L133

because `Destination directory /etc/sudoers.d does not exist`.

Sudo is not installed on my systems by default. Should `sudo` be added to

https://github.com/debops/debops/blob/2806b5517dfa69ba8eb08ea1c49b5726c75209ca/ansible/roles/debops.nginx/defaults/main.yml#L53

or should the task (_”Allow webadmins to control nginx system service using sudo“_) be skipped if `sudo` is not installed?

Possibly related: https://github.com/debops/debops/issues/165

myrdd:
fix #516

### DIFF
--- a/ansible/roles/debops.gitlab_runner/tasks/main.yml
+++ b/ansible/roles/debops.gitlab_runner/tasks/main.yml
@@ -210,7 +210,9 @@
     owner: 'root'
     group: 'root'
     mode: '0440'
-  when: gitlab_runner__vagrant_libvirt|bool
+  when: (ansible_local|d() and ansible_local.sudo|d() and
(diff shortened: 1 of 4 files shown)


drybjed (approved):
Nicely done! :+1:

PR #536, opened 2018-11-15
[fix] install acl package if used

### LINKED ISSUE #517: acl required but not installed (debops.gitlab, debops.users)
Some roles use the `acl` ansible module:

```
> rgrep -l 'acl:' ansible/roles/*/tasks/
ansible/roles/debops.gitlab/tasks/main.yml
ansible/roles/debops.users/tasks/users.yml
```

but ~`apt`~ `acl` is installed only in these roles:

* `debops.apt_install`
* `debops.pki`
* `debops.preseed`

IMHO this is an issue. Still, I could simply put `debops.apt_install` at the top of each playbook. How should this be solved? :)

myrdd:
fix #517 

I'm not sure about `debops.users` — whether this is the way to be implemented, and if it's implemented correctly. (I did not test the changes.) Also note that I've put the installation task _before_ `DebOps pre_tasks hook`.


drybjed (approved):
Implementation in `debops.users` looks fine. Thanks!

PR #539, opened 2018-11-15
Add explicit --decrypt command to padlock-init

timitos:
It has been done to all other calls already.


drybjed (approved):
Sounds good, thanks.

PR #550, opened 2018-11-21
Improvements in 'debops.lxc', switch to systemd LXC instances

drybjed:


scibi:
@drybjed could you add a way to specify container-specific systemd options? Example use case: Stretch uses LXC 2.0. It doesn't support user limits (`ulimit`). Support is added in 2.1. In previous versions user limits are inherited form parent process so they could be specified in systemd service configuration. In my case I need to change `max locked memory` for Elasticsearch.

drybjed:
@scibi Sure, I can add that before merging this PR. Good idea.

PR #228, opened 2018-03-16
first attempt to deploy pre and up hooks for review

ser:
@drybjed please confirm i am on the right track

PR #553, opened 2018-11-23
540 evilham tinc up down

### LINKED ISSUE #540: debops.tinc: Improve tinc-up/tinc-down
Hello,

I just spent some fun time debugging networking issues :-).
(Yes, it was actually fun)

After (manually) playing quite a bit with `tinc-down` and `tinc-up`, I think they need a couple improvements:

* For starters, I guess nobody is using the (undocumented) `custom_tinc_up`. I guess that, because there is no equivalent `custom_tinc_down` but instead there is again `custom_tinc_up`. This one I think should be pretty clear: changing `tinc-down.j2` and documenting both `custom_tinc_(up|down)`. What would you think?
* Another one would be that I am missing some kind of post-what-debops-helps-me-set-up hook/variable. Something like `post_tinc_up`. I know I could use `custom_tinc_up`, but that would pretty much mean copy pasting the rest of the script and just adding the one line.
* The above probably implies that some kind of `pre_tinc_down` would be nice, for symmetry's sake.

Background: I just spent quite a bit trying to get IPv6 connectivity on an IPv4-only environment. Turns out, `tap` devices appear to have `accept_ra` set to `0`, so advertisements are ignored :-D.

Basically, I'd like to be able to add a `sysctl net.ipv6.conf.${INTERFACE}.accept_ra=1` line after everything else.

(As usual, just asking for input and would gladly implement myself!)

evilham:
This fixes #540

### DIFF
--- a/ansible/roles/debops.tinc/templates/etc/tinc/network/tinc-up.j2
+++ b/ansible/roles/debops.tinc/templates/etc/tinc/network/tinc-up.j2
@@ -3,6 +3,7 @@
 # {{ ansible_managed }}
 
 custom_tinc_up="{{ item.value.tinc_up|d() }}"
+post_tinc_up="{{ item.value.post_up|d() }}"
 
 tinc_mode="{{ item.value.tinc_options.Mode|d() }}"
 
@@ -11,6 +12,14 @@ hwaddr_prefix="{{ item.value.hwaddr_prefix | d(tinc__hwaddr_prefix) }}"
 
 link_type="{{ item.value.link_type|d() }}"
 link_metric="{{ item.value.metric|d() }}"
+
+{% set tinc__tpl_accept_ra = '1'
+       if item.value.accept_ra|d(True) == True
+       else item.value.accept_ra  %}
+accept_ra="{{ tinc__tpl_accept_ra
+              if tinc__tpl_accept_ra in ['0', '1', '2']
+              else '' }}"
+
 
 readarray -t interface_addresses << EOF
 {% set tinc__tpl_addresses = [] %}
@@ -103,5 +112,13 @@ else
                 "${INTERFACE}" &
         fi
 
+    fi
+
+    if [ -z "${accept_ra}" ]
+        sysctl net.ipv6.conf.${INTERFACE}.accept_ra=${accept_ra}
+    fi
+
+    if [ -z "${post_tinc_up}" ]
+        eval "${post_tinc_up}"
     fi
 fi
--- a/ansible/roles/debops.tinc/templates/etc/tinc/network/tinc-up.j2
+++ b/ansible/roles/debops.tinc/templates/etc/tinc/network/tinc-up.j2
@@ -13,12 +13,14 @@ hwaddr_prefix="{{ item.value.hwaddr_prefix | d(tinc__hwaddr_prefix) }}"
 link_type="{{ item.value.link_type|d() }}"
 link_metric="{{ item.value.metric|d() }}"
 
+read -r accept_ra << EOF
 {% set tinc__tpl_accept_ra = '1'
        if item.value.accept_ra|d(True) == True
        else item.value.accept_ra  %}
-accept_ra="{{ tinc__tpl_accept_ra
-              if tinc__tpl_accept_ra in ['0', '1', '2']
-              else '' }}"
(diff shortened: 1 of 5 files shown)


ansible/roles/debops.tinc/templates/etc/tinc/network/tinc-up.j2
```diff
@@ -12,6 +13,14 @@ hwaddr_prefix="{{ item.value.hwaddr_prefix | d(tinc__hwaddr_prefix) }}"
 link_type="{{ item.value.link_type|d() }}"
 link_metric="{{ item.value.metric|d() }}"
 
+{% set tinc__tpl_accept_ra = '1'
+       if item.value.accept_ra|d(True) == True
```
drybjed, line 18: You cannot use Jinja directly in shell scripts, they are checked using `shellcheck` and that breaks the test. Check a bit below to see how you can wrap Jinja in a variable.

docs/ansible/roles/debops.tinc/examples.rst
```diff
@@ -0,0 +1,60 @@
...
+---------------------
+
+Obtain IPv6 connectivity while in an IPv4-only network.
+
+This assumes a ``[tinc_dualstack]`` group with IPv4-accessible hosts and a
+``[tinc_ipv4only]`` group for hosts that need a tunnel.
+
+Hosts in ``[tinc_dualstack]`` should have the `tap` interface bridged via
+`br0` to an interfaces that is receiving router advertisements.
+See `debops.ifupdown` for information on how to do that and `debops.radvd`
```
drybjed: You could refer to the roles directly in the RST documentation, like this:
```
:ref:`debops.ifupdown`
:ref:`debops.radvd`
```

drybjed (changes requested):
Nicely done, although you should fix the script issue, you could also use proper RST references to other roles.

Try running `make test` locally to check various things before pushing them to GitHub, this should save you from multiple commits to check against Travis.

ansible/roles/debops.tinc/templates/etc/tinc/network/tinc-up.j2
evilham: OH! That's what that's for. Thank you!

evilham:
I think this could also close #228?
That talks about pre-up, but I'd think pre-up and post-down don't make much sense for tinc (and can be easily added later if really needed).

ansible/roles/debops.tinc/templates/etc/tinc/network/tinc-up.j2
```diff
@@ -115,7 +117,7 @@ else
     fi
 
     if [ -n "${accept_ra}" ]; then
-        sysctl net.ipv6.conf.${INTERFACE}.accept_ra=${accept_ra}
+        echo -n "${accept_ra}" > "/proc/sys/net/ipv6/conf/${INTERFACE}/accept_ra"
```
drybjed: Why this change? I think that using `sysctl` command for this instead of `echo`ing the value directly is better by using the "official" API. Just quote the `net.ipv6...$${accept_ra}` and you should be good.

I assume that the value is removed automatically by the kernel when the interface goes out?

evilham: I did it to please `shellchecker`, which was complaining about:

    sysctl "net..accept_ra"="${accept_ra}"

But... it didn't occur to me that I could just quote the whole thing! Thank you.

Since the Kernel maintains the `proc` FS, the value is removed when the interface is gone, yes.

drybjed (approved):
Nicely done, thanks!

ser:
Amazing job, @evilham - really thanks for digging into it. Now I do not have to use workarounds related to not typical internal network structure I have.

evilham:
Thanks @ser, if this doesn't resolve what you were trying to solve with #228:
I'd be curious to know why you need pre-up and just reply over there and I'll integrate it in the same fashion as this PR.

ser:
Hi, my problem with tinc was related to netplan.io which is a total rubbish. It is not able to bring up a bridge interface and I had to make a workaround to do it via tinc startup. But after netplan failed me several times I decided to remove it from my servers for good and workaround shortly won't be needed.

PR #554, opened 2018-11-23
Check that shells exist before changing them.

evilham:
If *someone* wants to, say, change the default shell to [fish](https://fish.sh) and forgets computers are computers and will do what you tell them instead of what you want, saying something like:

    users__accounts:
      - name: *someone*
        shell: 'fish'

Will result in that entry in `/etc/passwd` and that means that no login is possible.

This PR checks shell existence for both `debops.root_account` and `debops.users` roles.

### DIFF
--- a/ansible/roles/debops.root_account/tasks/main.yml
+++ b/ansible/roles/debops.root_account/tasks/main.yml
@@ -15,6 +15,16 @@
   changed_when: False
   check_mode: False
 
+- name: Check if preferred shell exists
+  stat: path="{{ root_account__shell }}"
+  register: root_account__register_shell
+  when: root_account__enabled|bool and root_account__shell|d(False)
+
+- name: Fail if setting a shell that does not exist
+  fail:
+    msg: "Trying to set a shell that does not exist, this can lock you out!"
+  when: root_account__enabled|bool and root_account__shell|d(False) and not root_account__register_shell.stat.exists
+
 - name: Enforce root system group
   group:
     name: 'root'
--- a/ansible/roles/debops.users/tasks/users.yml
+++ b/ansible/roles/debops.users/tasks/users.yml
@@ -24,38 +24,20 @@
   getent:
     database: 'group'
 
-- name: Get preferred shells
-  set_fact:
-    users__register_shell_path: '{{ item.shell|d() }}'
-  with_flattened:
-    - '{{ users__default_accounts }}'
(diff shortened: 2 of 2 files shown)


drybjed:
OK, but since *someone* now remembers that this is how the role works, he/she will not make this mistake again. Right?

I agree that this is useful feature for the `debops.root_account` role, since if someone does lock himself/herself out on their normal UNIX account, they could use the `root` account to fix that. But I don't see this really happening all that often with normal user accounts. I don't think that changes to `debops.users` are needed. Do you agree?

evilham:
At first I thought the same, but no, these changes are needed in both places for this shoot-yourself-in-the-foot mistake to be avoided.

E.g. this also changes root account:
https://github.com/debops/debops/blob/master/ansible/roles/debops.users/tasks/dotfiles.yml#L49

drybjed:
In that case, probably better idea is to move the `root` dotfiles installation to the `debops.root_account` role so that the two are decoupled properly.

evilham:
Mmm, the problem goes a bit deeper then, because `resources.yml` and `sshkeys.yml` also affect the `root` account in the `debops.users` role.

It'd be weird to duplicate all that logic in `debops.root_account` which is probably why `users__root_accounts` exists.

Not sure what the best thing would be :-) also I dont' see the harm in checking for existing shells for users; I mean, shouldn't a user that does not have a shell have `/usr/sbin/nologin` as shell anyway?

drybjed:
Hmm, very well. However, I would do this a bit differently, saving some round trips:
```yaml
- name: Check if defined shells exist
  stat:
    path: "{{ item }}"
  loop: '{{ ((users__default_accounts + users__admin_accounts
             + users__accounts + users__group_accounts + users__host_accounts
              + users__dependent_accounts) | selectattr("shell", "defined")
              | map(attribute="shell") | unique | list)
             + ([ users__default_shell ] if users__default_shell|d() else []) }}'
  register: users__register_shell_stats

- name: Fail if a defined shell does not exist
  fail:
    msg: "Trying to set a shell that does not exist, this can lock you out!"
  loop: '{{ users__register_shell_stats.results }}'
  when: not item.stat.exists
```
What do you think?

evilham:
That is so much better and it works well :-), thank you!

I only added `users__root_accounts` to the lists that are checked.

drybjed (approved):

PR #558, opened 2018-11-27
Fix problem while cloning over ssh. 

rei-ifesca:
Gitlab was not able to add the ssh keys of the users in the ~/.ssh directory.
Therefore add task to set the correct directory permission for ~/.ssh.

### DIFF
--- a/ansible/roles/debops.gitlab/tasks/main.yml
+++ b/ansible/roles/debops.gitlab/tasks/main.yml
@@ -45,6 +45,14 @@
     state: 'present'
   when: gitlab_support_filesystem_acl|bool
 
+- name: Allow GitLab user to the ~/.ssh directory
+  file:
+    path: '{{ gitlab_home + "/.ssh" }}'
+    owner: '{{ gitlab_user }}'
+    group: '{{ gitlab_group }}'
+    mode: 0644
+    recurse: true
+
 - name: Configure ~/.gitconfig
   template:
     src: 'var/local/git/gitconfig.j2'
--- a/ansible/roles/debops.gitlab/tasks/main.yml
+++ b/ansible/roles/debops.gitlab/tasks/main.yml
@@ -50,7 +50,7 @@
     path: '{{ gitlab_home + "/.ssh" }}'
     owner: '{{ gitlab_user }}'
     group: '{{ gitlab_group }}'
-    mode: 0700
+    mode: '0700'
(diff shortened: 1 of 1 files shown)


ansible/roles/debops.gitlab/tasks/main.yml
```diff
@@ -45,6 +45,14 @@
     state: 'present'
   when: gitlab_support_filesystem_acl|bool
 
+- name: Allow GitLab user to the ~/.ssh directory
+  file:
+    path: '{{ gitlab_home + "/.ssh" }}'
+    owner: '{{ gitlab_user }}'
+    group: '{{ gitlab_group }}'
+    mode: 0644
+    recurse: true
```
drybjed: I see this issue crop up multiple times now, I wonder when `~/.ssh/` is created by user other than `git` during GitLab deployment...

I'm not sure if `recurse: true` is needed here, are the files inside of this directory not owned by `git` user?

The mode is completely bonkers, this has to be `0700`, otherwise OpenSSH server will refuse to work. Also, you should quote the mode, otherwise it will be interpreted as octal number.

drybjed (changes requested):

ansible/roles/debops.gitlab/tasks/main.yml
```diff
@@ -45,6 +45,13 @@
     state: 'present'
   when: gitlab_support_filesystem_acl|bool
 
+- name: Allow GitLab user to the ~/.ssh directory
+  file:
+    path: '{{ gitlab_home + "/.ssh" }}'
+    owner: '{{ gitlab_user }}'
+    group: '{{ gitlab_group }}'
+    mode: '0700'
```
drybjed, line 54: You also need to add to this task:
```
state: 'directory'
```
Otherwise if the directory does not exist, Ansible will exit with an error.

ansible/roles/debops.gitlab/tasks/main.yml
rei-ifesca: Apparently too much copy paste.
You are right with the mode, nevertheless with the 0644 mode it still worked. That's because ssh only searches there for ``authorized_keys``, thus "ignoring" the wrong settings.

My intention with recurse was to change the owner:group of any file inside the dir, if any.

drybjed (approved):

PR #566, opened 2018-12-04
[debops.owncloud] Fix memcache Redis configuration for owncloud 10

pedroluislopez:
See: https://doc.owncloud.org/server/10.0/admin_manual/configuration/server/caching_configuration.html#redis-label

### DIFF
--- a/ansible/roles/debops.owncloud/defaults/main.yml
+++ b/ansible/roles/debops.owncloud/defaults/main.yml
@@ -815,8 +815,8 @@ owncloud__role_config:
   'updatechecker': '{{ True if (owncloud__variant in ["nextcloud"]) else False }}'
 
   'memcache.local':
-    state: '{{ "present" if (owncloud__apcu_enabled|bool) else "absent" }}'
-    value: '\\OC\\Memcache\\APCu'
+    state: 'present'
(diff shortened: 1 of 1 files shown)


ansible/roles/debops.owncloud/defaults/main.yml
```diff
@@ -815,8 +815,8 @@ owncloud__role_config:
   'updatechecker': '{{ True if (owncloud__variant in ["nextcloud"]) else False }}'
 
   'memcache.local':
-    state: '{{ "present" if (owncloud__apcu_enabled|bool) else "absent" }}'
-    value: '\\OC\\Memcache\\APCu'
+    state: 'present'
+    value: '{{ "\\OC\\Memcache\\Redis" if (owncloud__redis_enabled|bool) else "\\OC\\Memcache\\APCu" }}'
```
drybjed, line 819: Sounds good, however you still want to disable it if `owncloud__apcu_enabled` is set to `False`, so something like this:
```yaml
'memcache.local':
  state: '{{ "present" if (owncloud__apcu_enabled|bool or owncloud__redis_enabled|bool) else "absent" }}'
```

drybjed (approved):

PR #568, opened 2018-12-04
debops.netbase: Fix /etc/hosts generation for IPs with single hostname

scibi:


drybjed (approved):
Thanks, not sure how I missed that. :+1:

PR #569, opened 2018-12-04
debops.ifupdown: change escaping in iface service to allow interfaces names with minus sign

scibi:


drybjed (approved):
Thanks, looks good.

drybjed:
NB: this issue exists in the `ifup@.service` unit [included in Debian Stretch](https://sources.debian.org/src/ifupdown/0.8.19/debian/ifup%40.service/#L14), and [doesn't seem to be fixed in Debian Sid](https://sources.debian.org/src/ifupdown/0.8.34/debian/ifup%40.service/#L14) either. Perhaps this should be brought to the `ifupdown` package maintainers for evaluation?

PR #572, opened 2018-12-06
[debops.apt] Allow sources to be overridden in inventory.

tumd:
Before this, there was no good way to override apt__security_sources
and apt__default_sources, without copying the entire variables to your
inventory.
This may be beneficial when you want to use a local mirror instead
of the public mirrors.

Example inventory configuration;
```yaml
apt__original_sources: []

apt__group_security_sources:
  - uri:          'http://deb.domain.fqdn/debian-security'
    comment:      '{{ "Local " + apt__distribution + " Security repository" }}'
    type:         '{{ apt__source_types }}'
    suite:        '{{ apt__distribution_release + "/updates" }}'
    components:   '{{ apt__distribution_components }}'
    distribution: 'Debian'
    state:        'present'
  - uri:          'http://deb.domain.fqdn/ubuntu-security'
    comment:      '{{ "Local " + apt__distribution + " Security repository" }}'
    type:         '{{ apt__source_types }}'
    suite:        '{{ apt__distribution_release + "-security" }}'
    components:   '{{ apt__distribution_components }}'
    distribution: 'Ubuntu'
    state:        'present'

  - uris:
      - 'https://deb.domain.fqdn/debian-security'
      - 'http://deb.domain.fqdn/debian-security/'
      - 'https://deb.domain.fqdn/debian-security/'
    state: "absent"

apt__group_sources:
  - uri:          'http://deb.domain.fqdn/debian'
    comment:      '{{ "Local " + apt__distribution + " repositories" }}'
    distribution: 'Debian'
  - uri:          'http://deb.domain.fqdn/ubuntu'
    comment:      '{{ "not-so-Official " + apt__distribution + " repositories" }}'
    distribution: 'Ubuntu'
```

### DIFF
--- a/ansible/roles/debops.apt/defaults/main.yml
+++ b/ansible/roles/debops.apt/defaults/main.yml
@@ -356,6 +356,19 @@ apt__original_sources:
                          else "absent" }}'
 
                                                                    # ]]]
+# .. envvar:: apt__default_sources_present [[[
+#
+# Define if the default APT package sources for specific distributions should
+# be used or not.
+# If package sources are defined in specific group, host or on all hosts in
+# Ansible inventory, the default sources will be disabled.
+apt__default_sources_present: '{{ "True"
+                                  if ( not apt__group_sources and
+                                       not apt__host_sources and
+                                       not apt__sources )
+                                    else "False" }}'
+
+                                                                   # ]]]
 # .. envvar:: apt__default_sources [[[
 #
 # List of the default APT package sources for specific distributions. These are
@@ -406,6 +419,32 @@ apt__default_sources:
     distribution: 'Devuan'
 
                                                                    # ]]]
+# .. envvar:: apt__security_sources_state [[[
+#
+# Enable (if ``present``) or disable (if ``absent``) the default APT
+# repositories proving security updates.
+# If a local security mirror is defined in Ansible inventory group or host,
+# this will disable the default security repositories.
+apt__security_sources_state: '{{ "present"
+                                  if ( not apt__group_security_sources and
+                                       not apt__host_security_sources )
+                                  else "absent" }}'
+
+                                                                   # ]]]
+# .. envvar:: apt__group_security_sources [[[
+#
+# List of APT repositories that provide security updates for hosts in
+# specific Ansible inventory group.
+apt__group_security_sources: []
+
+                                                                   # ]]]
+# .. envvar:: apt__host_security_sources [[[
+#
+# List of APT repositories that provide security updates for hosts in
+# specific Ansible inventory host.
+apt__host_security_sources: []
+
+                                                                   # ]]]
 # .. envvar:: apt__security_sources [[[
 #
 # List of APT package sources that provide security updates. This list is used
@@ -419,7 +458,7 @@ apt__security_sources:
     suite:      '{{ apt__distribution_release + "/updates" }}'
     components: '{{ apt__distribution_components }}'
     distribution: 'Debian'
-    state:      '{{ "present"
+    state:      '{{ apt__security_sources_state
                     if (apt__distribution_release|d("unknown")
                         in apt__distribution_release_map[apt__distribution]|d([]))
                     else "absent" }}'
@@ -451,7 +490,7 @@ apt__security_sources:
     suite:      '{{ apt__distribution_release + "-security" }}'
     components: '{{ apt__distribution_components }}'
     distribution: 'Ubuntu'
-    state:      '{{ "present"
+    state:      '{{ apt__security_sources_state
                     if (apt__architecture in ["amd64", "i386"])
                     else "absent" }}'
 
@@ -461,7 +500,7 @@ apt__security_sources:
     suite:      '{{ apt__distribution_release + "-security" }}'
     components: '{{ apt__distribution_components }}'
     distribution: 'Ubuntu'
-    state:      '{{ "present"
+    state:      '{{ apt__security_sources_state
                     if (apt__architecture not in ["amd64", "i386"])
                     else "absent" }}'
 
@@ -471,7 +510,7 @@ apt__security_sources:
     suite:      '{{ apt__distribution_release + "-security" }}'
     components: '{{ apt__distribution_components }}'
     distribution: 'Devuan'
-    state:      '{{ "present"
+    state:      '{{ apt__security_sources_state
                     if (apt__distribution_release|d("unknown")
                         in apt__distribution_release_map[apt__distribution]|d([]))
                     else "absent" }}'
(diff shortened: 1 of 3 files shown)


ansible/roles/debops.apt/defaults/main.yml
```diff
@@ -355,6 +355,19 @@ apt__original_sources:
...
+#
+# Define if the default APT package sources for specific distributions should
+# be used or not.
+# If package sources are defined in specific group, host or on all hosts in
+# Ansible inventory, the default sources will be disabled.
+apt__default_sources_present: '{{ "True"
+                                  if ( not apt__group_sources and
+                                       not apt__host_sources and
+                                       not apt__sources )
+                                    else "False" }}'
```
drybjed: A common practice in DebOps roles is to provide a set of default configuration for a role, as well as allow for the Ansible inventory variables or optionally role dependent variables to modify or override the defaults. However I try to avoid a situation that setting up inventory variables automatically disables entire default configuration variables provided by the role, which is the case here.

While testing these changes I noticed that the default APT sources were unexpectedly removed from the configuration - it was due to my inventory configuration which added a local Debian mirror as an APT source repository in addition to the `deb.debian.org` mirror. Because APT supports multiple mirrors at once, I think that adding the repositories via inventory variables in addition to the default, instead of essentially replacing the defaults with inventory variables, is a better action to take without an explicit decision by the user.

I would change this variable to `apt__default_sources_state`, `present` and `absent` as available states, `present` by default.  This variable would allow for easy deactivation of default sources when the user desires it, and it has similar function to `apt__security_sources_state` variable.

ansible/roles/debops.apt/defaults/main.yml
```diff
@@ -405,6 +418,32 @@ apt__default_sources:
...
+# .. envvar:: apt__security_sources_state [[[
+#
+# Enable (if ``present``) or disable (if ``absent``) the default APT
+# repositories proving security updates.
+# If a local security mirror is defined in Ansible inventory group or host,
+# this will disable the default security repositories.
+apt__security_sources_state: '{{ "present"
+                                  if ( not apt__group_security_sources and
+                                       not apt__host_security_sources )
+                                  else "absent" }}'
```
drybjed: This sounds good to me, however I wonder how to justify automatic disabling the default security sources when group or host security sources are specified, as opposed to not doing it with normal sources using `apt__default_sources_present` variable.

Traditionally Debian [provides one central security repository](https://unix.stackexchange.com/questions/4255/why-is-there-a-separate-package-repository-for-debian-security-updates) which justifies the split in the `debops.apt` role between normal sources and security sources. If you want to use your own local mirror of `security.debian.org`, I would presume that happens due to proxy requirements, or that you want to vet the updated packages before applying them in production. In that case disabling the default security sources automatically makes sense and would be expected, unlike the normal APT sources.

ansible/roles/debops.apt/defaults/main.yml
```diff
@@ -419,7 +458,7 @@ apt__security_sources:
     suite:      '{{ apt__distribution_release + "/updates" }}'
     components: '{{ apt__distribution_components }}'
     distribution: 'Debian'
-    state:      '{{ "present"
+    state:      '{{ apt__security_sources_state
```
drybjed: I don't really like that the state of the repository sources is now determined in two places instead of one (the `item.state` parameter and `apt__security_sources_state` default variable). But with current design of the `debops.apt` role I don't see a better way to handle this. It might not even be possible to do it differently when we take into account that the list has to be differentiated between different OS distributions and releases.

You don't need to change this - it's just a comment on state of the role in general.

ansible/roles/debops.apt/defaults/main.yml
```diff
@@ -485,8 +524,10 @@ apt__security_sources:
 apt__combined_sources: '{{ apt__sources +
                            apt__group_sources +
                            apt__host_sources +
-                           apt__original_sources +
-                           apt__default_sources +
+                           apt__original_sources }} +
+                           {{ apt__default_sources if apt__default_sources_present else [] }} +
+                           {{ apt__host_security_sources +
+                           apt__group_security_sources +
```
drybjed, line 526: This is an unexpected syntax. The entire value should be enclosed in one pair of `{{ }}` parentheses, basically something like this:
```
apt__combined_sources: '{{ apt__sources +
                           ...
                           apt__original_sources +
                           (apt__default_sources if (apt__default_sources_present|bool) else []) +
                           apt__host_security_sources +
                           ...
                           apt__security_sources }}'
```
Additionally, why didn't you set the status of APT default sources similarly to how you defined it in APT security sources? Both methods seem feasible to do the same thing, there keeping both configurations the same would make things consistent.

ansible/roles/debops.apt/templates/etc/ansible/facts.d/apt.fact.j2
```diff
@@ -15,7 +15,9 @@ import os
 {% set apt__tpl_default_sources = [] %}
 {% set apt__tpl_default_sources_map = {} %}
 {% set apt__tpl_source_distributions = {} %}
-{% for repo in apt__security_sources %}
+{% for repo in ( apt__host_security_sources,
+                 apt__group_security_sources,
+                 apt__security_sources ) %}
```
drybjed, line 20: With this change, `apt__tpl_security_sources` variable in the fact script template ends up being empty. That's because now the `for repo` loop operates on list of lists. You should add the lists together instead of making a set of nested lists here - `(apt__host_security_sources + apt__group_security_sources + apt__security_sources)`.

drybjed (changes requested):
Thanks for the PR! These are some good ideas, but I see a need for small refactoring, can you check the comments to specific parts of the commit?

Also, good explanation in the commit message, however I think that it should be added to the `debops.apt` documentation as well, for better discoverability by the users. Probably the "Getting Started" section of the role documentation is a good place for it.

ansible/roles/debops.apt/defaults/main.yml
tumd: Ah, that's the syntax I was after. Thanks!

The reason I put this logic here in the first place was because I didn't want to mess with the local fact `default_sources_map`. I noticed this local fact is used in debops.apt_cacher_ng, and should always contain the  mirrors from `apt__default_sources`.
Allowing change of the 'state' of the items in `apt__default_sources` to `absent` would cause `ansible_local.apt.default_sources_map` to be empty. Which probably should be avoided.

ansible/roles/debops.apt/defaults/main.yml
tumd: Yes, I agree. It's probably a disservice putting that logic there. 

However, I'm not sure I agree regarding changing the state-values from bool to `present`/`absent`, because it don't actually change the '`item.state`'. See https://github.com/debops/debops/pull/572#discussion_r240175949.

drybjed: I was aiming more for similarity with `apt__security_sources_state`, but if you think the functionality is different, fair enough. In that case, I would suggest a rename to `apt__default_sources_enabled` which, I think, more adequately describes this variable purpose, `_enabled` is also consistenly used in other DebOps roles for boolean variables that enable/disable parts of the role functionality without changing the state on the managed hosts.

tumd:
Not sure I got the rebase right, but I think so.
Let me know if  you find any issues! :)

drybjed (approved):

PR #575, opened 2018-12-10
Fix Py3 type exception.

cultcom:
a minor fix for type mismatch "str" vs. "bytes".


drybjed (approved):
Thanks for the fix! :+1:

PR #578, opened 2018-12-14
Add default config parameters and doc for the nginx_http_server_names_hash

rei-ifesca:


rei-ifesca:
We encountered a problem regarding the server_name_hash and required to change this parameter in nginx. Thus having a documentation about this setting would be nice ;)

drybjed (approved):
Sounds good, thanks. The variables in `nginx.conf` template should really be exposed in the defaults, but `debops.nginx` itself requires a through cleanup and restructuring. I will get there eventually. :-)

Issue #580, opened 2018-12-27
[debops.netbox] migrating to python3


mikenowak:
As per the https://netbox.readthedocs.io/en/latest/installation/migrating-to-python3/, starting with netbox v2.5, netbox no longer supports python 2.

This prevents netbox from starting at the moment since we default to `netbox__git_version: 'master'`

I don't really have the capacity to work on this at the moment, but in case somebody needs to revert to a working version, the following needs to be adjusted as follows:

```
netbox__git_version: 'v2.4.9'
```

drybjed:
@mikenowak Thanks for the bug report. I'm aware of the issue with NetBox, hadn't had an opportunity to test some kind of transition yet, but I'm planning on doing so.

I'm not sure how to approach this yet. Probably going "straight" and just switching to Py3 `virtualenv` in the role would leave existing installations broken and require a clean reinstall of NetBox, so I don't really want to do that. Role would probably have to check if existing py2 `virtualenv` is set up and then do migration to py3 `virtualenv`. I'm not sure yet if existing virtual environments can be upgraded, or you should just create a new `virtualenv` with py3 and install NetBox there as usual - that probably could work, depending on the role checking if an existing database is set up and not clobbering it.

In the meantime, perhaps DebOps should switch to the NetBox version you specified. Then I can release a new DebOps v0.8.1 which I'm planning to do before New Years hits, and after that we can research some kind of migration steps. What do you think about this? Want to do a PR to change the default version of NetBox for now?

mikenowak:
hi @drybjed, I totally agree and I am also worried about leaving garbage behind as part of this upgrade.

In the meantime, please see the PR as requested.

mikenowak:
my bad, wasnt meant to close this one.

drybjed:
@mikenowak Can you check if recent changes in `debops.netbox` solve this issue for you?

mikenowak:
While I acknowledge that the new virtualenv has been installed, and old removed, it quite did not solve the issue. 

```
Jan 09 01:42:24 web1 systemd[1]: Started Green Unicorn - Python application server for netbox.
Jan 09 01:42:24 web1 gunicorn@netbox[2761]: Traceback (most recent call last):
Jan 09 01:42:24 web1 gunicorn@netbox[2761]:   File "/usr/bin/gunicorn3", line 6, in <module>
Jan 09 01:42:24 web1 gunicorn@netbox[2761]:     from pkg_resources import load_entry_point
[...]

PR #585, opened 2018-12-29
[debops.netbox] Create new Python 3 virtualenv

tumd:
Netbox v2.5 and newer no longer supports Python 2.
This creates a new vituealenv directory based on Python 3 when
`netbox__git_version` is v2.5+ or master/develop.
This *should* make a clean upgrade of existing Netbox installations.

I'm not sure this is the best way to handle this kind of situation. I'm not that 'at home' with python virtualenvs, and couldn't find a way to upgrade the existing virtualenv to python 3.
This also most likely needs a bit more testing.

### DIFF
--- a/ansible/roles/debops.netbox/defaults/main.yml
+++ b/ansible/roles/debops.netbox/defaults/main.yml
@@ -40,15 +40,15 @@ netbox__domain: '{{ ansible_local.core.domain
 netbox__base_packages:
   - 'git'
   - 'build-essential'
-  - 'python-dev'
-  - 'python-pip'
+  - 'python{{ netbox__virtualenv_version }}-dev'
+  - 'python{{ netbox__virtualenv_version }}-pip'
   - 'libxml2-dev'
   - 'libxslt1-dev'
   - 'libffi-dev'
   - 'graphviz'
   - 'libpq-dev'
   - 'libssl-dev'
-  - '{{ "python-virtualenv"
+  - '{{ "python{{ netbox__virtualenv_version }}-virtualenv"
         if (ansible_distribution_release in [ "wheezy", "precise" ])
         else "virtualenv" }}'
 
--- a/ansible/roles/debops.netbox/defaults/main.yml
+++ b/ansible/roles/debops.netbox/defaults/main.yml
@@ -170,7 +170,7 @@ netbox__virtualenv_version: '{{ ""
 # .. envvar:: netbox__virtualenv [[[
 #
 # Path where the NetBox ``virtualenv`` directory will be stored.
-netbox__virtualenv: '{{ netbox__lib + "/virtualenv" + netbox__virtualenv_version }}'
+netbox__virtualenv: '{{ netbox__lib + "/virtualenv" }}'
 
                                                                    # ]]]
 # .. envvar:: netbox__virtualenv_env_path [[[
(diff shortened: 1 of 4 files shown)


drybjed:
@tumd thanks, it looks OK to me, although I'll test it a few times with existing and new NetBox installation before merging, to be sure that it works as expected.

tumd:
Don't merge.
This doesn't work in the current state.
Will have another look at it in the upcoming days.

tumd:
With the last commit I changed the behavior slightly. The old virtualenv will now be removed and recreated if it's outdated (i.e. python2 based).
I'm not entirely happy with how `netbox__virtualenv_version` is used, but I wasn't sure how else the version-change of packages and others should be handled.

Please try it out and let me know what you think.

ansible/roles/debops.netbox/defaults/main.yml
```diff
@@ -40,15 +40,15 @@ netbox__domain: '{{ ansible_local.core.domain
 netbox__base_packages:
   - 'git'
   - 'build-essential'
-  - 'python-dev'
-  - 'python-pip'
+  - 'python{{ netbox__virtualenv_version }}-dev'
+  - 'python{{ netbox__virtualenv_version }}-pip'
```
drybjed: The `netbox.yml` playbook already contains the `debops.python` role which is required by the `debops.gunicorn` role. You could move the `python*` APT packages to the `netbox__python__dependent_packages2` and `netbox__python__dependent_packages3`, so that the `debops.python` role itself installs the desired version (see how it's done in `debops.gunicorn` role for an example). You would probably only need to install `python-dev`/`python3-dev` packages since `pip` and `virtualenv` is already included in `debops.python` role by default.

drybjed (commented):
Changes look promising, but I haven't tested them yet, will do during the weekend.

I gave you a possible better solution for installing APT `python` packages via `debops.python` role. You could try implementing it if you want, or we can do that later after the PR is merged.

ansible/roles/debops.netbox/defaults/main.yml
tumd: This should now be implemented.

tumd:
The changes has been tested both from a clean install and by upgrading a existing netbox installed using the current debops/master (with sample-data only). Seems to be working.

The existing tasks handling db-migration and such seem to be working too.
However, netbox provides their own [upgrade.sh](https://github.com/digitalocean/netbox/blob/develop/upgrade.sh) to handle upgrades, which pretty much do the same thing as many of the tasks in this role. After a quick look in the git-history, it looks like the role was created before the upgrade.sh-script existed upstream.
Would it be a good idea to try to utilize this script instead of the tasks?

drybjed:
In the past, the `upgrade.sh` script used `sudo` commands internally, therefore it was unsuitable for installation managed by DebOps in a virtualenv. This has been fixed about a month ago when use of `sudo` inside of the script was dropped. So I guess now the `debops.netbox` role could be switched over to it after checking that everything works as expected. But, I would wait with this change a bit to let people update to newer NetBox version, and perhaps do it after new DebOps version is released.

drybjed (approved):
I tested these changes on an existing NetBox installation and upgrade went smoothly, everything seems to be in order. Nicely done! :+1:

tumd:
Sweet!
This probably also close #580

PR #587, opened 2019-01-01
change test dir for docker-entrypoint

ghost:
Allow a user to mount a persistent volume to `/home/ansible/src` without short circuiting the initial project creation built into docker entry point

this will allow a user to edit src files outside the container using their favorite IDE.

```
docker run -it --rm \
	-v "$(pwd)":/home/ansible/src \
	debops/debops
```


drybjed (approved):
Sounds good, thanks.

PR #593, opened 2019-01-06
fix PKGBUILD

Cadair:
This is just a simple fix to the build step of the arch package :)

I have copied this into the AUR as debops-git.


drybjed (approved):
Sounds good, thanks!

PR #596, opened 2019-01-08
[WIP] Adds TrustedUserCAKeys to debops.sshd template

davemcphee:
Hi all,

I'd like to create a PR to add TrustedUserCAKeys & HostCertificate config entries / variables to the debops.sshd role, but there's an issue I'm unsure of how to approach: who's responsibility is it to ensure these files exist? Should these cert files be deployed as static files, or perhaps should the variable `sshd__trusted_user_ca_key_files` be a list of strings, containing keys, which are then assembled into a single file by ... ? 

Any guidance much appreciated, and will be converted to code, where possible. 

Thanks!


drybjed:
I think that a YAML list variable which is then templated into the final file on the server should be fine - it's just one additional task in the role, could be activated conditionally when the list of keys is not empty. `sshd_trusted_user_ca_keys` sounds better, IMO, it doesn't imply that this is a list of files. You should specify the final file in a default variable, in case the users want to rename it.

davemcphee:
How does the TrustedUserCAKeys part look so far? keys in a list, assembled into a template, everything conditional on list length > 0 and `sshd__trusted_user_ca_keys_file` being defined; otherwise, no action. 

If the logic / ansible parts pass muster, anything else I should add? 

If this works, I'll add host certificates as well.

drybjed:
Yes, the changes look good. Although I don't really like the `pycharm` stuff in a separate commit, it will be included in DebOps monorepo that way, even as removed... Perhaps you could squash the commits together? That should do the trick.

davemcphee:
Will do - I'll probably close this PR and delete the branch, and start clean once I'm done. 

I originally wanted to add `HostCertificate` options to this role as well, but on planning this, decided it was too difficult. HostCerts are-per host, so we'd have to store 100's of them as variables. Or we could generate them, but that's prob way outside the domain of this role

What about a variable `sshd__scan_for_host_certs`? Default False, but if True, it looks for any files matching `/etc/ssh/ssh_host_*_key-cert.pub` on a host, and if found, adds them to sshd_config?

drybjed:
I think that you can only add one `HostCertificate` parameter in the configuration? Not sure what scanning for multiple key-certs would accomplish, unless you are talking about a different `sshd_config` parameter?

Creating a new clean PR sounds good.

davemcphee:
I think multiple `HostCertificate` are allowed, one per `HostKey` ? May you've been using rsa and want to transition seamlessly to ed25519, so you provide both? 

My thinking is that by scanning for available `HostCertificate` files, the sshd role offloads the provisioning of host certificates - if they're there, they're added to sshd_config. Who knows how they got there. Does this make sense, or not deterministic enough? 

I also added a check to make sure a found host certificate file matches a host key file, and that the key type matches what's defined in the `sshd__host_keys`. Eg.: you've disabled dsa, so let's not add a certificate for a dsa key to the config, even if one exists.

drybjed:
I see. You're right, `sshd_config` supports multiple instances of a parameter. In that case, OK, changes look good to me.

davemcphee:
Cool, closing this PR and starting fresh

PR #600, opened 2019-01-09
Fix various issues found by 'ansible-lint' in DebOps roles

drybjed:


ypid:
Give me a few days :)

drybjed:
@ypid Actually, I would want to finish fixing `ansible-lint` issues (317 lines too long left) and prepare next release of DebOps soon. This is so that on Galaxy all of the issues detected currently by the import process are fixed.

I tested these changes multiple times while cleaning up the code, and I'm confident that there are no changes vs the current code. I'll wait with merging this PR while cleaning up the long lines, it should take a day or two. Of course, `ansible-lint` will be then added to the test pipeline on Travis-CI.

drybjed:
@ypid All fixes for issues found by `ansible-lint` are done. I plan to merge this PR on Sunday. :-)

https://xkcd.com/1070/

ansible/roles/debops.ferm/tasks/main.yml
```diff
@@ -23,7 +25,7 @@
     state: 'directory'
     owner: 'root'
     group: 'adm'
-    mode: '2750'
```
ypid, line 26: This should probably have one `0` less :)

ansible/roles/debops.gitusers/tasks/git-shell.yml
```diff
@@ -37,19 +38,20 @@
...
 
 - name: Create root directory for user websites
   file:
-    path: '{{ gitusers_default_www_prefix }}/{{ item.0.group | default(item.0.name + gitusers_name_suffix) }}/{{ item.1 }}'
+    path: '{{ gitusers_default_www_prefix }}/{{ item.0.group | default(item.0.name
+                                                                       + gitusers_name_suffix) }}/{{ item.1 }}'
     state: 'directory'
     owner: 'root'
     group: '{{ item.0.group | default(item.0.name + gitusers_name_suffix) }}'
-    mode: '2775'
```
ypid, line 52: Same. I am guessing some `git ls-files -z "$(git rev-parse --show-toplevel)" | xargs --null -I '{}' find '{}' -type f -print0 | xargs --null sed --in-place --regexp-extended 's/mode: .*.../\1/g;'` magic :)

ansible/roles/debops.opendkim/tasks/main.yml
```diff
@@ -149,6 +152,6 @@
     state: 'directory'
     owner: '{{ opendkim__user }}'
     group: '{{ opendkim__postfix_group }}'
-    mode: '2750'
+    mode: '02750'
```
ypid, line 155: Same.

ansible/roles/debops.rails_deploy/tasks/deploy.yml
```diff
@@ -139,20 +150,25 @@
...
+  with_items: '{{ rails_deploy_post_restart_shell_commands }}'
+  become: True
   become_user: '{{ rails_deploy_service }}'
 
 - name: Disable the temporary deploy page
-  command: rm -f public/index.html chdir={{ rails_deploy_src }}
+  command: rm -f public/index.html
+  args:
+    chdir: '{{ rails_deploy_src }}'
+    warn: False
```
ypid: This can also be expressed as file: state=absent. Then you also don’t need the `warn: False` ;-)

ansible/roles/debops.rails_deploy/tasks/deploy_keys.yml
```diff
@@ -14,15 +14,20 @@
         rails_deploy_register_deploy_key is defined
 
 - name: Transfer the deploy key to Github
-  command: "curl --silent --header 'Authorization: token {{ rails_deploy_git_access_token }}' --data '{{ rails_deploy_key_data | to_nice_json }}' https://api.github.com/repos/{{ rails_deploy_git_account }}/{{ rails_deploy_git_repo }}/keys"
+  command: "curl --silent --header 'Authorization: token {{ rails_deploy_git_access_token }}'
+                 --data '{{ rails_deploy_key_data | to_nice_json }}'
+                 https://api.github.com/repos/{{ rails_deploy_git_account }}/{{ rails_deploy_git_repo }}/keys"
+  args:
+    warn: False
```
ypid, line 22: It should be possible to not depend on `curl` here and do it with a Ansible module.

ansible/roles/debops.rails_deploy/tasks/deploy_keys.yml
```diff
@@ -31,7 +36,12 @@
         not 'github' in rails_deploy_git_host
 
 - name: Transfer the deploy key to Gitlab
-  command: "curl --insecure --header 'PRIVATE-TOKEN: {{ rails_deploy_git_access_token }}' --data '{{ rails_deploy_key_data | to_nice_json }}' https://{{ rails_deploy_git_host }}/api/v3/projects/{{ rails_deploy_register_gitlab_response.json.id }}/keys"
+  command: "curl --insecure --header 'PRIVATE-TOKEN: {{ rails_deploy_git_access_token }}'
+                 --data '{{ rails_deploy_key_data | to_nice_json }}'
+                 https://{{ rails_deploy_git_host }}/api/v3/projects/{{
+                   rails_deploy_register_gitlab_response.json.id }}/keys"
+  args:
+    warn: False
```
ypid, line 46: It should be possible to not depend on `curl` here and do it with a Ansible module. Also the `--insecure` gives me nightmares.

ansible/roles/debops.sftpusers/tasks/sites.yml
```diff
@@ -20,7 +20,7 @@
     state: 'directory'
     owner: 'root'
     group: '{{ item.group | default(item.name + sftpusers_name_suffix) }}'
-    mode: '2775'
```
ypid, line 23: Please double check the mode parameter in this file.

ypid (changes requested):
LGTM. Thanks for your work! A few comments are attached.

ansible/roles/debops.ferm/tasks/main.yml
```diff
@@ -23,7 +25,7 @@
...
+    mode: '02750'
```
ypid, line 28: This should probably have one 0 less :)

ansible/roles/debops.rails_deploy/tasks/deploy.yml
drybjed: I know, but to fix `ansible-lint` issues in this PR I wanted to do as few changes to the existing code as possible, since it's a large set of patches. Tasks like this can be converted later to proper Ansible modules.

ansible/roles/debops.rails_deploy/tasks/deploy_keys.yml
drybjed: This is the same reason as with the `rm` and the `command` module - my aim here was to fix the `ansible-lint` error with minimum changes to the existing code.

ansible/roles/debops.rails_deploy/tasks/deploy_keys.yml
drybjed: Agreed, the `rails_deploy` role is unfortunately neglected, I'm not using it personally. Perhaps some kind of survey or poll to find what roles are popular among users could be interesting to do. Food for thought.

ansible/roles/debops.ferm/tasks/main.yml
drybjed: This is a fix for the E202 error in `ansible-lint`. The `file` Ansible module seems to work as expected with this leading zero.

ansible/roles/debops.sftpusers/tasks/sites.yml
drybjed: The value with leading zero works fine with the `file` Ansible module.

ansible/roles/debops.gitusers/tasks/git-shell.yml
drybjed: The leading zero in the value works fine with the `file` Ansible module.

ansible/roles/debops.opendkim/tasks/main.yml
drybjed: The leading zero in the value works fine with the `file` Ansible module.

drybjed:
@ypid OK, answered all your comments, waiting for response. :-)

ansible/roles/debops.rails_deploy/tasks/deploy.yml
ypid: Then maybe add a FIXME note to the files. My reasoning is that `ansible-lint` makes you aware of things that can be improved. When you do as little changes as possible if just feels like you hack the linter in ways that it will stop complying for example by disabling warnings while there is a proper way to do it. This is a trivial fix/change.

ansible/roles/debops.rails_deploy/tasks/deploy_keys.yml
ypid: Same as I explained with the `rm` and the `command` module except that I am more on your side here because this task is more complex than a simple file: state=absent. Maybe also add FIXME notes so that people working on this role the next time and test it can fix the task in one go.

ansible/roles/debops.rails_deploy/tasks/deploy_keys.yml
ypid: Fair enough. Maybe a FIXME note can be added for now as this really needs fixing.

ansible/roles/debops.ferm/tasks/main.yml
ypid: Hm, but is the leading 0 needed. It looks wired. In all the other places we use 4 digits. My point here is consistency in the code base.

ypid:
@drybjed Done that, sorry for the delay ;-)

ansible/roles/debops.ferm/tasks/main.yml
drybjed: In other cases with 4 digits the attributes are usually `0755` or `0644` which is "natural". In this case and similar, the role sets SGID attribute, but the notation, at least in Ansible, still hinges on `0` being the lead digit which signifies an octal number. Remember when in some cases an unquoted attribute mask resulted in a weird permissions because the number was interpreted wrong by the `file` module? I bet that unquoted number with `0` as a prefix would be interpreted correctly as an octal. Using `0` for octal numbers for me seems more consistent than using 4 digits for attribute values, at least when interpreting values correctly by the software.

ansible/roles/debops.rails_deploy/tasks/deploy_keys.yml
drybjed: The whole `debops.rails_deploy` role needs a major overhaul at some point, since lots of the other roles that it depends on changed over time. We will get to it eventually. :-)

drybjed:
@ypid I added the `FIXME` notes to `debops.rails_deploy` tasks, and used the `file` module instead of simple `rm` command.

ypid:
@drybjed Thanks! Can you comment on the file module and mode parameter discussion?

drybjed:
@ypid I did: https://github.com/debops/debops/pull/600#discussion_r247350928

ansible/roles/debops.ferm/tasks/main.yml
ypid: I like to use 4 digits for the mode to be explicit so that it is clear that I did not forget a special permission. Interestingly the only use of 5 digits currently in master is apparently from me in the `debops.apt_cacher_ng` role (`ag 'mode:[^0-9\n]*[2-7][0-7]{3}'`).

The mode parameters with 4 digits not starting with 0 currently in master (that you fix in this PR), it has not caused issues because they are all quoted. I am still in favor of your point that you want to avoid permission issues. This only stands against that we have not seen/used this before DebOps really.

Thanks for explaining. Lets go with 5 digits in case the special file/dir permissions are used.

Using `ag 'mode:[^0-9]*[2-7][0-7]{3}'` I found one additional instance of this in `ansible/roles/debops.icinga_web/tasks/main.yml` which was not found by `ansible-lint`. Can you update this also?

ypid:
Ah right. GitHub UI changes confused me. Regular comments just appear but the review discussion does not auto reload. Never mind.

ansible/roles/debops.ferm/tasks/main.yml
drybjed: @ypid Thanks, done.

ypid (approved):
LGTM, thanks @drybjed for your great work and working out the comments I gave!

PR #603, opened 2019-01-12
[debops.unattended_upgrades] Fix Unattended-Upgrade::Origins-Pattern section in unattended-upgrades.j2 template

gajdusek:
Template unnattended-upgrades.j2 tries to set outer variables in Jinja
for..in loops in order to track if distribution has been found in
unattended_upgrades__origins_lookup list which is used to map distribution
to list of origin patterns.

However, because assignments inside loops do not outlive loop scope
all matching origins are included whereas only the first match should.

For instance in default configuration this results in

Unattended-Upgrade::Origins-Pattern {
	"o=Debian,n=${distro_codename},l=Debian-Security";
	"o=${distro_id},n=${distro_codename},l=${distro_id}-Security";
};

instead of mere

Unattended-Upgrade::Origins-Pattern {
        "o=Debian,n=${distro_codename},l=Debian-Security";
};

This commit fixes this by using namespaced variables introduced in
Jinja 2.10. which can be modified inside loops.

### DIFF
--- a/ansible/roles/debops.unattended_upgrades/templates/etc/apt/apt.conf.d/50unattended-upgrades.j2
+++ b/ansible/roles/debops.unattended_upgrades/templates/etc/apt/apt.conf.d/50unattended-upgrades.j2
@@ -5,23 +5,22 @@
 // upgraded.
 Unattended-Upgrade::Origins-Pattern {
 {% set unattended_upgrades__tpl_origins = [] %}
-{% set unattended_upgrades__tpl_security_origins_found = False %}
-{% set unattended_upgrades__tpl_release_origins_found = False %}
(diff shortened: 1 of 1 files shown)


drybjed (approved):
Thanks! I tested the PR and it works as expected. I didn't know about `namespace()` in Jinja 2.10, this will make things interesting. :-)

drybjed:
It seems that I have been too hasty with merging this package. The issue is that `python-jinja2` package in Debian Stretch is only at 2.8, therefore it lacks the `namespace()` functionality from Jinja 2.10. This results in DebOps (well, `debops.unattended_upgrades` role, but it's part of the common playbook and applied everywhere) not being usable on Debian Stretch, which I want to avoid. I'm reverting this PR for now, we can get back to it later.

PR #604, opened 2019-01-13
Remove ansible dependency in the 'debops' script.

tumd:
All interaction with ansible are done with python's subprocess,
which finds ansible in the env. PATH.
The version-check in 'debops' script however, previously relied on
the ansible python-library, which isn't guaranteed to be the same
version as ansible found in the PATH.

This change will make the version-check with the printout from
'ansible-playbook --version' command instead.

The other nice benefit from this is that debops doesn't have a hard
dependency on ansible, making it much lighter.

Testing this locally with different ansible-versions can quite easily be done with `pipenv`;

- `cd` into this branch-checkout.
- `pipenv install --python 2.7 -e . ansible~=1.9` to build the env
- `pipenv shell` to enter the env

I think debops.ansible_plugins could need some extra testing for this. Because it fails on me with both ansible 1.9 and 2.0.2.0 installed with above pipenv using the below test playbook.

```yaml
---
- name: Test filters
  hosts: 'localhost'

  vars:
    scaffold:
      - name: 'test'
        value: 'value'

  roles:
    - role: 'debops.ansible_plugins'

  tasks:
    - name: Test filtering
      debug:
        msg: '{{ scaffold | parse_kv_config }}'
```


drybjed:
@tumd Thanks, I tested this PR and it seems to work fine. However that got me thinking... Currently, `debops` Python package depends on `netaddr`, `python-ldap`, `passlib` and `dnspython` Python packages. However, these are not required by the `debops` scripts themselves, but by Ansible due to usage od LDAP, and other features in DebOps roles. After the merge of this PR, these modules won't be needed in a given `debops` virtualenv, and furthermore they will have to be installed separately in whatever environment Ansible runs for their features to work.

Edit: Oops, a misclick. :-) Anyway, I'll create a second PR that includes this PR and some reorganization that makes the `ansible` Python package as well as dependencies related to it optional in the `debops` package. This should allow installation either only `debops` or `debops` + `ansible` in the same `virtualenv`. I'll update the installation documentation as well.

PR #606, opened 2019-01-13
[debops.unattended-upgrades] Fix jinja scope issue

tumd:
I saw #603 and remembered I had done a fix for this some time ago in an other project.
This should be backwards compatible with older jinja.

I haven't studied the code in depth, but I think I got my local merge correct. And it looks OK in my tests.


drybjed (approved):
Thanks! I tested it and changes seem to work as intended. :-)

PR #609, opened 2019-01-13
[bsdops.apt] fix: add `when: apt__enabled` to all tasks

AnotherKamila:
It was missing for two tasks before.

(cherry picked from anotherkamila/debops@12856ca0)

---
This change is [Reviewable](https://reviewable.io/reviews/debops/debops/609)

### DIFF
--- a/ansible/roles/debops.apt/tasks/main.yml
+++ b/ansible/roles/debops.apt/tasks/main.yml
@@ -20,6 +20,7 @@
     owner: 'root'
     group: 'root'
     mode: '0755'
+  when: apt__enabled|bool
 
 - name: Save APT local facts
   template:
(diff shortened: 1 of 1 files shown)


drybjed:
This is done deliberately. The scripts in `/etc/ansible/facts.d/` mean that a given role was applied to the host and are easy way to see what roles were run against one. You can create a "stub" `debops.apt` role in the `ansible/roles/` directory which should completely override the one in DebOps; that way the fact from this role will not show up on FreeBSD hosts with the `common` playbook applied. The same applies to other similar roles, like `debops.apt_*`, and others related specifically to Debian.

AnotherKamila:
Ah, okay, that makes sense. Sorry for the newbie mistake.

And thanks for your quick and helpful reply!

PR #611, opened 2019-01-15
[debops.ifupdown] Adding support for purging netplan in favour of ifupdown

mikenowak:
Netplan is a default network manager on Ubuntu Bionic (18.04), so these of us who would rather use  ifupdown for whatever reason, need to purge `netplan.io` and `nplan` packages manually, so that they would not conflict with ifupdown. This PR adds support for this.


drybjed (approved):
Good idea, thanks.

PR #612, opened 2019-01-15
[debops.snmpd] Adding support for per group and per host custom options

mikenowak:
I have a use case where I add some custom options to the whole of the inventory, and now need to define some additional options on per group basis. This PR facilitates this.


drybjed (approved):
Sounds good, although this pattern is usually used with the custom filter used in DebOps to intelligently merge lists of YAML dictionaries. The `debops.snmpd` role will be overhauled in due time. :-)

PR #613, opened 2019-01-16
Purge some "with_flattened" for apt module

gardouille:
In order to fix some deprecation warning like : 
```
[DEPRECATION WARNING]: Invoking "apt" only once while using a loop via squash_actions is deprecated. Instead of using a loop to supply 
multiple items and specifying `name: {{ item }}`, please use `name: [u'{{ apt__base_packages }}', u'{{ apt__packages }}']` and remove 
the loop. This feature will be removed in version 2.11. Deprecation warnings can be disabled by setting deprecation_warnings=False in 
ansible.cfg.
```

### DIFF
--- a/ansible/roles/debops.ntp/tasks/install.yml
+++ b/ansible/roles/debops.ntp/tasks/install.yml
@@ -2,14 +2,12 @@
 
 - name: Install required packages
   apt:
-    name: '{{ item }}'
+    name: '{{ "chrony" if (ntp__daemon == "chrony") else ""
+            + "ntpdate" if (ntp__daemon == "ntpdate") else ""
+            + "ntp" if (ntp__daemon == "ntpd") else ""
+            + "openntpd" if (ntp__daemon == "openntpd") else "" }}'
     state: 'present'
     install_recommends: False
-  with_flattened:
-    - [ '{{ "chrony" if (ntp__daemon == "chrony") else [] }}' ]
-    - [ '{{ "ntpdate" if (ntp__daemon == "ntpdate") else [] }}' ]
-    - [ '{{ "ntp" if (ntp__daemon == "ntpd") else [] }}' ]
-    - [ '{{ "openntpd" if (ntp__daemon == "openntpd") else [] }}' ]
   register: ntp__register_apt_install
   until: ntp__register_apt_install is succeeded
 
--- a/ansible/roles/debops.ntp/tasks/main.yml
+++ b/ansible/roles/debops.ntp/tasks/main.yml
@@ -6,23 +6,20 @@
 
 - name: Install required packages
   apt:
-    name: '{{ item }}'
+    name: [ 'tzdata' ]
     state: '{{ "present" if (ansible_local|d() and ansible_local.ntp|d()) else "latest" }}'
     install_recommends: False
-  with_items: [ 'tzdata' ]
   register: ntp__register_apt_tzdata
   until: ntp__register_apt_tzdata is succeeded
 
 - name: Ensure that alternative daemons/programs are not installed
   apt:
-    name: '{{ item }}'
+    name: '{{ "chrony" if (ntp__daemon != "chrony") else ""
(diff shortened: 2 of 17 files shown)


ansible/roles/debops.ntp/tasks/install.yml
```diff
@@ -2,14 +2,12 @@
 
 - name: Install required packages
   apt:
-    name: '{{ item }}'
+    name: '{{ "chrony" if (ntp__daemon == "chrony") else ""
+            + "ntpdate" if (ntp__daemon == "ntpdate") else ""
+            + "ntp" if (ntp__daemon == "ntpd") else ""
+            + "openntpd" if (ntp__daemon == "openntpd") else "" }}'
```
drybjed, line 8: You should probably use YAML lists for these:
```yaml
    name: '{{ ["chrony"] if (ntp__daemon == "chrony") else []
￼	            + ["ntpdate"] if (ntp__daemon == "ntpdate") else []
￼	            + ["ntp"] if (ntp__daemon == "ntpd") else []
￼	            + ["openntpd"] if (ntp__daemon == "openntpd") else [] }}'
```

ansible/roles/debops.ntp/tasks/main.yml
```diff
@@ -6,23 +6,20 @@
...
   register: ntp__register_apt_tzdata
   until: ntp__register_apt_tzdata is succeeded
 
 - name: Ensure that alternative daemons/programs are not installed
   apt:
-    name: '{{ item }}'
+    name: '{{ "chrony" if (ntp__daemon != "chrony") else ""
+            + "ntpdate" if (ntp__daemon != "ntpdate" and not ntp__ignore_ntpdate|bool) else ""
+            + "ntp" if (ntp__daemon not in [ "ntpd", "openntpd"]) else ""
+            + "openntpd" if (ntp__daemon != "openntpd") else "" }}'
```
drybjed, line 20: Similar to the situation above.

drybjed (changes requested):
Thanks, but I think that the solution in `debops.ntp` will break. Check the comments.

ansible/roles/debops.ntp/tasks/install.yml
gardouille: I used an empty string cause an empty dict won't work for me.
With the requested changes, `debops service/ntp -l myhost` gave me this : 
```
TASK [debops.ntp : Install required packages] ******************************************************************************************
fatal: [vm203030.univ]: FAILED! => {"msg": "Unexpected templating type error occurred on ({{ \"chrony\" if (ntp__daemon == \"chrony\") else [] + \"ntpdate\" if (ntp__daemon == \"ntpdate\") else [] + \"ntp\" if (ntp__daemon == \"ntpd\") else [] + \"openntpd\" if (ntp__daemon == \"openntpd\") else [] }}): can only concatenate list (not \"str\") to list"}
```

drybjed (approved):
After testing it seems that the `debops.ntp` changes work fine. Thanks!

mikenowak:
@gardouille, @drybjed,

this commit indeed breaks the debops.ntp module on Ubuntu  18.04 where `systemd-timesyncd` is used instead of any of the above ntp daemons.

https://github.com/debops/debops/blob/2a0902e097786fa40f51569a999caee969b5198f/ansible/roles/debops.ntp/defaults/main.yml#L40-L47

The role fails as follows:

```
TASK [debops.ntp : Install required packages] *******************************************************
fatal: [host1.example.co.jp]: FAILED! => {"attempts": 3, "changed": false, "msg": "No package matching '' is available"} 
```

drybjed:
@mikenowak Thanks for the heads up. I managed to reproduce this issue on Debian as well, now working on a fix.

drybjed:
The issue will be fixed by https://github.com/debops/debops/pull/615/commits/06ff3272eeca7c9d4d37e70c25988e5402a59430.

PR #618, opened 2019-01-17
Gunicorn logrotate

cultcom:
Create gunicorn logrotate configs per instance.


drybjed (approved):
Nice solution, thanks!

PR #619, opened 2019-01-17
Pipe flatten install

gardouille:
Fix some issues related to #613 changes.


drybjed (approved):
Looks OK, but I said that the last time as well. ;-)

PR #626, opened 2019-01-20
[debops.netbox] Fix version-check in python3

### LINKED ISSUE #597: [debops.netbox] git version "master" breaks version comparison
https://github.com/debops/debops/blob/2b68f94371990be86266fe6d2be7953b4c262261/ansible/roles/debops.netbox/defaults/main.yml#L141

without a proper version I get this exception:
´´´
TASK [debops.netbox : Remove old python2 based virtualenv] *******************************************************************************************
fatal: [vm01]: FAILED! => {"msg": "The conditional check '( netbox__virtualenv_version == '3' and netbox__register_virtualenv_version.stat.lnk_target|d() == 'python2' )' failed. The error was: An unhandled exception occurred while templating '{{ \"\" if netbox__git_version|replace(\"v\", \"\") is version(\"2.5\", \"<\") else \"3\" }}'. Error was a <class 'ansible.errors.AnsibleFilterError'>, original message: Version comparison: '<' not supported between instances of 'str' and 'int'\n\nThe error appears to have been in '/home/wigge/.local/share/debops/debops/ansible/roles/debops.netbox/tasks/main.yml': line 62, column 3, but may\nbe elsewhere in the file depending on the exact syntax problem.\n\nThe offending line appears to be:\n\n\n- name: Remove old python2 based virtualenv\n  ^ here\n"}
´´´

tumd:
It's not pretty, but seems to work.
Tested with both python 2.7 and 3.7.

'netbox__git_version' should only match the regex_search if it looks
like a version-string. I.e. starts with an optional 'v'-character
following something that resembles a version.

Fixes #597

### DIFF
--- a/ansible/roles/debops.netbox/defaults/main.yml
+++ b/ansible/roles/debops.netbox/defaults/main.yml
@@ -158,7 +158,8 @@ netbox__git_checkout: '{{ netbox__virtualenv + "/app" }}'
 # created next to the old one.
 # Valid values are `3` or an empty string.
 netbox__virtualenv_version: '{{ ""
-                                 if netbox__git_version|replace("v", "") is version("2.5", "<")
+                                 if ( netbox__git_version | regex_search("^v?[0-9\.]+")
+                                   and netbox__git_version|replace("v", "") is version("2.5", "<") )
                                  else "3" }}'
 
                                                                    # ]]]


drybjed (approved):
Sounds good, thanks!

PR #627, opened 2019-01-21
Add options to nginx to enable and disable HTTP/2 and SPDY protocol support

rei-ifesca:
In some use-cases HTTP/2 support is not desired as the default connection protocol between the client and the nginx server.


drybjed (approved):
Looks good, thanks. :+1:

PR #630, opened 2019-01-21
padlock: optionally use external unlock command

kouk:
This change allows optionally using "custom" types of encrypted folders,
via a user-provided executable command, but only if the encrypted folder
path doesn't already contain an EncFS configuration file.

I've re-used the `PADLOCK_CMD` variable which seems to have remained in
the `debops` library (even if unused).

This was prompted because I was looking for a way to integrate
debops-padlock with [gocryptfs][]. In my case I have a gocryptfs
encrypted folder in `ansible/.encfs.secret/gocryptfs` and a script under
`ansible/.encfs.secret/padlock` like this:

    #!/bin/sh

    extpass="..."
    exec gocryptfs -extpass="$extpass" $1/gocryptfs $2

[gocryptfs]: https://github.com/rfjakob/gocryptfs "gocryptfs"

### DIFF
--- a/debops/__init__.py
+++ b/debops/__init__.py
@@ -197,12 +197,20 @@ def padlock_unlock(encrypted_path):
     configfile = os.path.join(encrypted_path, ENCFS_CONFIGFILE)
     crypted_configfile = configfile+'.asc'
 
-    if (not os.path.exists(keyfile) or
-            not os.path.exists(crypted_configfile)):
-        return False
+    # Location of an alternative executable that will mount the decrypted dir
+    unlock_cmd = os.path.join(encrypted_path, PADLOCK_CMD)
 
     # Cut the EncFS directory prefix to get the decrypted directory name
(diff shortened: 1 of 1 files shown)


drybjed:
Interesting, thanks. `gocryptfs` seems to be a nice alternative to EncFS and it's already included in Debian Stretch, which is a plus. The `debops-padlock` script could detect either of them and pick one, setting up te encrypted filesystem for secrets.

Can the `extpass` value be retrieved via `gpg`, similarly how currently EncFS password is stored in a GPG-encrypted keyfile?

drybjed (approved):

kouk:
Thanks! Actually yes, in my shell script I do just that, pass a gpg command
line as the `--extpass` parameter to gocryptfs. Initially I started to do a
refactor of debops-padlock to support both encfs and gocryptfs as first
class citizens, but I was a bit constrained by time and thought that this
might be an acceptable solution that I could complete in time. That said,
since you seem onboard with a refactoring I might give it another go when I
find some time.

PR #636, opened 2019-01-24
[lxc_ssh] bugfix binary data copy

TuxCoder:
This fixes the bug, if you want to copy a binary file over lxc_ssh with ansible.
I haven't tested it with python2, but it should be compatible.


drybjed:
@TuxCoder Thanks! Perhaps you should also post a similar PR to this script' [upstream repository](https://github.com/andreasscherbaum/ansible-lxc-ssh).

PR #641, opened 2019-01-30
Additional templates to rsyslog

ser:
Rsyslog default templates are very basic and in real world situations you need to set more of them. This pull request allows additional templates definitions in DebOps.


drybjed (approved):
Looks good, thanks! :+1:

PR #642, opened 2019-01-31
Quick Fix to recent rsyslog templates, sorry about that

ser:
too many "and"s, sorry


drybjed (approved):
Thanks for the quick fix.

PR #655, opened 2019-02-01
[debops.cron] Docs: mention job name limitations.

evilham:
Closes debops/ansible-cron#3 which suggested an assertion, but just
mentioning it in the docs is enough and simpler.

Sorry it took so long!


drybjed (approved):
Sounds good, thanks.

PR #660, opened 2019-02-04
[debops.pki] Improve portability when getting fqdn

Erethon:
This PR fixes a portability issue with the pki role when run from certain *BSD systems, in particular, the `-f` argument on `hostname` doesn't exist on OpenBSD, NetBSD and I'm guessing other BSDs as well that use the old `hostname` from 4.2BSD onwards.

The patch works on my Ansible controller which is an OpenBSD 6.4 host, as well as a FreeBSD host I tried it (where `hostname` has the `-f` option). From reading online the man pages of OpenBSD, FreeBSD, NetBSD, DragonFly and 4.3BSD, in all of them, by default `hostname` prints the FQDN, so I'm thinking this is a safe assumption to make for all BSDs.

### DIFF
--- a/ansible/roles/debops.pki/files/secret/pki/lib/pki-authority
+++ b/ansible/roles/debops.pki/files/secret/pki/lib/pki-authority
@@ -73,13 +73,21 @@ get_dnsdomainname () {
         dnsdomainname
     else
         local fqdn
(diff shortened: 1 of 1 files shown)


drybjed (approved):
Sounds good, thanks.

PR #667, opened 2019-02-07
Add Docker Registry support

### LINKED ISSUE #556: GitLab Registry integration
Hi,

we have manage to configure and run a Docker Registry within our GitLab server. This registry service can be accessed over port 5000 and 443, thanks to simple nginx virtual host proxy configurations ;)

Due to lack of good documentation on the Docker and GitLab side it was not easy to make each other talk... Nevertheless, we would like to return our knowledge to the community and extend the debops.gitlab role to support a docker registry.

Because of the circular depenencies between the docker registry container and gitlab and additonal dependencies to the gitlab role (e.g. debops.docker) I would like to discuss the best path for a nice PR ;)

I have the following ideas:
- Extend the gitlab role documentation and templates to configure the registry, including PKI rootbundle support
- Add a debops.dsocker_registry role to deploy a docker-registry container over compose with ready-to-go configuration to work together with gitlab, however also support a standalone docker registry service over TLS with simple HTTP auth.

Any opinions?

drybjed:
Fixes #556

### DIFF
--- a/ansible/roles/debops.docker_registry/defaults/main.yml
+++ b/ansible/roles/debops.docker_registry/defaults/main.yml
@@ -248,7 +248,7 @@ docker_registry__storage_dir: '{{ (ansible_local.root.data
 # .. envvar:: docker_registry__storage_mode [[[
 #
 # The UNIX permissions of the local filesystem storage directory.
-docker_registry__storage_mode: '0751'
+docker_registry__storage_mode: '0755'
                                                                    # ]]]
                                                                    # ]]]
 # Redis Server integration [[[
--- a/ansible/roles/debops.docker_registry/defaults/main.yml
+++ b/ansible/roles/debops.docker_registry/defaults/main.yml
@@ -202,7 +202,7 @@ docker_registry__allow: []
 # GitLab Container Registry integration is enabled, the internal HTTP Basic
 # Auth is disabled.
 docker_registry__basic_auth: '{{ False
-                                 if docker_registry__gitlab_enabled|bool
+                                 if docker_registry__token_enabled|bool
                                  else True }}'
 
                                                                    # ]]]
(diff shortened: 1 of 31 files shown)


ansible/roles/debops.docker_registry/defaults/main.yml
```diff
@@ -0,0 +1,629 @@
...
+
+                                                                   # ]]]
+# .. envvar:: docker_registry__gitlab_token_realm_url [[[
+#
+# The URL of the GitLab Container Registry API endpoint, passed to the Docker
+# clients. The clients can use this URL to request signed authentication tokens
+# from GitLab Container Registry service.
+docker_registry__gitlab_token_realm_url: '{{ ansible_local.gitlab.registry_token_realm_url
+                                             if (ansible_local|d() and ansible_local.gitlab|d() and
+                                                 ansible_local.gitlab.registry_token_realm_url|d())
```
rei-ifesca: Apparently the indentation here is not correct.

ansible/roles/debops.docker_registry/defaults/main.yml
```diff
@@ -0,0 +1,629 @@
...
+
+                                                                   # ]]]
+# .. envvar:: docker_registry__gitlab_token_issuer [[[
+#
+# The GitLab Container Registry token issuer string, included in the issued
+# tokens. This value needs to be the same in the GitLab Container Registry
+# configuration file.
+docker_registry__gitlab_token_issuer: '{{ ansible_local.gitlab.registry_token_issuer
+                                          if (ansible_local|d() and ansible_local.gitlab|d() and
+                                              ansible_local.gitlab.registry_token_issuer|d())
```
rei-ifesca: Apparently the indentation here is not correct.

ansible/roles/debops.docker_registry/defaults/main.yml
```diff
@@ -0,0 +1,629 @@
...
+
+  - name: '{{ docker_registry__fqdn }}'
+    filename: 'debops.docker_registry'
+    allow: '{{ docker_registry__allow }}'
+    auth_basic: '{{ docker_registry__basic_auth | bool }}'
+    auth_basic_realm: '{{ docker_registry__basic_auth_realm }}'
+    auth_basic_name: '{{ docker_registry__basic_auth_name }}'
+
+    options: |
+      client_max_body_size 0;
```
rei-ifesca, line 607: Do you want to allow to upload images with insane sizes? 
What do you think? It is better to set a limit, lets say of 5GB or 10GB? 
It would at least reduce abuse, however this docker-registry will be mostly run as a private service...

ansible/roles/debops.docker_registry/defaults/main.yml
```diff
@@ -0,0 +1,629 @@
...
+                             if (ansible_local|d() and ansible_local.core|d() and
+                                 ansible_local.core.domain|d())
+                             else (ansible_domain if ansible_domain else ansible_hostname) }}'
+
+                                                                   # ]]]
+# .. envvar:: docker_registry__backend_port [[[
+#
+# The TCP port on which the Docker Registry backend service listens for
+# connections.
+docker_registry__backend_port: '5070'
```
rei-ifesca, line 188: This is the port where the docker-registry daemon listens for HTTP connections, right?
So the same port number should be set in the gitlab defaults config file. There is the port set to 5005.

docs/ansible/roles/debops.gitlab/getting-started.rst
```diff
@@ -54,6 +54,25 @@ on the host, for example by the :ref:`debops.redis_server` Ansible role.
...
+the :ref:`debops.docker_registry` playbook to faciliate configuration
+synchronization between the two services. The installation process with Docker
+Registry role enabled will be:
+
+- Install PostgreSQL server, Redis Server
+- Install Docker Registry service
+- Install GitLab, using facts from Docker Registry role
+- Update Docker Registry service with facts from GitLab role
+
+This should avoid issue with circular dependencies in the configuration of both
```
rei-ifesca, line 71: What do you think of using using the "/env" sub-role to configure both services first, like the nullmailer role:
` - role: debops.nullmailer/env`
Instead of running the playbooks again, although due to the idempotency of ansible nothing should break :D

ansible/roles/debops.gitlab/defaults/main.yml
```diff
@@ -839,128 +839,124 @@ gitlab__piwik_url: ''
...
-# use port 443 and ACME TLS certificates.
+# If the Docker Registry is available on TCP port other than 443, you can
+# specify the port number here; it will be displayed in the GitLab Container
+# Registry section of the GitLab interface. Docker Registry published behind
+# a HTTPS reverse proxy does not need to advertise its port, 443 (HTTPS) will
+# be used automatically by Docker clients.
+gitlab__registry_host_port: '{{ ansible_local.docker_registry.host_port
+                                if (ansible_local|d() and ansible_local.docker_registry|d() and
+                                    ansible_local.docker_registry.host_port|d())
+                                else "5005" }}'
```
rei-ifesca, line 877: See my review regarding the registry listen port: https://github.com/debops/debops/pull/667/commits/83e42a69f4a97ed2a65215bbd9a9d28ef6920fad#diff-7f8f9d8ee0d63f8cbbdfe5b72cd893cbR188

rei-ifesca:
Good work! 👍 

What do you think, it is possible to move "old registry data" to the new registry data path and still have the old containers inside the new role?

Sadly I won't be able to test thid (huge) changes quickly, as our gitlab/registry is already running in production. We need to plan the migration

drybjed:
@rei-ifesca I found a couple of issues already, so the merge will probably happen during the weekend. I'll check if moving the data between directories changes anything, at least in the filesystem storage type.

drybjed:
@rei-ifesca I checked the data move between the directories, and it seems that it works as expected.

ansible/roles/debops.docker_registry/defaults/main.yml
drybjed: I'm not sure what you mean, what would be the correct indentation?

ansible/roles/debops.docker_registry/defaults/main.yml
drybjed: I'm not sure what you mean, what would be the correct indentation?

ansible/roles/debops.docker_registry/defaults/main.yml
drybjed: Sure, I suppose a 10G limit by default should be fine, although I would expose it as a separate default variable for easy change. I'll add a quick commit in my next PR for this.

ansible/roles/debops.docker_registry/defaults/main.yml
drybjed: I tend to avoid using ports suggested by upstreams that are not officially registered - usually it's something like `4000`, `9000`, etc. The '5070' port seems to be available in the `/etc/services` database, so I went with it.

docs/ansible/roles/debops.gitlab/getting-started.rst
drybjed: Don't forget that the roles are designed to manage standalone services - Docker Registry and GitLab - separately. Ansible local facts allow for combining the two roles together while still working separately just fine.

Another problem is that both services require information from each other to perform correct integration. The order DR/G/DR seems to me to be a better choice, especially that the GitLab playbook runs significantly longer than Docker Registry.

I'm not sure what could be moved to an `/env` role - you still need to put the configuration file somewhere, for that you need to create UNIX accounts, and quickly any other options are exhausted. The current setup seems to be the most sane enough without merging the two roles together in some way and losing the ability to set them up separately.

ansible/roles/debops.gitlab/defaults/main.yml
drybjed: If you don't use `debops.docker_registry` role to set up your Docker Registry, you probably have installed it manually or some other way, with default TCP ports. In that case, `debops.gitlab` role will try to connect to the service using the default ports suggested by upstream.

If you used the `debops.docker_registry`, the GitLab role should use the configuration defined by the Docker Registry role which could have used different TCP ports.

ansible/roles/debops.docker_registry/defaults/main.yml
rei-ifesca: In the browser display it just looked "wrong". I think this review element can be  closed as solved.

ansible/roles/debops.docker_registry/defaults/main.yml
rei-ifesca: In the browser display it just looked "wrong". I think this review element can be  closed as solved.

ansible/roles/debops.docker_registry/defaults/main.yml
rei-ifesca: A new variable sounds good.

ansible/roles/debops.docker_registry/defaults/main.yml
rei-ifesca: Ok, that's an argument I did not looked at. The port number itself is for me irrelevant, I was just confused about the different numbers.

docs/ansible/roles/debops.gitlab/getting-started.rst
rei-ifesca: Understood. Lets keep the current setup.

ansible/roles/debops.gitlab/defaults/main.yml
rei-ifesca: Ok

PR #668, opened 2019-02-08
Resources acl

gardouille:
Everything looks good this time and `make test` didn't return any related issue.

### DIFF
--- a/docs/ansible/roles/debops.resources/defaults-detailed.rst
+++ b/docs/ansible/roles/debops.resources/defaults-detailed.rst
@@ -340,7 +340,7 @@ to access to any new content:
 .. code-block:: yaml
 
    resources__paths:
-     - '/tmp/dir1'
+     - dest: '/tmp/dir1'
        acl:
          - default: True
            etype: 'group'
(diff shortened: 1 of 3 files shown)


docs/ansible/roles/debops.resources/defaults-detailed.rst
```diff
@@ -305,3 +320,79 @@ Create a custom :program:`cron` task that restarts a service daily:
...
+Examples
+~~~~~~~~
+
+Create a directory on all hosts and allow ``adm`` group and ``joe`` user
+to access to any new content:
+
+.. code-block:: yaml
+
+   resources__paths:
+     - '/tmp/dir1'
```
drybjed: This needs to be a valid YAML dictionary key - `dest`, for example. Otherwise the YAML parser will crash.

drybjed (changes requested):
Looks good, although one example needs to be updated to work as expected. Also, can you add a Changelog entry?

docs/ansible/roles/debops.resources/defaults-detailed.rst
gardouille: The example give complete RO access on the directory only for `adm` group now with two ACLs :)

drybjed (dismissed):
Nicely done, thanks! :+1:

drybjed (approved):

PR #669, opened 2019-02-08
 [debops.owncloud] Install PHP modules required for the twofactor app

ypid:
twofactor_rcdevsopenotp requires the ldap and soap PHP modules as of NC 14. Further versions of NC will require MFA/2FA. Allowing the twofactor_rcdevsopenotp app to be installed in a default DebOps setup seems like a sensible way to go.


drybjed (approved):
Sounds good. Will the 2FA be required to set up, or optional but recommended?

ypid:
> Will the 2FA be required to set up, or optional but recommended?

I guess it can only be optional for users, but I read something that Nextcloud wants to push 2FA further, maybe to use some features of Nextcloud. Thanks for merging!

PR #681, opened 2019-02-13
[debops.docker] More robust way to get docker daemon version

ypid:
If the daemon is not running, `docker version` returns:

> Cannot connect to the Docker daemon at unix:///var/run/docker.sock. Is the docker daemon running?

This in turn prevents the task "Configure Docker systemd options" from running which would have fixed the issue in my case.

### DIFF
--- a/ansible/roles/debops.docker/defaults/main.yml
+++ b/ansible/roles/debops.docker/defaults/main.yml
@@ -66,7 +66,7 @@ docker__upstream_key: '{{ "9DC858229FC7DD38854AE2D88D81803C0EBFCD88"
                                                                    # ]]]
 # .. envvar:: docker__upstream_packagename [[[
 #
-# The full docker packagename to be installed.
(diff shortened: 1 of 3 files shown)


drybjed (dismissed):
Indeed... It makes sense that the `docker` client cannot get the server's version if server is not running. I suppose using `dpkg` to check it is the next best thing.

drybjed:
@ypid You need to fix the line length in the fact script. Did you run `make test` in the root of the repository to check everything? Or at least Python code with `make pep8`.

drybjed:
Actually, I think that the previous version of the script was better. What you should do is specify the package name in a variable, and then use that variable in the `dpkg-query` call. It would better in the template as well, since you have the actual Python code separate from Jinja templating in variables.

Do you want to switch to the earlier version? I suppose that the current one will also work fine, and could be improved later. Your call.

drybjed (approved):

PR #682, opened 2019-02-13
[debops.logrotate] Fix generated logrotate configuration

ypid:
@drybjed I noticed that on my systems, invalid logrotate configuration was generated because the ending `}` was on the same line as the last directive. Can you reproduce this or do you think it is an issue with my setup?

This patch was not comprehensibly tested because I am not sure if I made a mistake.

### DIFF
--- a/ansible/roles/debops.logrotate/templates/etc/logrotate.conf.j2
+++ b/ansible/roles/debops.logrotate/templates/etc/logrotate.conf.j2
@@ -60,7 +60,7 @@
 {%     endif %}
 {%     if section.postrotate|d() %}
     postrotate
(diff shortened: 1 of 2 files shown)


drybjed:
What's your Jinja2 version? I noticed this as well on 2.8 from Debian Jessie, but after installing 2.10 via `pip --user` it works fine again. I'm not sure if fix for both can be done in the template.

Can you check with Jinja 2.10 first?

ypid:
OK thanks, I am closing this for now. I have still need to upgrade my config management qube to Stretch. Something prevented me when I last tried.

Issue #551, opened 2018-11-23
Postfix can not connect to opendkim


alinalexandru:
It seems that the socket path are not the same in postfix and opendkim

I made the following change in `/etc/default/opendkim`
```
old
#SOCKET="local:/var/run/opendkim/opendkim.sock"
new
SOCKET="local:/var/spool/postfix/opendkim/opendkim.sock"
```

I had this problem on  ubuntu 16 and ubuntu 18

drybjed:
On Debian it seems that's not an issue, so only Ubuntu related. And you also created https://github.com/debops/ansible-opendkim/issues/6 in the `debops.opendkim` repository.

I guess managing the `/etc/default/opendkim` configuration file by the role is inevitable then. I'll look into it in a few days, looks like it will have to be diverted as well.

PR #687, opened 2019-02-15
[debops.opendkim] Fix opendkim integration on ubuntu

### LINKED ISSUE #651: [debops.opendkim] Integration with postfix fails on ubuntu
It seems that the ansible role does not edit the /etc/default/opendkim on ubuntu when postfix is installed on the same host.

After I update the SOCKET line with **SOCKET="local:/var/spool/postfix/opendkim/opendkim.sock"** the integration was working.
I can fix the problem it is ok to update this file on ubuntu.

alinalexandru:
Will close #651 #551

### DIFF
--- a/ansible/roles/debops.opendkim/tasks/main.yml
+++ b/ansible/roles/debops.opendkim/tasks/main.yml
@@ -12,10 +12,16 @@
   notify: [ 'Check opendkim and restart' ]
   when: ansible_pkg_mgr == 'apt'
 
-- name: Divert original OpenDKIM configuration
+- name: Divert original OpenDKIM configuration from /etc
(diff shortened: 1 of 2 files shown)


drybjed:
The `/etc/default/opendkim` file is probably a Conffile which you should divert to avoid problems during upgrades. Also, it would be good if the configuration works with and without Postfix installed via `debops.postfix` role, ie. check the integration of the roles together.

alinalexandru:
I am working on that. I did this some small commits and PR to show to @LeonardoAndrei how is the process on contributing to open source

drybjed:
@LeonardoAndrei Do you know if the reload of the `opendkim.service` is enough to create the socket just after installation, or is full restart needed? Can you check?

PR #695, opened 2019-02-16
Space is required

ser:
Without space the next line is directly appending to the previous one, creating syntax error.


drybjed (dismissed):
Nice catch, thanks! I'll have to remember that.

drybjed (approved):

PR #698, opened 2019-02-19
[debops.slapd] Integrating TLS configuration in one step

ser:
- future versions of slapd do not allow setting TLS attributes as cn=admin,cn=config
- it is not required anymore to specify a TLS cipher suite


ser:
there was a problem with the submission account, it will be here again soon, closing for now

PR #699, opened 2019-02-19
[debops.slapd] Integrating TLS configuration in one step

ser:
- integrating TLS configuration into one step
- specification of cipher suite is not required


drybjed:
Thanks, I'll test this PR in a day or two to make sure it works as expected.

drybjed:
This should be fixed by #708. @ser, can you confirm the fix worked for you?

drybjed:
The `debops.slapd` role was redesigned, therefore this pull request does not make sense anymore. Closing.

Issue #684, opened 2019-02-14
[debops.stunnel] Failure to start on Debian Stretch


jstruebel:
When enabling the debops.stunnel service for a Debian Stretch host the stunnel4 service fails to start with the error "/etc/stunnel/\<service\>.conf:13: "options..= NO_SSLv2": Illegal TLS option".

Some googling of the error reveals that it is most likely due to using openssl v1.1.0 in stretch. Running the same setup against a Debian Jessie host works as expected. Per [this link](https://forge.univention.org/bugzilla/show_bug.cgi?id=45882) the NO_SSLv2 option was removed completely from that version of openssl. Other distributions that use this or later openssl versions likely have the same problem.

Since the NO_SSLv2 option is hard-coded into the stunnel config template, it needs to either be removed or somehow qualified based on the distribution or the openssl version installed. A third option would be to make those settings configured via the inventory with secure defaults.

What path would you recommend taking? If all of the distributions that debops is compatible with use openssl versions that disable SSLv2 by default then it is likely safe to just remove it. However, I'm a fan of setting the configuration to be as secure as reasonable, even if it is the default. Making those settings configurable is more flexible which I like, but it also opens up the possibility of a user inadvertently making their setup less secure. I'm willing to submit a pull request to fix this so long as it isn't too complicated, just want to get some feedback on what path you desire to go down.

drybjed:
Agreed, NO_SSLv2 will have to go since it breaks the service. Actually, the whole configuration file could be generated using data from the role defaults, most likely filtered through the `parse_kv_items` filter so that configuration could be augmented via inventory. But for now, a fix that removes NO_SSLv2 from the template and perhaps exposes the list of disabled SSL protocols via a default variable is fine. Go for it. :-)

PR #705, opened 2019-02-22
[debops.stunnel] Make options parameter for services configured via inventory

jstruebel:
Changed the SSL options parameter to be configured via inventory instead of hard-coded in the template. The default setting drops the "NO_SSLv2" option. This addresses #684.


drybjed (approved):
Nicely done, thanks!

PR #710, opened 2019-02-28
[debops.icinga_web] Allow DELETE method in HTTP requests

drebs:
Icinga Director REST API includes HTTP DELETE method for some
operations, and thus this method needs to be allowed by Icinga Web Nginx
server configuration:

https://github.com/Icinga/icingaweb2-module-director/blob/master/doc/70-REST-API.md


drybjed (approved):
Sounds good, thanks. :-)

PR #711, opened 2019-02-28
[debops.nginx] Move pre_tasks hook to top of main.yml

ghost:
I ran into an issue extracting the IP of a Docker container, setting it as a fact and then using it in the nginx.yml to configure Nginx as a proxy.

Example: Extract Docker IP from container, store as a fact and use inside nginx proxy configuration.

### docker-ip 
Shell script to find the IP address of the Docker container

```
#!/bin/bash
if [ "$#" -eq  "0" ]
  then
    echo "$0 <container name>"
    exit
fi
docker inspect -f '{{.Name}} - {{range .NetworkSettings.Networks}}{{.IPAddress}}{{end}}' $(docker ps -aq) | grep $1 | awk '{ print $3}'
```

​
### pre_main.yml
```
---
- name: Copy docker-ip.sh to /usr/local/bin
  copy:
    src: ../../commands/
    dest: "/usr/local/bin" 
    mode: 0755
    remote_src: no
  with_items:
    ['docker-ip.sh']  
  
- name: Find IP of xyz-container
  shell: "docker-ip xyz-container"
  register: command_output
​
- set_fact:
    xyz_container_ip: "{{ command_output.stdout }}"
​
- debug:
    msg: "{{ xyz_container_ip }}"
```

### nginx.yml
```
nginx__deploy_state: present

nginx__servers:
  
  - name: [ 'somedomain.com' ]
    filename: 'somedomain'
    type: 'proxy'
    proxy_pass: 'http://{{ xyz_container_ip }}:8000'
```

For the above to work (`proxy_pass`) the `DebOps pre_tasks hook` needs to come first, before the `Assert that no legacy options are used` task. Otherwise Ansible complains that the `xyz_container_ip` variable is missing.

Note: In above example you can't simply drop the `docker inspect` into cmd or shell as the {{ and }} conflict with Jinja templating and the solution isn't pretty.

### DIFF
--- a/ansible/roles/debops.nginx/tasks/main.yml
+++ b/ansible/roles/debops.nginx/tasks/main.yml
@@ -1,5 +1,9 @@
 ---
 
+- name: DebOps pre_tasks hook
+  include: '{{ lookup("task_src", "nginx/pre_main.yml") }}'
+  when: (nginx__deploy_state in [ 'present' ])
+
 - name: Assert that no legacy options are used
   assert:
     that:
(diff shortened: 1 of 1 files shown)


drybjed (approved):
Good idea, thanks! :+1:

PR #712, opened 2019-03-02
debops.lvm: Create file system when mount parameter is defined

ganto:
According to the documentation:
> By default, a filesystem specified in `lvm_default_fs_type` variable is created in all new Logical Volumes if `item.mount` is specified.

Unfortunately this doesn't work because in this case the default value for `fs` is empty which means `false`. This fix makes sure that a file system is created when a mount point is defined.


drybjed (approved):
Thanks for the fix! :+1:

PR #715, opened 2019-03-07
[debops.gitlab_runner] Added feature for tmpfs mounting

### LINKED ISSUE #595: [debops.gitlab_runners] Add support for mounting a directory in RAM
There is a very nice feature that allows to mount a directory in ram https://docs.gitlab.com/runner/executors/docker.html#mounting-a-directory-in-ram

One idea is to add 2 new parameters: `docker_tmpfs` and `docker_services_tmpfs`, each one to be a dictionary with key as directory and value as options. Those will be used at instance level.

LeonardoAndrei:
Close #595


drybjed (approved):
Nicely done, thanks.

PR #716, opened 2019-03-08
[debops.redis_s*] Fixed logic on non-systemd.

evilham:
This does not affect functionality on systemd systems and enables
use of the roles on non-systemd ones.


drybjed (approved):
Thanks! This actually fixes an uncertainity with the `or` conditions in this tasks. Without the brackets, only the first `or` condition that is `True` will trigger, with them all conditions are evaluated.

PR #719, opened 2019-03-09
[debops.ifupdown] Run ifup_systemd conditionally.

### LINKED ISSUE #718: [debops.ifupdown] Remove full dependency on systemd
Hello,

I just noticed `Check systemd version` would fail in `debops.ifupdown/tasks/ifup_systemd` on non-systemd systems.

This wasn't the case before and I wasn't able to pinpoint the exact cause of change in behaviour. 

However, since the file name is systemd-centric, I figured it shouldn't run at all when it's not the service manager/init system, having a conditional include in the `main` tasks file makes this more robust and also simplifies `ifup_systemd`.

Creating issue to foster further discussion in case the incoming PR is not deemed appropriate.

evilham:
Fixes #718.

No changes in behaviour on a systemd setup.
Allows usage of the role on non-systemd setups.

### DIFF
--- a/ansible/roles/debops.ifupdown/tasks/ifup_systemd.yml
+++ b/ansible/roles/debops.ifupdown/tasks/ifup_systemd.yml
@@ -24,7 +24,7 @@
 - name: Reload systemd services
   systemd:
     daemon_reload: True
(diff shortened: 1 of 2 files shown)


drybjed (approved):
Sure, sounds good. Thanks. :-)

PR #722, opened 2019-03-12
Refactored python for Dokuwiki role

LeonardoAndrei:
Hello @drybjed ! Can you tell me if these changes are ok? I can do the same for all the remaining roles mentioned earlier.

### DIFF
--- a/ansible/playbooks/service/apt_cacher_ng.yml
+++ b/ansible/playbooks/service/apt_cacher_ng.yml
@@ -27,6 +27,13 @@
         - '{{ apt_cacher_ng__ferm__dependent_rules }}'
         - '{{ nginx_ferm_dependent_rules }}'
 
+    - role: debops.python
+      tags: [ 'role::python', 'skip::python' ]
+      python__dependent_packages3:
+        - '{{ nginx__python__dependent_packages3 }}'
+      python__dependent_packages2:
+        - '{{ nginx__python__dependent_packages2 }}'
+
     - role: debops.nginx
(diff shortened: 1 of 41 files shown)


drybjed:
@LeonardoAndrei Yes, these are good changes. Although Travis reports some trailing spaces, which you should remove. DebOps monorepo has a test suite you can use to check the syntax, run:
```
make test
```
to check everything in the repository. You can run:
```
make help
```
to see what tests are currently implemented.

If you want to put the changes in separate commits, it would be nice if you tag each commit's title with the role name, that way it's easy to spot changes in roles you are invested in. For example, this commit could be named:
```
[debops.dokuwiki] Install Python APT packages via 'debops.python' role
```
Check [a blog entry about good `git` commit messages](https://chris.beams.io/posts/git-commit/) for some pointers.

Thanks for picking this up! :-)

drybjed:
@LeonardoAndrei Keep up the good work! Let me know when the PR is ready for a review.

drybjed:
@LeonardoAndrei Nice work so far, keep it up.

Now, because you added `debops.python` dependency to `debops.nginx` playbook, you also need to update it in all playbooks that use `debops.nginx` as a dependency as well, otherwise `debops.nginx` role will be broken in many different applications (LibreNMS, GitLab, RStudio Server...).

drybjed (approved):
@LeonardoAndrei Thanks for your work! :+1: There are still some `debops.nginx` dependencies missing but I'll add them before merging.

PR #725, opened 2019-03-20
[debops.cryptsetup] Fixes present state flag behavior

inthehack:
The state flag of the plaintext block devices controls if the block
devices must be opened or not and if the filesystem on it must be
mounted.

In case the filesystem is not managed by the role and the state flag is
set to `present', the plaintext block device should not be closed as it may be
used by other roles like debops.lvm.

Then, this commit adds an additional condition inside the teardown procedure
at the end of the role in order to implemente this use case.

In a further patch, the lvm-based use case should be better covered with
explicit configuration other than simply disable filesystem management.

### DIFF
--- a/ansible/roles/debops.cryptsetup/tasks/manage_devices.yml
+++ b/ansible/roles/debops.cryptsetup/tasks/manage_devices.yml
@@ -11,7 +11,7 @@
   assert:
     that:
       - item.state|d(cryptsetup__state) in
-        [ 'mounted', 'ansible_controller_mounted', 'unmounted', 'present', 'absent' ]
+        [ 'mounted', 'ansible_controller_mounted', 'unmounted', 'mapped', 'present', 'absent' ]
       - item.name is defined and item.name is string
       - (item.state|d(cryptsetup__state) == 'ansible_controller_mounted' and 'remote_keyfile' not in item) or
         item.state|d(cryptsetup__state) != 'ansible_controller_mounted'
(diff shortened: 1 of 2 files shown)


drybjed (dismissed):
Sounds reasonable to me. I'll wait for @ypid a day or two before merging, in case he sees any issues.

ypid:
Sounds also reasonable to me, thanks @jpeeters. I would like to check and test this in more detail. @drybjed Please wait with merging until Tuesday. Until then I should be able review it more.

ypid (changes requested):
This change is not in line with the docs of the role. Refer to https://docs.debops.org/en/master/ansible/roles/debops.cryptsetup/defaults-detailed.html#cryptsetup-devices-state-present So this would be a breaking change. I would suggest to introduce a new state called "mapped" instead which does what you want. Please update the docs as well ;-)

inthehack:
Ok I got i @ypid. Thx for the review. I will deal with this next week.

inthehack:
@ypid @drybjed I pushed a new commit that should implement proposed suggestions from @ypid. Let me know if this requires more improvements. This docs should be up to date too.

ypid:
@jpeeters Thanks for the update. I would like to review and test this a bit. Will probably take a few weeks.

ansible/roles/debops.cryptsetup/tasks/manage_devices.yml
```diff
@@ -269,7 +269,8 @@
     opts:   '{{ item.0.format_options | d(omit) }}'
   when: (item.1|d() and item.1.stat|d() and item.1.stat.exists|d() and
          (item.0.create_filesystem|d(item.0.manage_filesystem|d(True)) | bool) and
-         not (item.0.swap|d(False) | bool))
+         not (item.0.swap|d(False) | bool) and
+         not (item.state|d(cryptsetup__state) in ['mapped'])
```
ypid, line 273: Missing `)`.

ypid (changes requested):
Thanks for the update. I have an issue with this patch because it does not "compile" so I assume it is not so well tested. After I fixed the "compile error", the "Manage fstab and mount state of the plaintext device mapper targets" task fails. This task also needs to be excluded from the "mapped" state. I stopped my review of the tasks file at this point.

Regarding the docs, I know they are quite extensive and not easy to change. It would be great if you could read defaults-detailed.rst as a whole and see if any other options of the role might be affected by the new state. At a minimum the difference between manage_filesystem and the mapped state should be mentioned. Bonus points for a full example showing an actual use case for the `mapped` state. This role is quite complex and I feel it needs easy and practical examples for it to be usable.

PR #727, opened 2019-03-25
[debops.ferm]: Allow saddr and daddr to be present in the same rule

monsterry:


drybjed (approved):
Sounds good, thanks! :+1:

PR #728, opened 2019-03-27
Update default nameConstraints to allow subdomains

imrejonk:
'permitted;DNS:${config_domain}' only allows names which exactly match
${config_domain}. 'permitted;DNS:.${config_domain}' (notice the extra ".") only
allows expanded labels, but not ${config_domain} itself. Let's have the best of
both worlds by combining the two name constraints together, which allows both
${config_domain} and expanded labels.

OpenSSL throws `error 47 at 0 depth lookup: permitted subtree violation; error
hcert.pem: verification failed` when using this role with critical
nameConstraints. That's why I removed the 'critical' property. This might be
better for backwards compatibility as well. Modern software will still refuse
to accept the certificate when the name is outside the nameConstraints space.
For example, Mozilla Firefox 60.6.1esr-1~deb9u1 will fail to connect with
'SEC_ERROR_CERT_NOT_IN_NAME_SPACE', and curl 7.52.1-5+deb9u9 fails with '(60)
SSL certificate problem: permitted subtree violation'.

### DIFF
--- a/ansible/roles/debops.pki/defaults/main.yml
+++ b/ansible/roles/debops.pki/defaults/main.yml
@@ -530,6 +530,15 @@ pki_default_authority: 'domain'
 pki_ca_name_constraints: True
 
                                                                    # ]]]
+# .. envvar:: pki_ca_name_constraints_critical [[[
+#
+# Specify if the ``nameConstraints`` CA certificate extension should be
+# marked critical by default. Enabling this when using ``nameConstraints`` is
+# recommended by the CA/Browser Forum.
+# Ref: https://cabforum.org/wp-content/uploads/CA-Browser-Forum-BR-1.6.4.pdf
+pki_ca_name_constraints_critical: True
+
+                                                                   # ]]]
 # .. envvar:: pki_ca_domain [[[
 #
 # The DNS domain used by default for an Certificate Authority configuration if
--- a/ansible/roles/debops.pki/files/secret/pki/lib/pki-authority
+++ b/ansible/roles/debops.pki/files/secret/pki/lib/pki-authority
@@ -138,7 +138,7 @@ get_openssl_name_constraints_directive () {
     local name_constraints
     case "${config['name_constraints']}" in
         true|True)
-            name_constraints="nameConstraints         = critical, permitted;DNS:${config_domain}"
+            name_constraints="nameConstraints         = permitted;DNS:${config_domain},permitted;DNS:.${config_domain}"
             ;;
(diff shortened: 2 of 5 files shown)


drybjed:
The [Baseline Requirements for the Issuance and Management of Publicly-Trusted Certificates](https://cabforum.org/wp-content/uploads/CA-Browser-Forum-BR-1.6.4.pdf) (PDF) document released on March 16, 2019 contains the following:

> f. nameConstraints (optional)
> If present, this extension SHOULD be marked critical*.
> * Non-critical Name Constraints are an exception to RFC 5280 (4.2.1.10), however, they MAY be used until the Name Constraints extension is supported by Application Software Suppliers whose software is used by a substantial portion of Relying Parties worldwide.

It seems to me that use of the `critical` status is recommended by the CA/Browser Forum, but in specific cases it can be omitted, at least that's what the `SHOULD` keyword means according to [RFC 2119](https://www.ietf.org/rfc/rfc2119.txt).

I think that a proper solution would be to implement a parameter for the `pki-authority` script to optionally enable or disable the criticality of `nameConstraints` in the generated CA certificates, with the default being enabled. It will take some time, at least for me, to get around to that, so in the meantime I would go with this proposal and merge it for now. The certificates managed by DebOps PKI internal CA are meant to be for internal use, and they are not trusted publicly by default, so not having a critical `nameConstraints` shouldn't hurt.

Any other opinions in this matter? @ypid, do you agree with my assessment?

ypid:
I would go for controllability directly. @imrejonk can you give that a try?

imrejonk:
Ah, I was not familiar with that document. Looks like it's indeed better to have nameConstraints enabled by default.

I'm introducing a new variable for the debops.pki role: `pki_ca_name_constraints_critical`, with a default value of `True`. `pki_authorities` will have a corresponding attribute.

imrejonk:
About the OpenSSL error I got earlier: that appears to be caused by a syntax error of mine. Running the playbook twice works fine now, unlike the [problems](https://github.com/debops/debops/commit/04bdf274c2485bb7c3ffbd9ac144bae1943fa1ba) @drybjed had before. I think that error was thrown because the base domain was excluded from the nameConstraints, causing OpenSSL to complain (for good reason).

drybjed (dismissed):
Nicely done. :+1:

ypid (approved):
LGTM, thanks! I have not tested it through.

drybjed (approved):

PR #731, opened 2019-03-29
Add nginx_multi_accept option

rei-ifesca:
Enable a worker proccess accept all new connections at a time.

This PR solves an issue with nginx under low load with long polling HTTP connections, waiting for a response from the backend.

Setting ```multi_accept``` to ```on``` reduces the CPU load of the nginx processes.


drybjed (approved):
Sounds good, thanks. Do you know if enabling it by default has any drawbacks? I'll merge it as-is for now, though.

rei-ifesca:
It depends on your setup. I choosed 'off' as default just like the nginx documentation. 

If you have long polling connections of several seconds I would recommend set it on. 

Thanks for merging.

PR #735, opened 2019-04-01
[debops.ferm] Fix forced rule "filter_icmp_flood"

### LINKED ISSUE #734: [debops.ferm] Can't turn off hashlimit for rule "filter_icmp_flood"
~~I think this line should be replaced:
https://github.com/debops/debops/blob/master/ansible/roles/debops.ferm/defaults/main.yml#L455
from `enabled` to `rule_state`
Similar to the other default rules.~~

Edit:
I tried to add `ferm__filter_icmp: False`, like described in the documentation. But it has no effect. I can still find a `mod hashlimit` line in the corresponding file:

`/etc/ferm/rules.d/015_rule_filter_icmp_flood.conf`
```
# This file is managed remotely, all changes will be lost

domain (ip ip6) table filter chain INPUT {
    protocol icmp @subchain "hashlimit-filter_icmp_flood" {

        mod hashlimit    hashlimit 10/second
                         hashlimit-burst 10
                         hashlimit-mode srcip
                         hashlimit-name filter_icmp_flood
                         hashlimit-htable-expire 3600000

            ACCEPT;

        &log("ipt-hashlimit-filter_icmp_flood: ");

        DROP;
    }
}
```

simonsystem:
~~Will close https://github.com/debops/debops/issues/734.~~
Replaces the unused `enabled` attribute with a correct `rule_state` in the default rule "filter_icmp_flood". Now it's possible to remove the rule using:

    ferm__filter_icmp: False

### DIFF
--- a/ansible/roles/debops.ferm/defaults/main.yml
+++ b/ansible/roles/debops.ferm/defaults/main.yml
@@ -452,7 +452,7 @@ ferm__default_rules:
     type: 'hashlimit'
     weight_class: 'filter-icmp'
     protocol: 'icmp'
-    enabled: '{{ ferm__filter_icmp | bool }}'
+    rule_state: '{{ "present" if (ferm__filter_icmp | bool) else "absent" }}'
     hashlimit: '{{ ferm__filter_icmp_limit }}'
     hashlimit_burst: '{{ ferm__filter_icmp_burst }}'
     hashlimit_expire: '{{ ferm__filter_icmp_expire }}'


simonsystem:
Okay, maybe is misunderstood the use of `enabled`. Can anyone explain, why this rule is only disabled instead of completely deleted?
Also, the `enabled` attribute isn't documented anywhere.

drybjed:
You're right, it looks like the `rule_state` for the icmp filter rule was missed somehow during the rewrite. Thanks for the bug report, I'll fix it right away.

PR #745, opened 2019-04-12
[debops.logrotate] keep newlines in config

imrejonk:
These changes remove the Jinja 'strip' operations from the logrotate templates. Logrotate does not like missing newlines and presented us with errors like these:

```
/etc/cron.daily/logrotate:
error: php7.0-fpm:prerotate, postrotate or preremove without endscript
error: found error in file php7.0-fpm, skipping
error: rsyslog:prerotate, postrotate or preremove without endscript
error: found error in file rsyslog, skipping
error: rsyslog-remote:prerotate, postrotate or preremove without endscript
error: found error in file rsyslog-remote, skipping
error: /etc/logrotate.conf:14 bad rotation count '1}'
error: found error in /var/log/wtmp , skipping
```

These errors were all caused by missing newlines. For example, 'endscript' and the curly bracket after the '1' must be on new lines.
I'm not sure why the 'strip' operation was in the templates in the first place. Maybe I'm overlooking something, but this seems to work for us.

### DIFF
--- a/ansible/roles/debops.logrotate/templates/etc/logrotate.conf.j2
+++ b/ansible/roles/debops.logrotate/templates/etc/logrotate.conf.j2
@@ -1,16 +1,16 @@
 # {{ ansible_managed }}
 
 {% if logrotate__options %}
-{{ logrotate__options -}}
+{{ logrotate__options }}
 {% endif %}
 {% if logrotate__group_options %}
-{{ logrotate__group_options -}}
+{{ logrotate__group_options }}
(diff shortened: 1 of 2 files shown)


drybjed:
This error is present in environments with Jinja <2.10, with Jinja 2.10 your modifications result in an additional empty line in each configuration section. But it seems like the solution that was used previously  is not really effective; I'll try to come up with a better one that doesn't use the Jinja stripping.

PR #757, opened 2019-04-18
[debops.opendkim] Add pyOpenSSL

imrejonk:
I was greeted with an error when I tried out the debops.opendkim role on a Debian 9 box (with Ansible 2.7.5) this morning:

```
failed: [myhostname -> localhost] (item={u'name': u'mail'}) => {"changed": false, "item": {"name": "mail"}, "msg": "the python pyOpenSSL module is required"}
```

Turns out the 'python-openssl' package was missing. This package contains the pyOpenSSL module needed to generate the DKIM keys. These changes add this package to the `opendkim__base_packages` variable.

### DIFF
--- a/ansible/playbooks/service/opendkim.yml
+++ b/ansible/playbooks/service/opendkim.yml
@@ -30,5 +30,12 @@
           config: '{{ opendkim__postfix__dependent_maincf }}'
       when: opendkim__postfix_integration|bool
 
+    - role: debops.python
+      tags: [ 'role::python', 'skip::python' ]
+      python__dependent_packages3:
+        - '{{ opendkim__python__dependent_packages3 }}'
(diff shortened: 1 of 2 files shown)


drybjed:
Thanks. Perhaps this was not encountered before because some other role installed this APT package.

However, it would be better to handle this via role dependent variables and `debops.python` role. This role supports installing Python 2.7 and Python 3.x packages conditionally, with aim to disable Python 2 entirely in the future. You can check some other DebOps roles to see how it's done. If you're not up to it, I can implement that in a few days.

imrejonk:
Ah right, yes that would be better. I'll give it a shot.

imrejonk:
How's this?

drybjed (approved):
Looks great, thanks!

PR #763, opened 2019-05-01
[debops.nginx] Only reference named location when we will generate it

ypid:
Prevents error: could not find named location "@redirect-acme-challenge"

### DIFF
--- a/ansible/roles/debops.nginx/templates/etc/nginx/snippets/acme-challenge.conf.j2
+++ b/ansible/roles/debops.nginx/templates/etc/nginx/snippets/acme-challenge.conf.j2
@@ -17,7 +17,9 @@ location ^~ /.well-known/acme-challenge/ {
 location @well-known-acme-challenge {
         root {{ nginx_acme_root }};
         default_type "text/plain";
+{% if nginx_acme_domain %}
         try_files $uri @redirect-acme-challenge;
+{% endif %}
 }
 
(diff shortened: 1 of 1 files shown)


ansible/roles/debops.nginx/templates/etc/nginx/snippets/acme-challenge.conf.j2
```diff
@@ -17,7 +17,9 @@ location ^~ /.well-known/acme-challenge/ {
 location @well-known-acme-challenge {
         root {{ nginx_acme_root }};
         default_type "text/plain";
+{% if nginx_acme_domain %}
         try_files $uri @redirect-acme-challenge;
+{% endif %}
```
drybjed, line 22: Shouldn't the `try_files $uri;` be present if the `nginx_acme_domain` is empty? Because in that case the configuration section will not have `try_files` at all, not sure what `nginx` does in this case.

ypid: I tried that but try_files requires two or more arguments. It worked in my test on stretch without try_files.

drybjed: Very well then, merging.

drybjed (approved):

PR #764, opened 2019-05-01
[debops.owncloud] Small fixes

ypid:
I don’t fully understand the error to be honest. `owncloud__do_autosetup` is defined. ansible 2.7.5

```
fatal: [host.example.org]: FAILED! => {"msg": "The conditional check 'owncloud__do_autosetup|bool' failed. The error was: error while evaluating conditional (owncloud__do_autosetup|bool): 'owncloud__do_autosetup
' is undefined\n\nThe error appears to have been in '/home/user/.ansible/ypid-ansible-common/submodules/debops/ansible/roles/debops.owncloud/tasks/setup_owncloud.yml': line 153, column 3, but may\nbe elsewhere i
n the file depending on the exact syntax problem.\n\nThe offending line appears to be:\n\n# Run occ commands as specified in the inventory [[[\n- name: Run occ commands as specified in the inventory\n  ^ here\n"
}
```

### DIFF
--- a/ansible/roles/debops.owncloud/defaults/main.yml
+++ b/ansible/roles/debops.owncloud/defaults/main.yml
@@ -417,7 +417,6 @@ owncloud__variant_name_map:
 #
 # * ownCloud ``10.0``
 #
-# * Nextcloud ``13.0``
 # * Nextcloud ``14.0``
 # * Nextcloud ``15.0``
 #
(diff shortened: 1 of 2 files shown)


drybjed (approved):
Sounds good. I'll merge it and let's see how the GitLab CI tests go.

PR #769, opened 2019-05-02
Add 'debops.yadm' role

drybjed:


drybjed:
The error on Travis seems to be a temporary one, related to Docker container build. I'll ignore it since it shouldn't have an impact on the contents of the PR.

PR #770, opened 2019-05-03
[debops.roundcube] Always fetch JS, fixes upgrades

amette:
This needs to be run on upgrades.

This can also fix broken installs, which actually was the original issue that got me into debugging this.

Also it doesn't hurt to fetch the JS, should be totally idempotent as long as the previous run was successful also.


drybjed:
Sounds good, but removing this parameter will make the task non-idempotent. Can you add a check for changes, for example with `changed_when:` and `register:`?

PR #771, opened 2019-05-06
Redesigned dotfiles management

drybjed:


drybjed:
The issue again seems to be with a Docker container build. I'll ignore the error in this case, since the PR does not modify the Dockerfile.

PR #773, opened 2019-05-13
Small doc changes to debops.pki.

htgoebel:
I hope I got the commit messages right.

### DIFF
--- a/docs/ansible/roles/debops.pki/acme-integration.rst
+++ b/docs/ansible/roles/debops.pki/acme-integration.rst
@@ -198,7 +198,7 @@ Names`_), which does not include the ``example.com`` apex (root) domain.
         # Can also include different domains like 'mail.example.org'
         # in the same realm.
         acme_domains: [ 'mon.example.com' ]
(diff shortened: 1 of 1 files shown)


docs/ansible/roles/debops.pki/acme-integration.rst
```diff
@@ -199,6 +200,10 @@ Names`_), which does not include the ``example.com`` apex domain.
         acme_domains: [ 'mon.example.com' ]
         # acme_ca: 'le-staging'
```
drybjed: You could also update this example, since the `le-staging` version is not usable anymore. Replace this with `le-staging-v2`.

drybjed (changes requested):
Looks good. :+1: Do you want to also update the commented out example?

docs/ansible/roles/debops.pki/acme-integration.rst
htgoebel: Done.

drybjed (approved):

PR #777, opened 2019-05-13
[debops.authorized_keys] Fix bug when primary group != name

ypid:


drybjed (approved):
Good idea, and nice implementation. I didn't know that using UID/GID instead of the user/group names will work via Ansible (or at all), but I guess it does.

This could be a security issue when shared groups are used, for example `users` instead of each user having a primary group. But I'm not sure there's anything we could do in this case, perhaps using ACLs instead to give access to the user accounts.

tumd:
I haven't looked into this hypothesis in depth, but does the read-only mode actually work with "StrictModes yes" set in `sshd_config`? At least it doesn't seem to work with sshd 7.9p1 in Buster. :/

drybjed:
@tumd: I just checked on Debian Stretch and the current setup seems to work fine with StrictModes yes, not sure about Debian Buster. Quick google search doesn't show any reports about this happening.

Could you check with `LogLevel DEBUG` set in `/etc/ssh/sshd_config` what's happening behind the scenes?

tumd:
If the authorized_keys-file isn't owned by the user trying to login, it will fail with "Permission denied".
@drybjed: Here you go;

```
# ls -l /etc/ssh/authorized_keys/repo
-r-xr-x--- 1 root repo 1663 May 19 21:31 /etc/ssh/authorized_keys/repo
# grep "/etc/ssh/authorized_keys/repo" /var/log/auth.log
May 20 19:32:05 mort sshd[25348]: debug1: trying public key file /etc/ssh/authorized_keys/repo
May 20 19:32:05 mort sshd[25348]: debug1: Could not open authorized keys '/etc/ssh/authorized_keys/repo': Permission denied
# chown repo /etc/ssh/authorized_keys/repo
# ls -l /etc/ssh/authorized_keys/repo
-r-xr-x--- 1 repo repo 1663 May 19 21:31 /etc/ssh/authorized_keys/repo
# grep "/etc/ssh/authorized_keys/repo" /var/log/auth.log
May 20 19:32:05 mort sshd[25348]: debug1: trying public key file /etc/ssh/authorized_keys/repo
May 20 19:32:05 mort sshd[25348]: debug1: Could not open authorized keys '/etc/ssh/authorized_keys/repo': Permission denied
May 20 19:35:05 mort sshd[25357]: debug1: trying public key file /etc/ssh/authorized_keys/repo
May 20 19:35:05 mort sshd[25357]: debug1: /etc/ssh/authorized_keys/repo:2: matching key found: RSA SHA256:...
```

Edit: I might have screwed up a bit. Investigating a bit more. :)

tumd:
Never mind my previous rambling. :)
The created user 'repo' didn't have it's own primary group set. Instead it had 'users' set as it's primary group. Setting the primary group to 'repo' solves it, and a read-only authorized_keys work as expected.

Something must have changed in either the user_module in Ansible or `useradd` in Buster since I last ran this old role of mine. Now a days apparently the 'group'-option must be set in user_module, else the primary group will become 'users'.

drybjed:
Good to hear you found the cause. BTW, the change in this PR should have solved this for you, since role now checks the primary group of a given UNIX account and sets that as the file group. So, the file should have been owned by `root:users` for authentication to work.

I'm currently working on updated `debops.users` role which relies on `libuser` to manage the accounts. Because of that the behaviour of the role will change - it will create private group for a given user automatically unless `item.group` parameter is specified as some existing UNIX group. The changes will be merged within a day or two.

PR #778, opened 2019-05-13
Various minor fixes

ypid:


ansible/roles/debops-contrib.kodi/tasks/main.yml
```diff
@@ -30,3 +30,18 @@
...
+    owner: 'root'
+    group: 'root'
+    mode: '0644'
+  when: (kodi__polkit_action != "")
+
+- name: Remove polkit configuration
+  file:
+    path: '/etc/polkit-1/localauthority/50-local.d/kodi-actions.pkla'
+    state: 'absent'
+  when: (kodi__polkit_action == "")
```
drybjed: You could try to replace that with:
```yaml
when: not kodi__polkit_action|d()
```
It should have the same result.

ypid: Thanks for the hint. Fixed.

drybjed (approved):
Looks good, thanks for fixing so many misspellings. :-)

ypid:
You are welcome ;-)

PR #780, opened 2019-05-19
removed hardcoded deploy_path_mode

damko:
I noticed that deploy_path_mode was hard-coded 0750 so I added a variable

### DIFF
--- a/ansible/roles/debops.owncloud/defaults/main.yml
+++ b/ansible/roles/debops.owncloud/defaults/main.yml
@@ -548,6 +548,9 @@ owncloud__temp_path: ''
 # Where the ownCloud instance will be deployed (web root).
 owncloud__deploy_path: '{{ owncloud__app_home }}'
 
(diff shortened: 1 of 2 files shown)


ansible/roles/debops.owncloud/defaults/main.yml
```diff
@@ -548,6 +548,9 @@ owncloud__temp_path: ''
 # Where the ownCloud instance will be deployed (web root).
 owncloud__deploy_path: '{{ owncloud__app_home }}'
 
+# .. envvar:: owncloud__deploy_path_mode
+#
+owncloud__deploy_path_mode: '0750'
```
drybjed, line 554: Could you add a description of this variable in the comment section as well? it will look better on the documentation page.

damko: sure

drybjed (approved):
Looks good, thanks.

PR #781, opened 2019-05-19
[debops.rsyslog] Don't rely on SysV init script

tumd:
Since Buster, rsyslog has a new helper script to close open log files.
Same for Ubuntu since at least 16.04.


drybjed (approved):
Sounds good, thanks for the update. :+1:

PR #782, opened 2019-05-21
Modernize UNIX account management roles

drybjed:


ypid:
Nice work. I am still a bit worried about the duplicated documentation (and code) between `debops.users` and `debops.system_users`. Ref: https://github.com/debops/debops/pull/788

drybjed:
I tried to implement the LDAP support from `debops.system_users` (the user/group prefix, different home directory root) in the `debops.users` itself so that only one role would exist, but having that functionality activated for all accounts in the role in the LDAP environment felt wrong to me. I know that lots of people use 'debops.users' to manage application accounts on hosts (myself included) via Ansible inventory and I felt that having the usernames and home paths change in an unexpected way might be confusing and could complicate things.

The `debops.users` role is meant to be used in small environment, in larger environments its role should be taken over by LDAP. In contrast to that, the `debops.system_users` is meant to be used in both small and large environments with LDAP to manage local admin accounts but not much else.

Actually the `item.chroot` and `item.admin` parameters were implemented fairly late in the redesign. I'm thinking if the roles could be combined with LDAP functionality somehow toggleable with similar parameter, but I don't really see it working out. The "local admin account" / "regular user account" split seems to be more sensible to me. Thoughts?

ypid:
Sounds reasonable. I was more refering to the documentation ;-) It seems you copied big parts. Maybe just refer to parts which are the same?

drybjed:
I thought about it but there's no precedence for that yet, and I felt that without explicit examples that reference the correct variables, users might use the wrong ones and get confused which would result in issue reports and duplication of content to stop them. So I just went ahead and duplicated most of it.

ypid:
Fair enough. I will just fix typos in two places then ;-)

Keep up the awesome work! I am reviewing closely :-)

PR #784, opened 2019-05-21
[debops.docker_gen] Update docker_gen to 0.7.4

imrejonk:
0.7.4 is the latest docker-gen release, supporting IPv6 and docker networks. These changes also update the nginx template in order to support docker networks.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -143,6 +143,10 @@ Changed
     the default Stretch environment. The downloaded ``.deb`` package will be
     verified using the RStudio Inc. GPG signing key before installation.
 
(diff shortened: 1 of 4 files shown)


drybjed:
Thanks! Can you also update the `ansible/roles/debops.docker_gen/meta/watch-docker-gen` file so that the `make versions` command reports the version installed by the role? You could also add a mention of the version change in the changelog "Changed" section, there are a few similar entries there already.

imrejonk:
Well that's a nifty tool, glad we're using it :)
How's this?

drybjed (approved):
Nicely done, thanks!

PR #785, opened 2019-05-21
[multiple roles] add NCSC-NL cipher suite

imrejonk:
The Dutch National Cyber Security Centre published updated TLS
guidelines last month. This version was created in collaboration with
the national communication security agency, with contributions from
numerous organizations and individuals.

One of the changes in this version is the cipher suite selection. This
commit adds the 'good' cipher suite from the document to debops.apache,
debops.dovecot and debops.nginx.

I'm not sure if this is the best way to add new ciphers, especially since they have to be added in multiple places. Can't we do this with Ansible facts or something?


drybjed (approved):
Very interesting, thanks. And yes, these should be centralized somehow. I'll probably create a set of local facts in the `debops.pki` role with cipher lists for `openssl` and `gnutls` libraries, as well as TLS protocols to use. Then other roles could be hooked up to these facts automatically. But that's something for the next release, for now this PR will do.

PR #787, opened 2019-05-21
[debops.unattended_upgrades] Use random-but-idempotent VMs reboot times

ypid:


drybjed (approved):
Interesting idea, thanks.

Issue #792, opened 2019-05-22
Create a 'debops.borgbackup' role


ypid:
Hi

I am working on a role for [BorgBackup](https://www.borgbackup.org/) which is intended for backing up the whole system. It currently lives here https://github.com/ypid/debops/tree/add/debops.borgbackup. The role should already be somewhat major (I already use it in production) and I will add extensive examples to the docs of the role soon. It tightly plugs into other DebOps roles to configure a cluster-wide automated backup. I guess when some docs are added we can merge it into the mono repo. I currently don’t have the time to get the role fully integrated into DebOps unfortunately.

drybjed:
Sounds like a good idea. I'll look the role over in a bit, although you could start by creating a branch in forked `debops/debops` repository. And I would call the role `debops.borgbackup` if it's going straight to the monorepo. :-)

ypid:
Thanks for your input. I removed the individual repo and added the role to https://github.com/ypid/debops/tree/add/debops.borgbackup. My idea was originally to merge the git repo into the mono repo later with history like the monorepo import was done. Anyway. This works for me.

I will need to rebase https://github.com/ypid/debops/tree/add/debops.borgbackup ontop of https://github.com/debops/debops/pull/794 to make the role work.

tumd:
Quite some time ago I started work on a role for setting up backup of hosts to one or many backup-servers using borg. I have put quite some work into the role, but it's currently not in a working state, and have been untouched for some time.

The role it self mainly does it's job on the client;
- preparing the binaries
- installing a quite simple but helpful borg-wrapper shell-script that mainly reads one (of potentially many) config-files. One config for each backup-server.
- sets up a client-unique user-account & ssh on one or more predefined backup-servers with help of `delegate_to` from the client.
- initializes the repo on each backup-server.
- exposes information of each initialized repo via local fact-scripts.

The role is currently not published anywhere. But let me know if it may be of any interest and I'll publish it somewhere.

ypid:
@tumd Have you peeked at my implementation yet? Maybe our two roles can merge so that we have one that works for all of us? Anything that can be improved in my role? My work has a similar feature set except:

* preparing the binaries. (No need because package managers are preferred, ref: https://docs.debops.org/en/master/dep/dep-0003.html)
* installing a quite simple but helpful borg-wrapper shell-script that mainly reads one. ("Not invented here"? There is already a perfectly fine and healthy Python project which does this – borgmatic).
[...]

PR #794, opened 2019-05-22
[debops.ansible_plugins] Refactor debops_filter_plugins.py code and add integration tests

### LINKED ISSUE #776: [debops.ansible_plugins] Rework and unit test filter_plugins/debops_filter_plugins.py
I started using `parse_kv_items` and `parse_kv_config` heavier and need to extend them as I want them to support more nesting levels. The functions share some code and are not unit tested. This makes it difficult to extend them. I am in the process of reworking them.

Maybe give me a few days until I have extended, unit tested and tested them with regular DebOps usage :)

ypid:
This is a start. Further refactoring could be done as there is still quite some redundant code. It was not easy for me without tests so I needed to write some first.

@drybjed If you could do me one favor it would be less redundant code. That is all I could ask for. OK, maybe 100 % code coverage by unit testing but lets not set the bar too high for now ;-)

Required for: #792
Closes: #776

### DIFF
--- a/ansible/roles/debops.ansible_plugins/filter_plugins/debops_filter_plugins.py
+++ b/ansible/roles/debops.ansible_plugins/filter_plugins/debops_filter_plugins.py
@@ -102,7 +102,8 @@ def _parse_kv_value(current_data, new_data, data_index):
                             'weight': dict_element.get('weight', 0),
                             'state': 'present'})
 
-                        dict_element['real_weight'] = _get_real_weight(dict_element)
+                        dict_element['real_weight'] = _get_real_weight(
+                                dict_element)
                         dict_value[element] = dict_element
                         current_data['value'] = dict_value
 
@@ -131,7 +132,8 @@ def _parse_kv_value(current_data, new_data, data_index):
                                     dict_element.get('weight', 0))) +
                                 int(dict_element.get('weight', 0)))
 
-                        dict_element['real_weight'] = _get_real_weight(dict_element)
+                        dict_element['real_weight'] = _get_real_weight(
+                                dict_element)
 
                         # Include any unknown keys.
                         for key in element.keys():
@@ -174,13 +176,17 @@ def parse_kv_config(*args, **kwargs):
                                  else {})
 
                 if element.get('state', 'present') == 'append':
-                    current_param['state'] = current_param.get('state', 'present')
+                    current_param['state'] = current_param.get(
+                            'state', 'present')
                 else:
                     current_param['state'] = (
-                        element.get('state', current_param.get('state', 'present')))
+                        element.get('state', current_param.get(
+                            'state', 'present')))
 
                 if (current_param['state'] == 'init' and
-                        (element.get('state', 'present') != 'init' and _check_if_key_in_nested_dict('value', current_param))):
+                    (element.get('state', 'present') != 'init' and
+                        _check_if_key_in_nested_dict(
+                        'value', current_param))):
                     current_param['state'] = 'present'
 
                 current_param.update({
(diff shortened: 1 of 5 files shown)


ypid:
@drybjed Would it make sense to use a newer Ubuntu version for CI testing?

I do that in another project of mine: https://github.com/opening-hours/opening_hours.js/blob/master/.travis.yml

drybjed:
@ypid Sure, switching to `xenial` or `bionic` on Travis sounds like a good idea, I guess that's available now?

ypid:
Finally passing.

At least xenial works for me. Currently we use 14.04 in DebOps which was released over 5 years ago – ancient :wink:

ansible/roles/debops.ansible_plugins/filter_plugins/debops_filter_plugins.py
```diff
@@ -430,3 +422,434 @@ def filters(self):
...
+            expected_data = yaml.safe_load(textwrap.dedent('''
+            name: local
+            value:
+              beta:
+                id: 0
+                name: beta
+                real_weight: 0
+                state: present
+                weight: 0
+            '''))
```
drybjed, line 477: I think that this might be a bug. From the top of my head I don't remember any roles that depend on this data structure, but I would think that this should be a list, and not a dictionary, since the `value`'s value is a YAML list. Then the "value" can be extracted from each dictionary `name` key, which is done in a bunch of roles already.

ansible/roles/debops.ansible_plugins/filter_plugins/debops_filter_plugins.py
```diff
@@ -24,44 +25,75 @@
...
-                current_data.update({'value': new_value})
+                current_data['value'] = new_value
 
-            if (old_value is not None and old_state in ['comment'] and
-                    current_data['state'] != 'comment'):
-                current_data.update({'state': 'present'})
+            # TODO(drybjed): This never evaluates to true.
+            #  if (old_value is not None and old_state in ['comment'] and
+            #          current_data['state'] != 'comment'):
+            #      current_data['state'] = 'present'
```
drybjed, line 87: I need to write more comments in my code. :-) This IIRC was meant to activate a given configuration entry that is commented out in role defaults if somebody defined an entry with the same `name` and some value, basically switching the state from `comment` to `present` without explicitly defining the `state` parameter. I suspect that this might be too much magic, so perhaps this should be removed altogether. Let's leave it commented out now to see if anything breaks.

drybjed (commented):
@ypid Awesome work! I need to learn more about unit tests and how that should be done. Including tests in a script is a neat trick. It seems that the script has a severe lack of comments, I think that I have some bias against adding lots of comments to not slow down the interpreter.

I'll test the PR in my current setup and see if anything breaks.

ansible/roles/debops.ansible_plugins/filter_plugins/debops_filter_plugins.py
ypid: Works for me. Where is the difference between the `init` and the `comment` state then?

ypid:
> I think that I have some bias against adding lots of comments to not slow down the interpreter.

I had the same fear once but I am pretty sure that this does not matter with modern computers. Feel free to add thousands of comment lines. At least Python won’t really mind.

drybjed:
Looks like [all roles passed successfully](https://gitlab.com/drybjed/debops/pipelines/62973097) the tests on GitLab. @ypid, let me know if you want to change anything else or is this PR ready for merge.

ansible/roles/debops.ansible_plugins/filter_plugins/debops_filter_plugins.py
drybjed: Difference is that if an entry has an `init` state, it would not show up in the generated config file at all unless it was redefined in the Ansible inventory. With the `comment` state it will show up, but commented out, unless it is redefined in the inventory. The difference is mostly cosmetic I suppose, for example if you want to keep some of the commented out parameters in the Postfix `main.cf` configuration file unless they are changed by the user.

ypid: Thanks. I only used `init` for now.

ypid:
Ready for merging.

drybjed (approved):
All right, let's do this and see what breaks. Thanks for your work, @ypid. :-)

PR #796, opened 2019-05-23
[debops.slapd] Re-introduce slapd__default_services

monsterry:
It is essential to be able to configure this setting if you set
olcServerId or want to bind to a specific ip address


drybjed:
Sounds good, however the `/etc/default/slapd` file is a Conffile, tracked by Debian, so it will have to be diverted and role will have to template it. I'll try to add this in the next few days.

Do you think that `ldaps:///` port should be enabled by default? It's apparently deprecated, and `ldap:///` + StartTLS should be used instead.

monsterry:
Perfect bugfix. Thanks!.

PR #807, opened 2019-05-28
Fix `yadm pull`-output to match on buster

tumd:
Using a newer git-package from stretch-backports or buster
and performing `git pull` in an already up-to-date git-repository,
returns no dashes;
'Already up to date.'


drybjed (approved):
Awesome, thanks! This sounds like an idempotency issue, so I'll put it in the stable release as well.

PR #809, opened 2019-05-29
Small tweaks and fixes in custom Ansible plugins

drybjed:


ypid:
> Don't duplicate 'options'

This breaks my use case from #792. Do you have more background for this change?

My use case: Search for "keep_within" in https://github.com/ypid/debops/blob/add/debops.borgbackup/ansible/roles/debops.borgbackup/defaults/main.yml

drybjed:
This change should just ensure that the `options` key is always merged, even if it's not specified in the `merge_keys` parameter. I suppose that if it would be better to be able to disable this entirely, I can go for it - it would just mean that `options` would need to be specified explicitly if `merge_keys` is defined. Do you think that's a better approach?

As for your use case, care to explain it a bit more? All I see is that all of the various options are defined on a sub-list on the `config` dictionary. I don't see any other entries on the `config` level, from what I'm seeing for now you could move all of the different sub-entries to the `config` level and it would work fine. Do you plan to add some more entries besides `config`?

ypid:
I added docs for the BorgBackup role. Ref: #835

> I suppose that if it would be better to be able to disable this entirely, I can go for it - it would just mean that options would need to be specified explicitly if merge_keys is defined. Do you think that's a better approach?

That script still looks a bit scary to me. I honestly don’t know exactly what would be the best approach. I initially wanted to fully rewrite the filter and remove all the redundancy but stopped because of time constraints when my use case was working and I added a few unit tests to improve the situation. Maybe you can check the new role and see what the best approach would be.

> Do you plan to add some more entries besides config?

I don’t right now, at least not as role defaults but I can imaging situations where users want to add more. This can come in handy to only backup parts of the system for faster partly backups.

PR #816, opened 2019-06-02
[debops.gitlab_runner] Fix loop variable typo

tumd:
Fixes docker_services as a list.


drybjed (approved):
Good catch, thanks! This goes to the stable release as well.

PR #819, opened 2019-06-04
[debops.gitlab_runner] use = instead of : when setting the tmpfs

alinalexandru:
There was a typo when we added this feature.


drybjed (approved):
Right, good catch. It's TOML, not JSON.

PR #821, opened 2019-06-04
[debops.users] Readd users__default_shell which was removed in v1.0.0

ypid:
@drybjed Am I missing something? I think this option is still useful.


drybjed (approved):
Sounds good. I removed it when the [`libuser` support was implemented](https://github.com/debops/debops/commit/dcdd7423e9c9840ffe2e99b1c93efe37d4fbccb5#diff-2b27b1a5680d1c62dd6447713561fea4L54) because `libuser` setting `/bin/sh` if no shell was specified sounded OK to me - the `users__default_shell` was a fix for an empty shell if not specified. But if a default shell setting is still useful, why not. :-)

PR #827, opened 2019-06-10
[debops.nginx] Support to disable logging per Nginx server

ypid:
I was surprised to find that the role does not actually support to disable logging yet.


drybjed (approved):
Sounds good, thanks. :-)

PR #832, opened 2019-06-12
Use up-to-date SPDX license identifiers

ypid:
Fixes incorrect usage of GPL-3.0-or-later.

```shell
git ls-files -z "$(git rev-parse --show-toplevel)" | xargs --null -I '{}' find '{}' -type f -print0 | xargs --null sed --in-place --regexp-extended 's/GPL-3.0/GPL-3.0-only/g;'
```


drybjed (approved):
Sounds good. I'll backport this PR to the `stable-1.0` branch so it's covered as well.

PR #608, opened 2019-01-13
use ansible_python['executable'] as Python path

AnotherKamila:
As not-quite-proposed in AnotherKamila/bsdops@b480c463 .

---
This change is [Reviewable](https://reviewable.io/reviews/debops/debops/608)

PR #835, opened 2019-06-15
[debops.borgbackup] Add debops.borgbackup role

### LINKED ISSUE #792: Create a 'debops.borgbackup' role

ypid:
Closes: #792

Do not merge yet, it is not functional currently. Depends on discussion started in: https://github.com/debops/debops/pull/809#issuecomment-500164549


ansible/roles/debops.borgbackup/templates/etc/ansible/facts.d/mandos.fact.j2
```diff
@@ -0,0 +1,13 @@
+#!/usr/bin/python{{ '2' if (ansible_python_version is version_compare('3.5', '<')) else '3' }}
```
tumd: See #608
Also is the filename correct? mandos?

TuxCoder:
if you like to, I created also a borg backup role for debops, sadly never upstream it, it also get a bit old and would need polishing, you can take a look, probably you find something interesseting.

https://git.o-g.at/ansible/debops.borg_client
https://git.o-g.at/ansible/debops.borg_server

ypid:
Thanks @TuxCoder on short peek, only debops.borg_client seems to use `ionice` and `nice` that I have not checked if my role covers this.

This PR needs work which I can hopefully do in the next months since nobody else has jumped in yet.

docs/ansible/roles/debops.borgbackup/inventory/group_vars/debops_service_borgbackup.yml
```diff
@@ -0,0 +1,8 @@
+---
+# vim: foldmarker=[[[,]]]:foldmethod=marker
+
+sudo__group_sudoers:
+
+  - name: 'backup-mgt-borgmatic'
+    raw: |
+      backup-mgt ALL = (ALL) NOPASSWD: /usr/local/bin/borgmatic --create --stats
```
TuxCoder: I think the path should be dynamic, ether if it is install through package manager or though pip

docs/ansible/roles/debops.borgbackup/inventory/host_vars/backup.example.net/default.yml
```diff
@@ -0,0 +1,13 @@
+---
+# vim: foldmarker=[[[,]]]:foldmethod=marker
+
+# cron [[[
+
+cron__host_jobs:
+
+  'backup':
+    job: '/usr/local/bin/borgmatic --create --stats >> /var/log/backup.log 2>&1'
```
TuxCoder: I think the path should be dynamic, ether if it is install through package manager or though pip

TuxCoder (commented):
@ypid nice to see that you also implement the virtualenv method for older systems, but I'm not sure why you need it on buster, there is version 1.1.9-2 currently stable, I know you don't wanna use the stretch version, but I think buster should be new enough.
Can you describe your reason for that? IMHO is version 1.1.x new enough with the new encryption method, and other breaking changes.
What kind of feature do you need from the newest version?

Also borgbackup looks over the last 10 years very stable, but I normally don't like to install something from pip that can change on the next deploy, without any notice.

TuxCoder:
Another question:
why do you use a different user for each host?

I managed it like this: (user was created by the server role, then different sub folders)
https://git.o-g.at/ansible/debops.borg_client/src/branch/master/tasks/main.yml#L81


Else looks good, I think --append-only is not needed anymore in the newer versions as it has an config entry, but I haven't tested this new config entry yet.
Do you have currently some append only restrictions? Also not sure if Borgmatic does something automatically.

docs/ansible/roles/debops.borgbackup/inventory/group_vars/debops_service_borgbackup.yml
ypid: Read `SUDOERS(5)` ;)

As you said already, installing via pip is only done up until buster. After that, it will by installed from Debian. If it where not a workaround, `borgbackup__bin_path` could be used.

docs/ansible/roles/debops.borgbackup/inventory/host_vars/backup.example.net/default.yml
ypid: I would like to wait for bullseye to fix this. Does that work for you?

ypid:
> why do you use a different user for each host?

[Principle of least privilege](https://en.wikipedia.org/wiki/Principle_of_least_privilege), [Defense in depth ](https://en.wikipedia.org/wiki/Defense_in_depth_(computing))
> I managed it like this: (user was created by the server role, then different sub folders)

Possible but that still requires on `borg` enforcing this restriction.

> --append-only is not needed anymore in the newer versions as it has an 

Not sure what you mean. I find borgbackup’s `--append-only` interesting obviously, but that is not something that is used in this role yet. But it would allow that clients can only append and a dedicated backup controller does the data retention.

> Also borgbackup looks over the last 10 years very stable, but I normally don't like to install something from pip that can change on the next deploy, without any notice.

I feel the same as you. I also don’t like the idea that on next deployment, something other would happen as defined. This is not something I would accept easily. That is why I hash locked a specific git commit :) Did you overlook that? Note that the thing the role installs with pip is `borgmatic` not `borgbackup`.

> but I'm not sure why you need it on buster, there is version 1.1.9-2 currently stable, I know you don't wanna use the stretch version, but I think buster should be new enough.

As said, the thing the role installs with pip is `borgmatic` not `borgbackup`.

But in case you wonder about borgmatic, which is also in Debian. That version is the first which includes `validate-borgmatic-config` (ref: https://projects.torsion.org/witten/borgmatic/pulls/161), something we want to have.

> Else looks good, I think

This role was basically ready for merging in 2019. Only me insisting on having the configuration format of the role very flexible kept it from getting considered for merging. Note that it still collides with #809. Something I have not thought of when I rebased. So when you test this PR, be sure to revert #809 or better, fix the collision :)

PR #836, opened 2019-06-17
Update lxc-net.j2

chekolyn:
Template for lxc-net.j2 is failing with this message unsupported operand type(s) for +: 'bool' and 'str'"


drybjed (approved):
Thanks. It seems to be caused by [the issue with the `ipaddr` filter in Ansible](https://github.com/ansible/ansible/issues/55570), hopefully it will be fixed soon. I'll backport this fix to the current stable release as well.

PR #842, opened 2019-06-24
[debops.sshd] Add support for 'sshusers' in LDAP

pedroluislopez:
Init LDAP directory playbook now create a LDAP sshusers group.


drybjed (approved):
Looks good, thanks. I'll include it in the stable release as well.

PR #851, opened 2019-06-26
debops.nginx/defaults-detailed.rst: Document frame_options.

apapsch:
This option seems to be missing from the docs.


drybjed (approved):
Thanks for updating the documentation! :-)

PR #866, opened 2019-07-05
[debops.dovecot] Add local facts

amette:
Without these Postfix doesn't configure SASL auth.

### DIFF
--- a/ansible/roles/debops.dovecot/tasks/main.yml
+++ b/ansible/roles/debops.dovecot/tasks/main.yml
@@ -19,6 +19,27 @@
   register: dovecot__register_packages
   until: dovecot__register_packages is succeeded
 
+- name: Make sure Ansible local facts directory exists
+  file:
+    dest: '/etc/ansible/facts.d'
+    state: 'directory'
+    owner: 'root'
+    group: 'root'
+    mode: '0755'
+
+- name: Configure Dovecot local facts
+  template:
+    src: 'etc/ansible/facts.d/dovecot.fact.j2'
+    dest: '/etc/ansible/facts.d/dovecot.fact'
+    owner: 'root'
+    group: 'root'
+    mode: '0755'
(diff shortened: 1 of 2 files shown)


ansible/roles/debops.dovecot/tasks/main.yml
```diff
@@ -19,6 +19,27 @@
   register: dovecot__register_packages
   until: dovecot__register_packages is succeeded
 
+- name: Make sure Ansible local facts directory exists
+  file:
+    dest: '/etc/ansible/facts.d'
+    state: 'directory'
+    owner: 'root'
+    group: 'root'
```
drybjed, line 27: You should remove the `owner` and `group` keywords from the tasks, since the role will be executed as `root` anyway.

ansible/roles/debops.dovecot/templates/etc/ansible/facts.d/dovecot.fact.j2
```diff
@@ -0,0 +1,26 @@
...
+# {{ ansible_managed }}
+
+from __future__ import print_function
+from json import load, loads, dumps
+from sys import exit
+import subprocess
+import signal
+import os
+
+output = {'installed': True}
```
drybjed, line 12: This fact could be more dynamic by checking the existence of the `dovecot` command during execution. Check the fact of the `debops.ansible` role for a good example.

drybjed (changes requested):
Thanks! Looks like there are issues with the fact script, check the Travis-CI status for details. You could try running the `make test` command in the DebOps monorepo root directory to run the Travis-CI tests locally, after installing needed packages.

amette:
Cool! Thanks for the feedback. I'll look into it in the next days.

drybjed:
The Ansible local facts have been added to the `dovecot` role in a different PR, closing this one since it is now redundant.

PR #867, opened 2019-07-05
[debops.dovecot] Allow configuration of sieve directories

amette:


drybjed (approved):
Sounds good, thanks! :-)

Issue #863, opened 2019-07-03
gitlab_runner: 81 configuration variables


drzraf:
It's currently way simpler to copy/paste a dozen of shell cmds than going after configuration variables for this role. Seriously seems something is wrong about its "configuration".

1. The `big-static-config-pitfall.j2` : Hardcoding the whole file as a j2 template makes this role incompatible with any **external** and **modular** modification. A better alternative is:
   1. See that [toml is supported by augeas](http://augeas.net/docs/references/lenses/files/toml-aug.html) (could have been an Ansible `ini_file`-like plugin btw)
   2. See that there **is** an Ansible Augeas module
   3. Then Provide helpers as building-blocks for **modular** configuration. For example:
```yaml
  # runner_cache.yml
  toml_file: item="{{ item.n }}" value="{{ item.v }}" section="{{ item.s }}"
  with_items:
    - { n: cache_type, v: "{{ gitlab_runner__cache_type }}", s: runners.cache }
    - ... cache_server_address, cache_access_key, ...
```
2. Split variables across `defaults/main/*.yml` directory, organized by topic.
3. Finally drop variables related to edge-cases (at least strip them from the main documentation). Advanced users **will** find them or implement them (something which becomes possible once we move away from the `big-static-config-pitfall.j2`)
4. From the 61 variables used in the runner configuration template, 55 are used **exclusively** for it and most of them could be given sane default offering a **clear separation between optional and mandatory configuration**
5. Then we are left with:
   - gitlab_runner__additional_groups
   - gitlab_runner__api_fqdn
   - gitlab_runner__api_token
   - gitlab_runner__apt_key
   - gitlab_runner__apt_repo
   - gitlab_runner__apt_upstream
   - gitlab_runner__base_packages
   - gitlab_runner__combined_tags
   - gitlab_runner__comment
   - gitlab_runner__custom_files
   - gitlab_runner__default_tags
   - gitlab_runner__domain
   - gitlab_runner__fqdn
   - gitlab_runner__group
   - gitlab_runner__group_custom_files
   - gitlab_runner__group_tags
   - gitlab_runner__home
   - gitlab_runner__host_custom_files
   - gitlab_runner__host_tags
   - gitlab_runner__packages
   - gitlab_runner__run_untagged
   - gitlab_runner__shell
   - gitlab_runner__shell_tags
   - gitlab_runner__ssh_generate
   - gitlab_runner__ssh_generate_bits
   - gitlab_runner__ssh_install_to
[...]

PR #877, opened 2019-07-08
gitlab_runner refactor

drzraf:
* Fixes issue #863

* A Toml helper allows YAML-based configuration of gitlab-runner config.toml **directly** following the [official documentation configuration](https://docs.gitlab.com/runner/configuration/advanced-configuration.html) structure.
* It's extended with a `per-runner` `underscore-prefixed` keys defining specific registration parameters as passed to the `gitlab_runner` module.
* Docker, gitlab.com and custom gitlab instance supported as first-class citizens
* Vagrant supports dropped for now (feel free to reintroduce it)
* Still based on a _big-template-pitfall.j2_ strategy but gets rid of former `config-hell`.
* Works ideally in conjunction with https://github.com/ansible/ansible/pull/58840

I don't expect to spend time in reworking this PR. It's just offered here as-is. Let's hope to avoid the over-engineering-process-trap and that you can get the best of it.

`278 insertions(+), 1529 deletions(-)`

Regards


drybjed:
Thanks! Although I wouldn't call this PR a [refactor](https://en.wikipedia.org/wiki/Code_refactoring), more like a redesign from the ground up.

I would say that you still have some ways to go with this before it can be considered for merging. I would start by fixing all the Travis-CI issues encountered so far. DebOps monorepo has a built-in test suite which you can use by running `make test` in the root of the repository. This should help making the Travis-CI tests pass easier, without the need to push on each change.

Dropping support for the `shell` executor, and in consequence Vagrant, is a problem because it is used to set up [GitLab CI test pipeline for DebOps itself](https://gitlab.com/debops/debops/pipelines). Without this functionality present the validity of the project itself could be easily compromised. Is there any chance to bring it back within your proposed redesign?

There are other things in this PR that need addressing, but I would like to see these two things resolved first. Since you mentioned that you don't plan to commit more time for this PR, perhaps somebody else will be willing to pick it up. We will see.

drybjed:
The role will be reworked in a different direction in the future. Closing this issue.

drzraf:
Details/follow-up issue?

drybjed:
I'll go through the PR in detail to explain why I didn't accept it.

- The `toml.py` filter plugin: the idea to add the `to_toml` and `from_toml` filters is very good, generating configuration via an existing library instead of making a custom Jinja template is usually a better option. Of course if we care about comments that might be a blocker if the library does not support them. However in your implementation you put some additional filters related specifically to `gitlab-runner`. Extracting the TOML filters to a separate Python script and adding it to the `ansible_plugins` role would be a much more robust solution and would allow to use them in other roles in the future.

- The `COPYRIGHT` file: You removed existing authors and replaced them with yourself instead. Why? The role has a history and the `COPYRIGHT` file is used to keep track of who worked on a given role and to give the authors recognition. If you had just added yourself to the existing list that would have been perfectly fine.

- The role default variables: I'm fine with removing all of the existing variables, that's commonly done when the role is refactored. However you didn't try to preserve existing functionality at all - the only supported execution environment that's left is Docker, with hard-coded number of concurrent threads, the configuration is focused on external third party services and if users want to modify just one part of the proposed configuration, they have to copy the whole YAML dictionary over to the inventory. With DebOps, I aim for the project to be focused on internal services and I don't want to force users to copy 30+ lines of variables to the inventory to change one parameter. That's why the pattern of multiple default variables which define lists, that are combined and parsed by the custom `parse_kv_items` and `parse_kv_config` filter plugins was created. In the configuration, you are also referencing variables that are not defined elsewhere like `gitlab_runner__gitlab_project` or `gitlab_runner__cache_bucket_name` which would mean that Ansible would complain if they are not defined in the Ansible inventory and role would be broken - it would not pass the GitLab CI pipeline used by DebOps for validation.

- The `package_domain-keep-ssh-host-keys` patch: I'm not sure yet if this patch is still needed or not, but it does not have to be applied in Buster and beyond, it would be nice to leave it in the role and just apply it conditionally on older Debian releases instead. This makes the role usable on older versions of the OS if somebody still uses them for some reason. DebOps tracks Debian Stable, but functionality for older OS releases is not removed right away without a reason.

- The `05-install` tasks: you replaced use of the `keyring` role with managing the GPG keys and APT repositories directly. The `keyring` role gives much more flexibility in how GPG keys are retrieved by Ansible and gives users the option of selecting the source (keyserver, local directory, URL, YAML text block), that's why I switched all of the roles to use it to have one central place for these things. Also, the GPG keys, and APT repository URLs are essentially hardcoded and cannot be changed without changing the code in the DebOps monorepo. This should have been exposed in the default variables, that way users can always change the keys or APT mirrors to their preferred ones via the Ansible inventory without the need to touch the monorepo code. Which then means that the DebOps codebase can be updated without the need for merges and manual conflict resolution.

- Again, the `05-install` tasks: You install Docker in the `gitlab_runner` role without even an obvious way to disable this (I guess the `with_docker` variable could be used for that but it's not exposed anywhere). DebOps already has a `docker_server` role which installs and configures the Docker instance. You should have just pointed to it in the documentation and teach the `gitlab_runner` role to detect its installation and reconfigure itself as needed. That way the users can easily compose the execution environment in whatever way they wish.

- The `10-cert` tasks: You create an entire CA and self-signed certificate on the `gitlab_runner` host, with hard-coded `ae.vg` and `runner.ae.vg` FQDN in the certificates. Are you aware that DebOps includes the `pki` role which has support for internal CA, Let's Encrypt, external CA, and provides an extensive environment in the form of PKI realms which Ansible roles can use to automatically configure the relevant certificates and private keys? This also ensures that all hosts in the cluster managed by DebOps trust each other and there are no issues with self-signed certificates served by different services.

- The `50-register-runners` tasks: use of the `gitlab_runner` module is interesting - when the original role was written that runner was not yet in the Ansible codebase. I will definitely consider using it when the role is rewritten as long as it allows for idempotency. We will see.

- The `main` tasks: You completely removed the use of the custom Ansible fact script. These scripts can be used as reference about what roles were applied on a given host, so at least a simple Python script that returns information that GitLab Runner is installed would still be beneficial.

Your proposed role seems to be designed as completely standalone and self-contained. It's not bad if you want to publish it on Ansible Galaxy for other people to learn from and customize for their needs, but including it in DebOps would make GitLab Runner not integrate with existing infrastructure at all. To mitigate this, the role would have to be reworked again to enable that integration. So, merging the PR in its current form would be counter-productive for existing project users and developers.

drzraf:
Thank you for the very detailled reply. I understand your reasons.

I tend to think that matching all use-cases in a role dis-balances the project in a sense of being:
- more configurable (but also more complex to configure, document, and at risk of clustering with rarely used (or simply "marginal") options (custom gpg, ssl, ssh setup, Vagrant, ...)
- ...and that this complexity, to be dealt with, implies more reliance on other libraries/functions/toolkit and lowers individual roles self-containment

That being said, I wish you good luck on that road.

drzraf:
>     * Again, the `05-install` tasks: You install Docker in the `gitlab_runner` role without even an obvious way to disable this.

No needless option => If at least one of the runner needs docker, then it's installed. Otherwise it's not.

>     * The `10-cert` tasks: You create an entire CA and self-signed certificate on the `gitlab_runner` host, with hard-coded FQDN in the certificates.

Fixed. Now only generated if needed, on a per-runner basis (and, obviously, no hardcoded domain anymore)

>     * The role default variables: [...] hard-coded number of concurrent threads, [...] they have to copy the whole YAML dictionary over to the inventory

The configuration represent the bare minimum to configure. It's structured (contrary to existing template-based configuration) and obvious since it matches official structure and names. It's expected to be overwritten. Importing a role's default variable is not very well supported by Ansible, but that's a core issue that can be worked-around by users via `set_facts`.

drybjed:
Computer systems by their very nature are inherently complex. When you look at it, all this, even the text you are reading right now, is just electical signals in a medium. But we humans managed to take ownership of this complexity and manage it through a simple idea - encapsulation and interfaces.

We encapsulated electicity in special hardware with a special interface - an [Instruction Set Architecture](https://en.wikipedia.org/wiki/Instruction_set_architecture). ISA complexity was encapsulated using [machine code](https://en.wikipedia.org/wiki/Machine_code). But of course writing programs as ones and zeroes is ridiculus, so we enacapsulated that using an [assembly language](https://en.wikipedia.org/wiki/Assembly_language). It didn't take long for higher-level languages to develop, such as the [C language](https://en.wikipedia.org/wiki/C_(programming_language)) which abstracted the assembly complexity and allowed for software portability independent from the underlying architecture. But that still wasn't enough because C is too low level for some problems, so [Guido van Rossum](https://en.wikipedia.org/wiki/Guido_van_Rossum) designed the [Python programming language](https://en.wikipedia.org/wiki/Python_(programming_language)) which is implemented by [CPython](https://en.wikipedia.org/wiki/CPython) interpreter, written in C. And finally, Michael DeHaan came along and invented [Ansible](https://en.wikipedia.org/wiki/Ansible_(software)) written in Python. Ansible lets us encapsulate complexity of other software and hardware systems using tasks, roles and finally collections. So, as you can see - the most sensible, natural way to deal with complex systems is to wrap them using an abstraction and provide an interface to that abstraction. We are doing this for decades and that method brought us here today.

Where am I going with this? Software is complex, and treating it like it isn't will only work for a specific use case and for a limited time. GitLab Runner has [200+ configuration variables](https://docs.gitlab.com/runner/configuration/advanced-configuration.html) (I think, I tried to count them but I gave up, it's just a short essay, forgive me) which allow you to deploy variety of execution environments, configure caching, authentication and other stuff. Now, how to deal with all this complexity? The answer should be obvious, we encapsulate it in an Ansible role, a natural "interface" Ansible provides. There are [about 60 Ansible roles for `gitlab-runner` on Ansible Galaxy](https://galaxy.ansible.com/search?deprecated=false&keywords=gitlab_runner&order_by=-relevance&page=1&page_size=100) right now. Most of them probably implement the configuration to a varying degree of complexity, depending on the creator's needs.

But there's one wrinkle - GitLab Runner does not live in a vacuum. It's a software distributed in some way, it's a daemon which has to talk to other services on the host and over the network. It exists in a specific environment which has its own levels of complexity. How can we deal with this? The answer seems obvious - add all the additional bits like APT package installation or code compilation, configuration of all those different services as needed, to the GitLab Runner role and be done with it.

This, I believe, is where the mistake was made - for the longest time, years - Ansible Galaxy offered only a single "artifact" for distributing encapsulated complexity of systems: as an Ansible roles. So people used what they could, and since dealing with multiple roles that could encapsulate different systems and connect them together via abstractions was hard, they instead put just enough complexty into the individual roles to deal with the most complicated cases, like X.509 certificate creation.

DebOps [from the very start](https://github.com/debops/debops/tree/eb42149555) was designed to work on a higher level than roles, to connect them together using Ansible playbooks. So what if in 2013 playbooks couldn't be distributed as easily as roles via Ansible Galaxy? I went with this design anyway and it evolved over time to what we have today. The roles are units of arbitrary complexity that is abstracted using defined interfaces connected together through playbooks.

It's true that in this project roles can be unreasonably complex - the [`nginx` role](https://github.com/debops/debops/tree/master/ansible/roles/nginx) role is a common example, with its horrendous and convoluted configuration templates and lots and lots of variables. But the role encapsulates `nginx` webserver so effectively that 19 other roles in DebOps (`docker_registry`, `dokuwiki`, `etesync`, `etherpad`, `gitlab`, `icinga_web`, `kibana`, `librenms`, `minio`, `mosquitto`, `netbox`, `owncloud`, `phpipam`, `phpmyadmin`, `rabbitmq_management`, `reprepro`, `roundcube`, `rstudio_server`, `sks`) use it to configure the webserver for their application. When I write a new role for a web application, I don't have to think how to set up an `nginx` webserver for it - all I need to do is write a simple set of YAML dictionaries that tell the `nginx` role how to configure the service for that particular application. That's how you encapsulate complexity in Ansible.

When you write about "lowering individual role's self-containment", you miss this entire paradigm shift, where Ansible roles are just building blocks connected together on a higher level of abstraction - playbooks. DebOps roles don't live in a vacuum - there are 160+ other roles in the project, most of them present easy to use interfaces to their underlying services, and when you connect them together, you are able to build and manage systems of higher and higher complexity down the road, that work together on separate hosts, clusters, entire data centers even. And the individual `gitlab_runner` role in DebOps can rely on other roles to deal with other services and systems, and doesn't have to include them - this in turn allows the role to be focused more on managing its service in a generalized way, encapsulating that service complexity.

Hopefully, with introduction of Ansible Collections as a higher level artifact that can include modules, playbooks, plugins and other stuff, this entire misguided (in my opinion) paradigm of creating roles that deal with services using bare minimum of configuration and include bits from other systems that are required by an application to function properly, will come to pass. Ansible roles can be designed for high levels of automation where users can trust them to configure the applications intelligently in a given environment, I hope that DebOps proves this. The change will probably take years, but we will get there, eventually.

PR #879, opened 2019-07-10
[debops.redis_server] Make `redis_server__auth_password` optional

carlalexander:
This is an issue that encountered trying to update my DebOps project. Previously, leaving `redis_server__auth_password` as an empty string would configure the Redis server without a password. Redis doesn't seem to like this anymore and the server won't start.

I added a check for `redis_server__auth_password` when generating `redis_server__default_base_options`. If `redis_server__auth_password` is `False`, it ignores the option which lets you configure the Redis server without a password. Not sure if that's the ideal fix, but it seems to fit the role structure.

### DIFF
--- a/ansible/roles/debops.redis_server/defaults/main.yml
+++ b/ansible/roles/debops.redis_server/defaults/main.yml
@@ -193,9 +193,11 @@ redis_server__default_base_options:
 
   - name: 'masterauth'
     value: '{{ redis_server__auth_password }}'
+    state: '{{ "present" if redis_server__auth_password|d() else "ignore" }}'
 
   - name: 'requirepass'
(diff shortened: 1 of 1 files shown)


drybjed:
I wouldn't use `False` here, because this changes the type of the variable from a string to a boolean, which I try to avoid. Instead, set:
```yaml
redis_server__auth_password: ''
```
So, an empty password string, and change the condition in the PR to be:
```yaml
state: '{{ "present" if redis_server__auth_password|d() else "ignore" }}'
```
This should do the trick. Can you check if that works for you, and update the PR?

carlalexander:
Updated the PR and tested. Works fine 😄

drybjed (approved):

PR #893, opened 2019-07-15
Adjust Debian releases in reprepro role as Buster became Debian stable.

racke:


drybjed (approved):
Sounds good, thanks! The role needs to be refreshed at some point to make management of new repositories eaiser. But that's for another time.

PR #898, opened 2019-07-16
Make dist-upgrade.yaml capable of upgrading /etc/apt/sources.list.d/

noqqe:
As I am only adding repositories in /etc/apt/sources.list.d/, I found it could be useful to detect upgrades in that directory as well

### DIFF
--- a/ansible/playbooks/tools/dist-upgrade.yml
+++ b/ansible/playbooks/tools/dist-upgrade.yml
@@ -96,13 +96,22 @@
       stat:
         path: '{{ dist_upgrade_lockfile }}'
       register: dist_upgrade_register_lockfile
+      
+    - name: Find all apt sources.list files
+      find:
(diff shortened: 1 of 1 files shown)


ansible/playbooks/tools/dist-upgrade.yml
```diff
@@ -96,13 +96,22 @@
       stat:
         path: '{{ dist_upgrade_lockfile }}'
       register: dist_upgrade_register_lockfile
+      
+    - name: Find all apt sources.list files
+      find:
+        paths: "/etc/apt/"
+        patterns: "*.list"
+        recurse: yes
+      register: repos
```
drybjed, line 105: Sounds good, although I would use a more descriptive register name, for example `dist_upgrade_register_source_lists`. Can you update the name in this and following task?

PR #902, opened 2019-07-18
[bootstrap] set become to true to avoid using -b on CLI

nqb:
- `debops.python/raw` use tasks that need privileges escalation
-  `service/core.yml` has `become` set to `False`
- Roles in third play need privilege escalation

I checked docs and there is no usage of bootstrap playbook with `-b` option.

### DIFF
--- a/ansible/playbooks/bootstrap-ldap.yml
+++ b/ansible/playbooks/bootstrap-ldap.yml
@@ -22,6 +22,7 @@
 - name: Bootstrap Python support on a host
   hosts: [ 'debops_all_hosts', 'debops_service_bootstrap' ]
   gather_facts: False
+  become: True
 
   roles:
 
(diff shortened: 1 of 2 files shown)


drybjed:
There's no `-b` specified because I usually run it against the `root` account directly. I was worried a bit about how the `raw` module will work with the `become` keyword when the host does not have Python installed yet, but it seems to do fine. Sounds good to me then.

Can you also add the `become` keyword in the `bootstrap-ldap.yml` playbook as well, for completeness?

nqb:
> There's no `-b` specified because I usually run it against the `root` account directly.

If you run `bootstrap` playbook in Vagrant environment in place of `common` playbook without `-b`, you will get an issue. However, Vagrant's post message tells you to use `common` playbook ;-) I suppose because you assume that Vagrant shell provisioning has complete all the steps of the bootstrap playbook.

Nonetheless, if server you want to bootstrap don't have root access allowed, it could be useful.

> Can you also add the `become` keyword in the `bootstrap-ldap.yml` playbook as well, for completeness?

Done, sorry I missed it.

drybjed (approved):

PR #904, opened 2019-07-18
[debops.gitlab] Allow excluding parts of backup

tumd:
Make it possible to add the SKIP-variable to the GitLab backup cronjob
and the pre-upgrade backup of GitLab.

I have however not tested the changes in `gitlab_ce_pre_upgrade.yml` to any great extent.


drybjed (approved):
Sounds good, thanks. Let's see how that works in practice. :-)

PR #906, opened 2019-07-19
[debops.ldap] Consider DNS NODATA response

### LINKED ISSUE #905: Ansible dig lookups: wildcard considerations
Not sure if this is really an issue, but we were kinda taken by surprise.

Some roles like `debops.icinga` and `debops.ldap` use an Ansible dig lookup to determine the hostnames of various services. `debops.ldap` for example queries `_ldap._tcp.<domain>/SRV` to figure out the hostname of the ldap server. if the DNS resolver returns an NXDOMAIN, the role will fallback to `ldap.<domain>`.

We didn't have an `_ldap._tcp.<domain>` SRV record. The `debops.ldap` role picked that up and used `ldap.<domain>` instead. That's all good! The problems started when we added wildcard MX and TXT records to our domain (`*.<domain>`) in order to more effectively combat email spoofing. This causes DNS resolvers to return not NXDOMAIN but NODATA answers for queries for the nonexisting `_ldap._tcp.<domain>/SRV` record. `debops.ldap` doesn't handle this well:

```
TASK [debops.ldap : Generate system-wide LDAP configuration] ********************************************
fatal: [stuxnet.bitsoffreedom.nl]: FAILED! => {"changed": false, "msg": "AnsibleUndefinedVariable: 'dict object' has no attribute 'value'"}
```

Maybe the dig lookup should check for NODATA responses as well?

imrejonk:
This change makes sure that the `debops.ldap` role falls back to using
'[ "ldap." + ldap__domain ]' as ldap__servers when the Ansible dig
lookup returns an empty (NODATA) response.

Closes: #905

### DIFF
--- a/ansible/roles/debops.ldap/defaults/main.yml
+++ b/ansible/roles/debops.ldap/defaults/main.yml
@@ -100,7 +100,8 @@ ldap__servers_srv_rr: '{{ q("dig", "_ldap._tcp." + ldap__domain + "./SRV", "flat
 # Fully Qualified Domain Name.
 ldap__servers: '{{ (ldap__servers_srv_rr | selectattr("target", "defined")
                     | map(attribute="target") | map("regex_replace", "\.$","") | list)
-                   if ("NXDOMAIN" not in ldap__servers_srv_rr)
+                   if ("NXDOMAIN" not in ldap__servers_srv_rr and
+                       ldap__servers_srv_rr[0])
                    else [ "ldap." + ldap__domain ] }}'
 
                                                                    # ]]]


ansible/roles/debops.ldap/defaults/main.yml
```diff
@@ -100,7 +100,8 @@ ldap__servers_srv_rr: '{{ q("dig", "_ldap._tcp." + ldap__domain + "./SRV", "flat
 # Fully Qualified Domain Name.
 ldap__servers: '{{ (ldap__servers_srv_rr | selectattr("target", "defined")
                     | map(attribute="target") | map("regex_replace", "\.$","") | list)
-                   if ("NXDOMAIN" not in ldap__servers_srv_rr)
+                   if ("NXDOMAIN" not in ldap__servers_srv_rr and
+                       ldap__servers_srv_rr[0])
```
drybjed, line 104: Hmm, that doesn't seem right... Can you show the value of `ldap__server_srv_rr` when the `NODATA` is returned? I would think that this would be either a string or a list, in which case `ldap__server_srv_rr[0]` would hold a True-like value.

imrejonk: It's an empty string inside a list ;)

```
TASK [debops.ldap : debug] *****************************************************************************************************************
ok: [infra.ciphermail.net] => 
  ldap__servers_srv_rr:
  - ''
```

drybjed: I see. I guess that's how Ansible `dig` lookup works then. Did you check if the `NXDOMAIN` and actually found SRV record cases work with this change as well?

imrejonk: I tested this change with NXDOMAIN and NODATA responses, as well as with multiple SRV records. Seems to work just fine.

drybjed (approved):
Thanks for the bugfix. I'll backport it to the stable release as well.

PR #912, opened 2019-07-19
[debops.docker_server] reverse condition to support next Debian releases

### LINKED ISSUE #911: [debops.docker_server] path to docker daemon for all distributions managed by DebOps
Next part of #820 and #909 

See @drybjed's comment:

> Now that I think about it, this will break on Bullseye and beyond. Perhaps you could invert the condition, use the old path on all the old Debian/Ubuntu releases, and otherwise use the new one? For Ubuntu, you can just stick to LTS releases (trusty, xenial, bionic).

nqb:
fixes #911 

I only tested under Buster with distribution packages.


drybjed (approved):
Thanks!

nqb:
Waiting merge to provide a more global fix.

drybjed:
@nqb: you can just push new commits to this branch, the PR will be updated automatically.

PR #921, opened 2019-07-22
[debops.nodejs] Add support for Debian buster and upstream repo

pedroluislopez:
Debian buster include a nodejs package.


drybjed:
@pedroluislopez Just an update: after checking this PR against Debian Buster I found that there are issues with `npm` installation there, not related to this particular PR. The `debops.nodejs` role will have to be updated to resolve them, I'll include this PR in the update (don't close it yet).

pedroluislopez:
Ok

PR #929, opened 2019-07-24
Make status and ping paths compatible with security limit extensions

pedroluislopez:


drybjed (approved):
Good idea, thanks. :+1:

PR #942, opened 2019-07-27
[debops.nginx] No longer limit HTTP methods

### LINKED ISSUE #935: [debops.nginx] default PHP-FPM config breaks REST APIs
The PHP-FPM configuration sections managed by `debops.nginx` contain this line:
```
limit_except GET HEAD POST { deny all; }
```
This limits the HTTP request methods accepted by nginx to GET, HEAD and POST. While this could be seen as a useful security feature and is probably harmless for most websites, it breaks parts of the HTTP core functionality required for REST APIs. For example, this week I was configuring a Passbolt server with debops.nginx, which seemed to work fine until I tried to register a new user. The last step in the user registration process involves the client sending a PUT request, which gets blocked by nginx and causes the client to hang.

The limit_except line can be disabled on a per-server basis, but is on by default. We should consider turning it off by default because of these kinds of unexpected side effects. The line can then be enabled per-server, and maybe we can add a variable for turning it on globally as well.

imrejonk:
HTTP request methods used to be limited to GET, HEAD and POST by
default on PHP-enabled websites. This caused REST API calls to break.
This commit changes that behaviour so that there are no limits applied
by default. The user can set their own request method whitelist in the
``item.php_limit_except`` parameter.

Closes #935

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -245,6 +245,9 @@ Changed
   be removed to allow the DNS to take over. If it doesn't, the configuration
   will be left intact with assumtion that the domain is configured locally.
 
+- [debops.nginx] The role will no longer default to limiting the allowed HTTP
+  request methods to GET, HEAD and POST on PHP-enabled websites.
+
 Removed
 ~~~~~~~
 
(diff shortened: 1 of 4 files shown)


drybjed (approved):
Sounds good, thanks!

PR #943, opened 2019-07-28
Fix idempotency issue when passing relative (e.g. 100%FREE) lvol size

### LINKED ISSUE #939: lvm: Error when using relative lvol size
I recently had an error when using a relative size in the logical volume definition. The following definition should be completely valid:
```yaml
lvm__logical_volumes:
  - lv: 'prometheus'
    vg: 'data'
    size: '100%FREE'
    mount: '/var/lib/prometheus'
    mount_opts: 'relatime,discard,nodev,nosuid,noexec'
```

However on consecutive runs, Ansible bailed out with the following error (using Ansible 2.8.3 and DebOps 1.0.6):
```
TASK [debops.lvm : Manage LVM Logical Volumes] ***********************************************************************************************************************************************************************************************
task path: /usr/lib64/python3.6/site-packages/debops/ansible/roles/debops.lvm/tasks/manage_lvm.yml:43                                                                                                                                         
failed: [prometheus.example.com] (item={'lv': 'prometheus', 'vg': 'data', 'size': '100%FREE', 'mount': '/var/lib/prometheus', 'mount_opts': 'relatime,discard,nodev,nosuid,noexec'}) => {"ansible_loop_var": "item", "changed": false, "item": {"lv": "prometheus", "mount": "/var/lib/prometheus", "mount_opts": "relatime,discard,nodev,nosuid,noexec", "size": "100%FREE", "vg": "data"}, "msg": "Sorry, no shrinking of prometheus to 0 permitted."}
```

Once I added the `shrink: no` attribute to the `lvol` task, the issue was resolved.

Now I was thinking about how to properly add this to a PR. I see two different ways:
* Make `shrink` a new item attribute of `lvm__logical_volumes` which the user can explicitly set. Default would be `no` (Ansible interestingly defaults this to `yes`). I was also thinking of setting `force: yes` when shrinking is explicitly requested.
* Or simply attach the `shrink` `lvol` attribute to the `force` `lvm__logical_volumes` attribute. So when `force` is false, `shrink` would be `false` too and vice versa.

What do you think? Or should I simply add `shrink` as a new `lvm__logical_volumes` and expect the user to dig into the [`lvol` documentation](https://docs.ansible.com/ansible/latest/modules/lvol_module.html) to find the right combination how to set these?

ganto:
Fixes #939

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -279,6 +279,9 @@ Fixed
   :file:`/etc/nsswitch.conf` file changes if DebOps is running against
   ``localhost``, to avoid breaking the existing user session.
 
+- [debops.lvm] Make sure logical volumes will only be shrinked when
+  volume item defines ``force: yes``.
+
 
 `debops v1.0.0`_ - 2019-05-22
 -----------------------------
--- a/ansible/roles/debops.lvm/tasks/manage_lvm.yml
+++ b/ansible/roles/debops.lvm/tasks/manage_lvm.yml
@@ -42,11 +42,12 @@
 
 - name: Manage LVM Logical Volumes
   lvol:
-    lv:    '{{ item.lv }}'
-    vg:    '{{ item.vg }}'
-    size:  '{{ item.size }}'
-    force: '{{ item.force | d(omit) }}'
-    state: 'present'
+    lv:     '{{ item.lv }}'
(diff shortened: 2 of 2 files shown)


drybjed (approved):
Thanks! :-)

PR #950, opened 2019-07-29
[debops.docker_server] Change TCP listen variable

### LINKED ISSUE #871: docker_server__tcp should default to False
debops.docker_server/getting-started.html says:

> If debops.pki was configured on the host, Docker will automatically listen on its TCP port for incoming TLS connections, which is by default blocked by the ferm firewall. If you don't use a firewall or have it disabled, you might want to set docker_server__tcp to False to disable this behavior.

For "security by default" I suggest `docker_server__tcp` to default to `False`, even if `debops.pki` was configured. Otherwise the docker server might e accessible from remote without the admin wittingly enabling this.

imrejonk:
This change turns listening on a TCP port off by default, even if
`debops.pki` is enabled. This is a security measure, it should prevent
administrators from unknowingly leaving the Docker TCP port open.

Closes #871

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -129,6 +129,9 @@ Changed
   preparation of adding a role that will provide client functionality like
   network and container management.
 
+- [debops.docker_server] The Docker server no longer listens on a TCP port by
+  default, even if :ref:`debops.pki` is enabled.
+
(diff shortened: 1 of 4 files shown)


drybjed (approved):
Excellent work, thanks! :+1:

PR #951, opened 2019-07-29
Do not map bad users to guest

debuglevel:
The Samba default is "map to guest = never", which is the more intuitive behaviour compared to the often used "bad user". If a client connects to a Samba server with "map to guest = bad user", the client does not get a hint about a bad user/password and thinks that he did everything fine and remembers this password. The client will then miss some directories which the guest does not have access to and is somehat irritated what is happening.


drybjed (approved):
Sounds good, thanks.

PR #952, opened 2019-07-30
[debops.reprepro] Default to no Component for update rules

gaudenz:
Don't set Component or UDebComponent for update rules if no component is
set in the configuration. Reprepro will then use the components from the
Distribution being updated. This is a saner default than using
reprepro_repository_components. The same logic is already used for pull
rules.

With the current code it's also not possible to not set Component or
UDebComponent on an update rule which makes updates from a repository
which does not have any udeb components hard.


gaudenz:
Sorry, I made a mistake while rebasing this onto master. The first version had an invalid template. Fixed now.

drybjed (approved):
Sounds like a good idea, thanks! :-)

PR #956, opened 2019-08-01
[debops.slapd] Fix regex for deriving base dn

drebs:
For a `slapd__domain` equals to "example.org", the previous regex would
generate `slapd__base_dn` like this:

  ["dc=exampledc=", "dc=orgdc="]

This would result in a final `slapd__basedn` of:

  "dc=exampledc=,dc=orgdc"

Changing the regex to "^(.*)$" fixes the issue for me.


drybjed:
@drebs I wasn't able to replicate the error in Debian Buster, so it might be something else than just a new OS release. I was using Ansible 2.8 but on Python 2.7, perhaps this might be a factor. Your change still seems to work, so I'll merge it, though.

drybjed (approved):

PR #957, opened 2019-08-02
Provisioning examples in Vagrant environment (WIP)

nqb:
Hello,

This is a WIP but I would like to get some review for now to be sure I'm not in a wrong direction.

What have been done:
- define a new Vagrant provisionner to install examples in Vagrant environment
- create a bootstrap example to introduce two ways of bootstrapping hosts:
  * one with current user of Ansible controller
  * one with a dedicated user (`ansible-admin`)
- start documentation

My idea is to provide other examples (not in this PR) like:
- common (with specific settings for nullmailer, sshd and unattended_upgrades)
- proxy-cacher
- preseed/ipxe/tftp
- webserver (with either a static site or one of the debops-provided
services) with Lets-Encrypt support (suggested by @htgoebel)

Each example directory will have same Vagrantfile. That way, you can have dedicated VMs for each example directory.

Regarding documentation, I'm not sure we can include inventory examples directly in DebOps' documentation. If yes, YAML files will need to be written like `defaults/*.yml` files in roles to be parsed by `yaml2rst`. Also, `lib/examples` is not part of `docs/` folder.


Vagrantfile
```diff
@@ -672,6 +672,32 @@ fi
...
+for example in ${examples}; do
+    rsync -a ${upstream_dir}/${example} ${projects_dir}
+    jane notify info "${upstream_dir}/${example} installed in ${projects_dir}"
+    rsync -a ${controller_dir}/ ${projects_dir}/${example}/
+    jane notify info "${projects_dir}/${example}/ synchronized with ${controller_dir}/"
+done
+jane notify success "All examples installed and synchronized in ${projects_dir}"
+
+wget -q -O $HOME/.ssh/id_rsa.pub https://raw.githubusercontent.com/hashicorp/vagrant/master/keys/vagrant.pub
+jane notify info "Insecure vagrant **public** key installed in $HOME/.ssh for bootstrap example"
```
drybjed: Is this really needed? The current `Vagrantfile` should alrady provision the insecure Vagrant key, it's necessary for communication between the nodes.

lib/examples/bootstrap/ansible/inventory/group_vars/node2/bootstrap_ansible_user.yml
```diff
@@ -0,0 +1,2 @@
+---
+ansible_user: ansible-admin
```
drybjed: I would quote the value here to be consistent.

lib/examples/bootstrap/ansible/inventory/group_vars/node2/bootstrap_system_users.yml
```diff
@@ -0,0 +1,15 @@
...
+system_users__self: False
+
+# dedicated admin account to manage servers
+system_users__accounts:
+
+  - name: 'ansible-admin'
+    group: 'ansible-admin'
+    comment: 'Ansible admin'
+    # home path is generated automatically
+    shell: '{{ system_users__fact_self_shell }}'
```
drybjed: This will use the shell you have set on your account on Ansible Controller. Perhaps for the sake of consitency and not relying ton the dynamic user configuration, you should use `/bin/bash` here instead? Otherwise the users will wonder where does this shell come from.

lib/examples/index.rst
```diff
@@ -0,0 +1,56 @@
...
+
+Provisioning examples in Vagrant
+--------------------------------
+
+After doing a :cmd:`vagrant up` in an example directory, you can provision
+examples with following command:
+
+..  code-block:: console
+
+    vagrant provision --provision-with=setup_examples
```
drybjed: Perhaps you could change the name of the provision script to `examples`, then the command would be:
```sh
vagrant provision --provision-with=examples
```
Seems a bit nicer to use and easier to remember.

drybjed (changes requested):
Interesting setup, looking forward for more documentation. There are some changes I'd like to see, look at the individual comments.

You can include files from anywhere in the monorepo in the documentation. The actual playbooks and the `COPYRIGHT` files are examples of this.

Vagrantfile
nqb: I would like to avoid this in order to keep examples independent from Vagrant setup.

Current vagrant provisioning provision vagrant insecure **private** key in `/home/vagrant/.ssh`, not public key. Consequently, when you want to bootstrap a host with another user than `vagrant` (here `ansible-admin`),  `debops.system_users` will not import any SSH key in `/home/ansible-admin/.ssh` because there is no agent available and no public SSH key in `/home/vagrant/.ssh` on controller.

lib/examples/index.rst
nqb: Agree, I try to follow name of current provisionners (like `setup_eatmydata`).

nqb: I got your point, you mean only change name of provisionner not name of script.

nqb:
> You can include files from anywhere in the monorepo in the documentation. The actual playbooks and the COPYRIGHT files are examples of this.

Thanks, I will take a look.

Vagrantfile
drybjed: The insecure public key is available in `/home/vagrant/.ssh/authorized_keys`, you should be able to grab it from there if the provision script is executed as the `vagrant` user - no need to hit an eternal URL every time a Vagrant VM is provisioned. Just copy the `authorized_keys` file to `id_rsa.pub` and you should be fine.

nqb: Nice idea, I did it.

However, in current `Vagrantfile`, this setup will work only if you are running `debian/buster64` box:

```ruby
        if ENV['VAGRANT_BOX'] || 'debian/buster64' == 'debian/buster64'
            subconfig.ssh.insert_key = false
```
From my point of view, there is no need for such test for master or nodes box.

Any reason about this current config ?

drybjed: I think we can drop this one, good catch. It will also help when `debian/stretch64` is used in some CI tests. I'll check it out in the CI pipeline to be sure.

nqb:
I addressed all issues.

I will work on documentation and update this PR.

PR #961, opened 2019-08-03
[debops.zerotier] - add new role

russelltsherman:
this role manages zerotier clients, attaching and removing hosts to existing zerotier networks
it does not take on the task of creating standalone zerotier controller

I imagine there is a bit of work to do here to make this consistent with other debops roles..
I look forward to guidance as to what changes need made


drybjed:
Do you plan to create the ZeroTier Controller role as well? Because without it, DebOps users are forced to use the public one, and I wouldn't want to pigeonhole them into that option.

russelltsherman:
i did not have immediate plans to add that.. 
but i understand your view and will resubmit when that aspect has been added.

PR #964, opened 2019-08-05
Add new role debops.discourse

htgoebel:
This install discourse without any container. Please review and give feedbak whether this meeds the quality requirements for inclusion,


ansible/playbooks/service/discourse.yml
```diff
@@ -0,0 +1,84 @@
...
+        - '{{ nginx__apt_preferences__dependent_list }}'
+        - '{{ nodejs__apt_preferences__dependent_list }}'
+        - '{{ postgresql__apt_preferences__dependent_list }}'
+        - '{{ ruby__apt_preferences__dependent_list }}'
+        - '{{ discourse__apt_preferences__dependent_list }}'
+
+    - role: debops.nodejs
+      tags: [ 'role::nodejs', 'skip::nodejs' ]
+      nodejs__upstream: True
+      nodejs__yarn_upstream: True
```
drybjed, line 31: With the recent changes in `debops.nodejs`, variables that enable upstream support shouldn't be needed, at least on Debian Buster. Can you check and confirm?

ansible/playbooks/service/discourse.yml
```diff
@@ -0,0 +1,84 @@
...
+- name: Manage Discourse service
+  hosts: [ 'debops_service_discourse' ]
+  become: True
+
+  environment: '{{ inventory__environment | d({})
+                   | combine(inventory__group_environment | d({}))
+                   | combine(inventory__host_environment  | d({})) }}'
+
+  roles:
+
```
drybjed, line 12: You are missing the `debops.keyring` role used by other roles to manage GPG keys and APT repositories. You should include it near the top of the playbook.

ansible/playbooks/service/discourse.yml
```diff
@@ -0,0 +1,84 @@
...
+      postgresql__dependent_extensions:
+        - '{{ discourse__postgresql__dependent_extensions }}'
+      postgresql__dependent_pgpass:
+        - '{{ discourse__postgresql__dependent_pgpass }}'
+      when: discourse__database == 'postgresql'
+
+    - role: debops.discourse/nginx
+      # Install discourse snippets required by nginx
+      # TODO: are these the correct tags?
+      tags: [ 'role::nginx', 'skip::nginx' ]
```
drybjed, line 69: At this point `nginx` might not be installed on the host. Perhaps this functionality should be added in `debops.nginx` to allow management of custom snippets in `/etc/nginx/snippets/` directory (IIRC) via role dependent variables.

ansible/roles/debops.discourse/README.txt
```diff
@@ -0,0 +1,442 @@
+Custom versions
```
drybjed, line 1: Perhaps this file could be moved to the `docs/ansible/roles/debops.discourse/` directory, named like `role-devlopment.rst` or something? Right now it will be confusing located inside of the role directory, I think.

ansible/roles/debops.discourse/defaults/main.yml
```diff
@@ -0,0 +1,686 @@
+---
+# .. vim: foldmarker=[[[,]]]:foldmethod=marker
+
+# debops.discourse default variables [[[
```
drybjed, line 4: You should remove the outermost `[[[` and `]]]` markers, so that the defaults file has one less level of folding we need to open in `vim`.

ansible/roles/debops.discourse/defaults/main.yml
```diff
@@ -0,0 +1,686 @@
...
+                                      ansible_local.core.distribution_release|d())
+                                  else ansible_distribution_release }}'
+
+                                                                   # ]]]
+# .. envvar:: discourse_release [[[
+#
+# Specify the Discourse release to install or manage. The role checks what release
+# is currently installed and can perform automatic upgrade if the installed
+# release is older than the specified. Downgrades are not supported.
+discourse_release: '{{ discourse__release_map[discourse__distribution_release] }}'
```
drybjed, line 47: This variable should be renamed to `discourse__release`, don't forget to update all of the references. I know that you based this on `debops.gitlab` role, that one hasn't been properly migrated to its own variable namespace yet.

You should also namespace other variables as well, to have consistent `discourse__` prefix.

ansible/roles/debops.discourse/defaults/main.yml
```diff
@@ -0,0 +1,686 @@
...
+discourse__database: '{{ ansible_local.discourse.database
+                      if (ansible_local|d() and ansible_local.discourse|d() and
+                          ansible_local.discourse.database|d())
+                      else ("postgresql"
+                            if (ansible_local|d() and ansible_local.postgresql is defined)
+                            else ("mariadb-is-deprecated"
+                                  if (ansible_local|d() and ansible_local.mariadb is defined)
+                                  else "no-database-detected")) }}'
+
+                                                                   # ]]]
```
drybjed, line 72: Does Discourse support the MariaDB database as well? Otherwise, this variable could probably be either removed, or its value simplified a bit to not check for MariaDB presence.

ansible/roles/debops.discourse/defaults/main.yml
```diff
@@ -0,0 +1,686 @@
...
+# .. envvar:: discourse__smtp_host [[[
+#
+# SMTP server to use.
+discourse__smtp_host: 'mail.{{ discourse_domain }}'
+                                                                   # ]]]
+# .. envvar:: discourse__smtp_port [[[
+#
+# SMTP server port to use.
+discourse__smtp_port: '{{ 25 }}'
```
drybjed, line 149: Just `'25'` should work, no need to wrap this in Jinja.

ansible/roles/debops.discourse/defaults/main.yml
```diff
@@ -0,0 +1,686 @@
...
+                                  else "no-database-detected")) }}'
+
+                                                                   # ]]]
+# Application FQDN and DNS addresses [[[
+# --------------------------------------
+
+# .. envvar:: discourse__instance_name [[[
+#
+# The name of this discourse instance.
+discourse__instance_name: 'discourse'
```
drybjed, line 79: You should mention here that the UNIX account and group will be based on this. Anything else?

ansible/roles/debops.discourse/defaults/main.yml
```diff
@@ -0,0 +1,686 @@
...
+                                                                   # ]]]
+                                                                   # ]]]
+# Internal application options [[[
+# --------------------------------
+
+# .. envvar:: discourse_unicorn_workers [[[
+#
+# Number of Unicorn workers to start.
+# If you only have 1GB RAM, use 2 workers. Otherwise increase to 3 or 4.
+discourse_unicorn_workers: 2
```
drybjed, line 355: How about using `ansible_processor_vcpus` as default here? It would simplify scaling - just set up a bigger VM or use more powerful host.

ansible/roles/debops.discourse/defaults/main.yml
```diff
@@ -0,0 +1,686 @@
...
+discourse_support_filesystem_acl: True
+                                                                   # ]]]
+                                                                   # ]]]
+# Discourse source code [[[
+# -------------------------
+
+# .. envvar:: discourse_git_repo [[[
+#
+# URL of the Discourse repository.
+discourse_git_repo: 'https://github.com/discourse/discourse.git'
```
drybjed, line 386: Discourse releases are... unsigned. *sigh* And the repository does not have the "Issues" page. But I guess that after so many releases upstream is completely uninterested it making GPG-signed releases, since they focus on Docker as a distribution method.

Does Discourse upstream use stable branches in the repository for updates? If not, perhaps you should consider deploying a specific `git` tag instead, upgrades could be done after reviewing that nothing breaks and new upstream releases can be tracked using the `watch` files.

ansible/roles/debops.discourse/defaults/main.yml
```diff
@@ -0,0 +1,686 @@
...
+
+# .. envvar:: discourse__base_packages [[[
+#
+# List of base APT packages required by Discourse.
+discourse__base_packages: [
+  'gawk', 'curl', # used by some helper ruby scripts
+  'jhead', 'jpegoptim', 'libjpeg-turbo-progs', 'optipng',
+  'gifsicle', 'pngcrush', 'pngquant',
+  'imagemagick-6.q16', 'libmagickcore-6.q16-6-extra',
+  'npm', # for installing nodes packages
```
drybjed, line 424: The `npm` package should be handled by the `debops.nodejs` role.

ansible/roles/debops.discourse/defaults/main.yml
```diff
@@ -0,0 +1,686 @@
...
+
+  - '{{ discourse__nginx_server }}'
+
+                                                                   # ]]]
+# .. envvar:: discourse__nginx__dependent_maps [[[
+#
+#
+discourse__nginx__dependent_maps:
+
+  - name: thescheme
```
drybjed, line 601: Hmm, `nginx` maps are global, perhaps this should be namespaced just to be sure. `discourse_scheme`?

ansible/roles/debops.discourse/nginx/tasks/main.yml
```diff
@@ -0,0 +1,27 @@
+---
+
+# FIXME: What is the proper way to add this file?
+# FIXME: Are these the correct tags? Is this the correct way to use tags?
```
drybjed, line 4: I think that this can stay for now, until `debops.nginx` role will have proper support for snippets.

ansible/roles/debops.discourse/tasks/helper-scripts.yml
```diff
@@ -0,0 +1,56 @@
...
+       RAILS_ENV=production sudo -H -E -u discourse bundle exec rbtrace "$@"
+
+- name: Create helper /usr/local/bin/stackprof
+  file:
+    path: /usr/local/bin/stackprof
+    mode: '0750'
+    contents: |
+       #!/bin/bash
+       cd /var/www/discourse && \
+       RAILS_ENV=production sudo -H -E -u discourse bundle exec stackprof "$@"
```
drybjed, line 56: These scripts will be in the global `$PATH`. What if user has Ruby, Rails, `rake`, etc. installed on the host and other applications beside Discourse? I wouldn't assume exclusive rights to the host by this role.

Perhaps these could be added in the `$HOME/.local/bin/` directory of the `discourse` user, thte `~/.bashrc` file on Debian Buster will include it automatically if it's found. In that case you don't need to use `sudo` of course.

ansible/roles/debops.discourse/tasks/main.yml
```diff
@@ -0,0 +1,79 @@
...
+  until: discourse__register_packages is succeeded
+
+- name: Install required node.js packages
+  npm:
+    name: '{{ item.key }}'
+    version: '{{ item.value or omit }}'
+    global: yes
+  with_dict: '{{ discourse__nodejs_packages }}'
+  register: discourse__register_nodes_packages
+  until: discourse__register_nodes_packages is succeeded
```
drybjed, line 22: I wonder if the `node` packages could be installed on the `discourse` UNIX account instead. Would that work as expected?

drybjed (changes requested):
Cooooool. Cool, cool, cool. I see that you based this on `debops.gitlab` role. There will probably lots of Travis-CI errors, you should consider testing this role via `make test` to find the issues quickier.

Very promising role. :-) I would target it for the next stable release, so at least November 2019, there's plenty of time to clean it up.

PR #972, opened 2019-08-08
Some small celanup.

htgoebel:
I suggest using **rebase** when merging.


drybjed:
@htgoebel With rebase, do you mean in general? If I remember correctly, rebase will destroy GPG signatures on commits, which I would like to avoid.

drybjed (approved):
Thanks for the fixes :-)

PR #974, opened 2019-08-08
[docs] Add pyOpenSSL to required Python packages

imrejonk:
Hey @drybjed, remember #757? I had the exact same problem today. Figured I'd take a closer look at what the `debops.opendkim` role actually does when it needs pyOpenSSL. Turns out it generates the DKIM private key on the Ansible Controller, not on the inventory host. Oops!

These changes revert my changes from #757 and add pyOpenSSL as a Python requirement to the DebOps installation guide. An errata for the errata...

### DIFF
--- a/INSTALL.rst
+++ b/INSTALL.rst
@@ -111,13 +111,21 @@ Ansible:
 
 .. __: https://bitbucket.org/ecollins/passlib/wiki/Home
 
+`pyOpenSSL`__
+  This is a Python wrapper for the OpenSSL library, available in the
+  ``python-openssl`` package. It's a requirement for :ref:`debops.opendkim` and
+  other roles that generate X.509 certificates or private keys on the Ansible
+  Controller.
+
+.. __: https://www.pyopenssl.org/
+
 You can install them using your distribution packages on Debian or
 Ubuntu by running the command:
 
 .. code-block:: console
 
    sudo apt install python-future python-ldap python-netaddr \
-                    python-dnspython python-passlib
+                    python-dnspython python-passlib python-openssl
 
 The missing Python dependencies will be automatically installed with the
 ``ansible`` and ``debops`` Python packages, however some of them, like the
--- a/ansible/playbooks/service/opendkim.yml
+++ b/ansible/playbooks/service/opendkim.yml
@@ -30,12 +30,5 @@
           config: '{{ opendkim__postfix__dependent_maincf }}'
       when: opendkim__postfix_integration|bool
 
-    - role: debops.python
-      tags: [ 'role::python', 'skip::python' ]
-      python__dependent_packages3:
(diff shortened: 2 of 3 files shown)


drybjed (approved):
Sounds good, thanks! This should probably be backported as well. I'll add the relevant package requirements in `setup.py` and `PKGBUILD`.

drybjed:
@imrejonk Actually, hold on. According to the [`openssl_privatekey` Ansible module documentation](https://docs.ansible.com/ansible/latest/modules/openssl_privatekey_module.html#requirements) it can use either `cryptography` or `pyopenssl` module. Ansible depends on `cryptography` so it should have been installed on the Ansible Controller in the first place... How did you install Ansible? Can you check if the `cryptography` Python module is available?

I think that when we figure this out, removing the dependency in `debops.opendkim` sounds good, but adding dependency on `pyopenssl` or `cryptography` to DebOps seems redundant - Ansible should have pulled that requirement already.

imrejonk:
Well this is strange. I installed Ansible with Pipenv, which also downloads `cryptography`, but the `debops.opendkim` role failed at the key generation task. It worked once I installed `pyOpenSSL`. Don't merge this yet, needs some more detective work...

imrejonk:
Looks like the `openssl_*` modules only work with `cryptography` since Ansible v2.8.0. From the [changelog](https://github.com/ansible/ansible/blob/stable-2.8/changelogs/CHANGELOG-v2.8.rst#v280):

> openssl_privatekey - now works with both PyOpenSSL and cryptography Python libraries. Autodetection can be overridden with select_crypto_backend option.

Elementary, my dear Watson! :male_detective:

Upgrading Ansible to 2.8.3 fixed the problem for me, but the Ansible version in Debian stable repositories (2.7.7+dfsg-1) throws this error:

```
TASK [debops.opendkim : Generate DomainKeys on Ansible Controller] **************************************
failed: [octopus.dcg.amsw.nl -> localhost] (item={'name': '20190810', 'domain': 'amsterdamwireless.nl'}) => changed=false 
  item:
    domain: amsterdamwireless.nl
    name: '20190810'
  msg: the python pyOpenSSL module is required
failed: [orbit.gwl.amsw.nl -> localhost] (item={'name': '20190810', 'domain': 'amsterdamwireless.nl'}) => changed=false 
  item:
    domain: amsterdamwireless.nl
    name: '20190810'
  msg: the python pyOpenSSL module is required
```

Not sure what the best course of action is here. Should we require Ansible 2.8+ or pyOpenSSL?

drybjed:
Thanks for finding out the reason. I want DebOps to be compatible with Ansible packaged in Debian for at least a year, so that users are not forced to install the upstream version. Let's add the dependency for now. I'll merge this PR and take care of the `PKGBULD` and `setup.py` files.

PR #976, opened 2019-08-09
Evilham improve devuan support

evilham:
There is currently an issue with Devuan Beowulf/testing (it's based on Debian Buster/stable), and it's related to lsb_release in there, so right now unattended-upgrades won't work in that release.

That requires more investigation and there's still at least some weeks until it becomes stable officially, so maybe it'll "fix itself" from debops' perspective.


drybjed (approved):
Thanks, I'll backport this to the stable release as well.

PR #1004, opened 2019-08-16
[debops.roundcube] Add role to role index

imrejonk:


drybjed:
You could also add a link in the Mail section, roles can be referenced multiple times here.

drybjed (approved):

PR #1006, opened 2019-08-16
Optional parameters for OpenSSL DH param generator

ckousik:
Allows passing optional parameters to the openssl dhparam generator.
This is useful for passing in options such as `-dsaparam` for faster
generation of keys.


drybjed (changes requested):
Cound you rename the variable to `dhparam__openssl_options`? It will be most likely set in the inventory, so abbrevations aren't really needed.

ckousik:
@drybjed I'm unable to request another review. Please take a look at this when convenient.

drybjed (approved):
Nice work, thanks! :-)

PR #1027, opened 2019-08-27
Use VAGRANT_HOSTNAME to set hostname of VM

nqb:
Permit to use a custom name for VM.


drybjed (approved):
Sounds good, thanks. I'll backport this as well.

PR #1033, opened 2019-08-30
Fix broken internal links to roles

rkandilarov:
Fix broken internal links from debops.tinc to debops.ifupdown and debops.dnsmasq


drybjed (approved):
Thanks for fixing the documentation. :-)

PR #1035, opened 2019-09-01
[debops.owncloud] Update dependent config

tumd:
- Disabled the internal debops.nginx maintenance page for Nextcloud,
  to use Nextcloud's own maintenance page instead.
- Updated various location regexp's to match upstream examples.
- Enabled PHP memory_limit because Nextcloud 16 now warns about it.

### DIFF
--- a/ansible/roles/debops.owncloud/defaults/main.yml
+++ b/ansible/roles/debops.owncloud/defaults/main.yml
@@ -1929,6 +1929,8 @@ owncloud__nginx__dependent_servers:
     name: '{{ owncloud__fqdn }}'
     root: '{{ owncloud__deploy_path }}'
     deny_hidden: False
+    ## Nextcloud has it's own maintenance page
+    maintenance: '{{ False if (owncloud__variant == "nextcloud") else True }}'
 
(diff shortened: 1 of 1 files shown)


drybjed (approved):
I tested the changes with an existing and clean installation, everything seems to work. Good job! I'll backport the changes to the stable releases as well.

ypid:
Thanks @tumd! There are a few more changes that should be done here to fully support Nextcloud 16. I am working on that and doing more testing.

In case somebody else is working on that let me know. @drybjed You merged without the changelog being updated? Are you planning to work on Nextcloud 16 support more?

drybjed:
@ypid: Currently we have 4 changelogs to manage - `git log`, the Changelog in the `master` branch, and the changelogs in the `stable-*` branches. Usually I don't mention changes that are not "visible" by the users in the `master` Changelog - in this case if users did not override the `nginx` and `php` dependent configuration from `debops.owncloud` in their inventory they shouldn't notice any changes - after applying the playbook the site should work as usual. Because of that I didn't mention them in the `master` Changelog file, however the changes are mentioned in the `stable-x.y` Changelogs.

I'm curretly not working on Nextcloud 16 support, so you can go ahead. After your changes, I might look into finally updating the LDAP configuration to use the new `debops.ldap` infrastructure.

PR #1037, opened 2019-09-03
[docs] fix double "The" in debops.ferm doc

nqb:
See https://docs.debops.org/en/master/ansible/roles/debops.ferm/guides.html#restrict-outgoing-traffic


drybjed (approved):
Thanks for fixing the docs. :-)

PR #1041, opened 2019-09-05
[docs] 'post-hook' is not a type

nqb:


drybjed:
Well, [`post-hook` can be used as a weight type](https://github.com/debops/debops/blob/master/ansible/roles/debops.ferm/defaults/main.yml#L373), so that configuration example should definitely work. Did you test it by chance?

nqb:
> Did you test it by chance?

Nope. Sorry, I'm a little bit confused by the ferm documentation. You can keep this PR opened for now and I will test it.

PR #1049, opened 2019-09-05
[debops.unattended_upgrades] Ensure valid reboot time generation

ypid:
Previously, if the user took this as an example and changed the parameters to the `random` Jinja2 filter, a time like "02:3" could have been generated.

With the default settings of the role, such a invalid time was not a concern because only the times from "02:40" to "02:54" (including start and end) where possible.

I have not tested how unattended_upgrades would behave with this invalid time.


drybjed (approved):
Sounds good, thanks. I'll backport it as well.

PR #1050, opened 2019-09-06
[debops.dhcpd] Secure dhcpd configuration

imrejonk:
The dhcpd.conf file used to be world-readable, which is a security
issue when nsupdate keys are stored in this file (see dhcpd_keys
variable). This change sets the file mode of the dhcpd configuration
files (/etc/default/isc-dhcp-server and /etc/dhcp/dhcpd.conf under
Debian) to 0640.


drybjed (approved):
Sounds good. Although in the future this will be more configurable, since you can import additional files in the `dhcpd` configuration, which could hold the keys securely.

PR #1051, opened 2019-09-07
Scripts: Use constants for external program names.

htgoebel:
This makes it much, much easier to replace the program
with one using an absolute path. This is necessary for
e.g. Guix to keep references to these external programs.


drybjed (approved):
Sounds good. I'll backport the changes to the stable branches as well.

PR #1053, opened 2019-09-08
[debops-defaults] If `view` is not available, try less, etc.

htgoebel:


drybjed (approved):
Good idea, thanks. I'll backport this as well.

@htgoebel I plan to make new stable releases today, do you have any other changes you would want to include?

htgoebel:
@drybjed There are no other changes in my pipeline (discourse is still WIP)

PR #1058, opened 2019-09-09
Fix DMZ ferm rule type with dport set

gaudenz:
The filter rule in the FORWARD chain is traversed after the DNAT rule
in the PREROUTING chain. Thus the filter rule has to match on the
dport value instead of the port value.

This issue was already fixed in commit 7e247611ccb56476a86766c2707d198da1346c32
for the old dmz rule template.


gaudenz:
BTW there are multiple files in `templates/etc/ferm/ferm.d` inside the debops.ferm role which are unused as far as I can see. Maybe they should be removed?

drybjed (approved):
Sounds good, thanks. I'll backport the fixes as well.

As for the additional files - sure, I suppose that they can be removed now. I kept them as reference, but I think all the rules should work with the new `rules.d/` version.

PR #1062, opened 2019-09-12
Add 'debops.tinyproxy' Ansible role

LeonardoAndrei:


ansible/roles/debops.tinyproxy/defaults/main.yml
```diff
@@ -0,0 +1,587 @@
...
+
+      - name: 'Port'
+        comment: |
+          
+          Port: Specify the port which tinyproxy will listen on.  Please note
+          that should you choose to run on a port lower than 1024 you will need
+          to start tinyproxy using root.
+           
+        raw: |
+          Port 8888
```
alinalexandru: Use the `tinyproxy__port` variable

ansible/roles/debops.tinyproxy/defaults/main.yml
```diff
@@ -0,0 +1,587 @@
...
+          
+          Allow: Customization of authorization controls. If there are any
+          access control keywords then the default action is to DENY. Otherwise,
+          the default action is ALLOW.
+
+          The order of the controls are important. All incoming connections are
+          tested against the controls based on order.
+           
+        raw: |
+          Allow 127.0.0.1
```
alinalexandru, line 299: Allow all the entries defined in `tinyproxy__allow`

drybjed:
Nice work! I've made a few suggestions over IRC, and you should first fix issues with the test suite. Looking forward to the fixes. :-)

LeonardoAndrei:
@drybjed told me to "use loop inside of the template itself, instead of with_items" (it is used in tasks to 'Generate tinyproxy configuration') but i don't know how to do that at all. 
That would be the last step before completing the PR.

drybjed:
@LeonardoAndrei You can do something like this in the template:
```jinja
 # {{ ansible_managed }}

{% for item in tinyproxy__combined_configuration | parse_kv_items %}
{%   if item.comment|d() %}
{{ item.comment | regex_replace('\n$','') | comment() }}
{%   endif %}
... rest of the code ...
{% endfor %}
```
And remove the loop statement from the task itself. That should do the trick.

ansible/roles/debops.tinyproxy/defaults/main.yml
```diff
@@ -0,0 +1,501 @@
...
+
+                                                                   # ]]]
+# .. envvar:: tinyproxy__port [[[
+#
+# Port number on which this Tinyproxy listens on.
+tinyproxy__port: '8888'
+
+                                                                   # ]]]
+# Configuration for other Ansible roles [[[
+# -----------------------------------------
```
drybjed: This section should go to the end of the `defaults/main.yml` file.

ansible/roles/debops.tinyproxy/defaults/main.yml
```diff
@@ -0,0 +1,501 @@
...
+tinyproxy__packages: []
+
+                                                                   # ]]]
+# .. envvar:: tinyproxy__allow [[[
+#
+# List of IP addresses or CIDR subnets which will be allowed to connect to the
+# Tinyproxy server in :command:`ip(6)tables` and TCP wrappers. If it's empty, remote
+# connections are not allowed.
+tinyproxy__allow: []
+
```
drybjed, line 34: I'm curious about the default use case of Tinyproxy within DebOps. With the current configuration the role configures the service as "private", where you have to specifically open the firewall for other users to access the service. But HTTP services to me are "public" by default, for example `debops.nginx` configures the webserver with this in mind. HTTP servers have additional means to restrict access, be it HTTP Basic Auth, per-IP access limits, and the like, so blocking them by default on the firewall level seems a bit excessive to me.

Of course it all depends on the use case you are aiming at with this role.

ansible/roles/debops.tinyproxy/defaults/main.yml
```diff
@@ -0,0 +1,501 @@
...
+  - name: 'tinyproxy'
+    port: '{{ tinyproxy__port }}'
+
+                                                                   # ]]]
+# Tinyproxy configuration file [[[
+# --------------------------------
+
+# The variables below define the contents of the
+# :file:`/etc/tinyproxy/tinyproxy.conf` configuration file.
```
drybjed, line 46: You could add there something like:
```
See :ref:`tinyproxy__ref_configuration` for more details.
```
so that in the documentation, users have a handy link to the reference page for these variables.

ansible/roles/debops.tinyproxy/defaults/main.yml
```diff
@@ -0,0 +1,501 @@
...
+
+      - name: 'User Group'
+        comment: |
+          ##
+          ## tinyproxy.conf -- tinyproxy daemon configuration file
+          ##
+          ## This example tinyproxy.conf file contains example settings
+          ## with explanations in comments. For decriptions of all
+          ## parameters, see the tinproxy.conf(5) manual page.
+          ##
```
drybjed: This header should be included in the template itself, since it is not relevant to the user and group configuration. And you could probably drop the "This example" paragraph, since it is still included in the diverted configuration file and the generated one is not an "example" anymore.

ansible/roles/debops.tinyproxy/defaults/main.yml
```diff
@@ -0,0 +1,501 @@
...
+        state: 'present'
+
+      - name: 'Port'
+        comment: |
+          Port: Specify the port which tinyproxy will listen on.  Please note
+          that should you choose to run on a port lower than 1024 you will need
+          to start tinyproxy using root.
+        raw: |
+          Port {{ tinyproxy__port }}
+        state: 'present'
```
drybjed, line 76: It looks like you use exclusively `raw` configuration scheme. It's fine of course, but there are many configuration options defined as `<name> <value>` pairs, which you could easily handle by specifying their value in the `value` parameter - that way these options could be easier to modify via inventory.

I suppose this could be implemented in the next iteration of the role, so no pressure about this one. :-)

ansible/roles/debops.tinyproxy/defaults/main.yml
```diff
@@ -0,0 +1,501 @@
...
+          The LAST matching rule wins the route decision.  As you can see, you
+          can use a host, or a domain:
+           name     matches host exactly
+           .name    matches any host in domain "name"
+           .        matches any host with no domain (in 'empty' domain)
+           IP/bits  matches network/mask
+           IP/mask  matches network/mask
+        raw: |
+          Upstream some.remote.proxy:port
+        state: 'comment'
```
drybjed, line 249: Since upstream definition will be the first go-to to configure after the role is applied, this section could be defined in separate set of configuration variables, for example `tinyproxy__*_upstreams`. Since the service uses just one configuration file, you would have to duplicate the templating code, but this would allow more flexibility in defining the upstreams.

Also, the above configuration seems to be from an older Tinyproxy version. I'm comparing this with Tinyproxy 1.10 included in Debian Buster and there the upstream configuration is a bit more extensive, with support for http and socks connections. What Tinyproxy version did you use as a base?

ansible/roles/debops.tinyproxy/defaults/main.yml
```diff
@@ -0,0 +1,501 @@
...
+        comment: |
+          ConnectPort: This is a list of ports allowed by tinyproxy when the
+          CONNECT method is used.  To disable the CONNECT method altogether, set
+          the value to 0.  If no ConnectPort line is found, all ports are
+          allowed (which is not very secure.)
+
+          The following two ports are used by SSL.
+        raw: |
+          ConnectPort 443
+          ConnectPort 563
```
drybjed, line 406: Port 563 is `nntps`, NNTP over SSL. I wonder how did it get included here by the upstream...

ansible/roles/debops.tinyproxy/tasks/main.yml
```diff
@@ -0,0 +1,41 @@
...
+  command: dpkg-divert --quiet --local --divert /etc/tinyproxy/tinyproxy.conf.dpkg-divert --rename /etc/tinyproxy/tinyproxy.conf
+  args:
+    creates: '/etc/tinyproxy/tinyproxy.conf.dpkg-divert'
+
+- name: Generate tinyproxy configuration
+  template:
+    src: 'tinyproxy.conf.j2'
+    dest: '/etc/tinyproxy/tinyproxy.conf'
+    owner: 'root'
+    group: 'root'
```
drybjed: You should drop the `owner` and `group` parameters from tasks, as long as they default to `root`:`root`. This can improve support for other platforms that use different UNIX group for `root`, for example FreeBSD.

docs/ansible/roles/debops.tinyproxy/getting-started.rst
```diff
@@ -0,0 +1,42 @@
...
+
+   [debops_service_tinyproxy]
+   hostname
+
+
+Example playbook
+----------------
+
+If you are using this role without DebOps, here's an example Ansible playbook
+that uses the ``debops.influxdb`` role:
```
drybjed: Don't forget to update the role name when you copy the documentation from something else. :-)

ansible/roles/debops.tinyproxy/templates/etc/ansible/facts.d/tinyproxy.fact.j2
```diff
@@ -0,0 +1,21 @@
...
+
+def cmd_exists(cmd):
+    return any(
+        os.access(os.path.join(path, cmd), os.X_OK)
+        for path in os.environ["PATH"].split(os.pathsep)
+    )
+
+
+output = loads('''{{ ({
+    "installed": true
```
drybjed, line 18: You should set this value dynamically using the `cmd_exists` function. In that case you don't need to wrap this in the `loads()` function, of course.

ansible/roles/debops.tinyproxy/templates/etc/ansible/facts.d/tinyproxy.fact.j2
```diff
@@ -0,0 +1,21 @@
...
+}) | to_nice_json }}''')
+
+print(dumps(output, sort_keys=True, indent=2))
```
drybjed: Set the indent to `4` since it's standardized now in other DebOps roles.

drybjed (changes requested):
Very nice work. There are some things I'd like to see changed, and some questions - see the comments below. Overall, the PR seems to be ready for merge, after some issues are resolved.

drybjed (approved):
Seems like the PR is ready for merge. I'll create a separate one with small changes and updated changelog. @LeonardoAndrei and @alinalexandru, thanks for your contribution to DebOps! :-)

PR #1066, opened 2019-09-14
Update Nextcloud support. Drop NC 14, add NC 16. Now default to NC 15.

ypid:
Following up on the work in #1035 by @tumd I did some more testing and updated the Nextcloud support of the role.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -104,6 +104,13 @@ General
   precompiled binaries from remote resources. See the role documentation for
   more details.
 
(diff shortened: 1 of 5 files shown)


drybjed (approved):
Nice work as always, thanks.

I'm not sure why the test status from Travis-CI does not show up on GitHub. Still, looks like [the test itself has passed successfully](https://travis-ci.org/debops/debops/builds/585006379), therefore I'll merge the PR regardless.

PR #1067, opened 2019-09-14
[debops.nginx] Ensure legacy PHP5 Nginx upstream is absent

ypid:


drybjed (changes requested):
Sounds good, although there's now an issue - if you apply `debops.php` and `debops.nginx` on a host, there's no default PHP `www-data` upstream anymore, it would have to be specifically enabled through the inventory.

I'm not see a good scenario where those two roles would have dependencies on each other; both of them can configure the relevant services standalone. I think that the PHP upstream could be conditional depending on presence of the `debops.php` facts.

This change can stay of course to remove the old PHP5-based upstream. But could you also add a new PHP upstream that depends on the version of installed PHP environment taken from facts?

ypid:
> This change can stay of course to remove the old PHP5-based upstream. But could you also add a new PHP upstream that depends on the version of installed PHP environment taken from facts?

All DebOps roles configure their own PHP FPM pools when I see it correctly. I think we should encourage this behavior also for roles outside of DebOps. Any reason to provide such a default PHP upstream?

drybjed:
@ypid The users could also configure PHP and nginx applications via the Ansible inventory. Although, I suppose that in such case they probably also configure their own PHP upstream in conjunction with the separate user account for the application. So OK, let's remove the default `www-data` upstream from `debops.nginx` and see how this affects things. There's still time to revert that change before the next stable release if need be.

drybjed:
@ypid Docs need to be updated as well, check the Travis-CI test output.

ypid:
> Docs need to be updated as well, check the Travis-CI test output.

Sorry for the delay, done that.

drybjed (approved):

PR #1069, opened 2019-09-16
[debops.icinga_db] Set fact interpreter to python3

imrejonk:
The icinga_db fact script fails when run with Python 2. Because of this, the icinga_web role does not set up the database backend correctly, resulting in a broken installation. This commit changes the interpreter to Python 3 which works fine.

This is the error I got when running the fact script with Python 2:
```
Traceback (most recent call last):
  File "/etc/ansible/facts.d/icinga_db.fact", line 50, in <module>
    output.update(get_config(config_file))
  File "/etc/ansible/facts.d/icinga_db.fact", line 25, in get_config
    config.read_string(config_string)
AttributeError: ConfigParser instance has no attribute 'read_string'
```

### DIFF
--- a/ansible/roles/debops.icinga_db/templates/etc/ansible/facts.d/icinga_db.fact.j2
+++ b/ansible/roles/debops.icinga_db/templates/etc/ansible/facts.d/icinga_db.fact.j2
@@ -1,4 +1,4 @@
-#!/usr/bin/env python3
+#!{{ ansible_python['executable'] }}
 
 # {{ ansible_managed }}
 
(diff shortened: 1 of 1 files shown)


drybjed:
I wonder if that should be a preferred approach for solving cases like this in the future. I know that Python3 will be the default in the future, but there are still old installations out there that might not have it.

@imrejonk: how about adding another set of `try/except` clause in this case that, when `read_string` is not available, reverts to using `readfp` from the Python2 version of ConfigParser? On Python3 it will hopefully will not be reached at all, and on Python2 it should allow the script to work fine in this case. Want to try implementing it? You can check https://github.com/debops/debops/commit/c5a8bd1a0960a92d7acf2f4ec1aa6c134c23508c to see what was used previously here.

imrejonk:
Well okay then, let's maintain these scripts for both Python 2 and 3 a little longer...
How do you like this solution? We are already using it in production :)

drybjed:
@imrejonk Much better, thanks. :-) Just fix the issue on Travis-CI and it will be ready for merge.

imrejonk:
Crap, forgot the first Except statement. Travis is unforgiving!

drybjed (approved):
Nice work, thanks. :+1:

PR #1075, opened 2019-09-23
[debops.gitlab_runner] Use new docker_server facts

imrejonk:
`debops.gitlab_runner` used to use the `debops.docker` local facts to
determine if the Docker daemon was installed. The `debops.docker` role
has been renamed to `debops.docker_server`. These changes make sure
that `debops.gitlab_runner` uses the new local facts.

A quick `grep` shows that this is the only other role where Docker local facts are used.


drybjed (approved):
Looks good, thanks. Perhaps there needs to be a checklist for role renames in the documentation that covers things like this.

PR #1077, opened 2019-09-26
Improved NextCloud/ownCloud LDAP integration

drybjed:


ypid (approved):
LGTM. Note that i don’t work with LDAP very much currently so my LDAP knowledge is not so in depth as yours. I also have not tested these LDAP changes. I only reviewed this PR to the best of my knowledge and checked/tested that no other functionally has been broken by your improvements.

PR #1085, opened 2019-09-30
[debops.resolvconf] Get upstream nameservers when using systemd-resolved

### LINKED ISSUE #1078: [debops.resolvconf] ansible_dns fact incorrect when using systemd-resolved
When using `systemd-resolved` as the system DNS resolver (the default on Ubuntu 18.04), `/etc/resolv.conf` only lists `127.0.0.53` instead of the real DNS servers being used, as it hooks the `glibc` DNS functions instead.

This leads to the `debops.resolvconf` facts script to only retrieving `127.0.0.53` as the `nameserver` in `ansible_dns`
https://github.com/debops/debops/blob/master/ansible/roles/debops.resolvconf/templates/etc/ansible/facts.d/resolvconf.fact.j2

So what are the effects of this? `debops.docker_server` uses `ansible_dns` to fill in the `/etc/docker/daemon.json` file. But `systemd-resolved` will [only listen on `lo0` and not respond to requests from inside containers](https://unix.stackexchange.com/questions/445782/how-to-allow-systemd-resolved-to-listen-to-an-interface-other-than-loopback). This means none of my containers can talk to DNS.

I have fixed this by setting `docker_server__dns_nameserver` to `null` which prevents DNS servers from being listed in `daemon.json` at all so that Docker can work out how to handle this on its own (which it does by correctly).

So, do we need to work out how to fetch DNS servers correctly when using `systemd-resolved`? I haven't yet looked into how to do that.

DouglasHeriot:
Fixes #1078

### DIFF
--- a/ansible/roles/debops.resolvconf/templates/etc/ansible/facts.d/resolvconf.fact.j2
+++ b/ansible/roles/debops.resolvconf/templates/etc/ansible/facts.d/resolvconf.fact.j2
@@ -13,6 +13,26 @@ def cmd_exists(cmd):
         for path in os.environ["PATH"].split(os.pathsep)
     )
 
+def read_resolv_file(file):
+    try:
+        with open(file, 'r') as f:
+            nameservers = []
+            domain = ''
+            search = []
+            for line in f:
+                if line.startswith('nameserver'):
+                    nameservers.append(line.strip().split()[1])
+                elif line.startswith('domain'):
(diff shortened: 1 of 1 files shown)


drybjed:
@DouglasHeriot Thanks for investingating the resolved data. I'm refactoring the code a bit to clean the script, stay tuned.

Issue #1061, opened 2019-09-11
[Vagrant] debops not installed if /vagrant is not mounted


nqb:
If you want to use DebOps `Vagrantfile` on a host system [like Windows](https://github.com/debops/debops/blob/40a22fc44c180d084af03f1622e21e871728697c/Vagrantfile#L816-L819) where `/vagrant` is not mounted, `vagrant` will fail and `debops` will not be install.

### Steps to reproduce
1. Disable mount of `/vagrant`
2. Run `vagrant up`

### Actual results
`vagrant` fail

### Expected results
`vagrant` should not fail and `debops` should be install.

### Proposal
`/vagrant` is only necessary if you want to build `debops` from sources. 

When `/vagrant` is not mounted, we could install `debops` from PyPI and use [code already present](https://github.com/debops/debops/blob/40a22fc44c180d084af03f1622e21e871728697c/Vagrantfile#L587-L593) to install master branch using `debops-update`.

drybjed:
Sounds good - we cannot really sync the workdir anyway, since symlinks will be broken. On Windows, Vagrant boxes can be treated as demos or to manage remote hosts; development will probably be done on Linux/macOS platforms anyway.

nqb:
I agree, I will provide a fix quickly.

Would it be possible to backport this change to stable releases ?

drybjed:
Of course, I don't see why not, as long as the changes are compatible.

nqb:
> We could install `debops` from PyPI and use [code already present](https://github.com/debops/debops/blob/40a22fc44c180d084af03f1622e21e871728697c/Vagrantfile#L587-L593) to install master branch using `debops-update`.

To be exact, we should use `debops-update` to checkout **current git branch**. I mean, if I'm on stable 1.1, we should use `debops-update` to install stable-1.1 branch.

drybjed:
You mean if you run `vagrant up` in DebOps monorepo checked out to `stable-1.1` branch, it should run `debops-update` and get the `stable-1.1` branch as well? Obvious way to do this is to pass the current branch/tag the monorepo is checked out from to inside of the Vagrant environment via env variables. I suppose that the `Vagrantfile` could run some `git` commands to get that information if possible and fall back gracefully to `master` oterwise.

I don't remember if `debops-update` supports cloning specific tag/branch directly at the moment, if not, implementing this would probably solve the need to check out the specific branch afterwards.

nqb:
[...]

PR #1087, opened 2019-10-01
Install debops from PyPI if /vagrant not mounted

nqb:
Related to #1061

### DIFF
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -21,8 +21,11 @@
 #     VAGRANT_NODE_BOX="debian/buster64"
 #         Specify the box to use for nodes.
 #
-#     ANSIBLE_FROM="debian" / ANSIBLE_FROM="pypi" / ANSIBLE_FROM="devel"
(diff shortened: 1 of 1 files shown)


drybjed:
Thanks! Looks good, however I'll test it in a few hours before merging.

drybjed (approved):
Looks like the changes work nicely. Thanks!

nqb:
The only drawback I saw is when you have run for example:
```
DEBOPS_FROM=pypi vagrant up
```
and after a halt of your VM, you run `vagrant up`. 

Provisioning will try to build `debops` from devel with an already installed `debops` from PyPI, we need to handle this case from my point of view.

drybjed:
How about:
```
export DEBOPS_FROM="pypi"
```
That should ensure a consistent environment for the duration of a given session.

nqb:
> How about:
> 
> ```
> export DEBOPS_FROM="pypi"
> ```

You mean on command-line when you run `vagrant up` for the first time ?

drybjed:
Yes, exactly. The variable will be available during a given session, so you don't need to specify it each time `vagrant` command is executed.

PR #1088, opened 2019-10-01
[debops-tools] Allow to checkout other branch than master

nqb:
Related to #1061 

This is my first attempt.

It works during initial checkout but I noticed unexpected behaviors when running `debops-update` after a successful checkout. I need to go deep into `update_git_repo` function.

@htgoebel, if you have time, I will appreciate your review ;-)

### DIFF
--- a/bin/debops-update
+++ b/bin/debops-update
@@ -157,44 +157,53 @@ def update_git_repository(path, branch, dry_run=False, remote_uri=False):
 
     current_branch = subprocess.check_output(
                 [GIT, 'rev-parse', '--abbrev-ref', 'HEAD']).decode('utf-8').strip()
-    print("Current branch checked out:", current_checkout_branch)
-    if dry_run:
-        subprocess.call([GIT, 'fetch'])
-        subprocess.call([GIT, 'diff', 'HEAD', 'origin/' + branch, '--stat'])
+    print("Current branch:", current_branch)
 
-    else:
-        # Fetch silently
-        # subprocess.call([GIT, 'fetch', '--quiet'])
-        subprocess.call([GIT, 'fetch'])
-
-        if current_branch != branch:
+    if current_branch != branch:
+        if dry_run:
+            subprocess.call([GIT, 'fetch'])
+            subprocess.call([GIT, 'diff', 'HEAD', 'origin/' + branch, '--stat'])
+        else:
             # checkout of new branch
             #subprocess.call([GIT, 'checkout', branch])
             subprocess.call([GIT, 'checkout', '--progress', branch])
+    else:
+        if dry_run:
+            subprocess.call([GIT, 'fetch'])
+            subprocess.call([GIT, 'diff', 'HEAD', 'origin/' + current_branch, '--stat'])
 
-        # Get the current sha of the head branch
-        current_sha = subprocess.check_output(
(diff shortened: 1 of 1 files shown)


nqb:
@drybjed and @htgoebel, tell me if I'm wrong but it looks like the part related to `debops-playbooks` should be deprecated.

drybjed:
@nqb Yes, that was to handle the old https://github.com/debops/debops-playbooks/ repository. I suppose that at this point the code can be removed entirely. You can do it in this PR if you want.

bin/debops-update
```diff
@@ -157,7 +160,7 @@ def update_git_repository(path, dry_run=False, remote_uri=False):
 
     if dry_run:
         subprocess.call([GIT, 'fetch'])
-        subprocess.call([GIT, 'diff', 'HEAD', 'origin', '--stat'])
+        subprocess.call([GIT, 'diff', 'HEAD', 'origin/'+branch, '--stat'])
```
drybjed: @nqb Perhaps you could reformat this change as
```python
'origin/' + branch
```
At first glance I thought that this was some weird way to pick a branch in `git`, not that `branch` was a Python variable.

nqb: Hi @drybjed,

I agree. But since this commit, I changed my idea on how to handle this case.

With current state of this PR, if you checkout a branch with `--branch stable-1.1` option, you will need to specify `--branch stable-1.1` when you will run `debops-update --dry-run` after a first checkout. If you didn't, you will get a diff between `stable-1.1` and `master`. That's not the case for merge operation because we use `SHA`.

So my idea is to use same code for `dry_run` and `merge` operations (use `SHA`) but handle cases where users use `--branch` after a first checkout.

Let me know if that sounds good for you.

drybjed: How about getting the currently checked out branch name with
```
git rev-parse --abbrev-ref HEAD
```
and using that in the diff and dry run?

drybjed:
@nqb Any progress with this PR? The next stable release is scheduled near the end of the November.

nqb:
Hi @drybjed,

I will focus on it to make it available for 1.2.0.

nqb:
I have one question regarding the way merge/pull operations need to be done by `debops-update`.

Currently, when we run `debops-update` for the second time (update), command will merge remote branch into local branch if SHA of local branch and SHA of remote branch differ. `debops-update` use `git rev-parse HEAD` and `git rev-parse FETCH_HEAD` to get SHAs.

I don't know how to handle following case:
```
# initial checkout on stable-1.1
debops-update --branch stable-1.1 .

# want to test something on master branch
debops-update .
```
I made code to checkout on `master` branch but getting `FETCH_HEAD` SHA return SHA of previous branch (here stable-1.1) and cause a merge conflict because `debops-update` is trying to merge `master` into `stable-1.1` (I guess).

I wonder if for this case, we should not simply make a `git pull`.

drybjed:
We only do this ourselves so that the changes can be previewed without merging them. If `git pull` can deal with already fetched changes, sure, I don't see why tihs couldn't be handled by `git` itself.

nqb:
@drybjed,

My `git` knowledge is limited. I'm not sure a `git pull` will always work even if you don't make changes in your local branch. I already saw cases where you need to make a merge into your local branch to be in sync with origin. But I can be wrong.

PR #1092, opened 2019-10-04
Fix owncloud occ programm interpreting the key and values of the command as parameters to itself.

reixd:
This problem occurs, when for example the LDAP service password for owncloud contains a dash ("-") as the first character :D Then the occ programm will interpret the password, which should be a value, as a parameter and then fail.
Therefore by putting the key and value in apostrophes solves the issue.

PR #1094, opened 2019-10-05
[debops.owncloud] Limit LDAP password characters

drybjed:
This patch specifies what characters can be used in the Nextcloud LDAP
password string. This should fix an error where a password with '-' at
the beginning is interpreted as the 'occ' script option.

Fixes #1092


drybjed:
Currently the string is quoted with `""` which indeed could result in something like a variable, for example `$adsd` to be expanded by the shell. Do you think that changing the quotes to `'` would solve this on the shell level? Also, the command itself is executed by the `command` Ansible module, that should give additional protection, right?

reixd:
> Do you think that changing the quotes to ' would solve this on the shell level?

Oh yes!

The shell interpretes stuff inside "", but not inside ''. 
Thus this way you can escape or hinder special chars or glob expansion.

drybjed:
OK then, I'll switch them over.

PR #1095, opened 2019-10-05
[debops.owncloud] Switch double quotes to single

drybjed:
This change should ensure that the strings provided by Ansible to the
shell are not interpreted by the shell by mistake - for example
a randomly generated password could contain a '$' character that could
be interpreted as a variable name.


reixd:
Thanks for the fix

PR #1101, opened 2019-10-07
[debops.owncloud] Improvements with host/group vars and doc

reixd:
- Add combined variables (just like in other roles).
- Add more expert config to the LDAP config

### DIFF
--- a/ansible/roles/debops.owncloud/defaults/main.yml
+++ b/ansible/roles/debops.owncloud/defaults/main.yml
@@ -1560,8 +1560,20 @@ owncloud__ldap_default_config:
 # The LDAP configuration options defined by the user.
 # See :ref:`owncloud__ref_ldap_config` for more details.
 owncloud__ldap_config: []
-
                                                                    # ]]]
+
+# .. envvar:: owncloud__ldap_group_config [[[
+#
+# This variable is intended to be used in a host inventory group of Ansible
+owncloud__ldap_group_config: []
+                                                                   # ]]]
+
+# .. envvar:: owncloud__ldap_host_config [[[
+#
+# This variable is intended to be used in the inventory of hosts.
+owncloud__ldap_host_config: []
+                                                                   # ]]]
+
 # .. envvar:: owncloud__ldap_combined_config [[[
 #
 # The variable that combines default and user LDAP configuration and is used in
--- a/ansible/roles/debops.owncloud/defaults/main.yml
+++ b/ansible/roles/debops.owncloud/defaults/main.yml
@@ -1599,12 +1599,6 @@ owncloud__ldap_combined_config: '{{ owncloud__ldap_default_config
                                     + owncloud__group_ldap_config
                                     + owncloud__host_ldap_config }}'
                                                                    # ]]]
-
-# .. envvar:: owncloud__ldap_user_object_classes [[[
-#
-# List of additional LDAP Object Classes to add to the user LDAP object.
-owncloud__ldap_user_object_classes: [ 'nextcloudAccount' ]
-                                                                   # ]]]
                                                                    # ]]]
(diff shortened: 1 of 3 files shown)


drybjed:
Thanks! Looks good, although comments for the LDAP config vriables could be a bit better worded. I'll test the configuration in a bit to make sure it works fine.

reixd:
Ok, I will rewrite the comments.
Do not merge yet

ansible/roles/debops.owncloud/defaults/main.yml
```diff
@@ -1533,17 +1570,46 @@ owncloud__ldap_default_config:
...
+# .. envvar:: owncloud__ldap_user_object_classes [[[
+#
+# List of additional LDAP Object Classes to add to the user LDAP object.
+owncloud__ldap_user_object_classes: [ 'nextcloudAccount' ]
+                                                                   # ]]]
+
+# .. envvar:: owncloud__ldap_group_object_classes [[[
+#
+# List of additional LDAP Object Classes to add to the user LDAP object.
+owncloud__ldap_group_object_classes: [ 'nextcloudGroup' ]
```
drybjed: These variables are not used anywhere else, care to explain their meaning? I think that they can be removed from defaults; these LDAP objectClasses only make sense in the LDAP directory itself.

reixd: The idea is to used them analogue to ```sshd__ldap_device_object_classes```, maybe from other ldap dependent roles.

drybjed: For that to be possible, in that hipothetical role you would have to include the `debops.owncloud` role as a dependency, so that hiphotetical LDAP role has access to these variables. I don't think that's what you want to do.

A better solution would be to expose these objectClasses through `owncloud.fact` script. That way other Ansible roles that would want to manipulate LDAP objects and add Nextcloud-related objectClasses to them could find out the names of the objectClasses to use. But this of course would only be available on the host with Nextcloud installed, where the role has been applied. I imagine that any Ansible roles that manage LDAP user accounts or groups, would be applied against the LDAP directory host itself, not random hosts in the cluster that use LDAP. So, exposing this kind of information through Ansible local facts would not be beneficial.

I think that the best way to do thiis would be to specify such objectClasses directly in that hipothetical LDAP user role. The objectClass names are known, and if the role is included in DebOps, if they change, the role will also be updated to match the changes. And if users use these roles outside of the DebOps environment, then they can always override the objectClass names through the inventory variables.

Do you agree?

reixd: > For that to be possible, in that hipothetical role you would have to include the `debops.owncloud` role as a dependency, so that hiphotetical LDAP role has access to these variables. I don't think that's what you want to do.

It is how many roles are working currently, however no, it's kind of "dirty".

> A better solution would be to expose these objectClasses through `owncloud.fact` script. 
> [...]
> So, exposing this kind of information through Ansible local facts would not be beneficial.

Agree.

> I think that the best way to do thiis would be to specify such objectClasses directly in that hipothetical LDAP user role. 
> 
> Do you agree?

Agree, I think its better this way. 
I was initially thinking to avoid duplicates to avoid problems during upgrades/updates. But this way is better.

WILCO

reixd: I will remove the objectClasses from this PR and push them in another one with the corresponding context.

drybjed (approved):
Looks good, thanks.

ypid:
Hey guys, I just reviewed this now as it was merged into master. Looks good. I have only one style remark. It is called "ownCloud". Please be consistent with the rest of the role ;-) No need to open a PR extra for this through :)

PR #1104, opened 2019-10-16
[docs] target stable-1.1 branch in quick start

bonesoul:
changed the git clone directive within vagrant quickstart -- now it target's stable-1.1 branch instead of master.

so that new users will have a stable experience.

### DIFF
--- a/docs/introduction/quick-start.rst
+++ b/docs/introduction/quick-start.rst
@@ -60,7 +60,7 @@ support, based on Debian Stretch:
 
 .. code-block:: console
 
-   git clone https://github.com/debops/debops ; cd debops
+   git clone -b stable-1.1 https://github.com/debops/debops ; cd debops
    vagrant up && vagrant ssh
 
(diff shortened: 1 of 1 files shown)


drybjed (approved):
Very well. But, on the next stable release this probably will have to be updated as well... I'll think about a solution - perhaps running:
```
git ls-remote --heads https://github.com/debops/debops | grep stable
```
will let you see the available stable branches, however that doesn't fit well with new DebOps users reading a quick start guide.

bonesoul:
how about an alias branch (ie named latest-stable) on git that always references to latest stable?

drybjed:
The `latest-stable` branch might be counter-productive. Users might expect a stable environment, and if they don't switch in time to their "current" stable branch, next stable release will swap the branch right under them to the next one which might break things. I'd like to avoid that.

Perhaps small explanation to how stable branches work in DebOps with a link to the release description could help here. But that can be added later, I'll merge this PR for now.

bonesoul:
okay sounds good

PR #1105, opened 2019-10-17
[debops.system_users] Add missing dependent variables to manage on the current playbook host

reixd:


ansible/roles/debops.system_users/defaults/main.yml
```diff
@@ -253,9 +269,11 @@ system_users__host_accounts: []
 system_users__combined_accounts: '{{ system_users__groups
                                      + system_users__group_groups
                                      + system_users__host_groups
+                                     + system_users__dependent_groups
```
drybjed: You probably would also want to flatten this variable as well.

drybjed (changes requested):
Well... From my point of view the `user` and `group` Ansible modules should be enough for Ansible roles to create the needed UNIX groups and accounts, but if users want to rely on this role to do that, why not. The `dependent` variables were not missing, they were intentionally left out to signify that this role is not designed for dependent operation. Oh well. :-)

Just flatten the `system_users__dependent_groups` variable and I'll merge it.

reixd:
>  The `dependent` variables were not missing, they were intentionally left out to signify that this role is not designed for dependent operation. 

I thought of keeping the role consistent, because other roles also provide "dependent services".

> Just flatten the `system_users__dependent_groups` variable and I'll merge it.

Done

drybjed (approved):

PR #1107, opened 2019-10-20
Add new role debops.firefly (unfinished)

ypid:
The role is unfinished, docs are missing. Basic functionality is provided however, the role should be able to setup a functional Firefly III instance.

I setup Firefly to play with it, however, I decided against it so I open this PR here in the hope that could be helpful. I don’t plan to work on finishing this role as I will not use Firefly III.

I am unsure how to handle this, as the PR would probably stay open for a long time. Maybe just merge it so that it can benefit from project wide improvements?

### DIFF
--- a/ansible/playbooks/service/firefly.yml
+++ b/ansible/playbooks/service/firefly.yml
@@ -1,0 +1,79 @@
+---
+
+- name: Setup and manage the personal finances manager Firefly
+  hosts: [ 'debops_service_firefly' ]
+  become: True
+
+  environment: '{{ inventory__environment | d({})
+                   | combine(inventory__group_environment | d({}))
+                   | combine(inventory__host_environment  | d({})) }}'
+
+  roles:
+
+    - role: debops.keyring
+      tags: [ 'role::keyring', 'skip::keyring', 'role::php',
+              'role::mariadb', 'role::postgresql',
+              'role::nginx', 'role::firefly' ]
+      keyring__dependent_apt_keys:
+        - '{{ php__keyring__dependent_apt_keys }}'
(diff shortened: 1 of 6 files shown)


drybjed:
Neat! Perhaps can you share your reasons for not using Firefly III yourself?

Looks like the application supports LDAP, Redis, multiple databases. It can be a nice example use case for DebOps, although we already have a few PHP-based applications already. The role seems to be about 70% done, all it needs is a better configuration template, integration with LDAP and Redis, and documentation.

ypid:
Nice that you appreciate the work :)

> Perhaps can you share your reasons for not using Firefly III yourself?

I did some rather deep comparison of personal accounting software and even checked all https://github.com/awesome-selfhosted/awesome-selfhosted#money-budgeting-and-management which could be used for personal accounting. Then I looked into the world of Plain Text Accounting.

To begin with, I was not so happy with a personal accounting software being a server application, but I wanted to give Firefly III a fair try. It is just not for me. I found other software suits my needs better. But other people have different needs and might find Firefly very suitable for them.

What I find is a killer feature of Firefly are the linked transactions. Why does no other software really have that? The closest I found was in plain text accounting (either beancount or hledger) and KMyMoney is to use tags for that. But come on, I want linked transactions.

Also, I find an Android app with full offline support very important. I did not test how good Firefly could handle that.

In the end, I found GnuCash and KMyMoney very suitable.

For tracking transactions on the go and later export them, I find My Expenses, Money Manager Ex or MoneyWallet interesting. I have not fully settled.

ypid:
To keep the list of open PRs actionable, I moved this unfinished work to https://github.com/debops/debops/wiki/DebOps-roles-in-development#ypid-roles.

PR #1108, opened 2019-10-22
[debops.mail_suite] Mail Suite with Virtual Mailboxes and LDAP user management

reixd:
This initial PR is part of an ongoing work to extend DebOps roles to support LDAP in order to install and configure a Mail Suite (Postfix SMTP, Dovecot IMAP, Roundcube Webclient).

- [X]  Extend debops.postfix to support Virtual Mailboxes over LDAP
- [ ]  Extend debops.dovecot to support user authentication over LDAP
- [ ]  Extend debops.roundcube to support user authentication over LDAP
- [X] Add debops.postfix documentation 
- [ ] Add debops.postfix guide on how to setup the LDAP environment
- [ ] Add debops.dovecot guide on how to setup the LDAP environment
- [ ] Add debops.roundcube guide on how to setup the LDAP environment
- [ ] Add ansible example inventory with working config and corresponding LDAP ldiff files

Please comment on this PR, if there need to additional changes made, if I forgot something or there's something wrong.


ansible/playbooks/service/postfix.yml
```diff
@@ -22,11 +22,23 @@
         - '{{ etc_aliases__secret__directories }}'
         - '{{ postfix__secret__directories }}'
 
+    - role: debops.system_users
+      tags: [ 'role::system_users', 'skip::system_users' ]
+      system_users__dependent_accounts:
+        - '{{ postfix__ldap_dependent_system_users
+              if (postfix__ldap_enabled|bool)
+              else [] }}'
```
drybjed, line 30: You shouldn't use the `debops.system_users` role as a dependent role in other DebOps roles, it was not meant to be used like that. The UNIX system group and UNIX account for the Postfix `vmail` user can be managed perfectly fine using existing `group` and `user` Ansible modules. The `debops.system_users` role is meant to be used primarly via the Ansible inventory and there's no need to drag it into each and every Postfix installation via the `postfix.yml` playbook.

ansible/playbooks/service/postfix.yml
```diff
@@ -22,11 +22,23 @@
...
+
     - role: debops.ferm
       tags: [ 'role::ferm', 'skip::ferm' ]
       ferm__dependent_rules:
         - '{{ postfix__ferm__dependent_rules }}'
 
+    - role: debops.ldap
+      tags: [ 'role::ldap', 'skip::ldap' ]
+      ldap__dependent_tasks:
+        - '{{ postfix__ldap__dependent_tasks }}'
```
drybjed, line 40: I don't think that LDAP configuration should be included directly in the `debops.postfix` role itself. This role was written to be a "core" of the Postfix installation which then can be augmented by other DebOps roles, for example `debops.postconf`, `debops.postscreen`. Since Postfix can support different Virtual Mail backends (PostgreSQL, MariaDB, LDAP, others), I believe that better way to do this would be to create `debops.postfix_ldap` role which then would contain all of the relevant LDAP configuration, and would use the `debops.postfix` role as a dependency to modify the Postfix configuration properly. With this setup, it would be easy for DebOps users to choose what kind of Postfix installation they want - plain UNIX accounts, some Virtual Mail backend, in the future some sort for mail filtering like `rspamd`, etc. Postfix and mail support in general is a complex beast and cramming everything into one Ansible role may quickly get things out of control.

ansible/roles/debops.postfix/defaults/main.yml
```diff
@@ -256,6 +271,351 @@ postfix__tls_dh512_param_file: '{{ ansible_local.dhparam[postfix__dhparam_set]
...
+# Currently the Virtual Mail support works only with LDAP enabled, in the future `mariaDB` could be enabled.
+#
+
+# .. envvar:: postfix__ldap_enabled [[[
+#
+# In order to enable Virtual Mail support, you need to enable LDAP authentication support manually.
+# It is disabled by default in order to define a single Mail Relay Server.
+# When enabled, `postfix` will configure mail-boxes locally and allow LDAP authenticated users to
+# send email over this instance.
+postfix__ldap_enabled: 'False'
```
drybjed: This value will be resolved to `True` by Ansible `bool` filter. You should use a `False` boolean here, not a string.

Then again, if/when the LDAP configuration is moved to a separate role, this variable might not even be needed anymore - adding a host to the `[debops_service_postfix_ldap]` Ansible inventory group would be enough to activate the support... On the other hand, additional toggle the enables/disables LDAP based on the `debops.ldap` Ansible local facts might be handy, so I would leave it in the new role as well.

ansible/roles/debops.postfix/defaults/main.yml
```diff
@@ -256,6 +271,351 @@ postfix__tls_dh512_param_file: '{{ ansible_local.dhparam[postfix__dhparam_set]
...
+# on the other hand it protects the mailboxes from external access.
+# Only vmail (and root) are allowed to access the mailboxes.
+postfix__virtual_mail_posix_user: 'vmail'
+                                                                   # ]]]
+# .. envvar:: postfix__virtual_mail_posix_uidnumber [[[
+#
+# Virtual Mail POSIX uidNumber
+#
+postfix__virtual_mail_posix_uidnumber: 5000
+                                                                   # ]]]
```
drybjed, line 314: You shouldn't specify static UID/GID allocations in a role, especially in the normal user range above >1000. Just make sure that the `vmail` UNIX group and account are created with the "system" parameter so that they have UID/GID below <1000.

In a single host setup you don't have to care what specific UID/GID the group and account has. In a distributed setup it might be better to define such account in LDAP itself so that it is consistent across hosts. For this, you probably would want to add a variable to enable/disable local account and group creation so that the relevant Ansible modules are skipped and don't break the run when they notice that the account and group is already present in LDAP.

Also, you shouldn't add an empty line at the end of the comment here, before the actual variable.

ansible/roles/debops.postfix/defaults/main.yml
```diff
@@ -256,6 +271,351 @@ postfix__tls_dh512_param_file: '{{ ansible_local.dhparam[postfix__dhparam_set]
...
+# .. envvar:: postfix__virtual_mail_posix_user [[[
+#
+# Virtual Mail POSIX username
+# For the accesses to the mailbox directories a separate user `vmail` (`Virtual Mail`) is created,
+# under which the accesses of Postfix, Dovecot and other components of the mail server should take place.
+# On the one hand this prevents mail server components from accessing sensitive system directories,
+# on the other hand it protects the mailboxes from external access.
+# Only vmail (and root) are allowed to access the mailboxes.
+postfix__virtual_mail_posix_user: 'vmail'
+                                                                   # ]]]
```
drybjed, line 308: I'm not sure how close Postfix Virtual Mail is tied to LDAP/MariaDB/PostgreSQL support. Perhaps the `vmail` setup could be moved into its own separate role, for example `debops.postfix_virtual`, that depends on `debops.postfix` to configure Postfix itself.

Virtual mail, especially the UNIX account part, would probably be useful for other roles as well, like `debops.dovecot`. You could expose information about the `vmail` account via Ansible local facts, and `debops.dovecot` role could use them to configure the mail backend.

ansible/roles/debops.postfix/defaults/main.yml
```diff
@@ -256,6 +271,351 @@ postfix__tls_dh512_param_file: '{{ ansible_local.dhparam[postfix__dhparam_set]
...
+# Specify zero or more `type:name` lookup tables, separated by whitespace or comma.
+# Tables will be searched in the specified order until a match is found.
+# Note: these lookups are recursive.
+postfix__virtual_alias_maps:
+  - '$alias_maps'
+  - 'ldap:/etc/postfix/ldap/virtual_aliases.cf'
+                                                                   # ]]]
+
+# LDAP authentication [[[
+# ^^^^^^^^^^^^^^^^^^^^^^^
```
drybjed, line 364: That's the wrong header marker for that particular file.

ansible/roles/debops.postfix/defaults/main.yml
```diff
@@ -256,6 +271,351 @@ postfix__tls_dh512_param_file: '{{ ansible_local.dhparam[postfix__dhparam_set]
...
+# .. envvar:: postfix__ldap_user_filter [[[
+#
+# The LDAP filter used to look up user accounts in the directory.
+postfix__ldap_user_filter: '(&
+                               (objectClass=inetOrgPerson)
+                               (objectClass=postfixUser)
+                             )
+                             (mailEnabled=TRUE)
+                             (|
+                               (authorizedService=postfix)
```
drybjed, line 565: I think that going with `smtpd` here instead of `postfix` would make this value more universal, since other DebOps roles that manage SMTP services could then use that filter value as well.

ansible/roles/debops.postfix/defaults/main.yml
```diff
@@ -1357,5 +1792,71 @@ postfix__ferm__dependent_rules:
...
+
+  - name: '{{ postfix__virtual_mail_posix_user }}'
+    group: '{{ postfix__virtual_mail_posix_group }}'
+    uid: '{{ postfix__virtual_mail_posix_uidnumber }}'
+    gid: '{{ postfix__virtual_mail_posix_gidnumber }}'
+    system: True
+    comment: 'Postfix Virtual Mail user'
+    shell: '/usr/sbin/nologin'
+    home: '{{ postfix__virtual_mailbox_base }}'
+    skeleton: null
```
drybjed: This is an interesting trick! Perhaps with this various UNIX account home directories created for system services by DebOps would not be littered with stuff from `/etc/skel/`. Thanks, I'll look into using it more.

ansible/roles/debops.postfix/tasks/ldap.yml
```diff
@@ -0,0 +1,59 @@
+---
+
+- name: Make sure 'postfix/ldap' directory exists
+  file:
+    dest: '/etc/postfix/ldap'
+    state: 'directory'
+    owner: 'root'
+    group: 'root'
```
drybjed, line 43: You shouldn't specify owner and group if these are `root:root`, that will be the default result. Doing it this way helps on different operating systems, like FreeBSD, where `root` UNIX group may not exist and `wheel` is usually used instead.

ansible/roles/debops.postfix/tasks/ldap.yml
```diff
@@ -0,0 +1,59 @@
...
+    dest:  '/etc/postfix/ldap/{{ item }}.cf'
+    owner: 'postfix'
+    group: 'postfix'
+    mode:  '0400'
+  loop:
+    - 'virtual_aliases'
+    - 'virtual_recipients'
+    - 'virtual_domains'
+    - 'virtual_senders'
+  no_log: '{{ postfix__no_log|d(True) }}'
```
drybjed: The `debops.postfix` role already implements and exposes a [mechanism for managing lookup tables](https://docs.debops.org/en/master/ansible/roles/debops.postfix/defaults-detailed.html#postfix-lookup-tables), you should use that instead of writing the lookup tables directly.

ansible/roles/debops.postfix/tasks/ldap.yml
```diff
@@ -0,0 +1,59 @@
...
+
+- name: "Make sure vmail home directory ({{ postfix__virtual_mailbox_base }}) exists"
+  file:
+    dest: '{{ postfix__virtual_mailbox_base }}'
+    state: 'directory'
+    owner: '{{ postfix__virtual_mail_posix_user }}'
+    group: '{{ postfix__virtual_mail_posix_group }}'
+    mode: '0755'
```
drybjed: The mailbox base is the home directory of the `vmail` UNIX account, right? That one should already exist, you don't need this particular task. Unless you want to make the directory mode more strict, which might be a good idea.

ansible/roles/debops.postfix/tasks/ldap.yml
```diff
@@ -0,0 +1,59 @@
...
+#mkdir -p /var/vmail
+#chown vmail:mail /var/vmail
+#chmod 770 /var/vmail
+#
+
+#addgroup --system --gid 5000 vmail
+#adduser --system --home /var/vmail --uid 5000 --gid 5000 --disabled-password --disabled-login -c "Virtual MailDir Handler" vmail
+#mkdir -p /var/vmail
+#chown -R vmail:vmail /var/vmail
+#chmod -R 770 /var/vmail
```
drybjed: This comment should be removed from the task list.

ansible/roles/debops.postfix/templates/ldap/postfix__virtual_ldap_header.j2
```diff
@@ -0,0 +1,21 @@
...
+##tls_ca_cert_file = {{ postfix__pki_ca_crt }}
+# Whether or not to request server's X509  certificate  and  check
+# its  validity  when  establishing SSL/TLS connections.
+# FIXME: set to yes again
+##tls_require_cert = no
+{% endif %}
+
+bind = yes
+bind_dn = {{ postfix__ldap_binddn }}
+bind_pw = {{ postfix__ldap_bindpw }}
```
drybjed, line 20: You will probably want to use this header file as an argument to manage the lookup tables directly instead of going through the `postfix__lookup_tables` variable. Actually I think that this might be better utilized by implementing `item.connection_config` parameter in the `debops.postfix` role itself so that it can be shared across many lookup tables easily. I'll look into it.

ansible/roles/debops.system_users/defaults/main.yml
```diff
@@ -205,6 +205,14 @@ system_users__group_groups: []
...
 system_users__host_groups: []
 
+                                                                   # ]]]
+# .. envvar:: system_users__dependent_groups [[[
+#
+# List of UNIX system groups to manage on the current playbook host. This variable is
+# meant to be used from a role dependency in :file:`role/meta/main.yml` or in
+# a playbook.
+system_users__dependent_groups: []
+
```
drybjed: These have been added in a separate PR and should be removed from this one.

ansible/roles/debops.slapd/files/etc/ldap/schema/debops/postfix.schema
```diff
@@ -0,0 +1,56 @@
+# postfix.schema - Postfix schema definition
+
```
drybjed, line 2: The schema you proposed is very basic, and I would like to include a much more expansive one, that supports mail routing via LDAP attributes, mail groups, etc. The closest I would see to a working schema definition is [50ns-mail.ldif](https://pagure.io/389-ds-base/blob/master/f/ldap/schema/50ns-mail.ldif) which is somewhat based on the [Netscape mailGroup schema](https://ldapwiki.com/wiki/Original%20Mailgroup%20Schema%20From%20Netscape), but of couse it would have to be rewritten in the `.schema` format for DebOps. The [draft-lachman-ldap-mail-routing-03.txt](https://tools.ietf.org/html/draft-lachman-ldap-mail-routing-03) RFC also has some interesting ideas, but I'm not sure how compatible they are with the Netscape schema. The [Enterprise mail routing with Postfix and LDAP](https://blog.felipe-alfaro.com/2006/05/24/enterprise-mail-routing-with-postfix-and-ldap/) might also be worth a read.

In the end I'm not ure if we can use the OIDs from `50ns-mail.ldif` schema and just extend it for our needs, or go with a separate OID namespace based on DebOps OID registration. Still, since messing with already imported LDAP schemas in OpenLDAP is hard bordering on impossible, this inclusion would probably be worth a separate pull request so that it can be worked on separately from the rest of Postfix-based code and thoughly engineered.

Looking at Dovecot and other core mail services during schema creation might also be worth it. Roundcube probably doesn't really need all that much unique objects or attributes, not sure about other services.

drybjed (changes requested):
@reixd Thanks for your work exploring the Postfix - LDAP integration! There's lots of good stuff in this pull request, however I would like the various tasks and variables set in `debops.postfix` to be moved to separate Ansible roles, so that the `debops.postfix` role can exist as a small "core" which other Ansible roles can use to manage Postfix in a composable way.

Please read the various comments below for my notes.

ansible/playbooks/service/postfix.yml
reixd: This is a good idea.
Regarding the different virtual mail backends I suggest to create a `debops.postfix_virtual` role. Then using variables choose the desired backend. It remains flexible, yet easy to understand

ansible/roles/debops.postfix/defaults/main.yml
reixd: This is really a dumb mistake. Just forgot to remove the ''. On my inventory I do have a boolean 😅

ansible/roles/debops.postfix/defaults/main.yml
reixd: Without a fixed uid/gid I had problems with the permissions. 
Maybe generate system user and fetch the uid/gid and store it as fact?

ansible/roles/debops.postfix/defaults/main.yml
reixd: Why?

ansible/roles/debops.postfix/defaults/main.yml
reixd: Agree

ansible/roles/debops.postfix/tasks/ldap.yml
reixd: Would that also work with a different role?

ansible/playbooks/service/postfix.yml
drybjed: That would result in variable names like `postfix_virtual__ldap_self_rdn`, so I dunno, kind of deeply nested for me. Different vmail backends have different methodology and configuration needs, I think that putting all that into single `debops.postfix_virtual` role could result in complicated mess. Separate roles for each backend seem better for me, especially that I don't see many setups where you use multiple backends at once on the same server.

ansible/roles/debops.postfix/defaults/main.yml
drybjed: This could be handled by using the `getent` Ansible module and getting the UID/GID of the specified user and group at runtime... But knowing the exact reason could probably give a better solution.

ansible/roles/debops.postfix/defaults/main.yml
drybjed: reStructuredText uses header markers to define the document structure, but they are expected to be the same on the same level. With this, you just put a few variables into a sub-section of the previous section. Sphinx will happily parse this, but that's not what you most likely want here.

ansible/roles/debops.postfix/tasks/ldap.yml
drybjed: Of course, that was the whole point to separate implementation details from other roles and focus them in `debops.postfix` role. Here's [definition of few lookup tables in `debops.postconf` role](https://github.com/debops/debops/blob/master/ansible/roles/debops.postconf/defaults/main.yml#L101-L181) and here's [how they are passed to the `debops.postfix` role](https://github.com/debops/debops/blob/master/ansible/roles/debops.postconf/defaults/main.yml#L214-L220). Note that you can still override the lookup tables via [these variables](https://github.com/debops/debops/blob/master/ansible/roles/debops.postconf/defaults/main.yml#L182-L201) separately from the `debops.postfix` role in the inventory. It's all done so that modifications through the Ansible inventory are possible.

reixd:
I will apply the small fixes and then close this PR, so I can start the new role with a clean sheat (branch).

reixd:
Pushed the small fixes and now closing this PR in order to create a better role.

PR #1113, opened 2019-10-25
[debops-tools] set yaml as default stdout_callback

nqb:
Really improve display of errors in tasks.


nqb:
If this change is too intrusive, we can also add it as a default parameter but commented (like for `callback_plugins`).

drybjed (approved):
Sounds like a good default, thanks. I checked the change a few days with different error cases and the messages are much clearer.

In the future I think that somewhat improved `unixy` callback plugin would also be interesting to add to DebOps. It gives Ansible output more shell-like look.

PR #1114, opened 2019-10-25
Small fixes for DebOps virtualization tools (Vagrant, Docker)

reixd:
When using the Docker containers and Vagrant VMs I found small issues that needed small fixes.

@drybjed  feel free to cherry pick changes if needed ;)

Cheers

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -58,7 +58,7 @@ ENV USER ansible
 
 # Add contents of the DebOps monorepo to the container
 # with the right permissions
-COPY --chown=ansible . .local/share/debops/debops
+COPY --chown=ansible:ansible . .local/share/debops/debops
 
 ENTRYPOINT ["/home/ansible/.local/share/debops/debops/lib/docker/docker-entrypoint"]
 CMD ["/bin/bash"]
--- a/Dockerfile
+++ b/Dockerfile
@@ -58,7 +58,7 @@ ENV USER ansible
 
 # Add contents of the DebOps monorepo to the container
 COPY . .local/share/debops/debops
-RUN chown -R ansible:ansible /home/ansible/.local
+RUN sudo chown -R ansible:ansible /home/ansible/.local
 
 ENTRYPOINT ["/home/ansible/.local/share/debops/debops/lib/docker/docker-entrypoint"]
 CMD ["/bin/bash"]
--- a/Dockerfile
+++ b/Dockerfile
@@ -22,6 +22,7 @@ RUN apt-get -q update \
     && DEBIAN_FRONTEND=noninteractive apt-get \
        --no-install-recommends -yq install \
        iproute2 \
+       iputils-ping \
        levee \
        openssh-client \
        python3-apt \
--- a/Dockerfile
+++ b/Dockerfile
@@ -57,8 +57,8 @@ WORKDIR /home/ansible
 ENV USER ansible
 
 # Add contents of the DebOps monorepo to the container
-COPY . .local/share/debops/debops
-RUN sudo chown -R ansible:ansible /home/ansible/.local
+# with the right permissions
+COPY --chown=ansible . .local/share/debops/debops
 
 ENTRYPOINT ["/home/ansible/.local/share/debops/debops/lib/docker/docker-entrypoint"]
 CMD ["/bin/bash"]
--- a/Dockerfile
+++ b/Dockerfile
@@ -58,6 +58,7 @@ ENV USER ansible
 
 # Add contents of the DebOps monorepo to the container
 COPY . .local/share/debops/debops
+RUN chown -R ansible:ansible /home/ansible/.local
 
 ENTRYPOINT ["/home/ansible/.local/share/debops/debops/lib/docker/docker-entrypoint"]
 CMD ["/bin/bash"]
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -31,7 +31,10 @@
 #         Specify the way to install debops.
 #
 #     VAGRANT_HOSTNAME="buster"
-#         Set a custom hostname after the box boots up.
+#         Set a custom hostname-prefix after the box boots up.
+#
+#     VAGRANT_HOSTNAME_MASTER="master"
+#         Set a custom hostname-prefix after the box boots up.
 #
 #     CONTROLLER=false
 #         Set to 'true' to set up a configuration with normal Diffie-Hellman
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -737,6 +737,14 @@ else
 end
 master_fqdn = master_hostname + '.' + VAGRANT_DOMAIN
 
+VAGRANT_HOSTNAME = (ENV['VAGRANT_DOTFILE_PATH'] || '.vagrant') + '/vagrant_hostname'
+if File.exist? VAGRANT_HOSTNAME
+      node_hostname_prefix = IO.read( VAGRANT_HOSTNAME ).strip
+else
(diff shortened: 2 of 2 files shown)


Vagrantfile
```diff
@@ -21,6 +21,9 @@
 #     VAGRANT_NODE_BOX="debian/buster64"
 #         Specify the box to use for nodes.
 #
```
nqb, line 23: It could be nice to add docs on:
- VAGRANT_HOSTNAME
- VAGRANT_HOSTNAME_MASTER

variables on top of the file.

nqb:
Something I've noticed when using Vagrant like this (before your changes):
```
VAGRANT_NODES=2 VAGRANT_HOSTNAME=debops vagrant up
```
is that `/etc/hosts` files doesn't contain short name of nodes. You had:
```
vagrant@debops:~$ cat /etc/hosts
127.0.0.1	localhost
127.0.1.1	debops.vagrant.test debops
[..]
172.28.128.5	debops-node1.vagrant.test debops-node1
```
and you should have:
```
172.28.128.5	debops-node1.vagrant.test debops-node1 node1
```
with `node1` as short name. This is caused by https://github.com/debops/debops/blob/master/Vagrantfile#L106.

@reixd, let me know if you want to adjust this here or I will open a separate issue.

reixd:
@nqb I would prefer you tackle this issue, I have other stuff to debug xD

Nevertheless this "hostname issue" all nodes can ping each other with fqdn or aliases.

nqb:
> @nqb I would prefer you tackle this issue, I have other stuff to debug xD

Ok @reixd 

> Nevertheless this "hostname issue" all nodes can ping each other with fqdn or aliases.

Yes I know but it's quite useful to run a playbook with a short name.

reixd:
@drybjed can you please restart the test suite. Apparently the last error was not caused by my changes.

Dockerfile
```diff
@@ -57,6 +58,7 @@ ENV USER ansible
 
 # Add contents of the DebOps monorepo to the container
 COPY . .local/share/debops/debops
+RUN sudo chown -R ansible:ansible /home/ansible/.local
```
drybjed: @reixd Don't use `sudo` here, Dockerfile already work with `root` permissions.

Edit: ah, sorry, that's after the `USER` statement. Hmmm, isn't `/home/ansible/.local` owned by `ansible` user at this point? Actually why would you even `chown` this, the Docker image is read-only...

reixd: I was using this image as the controller and got some issues with the permission under the given path. So I thought to fix it.
Yes its true, docker works as root, but after setting `USER ansible` not anymore, therefore the need to use `sudo`.

drybjed: OK, I checked in a Docker container and the `/home/ansible/.local/` directory and subdirectories are owned by `root` account, not `ansible`. I thought that by placing the `COPY` command after the `USER` command the files would be owned by `ansible:ansible`, I gues that changed at some point? Anyway, the solution would be to move the `COPY` command before the `USER` command, `RUN` a `chown` command afterwards without `sudo`, and then switch to `USER ansible`.

reixd: > Anyway, the solution would be to move the `COPY` command before the `USER` command, `RUN` a `chown` command afterwards without `sudo`, and then switch to `USER ansible`.

That would create many more image layers, making the docker image bigger.

RTFM helps, I just found out what's needed:  `COPY --chown=patrick hostPath containerPath`
See https://docs.docker.com/engine/reference/builder/#copy

drybjed:
@reixd I tested both Vagrant and Dockerfile changes. Dockerfile modifications look OK in the end, but the `~ansible/.local/share/debops` subdirectories are still owned by `root:root`, I assume that doesn't give you any issues? In that case I think that I'll cherry-pick the Dockerfile changes and make a new PR with just them.

The Vagrantfile changes result in the additional nodes not added to the `/etc/hosts` file. I think that it might be because the random suffix for the `master` host and the additional nodes is different. Can you explain a bit what your changes are meant to be? I think that you want to be able to rename the `master` host to something else but that breaks the node detection via Avahi.

nqb:
> The Vagrantfile changes result in the additional nodes not added to the `/etc/hosts` file. I think that it might be because the random suffix for the `master` host and the additional nodes is different. Can you explain a bit what your changes are meant to be? I think that you want to be able to rename the `master` host to something else but that breaks the node detection via Avahi.

Current code only allow you to change default prefix `debops` by setting `VAGRANT_HOSTNAME` to a different value. For example:
```
VAGRANT_HOSTNAME=buster VAGRANT_NODES=1 vagrant up
```
will create two VMS: 
- `buster`
- `buster-node1`

I supposed what @reixd wants is to have different name between `master` and `nodes`. For example:
```
VAGRANT_HOSTNAME_MASTER=controller VAGRANT_HOSTNAME=hosts VAGRANT_NODES=1 vagrant up
```
will create two VMS: 
- `controller`
- `hosts-node1`

**Important** : if nobody needs it in stable branch, I will prefer to keep these changes only on master branch. Thanks.

drybjed:
In that case, without somehow handling the detection of other VMs in the current cluster via Avahi, other nodes will not be added in the `/etc/hosts` files. Without using a randomized string to tag the nodes in the current Vagrant instance, the different clusters started in parallel could overlap and cause issues in host resolution and Ansible inventory.

Are you guys using the DebOps Vagrantfile to deploy hosts in production somehow? Because I never intended the current setup to be used like that, it was meant for testing different deployment setups but nevet to be used directly...

As for backporting the changes to the stable releases - that depends. If the current features are preserved, I don't see a reason why not do it; if the changes are breaking functionality, they won't be backported.

reixd:
> I supposed what @reixd wants is to have different name between `master` and `nodes`. For example:
> 
> ```
> VAGRANT_HOSTNAME_MASTER=controller VAGRANT_HOSTNAME=hosts VAGRANT_NODES=1 vagrant up
> ```
> 
> will create two VMS:
> 
> * `controller`
> * `hosts-node1`
> 

This is my intention, have the possibility to rename the prefix of the master/controller and the node VMs.
For example for my infrastructure I like to use `controller` instead of `master`. 
I thought my changes were less invasive...

reixd:
> Without using a randomized string to tag the nodes in the current Vagrant instance, the different clusters started in parallel could overlap and cause issues in host resolution and Ansible inventory.

Agree, the random suffix for the hostname is necessary.


> Are you guys using the DebOps Vagrantfile to deploy hosts in production somehow? Because I never intended the current setup to be used like that, it was meant for testing different deployment setups but nevet to be used directly...

NACK. Wanted to create a simple test environment with a controller and +3 VMs. Add the end it was easier for me to use docker-compose :D (can post the example docker environment if needed).

drybjed:
@reixd I think that if the random string was moved from the hostname to Avahi DNS records, then the custom hostnames would make sense. The random string probably would have to be read from the host environment configured by Vagrant and stored in a TXT record which is then retrieved somehow and used to detect other cluster nodes.

This would probably necessiate a major overhault of the Vagrantfile so I will not do it right away, but might be a good idea in the future. If you want to take a knack at it, be my guest. :-)

reixd:
I just wanted to try vagrant out, but it did not worked quite well, specially mounting the project dir....
The changes were thought as improvements and quite frankly I didn't expect this "funky" behavior with the hostnames. 
@drybjed you can cherry pick the lines 24-26 and then close this PR. 

Currently my necessity of vagrant nodes is not high enough to justify more time in this issue.

drybjed:
I cherry-picked the suggested commit, closing the PR as requested.

PR #1117, opened 2019-10-25
service/php: Allow specification of listen.backlog; document more

apapsch:
Allow specification in pool config, while defaulting to a debops defaults variable. The default value of 511 keeps the behavior that existed with the commented out directive.

Parts of the docs text is blatantly copied from the role `defaults/main.yml` or `pool.conf.j2`. The advice regarding sysctl parameters is useful if you use php-fpm as an async queue (which works fine actually, but you'll need some bookkeeper).


drybjed (approved):
Nice work, thanks! I'll backport these changes to the stable releases as well.

PR #1112, opened 2019-10-24
Security fixes for CVE-2019-11043 (nginx + php-fpm vulnerability)

drybjed:

PR #1118, opened 2019-10-25
[debops.owncloud] Fix when running playbook with --tags role::nginx

ypid:
`owncloud__do_autosetup` is set by a `set_fact` task which is not run with `role::nginx` for example.

fatal: [cloud.example.org]: FAILED! => {"msg": "The conditional check '(owncloud__do_autosetup|bool and not ansible_check_mode)' failed. The error was: error while evaluating conditional ((owncloud__do_autosetup|bool and not ansible_check_mode)): 'owncloud__do_autosetup' is undefined\n\nThe error appears to have been in '/home/user/.ansible/ypid-ansible-common/submodules/debops/ansible/roles/debops.owncloud/tasks/setup_owncloud.yml': line 130, column 3, but may\nbe elsewhere in the file depending on the exact syntax problem.\n\nThe offending line appears to be:\n\n\n- name: Set ownCloud apps configuration for each app\n  ^ here\n"}

Related to: #1112

### DIFF
--- a/ansible/roles/debops.owncloud/tasks/setup_owncloud.yml
+++ b/ansible/roles/debops.owncloud/tasks/setup_owncloud.yml
@@ -132,7 +132,7 @@
   loop_control:
     loop_var: 'owncloud__apps_item'
   with_dict: '{{ owncloud__apps_config_combined|d({}) }}'
-  when: (owncloud__do_autosetup|bool and not ansible_check_mode)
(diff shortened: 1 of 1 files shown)


drybjed (approved):
Thanks for the bugfix, I'll backport it to the stable releases as well, however new releases will be performed in a few days.

PR #1120, opened 2019-10-26
[debops.nginx] Fix custom location try_files bug

### LINKED ISSUE #1119: debops.nginx try_files doesn't work expected when there is custom location in the server block
Os: Ubuntu 18.04
Python: 3.6.8
Ansible: 2.8.6
Debops: 1.1.4

As the title explains; `try_files` variable doesn't work as expected when there is `location` block on the server configuration. However it correctly configure `try_files` when I commented out the `location` block. Otherwise it pushes `nginx_default_try_files`. Here is the the config I've located the bug on it;

```yaml
nginx_server_www_example_com:
  name: [ 'www.example.com', 'example.com' ]
  acme: True
  type: php
  redirect_from: True
  php_upstream: 'example_com'
  try_files: $uri $uri/ /index.php?$args
  options: |
    set $skip_cache 0;

    # POST requests and urls with a query string should always go to PHP
    if ($request_method = POST) {
        set $skip_cache 1;
    }

    if ($query_string != "") {
        set $skip_cache 1;
    }

    # Don't cache uris containing the following segments
    if ($request_uri ~* "/wp-admin/|/xmlrpc.php|wp-.*.php|/feed/|index.php|sitemap(_index)?.xml") {
        set $skip_cache 1;
    }

    # Don't use the cache for logged in users or recent commenters
    if ($http_cookie ~* "comment_author|wordpress_[a-f0-9]+|wp-postpass|wordpress_no_cache|wordpress_logged_in") {
        set $skip_cache 1;
    }
  location:
    '~ /purge(/.*)': 'fastcgi_cache_purge EXAMPLE_COM "$scheme$request_method$host$1";'

  php_options: |
      fastcgi_cache_bypass $skip_cache;
      fastcgi_no_cache $skip_cache;
      fastcgi_cache EXAMPLE_COM;
      fastcgi_cache_valid 60m;
      add_header X-FastCGI-Cache $upstream_cache_status;


nginx_http_extra_options: |
  fastcgi_cache_path /var/run/nginx-fastcgi-cache levels=1:2 keys_zone=EXAMPLE_COM:100m inactive=60m;
  fastcgi_cache_key "$scheme$request_method$host$request_uri";
  fastcgi_cache_use_stale error timeout invalid_header http_500;
  fastcgi_ignore_headers Cache-Control Expires Set-Cookie;
```

thiras:
I've tried to fix #1119 bug. I've saved the existing config if there is no `try_files` defined. Please let me know if something is wrong.

### DIFF
--- a/ansible/roles/debops.nginx/templates/etc/nginx/sites-available/php.conf.j2
+++ b/ansible/roles/debops.nginx/templates/etc/nginx/sites-available/php.conf.j2
@@ -90,7 +90,7 @@
                 limit_except {{ item.php_limit_except if item.php_limit_except is string else item.php_limit_except | join(' ') }} { deny all; }
 
 {%     endif %}
-                try_files $script_name =404;
+                try_files {{ (([ item.try_files ] if item.try_files is string else item.try_files) if item.try_files|d() else ["$script_name =404"]) | join(' ') }};
 
 {%     if (item.php_include is undefined or item.php_include|d()) %}
                 include {{ item.php_include | default('fastcgi_params') }};


drybjed:
The problem with this solution is that if you specify the `try_files` parameter, it will replace both in the `/` location, as well as in the PHP location blocks. But it's probably a good idea to allow modification of the `try_files` option in the PHP blocks, so I'll add a `item.php_try_files` parameter which will permit that.

Also, you made a pull request against the `stable-1.1` branch, you should do it against the `master` branch instead. :-)

PR #1123, opened 2019-10-28
[debops.nginx] Fix logic of root dirs and welcome page creation

tasosalvas:
## The issue
As it was, `debops.nginx` was skipping the creation of server root directories in a bunch of cases, specifically when  `item.owner`, `.group` or `.root` was set.

Subsequently it errored out if any servers with `.owner` set contained `welcome: True`, while trying to write to `{{ nginx_www }}/sites/{{ name[0] }}/public/`, while it should be pointing at `{{ nginx_www }}/{{ owner }}/sites/{{ name[0] }}/public/`.

## Stuff I did in `tasks/nginx_servers.yml`
[debops.nginx] Fix logic of root dirs and welcome page creation

- The welcome pages now handle `item.owner` and `item.root` as described in the documentation.
- The root dir creation task doesn't skip server entries that have `item.group` set anymore.
- Implemented `item.group` so it is actually used.
- Replaced a couple hardcoded instances of `/public/` with `{{ item.public_dir_name | d(nginx_public_dir_name) }}`

### root dirs and welcome pages
I think I've made the new root path expression as readable as I could. I did it first with string concatenations, but I like how `.format()` keeps the path template in its own line.

### The undocumented `item.group`
The first task of the role ("Create global webroot directories if allowed") only ran if a server had `item.group` undefined. Then it never bothered with it again. The documentation doesn't even mention it.

I have set it to take precedence over `item.owner` and `nginx_webroot_group` if defined. 

I just saw it and thought I'd fix it, so if it's unwanted I'm not invested in it or anything. If it's ok to keep, do tell me so I fix its documentation too.

---
I'm pretty new to ansible and just now finding my way around debops, so any feedback is appreciated. I won't be surprised if I'm missing something obvious.

I also spotted a bunch of hardcoded `/etc/nginx` references (instead of `{{ nginx_etc_path }}`. I thought I'd leave them for another PR, since they seem to span more files than `nginx_servers.yml` and I thought I'd introduce myself first.

So, yeah, hi!
Tasos Alvas

### DIFF
--- a/ansible/roles/debops.nginx/tasks/nginx_servers.yml
+++ b/ansible/roles/debops.nginx/tasks/nginx_servers.yml
@@ -1,13 +1,21 @@
 ---
-
 - name: Create global webroot directories if allowed
   file:
-    path: '{{ nginx_www }}/sites/{{ item.name
-                                    if item.name is string else item.name[0]
-                                      | d("default") }}/{{ item.public_dir_name | d(nginx_public_dir_name) }}'
+    path: '{{ item.root if item.root is defined else (
+          "{}/{}/sites/{}/{}".format(
+          nginx_www,
+          item.owner,
+          (item.name if item.name is string else item.name[0]) | d("default"),
+          item.public_dir_name | d(nginx_public_dir_name))
+          if item.owner is defined else
+          "{}/sites/{}/{}".format(
+          nginx_www,
+          (item.name if item.name is string else item.name[0]) | d("default"),
+          item.public_dir_name | d(nginx_public_dir_name))
+          )}}'
     state: 'directory'
-    owner: '{{ nginx_webroot_owner }}'
-    group: '{{ nginx_webroot_group }}'
+    owner: '{{ item.owner | d(nginx_webroot_owner) }}'
+    group: '{{ item.group | d(item.owner | d(nginx_webroot_group)) }}'
     mode: '{{ nginx_webroot_mode }}'
   with_flattened:
     - '{{ nginx__servers }}'
@@ -18,16 +26,25 @@
     - '{{ nginx_default_servers | d([]) }}'
     - '{{ nginx_internal_servers | d([]) }}'
     - '{{ nginx_dependent_servers | d([]) }}'
-  when: (nginx_webroot_create|bool and item.state|d('present') != 'absent' and
-         (item.root is undefined and item.owner is undefined and item.group is undefined))
+  when: (nginx_webroot_create|bool and item.state|d('present') != 'absent')
 
 - name: Create default welcome page if enabled
   template:
     src: '{{ item.welcome_template | d(nginx_welcome_template) }}'
-    dest: '{{ nginx_www }}/sites/{{ item.name
-                                    if item.name is string else item.name[0] | d("default") }}/public/index.html'
+    dest: '{{ item.root + "/index.html" if item.root is defined else (
+          "{}/{}/sites/{}/{}/index.html".format(
+          nginx_www,
+          item.owner,
+          (item.name if item.name is string else item.name[0]) | d("default"),
+          item.public_dir_name | d(nginx_public_dir_name))
+          if item.owner is defined else
+          "{}/sites/{}/{}/index.html".format(
+          nginx_www,
+          (item.name if item.name is string else item.name[0]) | d("default"),
+          item.public_dir_name | d(nginx_public_dir_name))
+          )}}'
     owner: '{{ item.owner | d(nginx_webroot_owner) }}'
-    group: '{{ item.owner | d(nginx_webroot_group) }}'
+    group: '{{ item.group | d(item.owner | d(nginx_webroot_group)) }}'
     force: '{{ item.welcome_force|d() | bool }}'
   with_flattened:
     - '{{ nginx__servers }}'
@@ -45,10 +62,20 @@
 - name: Copy 'normalize.css' CSS file for welcome page
   template:
     src: 'srv/www/sites/welcome/public/normalize.css'
-    dest: '{{ nginx_www }}/sites/{{ item.name
-                                    if item.name is string else item.name[0] | d("default") }}/public/normalize.css'
+    dest: '{{ item.root + "/normalize.css" if item.root is defined else (
+          "{}/{}/sites/{}/{}/normalize.css".format(
+          nginx_www,
+          item.owner,
+          (item.name if item.name is string else item.name[0]) | d("default"),
+          item.public_dir_name | d(nginx_public_dir_name))
+          if item.owner is defined else
+          "{}/sites/{}/{}/normalize.css".format(
+          nginx_www,
+          (item.name if item.name is string else item.name[0]) | d("default"),
+          item.public_dir_name | d(nginx_public_dir_name))
+          )}}'
     owner: '{{ item.owner | d(nginx_webroot_owner) }}'
-    group: '{{ item.owner | d(nginx_webroot_group) }}'
+    group: '{{ item.group | d(item.owner | d(nginx_webroot_group)) }}'
     force: '{{ item.welcome_force|d() | bool }}'
   with_flattened:
     - '{{ nginx__servers }}'
--- a/ansible/roles/debops.nginx/tasks/nginx_servers.yml
+++ b/ansible/roles/debops.nginx/tasks/nginx_servers.yml
@@ -1,22 +1,16 @@
 ---
 - name: Create global webroot directories if allowed
   file:
-    path: '{{ item.root if item.root is defined else (
-          "{}/{}/sites/{}/{}".format(
-          nginx_www,
-          item.owner,
-          (item.name if item.name is string else item.name[0]) | d("default"),
-          item.public_dir_name | d(nginx_public_dir_name))
-          if item.owner is defined else
-          "{}/sites/{}/{}".format(
-          nginx_www,
-          (item.name if item.name is string else item.name[0]) | d("default"),
-          item.public_dir_name | d(nginx_public_dir_name))
-          ) }}'
+    path: '{{ item.root | d(
+              "{}/sites/{}/{}".format(
+                nginx_www + ("/" + item.owner if item.owner | d() else ""),
+                (item.name if item.name is string else item.name[0]) | d("default"),
+                item.public_dir_name | d(nginx_public_dir_name))
+              ) }}'
     state: 'directory'
     owner: '{{ item.owner | d(nginx_webroot_owner) }}'
     group: '{{ item.group | d(item.owner | d(nginx_webroot_group)) }}'
-    mode: '{{ nginx_webroot_mode }}'
+    mode: '{{ item.mode | d(nginx_webroot_mode) }}'
   with_flattened:
     - '{{ nginx__servers }}'
     - '{{ nginx__default_servers }}'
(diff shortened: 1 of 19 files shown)


ansible/roles/debops.nginx/tasks/nginx_servers.yml
```diff
@@ -1,13 +1,21 @@
...
+          nginx_www,
+          (item.name if item.name is string else item.name[0]) | d("default"),
+          item.public_dir_name | d(nginx_public_dir_name))
+          )}}'
     state: 'directory'
-    owner: '{{ nginx_webroot_owner }}'
-    group: '{{ nginx_webroot_group }}'
+    owner: '{{ item.owner | d(nginx_webroot_owner) }}'
+    group: '{{ item.group | d(item.owner | d(nginx_webroot_group)) }}'
     mode: '{{ nginx_webroot_mode }}'
```
drybjed: Perhaps this line could be changed to
```
mode: '{{ item.mode | d(nginx_webroot_mode) }}'
```
to allow for for better control over directory attributes. Don't forget to add the `mode` parameter in the documentation.

ansible/roles/debops.nginx/tasks/nginx_servers.yml
```diff
@@ -18,16 +26,25 @@
     - '{{ nginx_default_servers | d([]) }}'
     - '{{ nginx_internal_servers | d([]) }}'
     - '{{ nginx_dependent_servers | d([]) }}'
-  when: (nginx_webroot_create|bool and item.state|d('present') != 'absent' and
-         (item.root is undefined and item.owner is undefined and item.group is undefined))
+  when: (nginx_webroot_create|bool and item.state|d('present') != 'absent')
```
drybjed: The change in the condition doesn't give per-server control over when the root directory should be created. How about adding `item.webroot_create` boolean here to allow that? Something like:
```
when: ((item.webroot_create|d(nginx_webroot_create))|bool and item.state|d('present') != 'absent')
```
That should do the trick. The new parameter needs to be documented as well.

drybjed (changes requested):
Hello, and welcome to DebOps! Awesome PR description, by the way. I added a few comments in the PR body with suggested changes.

Use of `.filter()` in the task list is an interesting approach, good to know that it works. Although I think that the templates could be simplified a bit, for example with:
```
dest: '{{ (item.root + "/normalize.css") if item.root|d() else (
          "{}/sites/{}/{}/normalize.css".format(
          nginx_www + (("/" + item.owner) if item.owner|d() else ''),
          (item.name if item.name is string else item.name[0]) | d("default"),
          item.public_dir_name | d(nginx_public_dir_name))
      )}}'
```
And with a bit more readable formatting, hopefully:
```
dest: '{{ (item.root + "/normalize.css")
          if item.root|d()
          else ("{}/sites/{}/{}/normalize.css".format(
                   nginx_www + (("/" + item.owner) if item.owner|d() else ''),
                   (item.name if item.name is string else item.name[0]) | d("default"),
                   item.public_dir_name | d(nginx_public_dir_name))
               ) }}'
```
What do you think? Don't forget about the spaces between `{{` and `}}` characters, you can see that in the Travis-CI suite.

The `debops.nginx` role is very old now and in a dire need of a rewrite to use all the new stuff I learned about Ansible over the years. But modernization like this is always welcome. :-)

tasosalvas:
Thanks for the quick reply!

- I fully agree with `item.webroot_create`, it actually crossed my mind as I was working on it.
- `item.mode` fits well with `.group` indeed.

I had kept the expressions split for readability, i.e. to make clear that those are the two cases the doc refers to, since it's a tangled mess of conditionals already. But the more I read your take the more I like it.

I'll get on it

tasosalvas:
Oops I have a typo. Fix coming.

ansible/roles/debops.nginx/tasks/nginx_servers.yml
```diff
@@ -1,14 +1,16 @@
 ---
-
```
drybjed, line 2: This line should stay as is, to be consistent with rest of DebOps YAML files.

drybjed:
Thanks for the changes. Unfortunately after testing the PR with `debops.dokuwiki` I noticed idempotency issues, currently re-running the test to see what exactly changes. I'll let you know.

tasosalvas:
Hm, weird. It's not running anything new. I thought vanilla ansible operations would meet our standard of idempotency. Might it have to do with `.root` no longer implicitly skipping dir creation? I guess now the role is overall a bit more eager to make dirs if the global setting allows.

drybjed:
No, the issue will be present with roles that create their own web directories, the `debops.dokuwiki` role does it. Then on the next run, `debops.nginx` changes the role owner/group/perms to the ones it thinks the user wants, and then `debops.dokuwki` changes them back.

That's where the mentioned `item.webroot_create` parameter comes in. With it set to `False` in the `debops.dokuwiki` configuration for the `debops.nginx` role, the idempotency issue is solved. Since we already know about it, perhaps you could add that parameter in `debops.dokuwki` configuration, [here](https://github.com/debops/debops/blob/master/ansible/roles/debops.dokuwiki/defaults/main.yml#L636). After that I'll merge this PR and make full test run on GitLab CI, this will show if any other roles have such issues present, these can be solved in a similar fashion.

Due to extensive changes, this PR will not be backported, so it won't affect stable releases.

tasosalvas:
Ok I see, this is pretty much what I was talking about in my guess about `.root`.

Is it worth considering to properly populate `debops.dokuwiki`'s `owner` and `group` and skip its own directory creation task? I guess that would be more work when the rest of the affected roles pop up, but the new options do give `debops.nginx` the capability to do so. Also I suppose any roles that create their own webroot also do more dir/file stuff, so it wouldn't benefit them that much semantically to be a task shorter.

Anyway, sounds like a grand tour of debops - anything web related will probably need at least the `.webroot_create` option set. I'll be happy to help.

Oh btw I've been looking into getting `make test` to work on my local machine, currently having some trouble with docker. I think it would be worth mentioning in the (otherwise excellent) `DEVELOPMENT.rst`, along with its dependencies.

I'll do the `.webroot_create`, but do tell me if you think it's cleaner chopping the dir creation off the roles instead.

drybjed:
> Is it worth considering to properly populate `debops.dokuwiki`'s `owner` and `group` and skip its own directory creation task? I guess that would be more work when the rest of the affected roles pop up, but the new options do give `debops.nginx` the capability to do so. Also I suppose any roles that create their own webroot also do more dir/file stuff, so it wouldn't benefit them that much semantically to be a task shorter.

Yes, that's another way to deal with it. Some roles that interact with unprivileged UNIX accounts of the applications (`debops.php`, `debops.mariadb`, `debops.postgresql`, `debops.keyring`, etc.) have tasks to create a barebones group, account and home directory which are enough to make them work; the rest of the configuration (shell, GECOS, etc.) is done in the application role. In the same spirit, the `debops.nginx` could create the directories with correct UNIX owner/group, but the relevant tasks that create the UNIX group and account would have to be added in the role. The aim is for the application roles like `debops.dokuwiki` to set up all relevant details for themselves, so that for example you could swap `debops.nginx` for a different `nginx` role and `debops.dokuwiki` role would still work fine.

> Anyway, sounds like a grand tour of debops - anything web related will probably need at least the `.webroot_create` option set. I'll be happy to help.

Sure, if you want you can add the `webroot_create` parameter in nginx configs of various application roles, that shouldn't break anything important. The `git` monorepo layout is perfect for this.

> Oh btw I've been looking into getting `make test` to work on my local machine, currently having some trouble with docker. I think it would be worth mentioning in the (otherwise excellent) `DEVELOPMENT.rst`, along with its dependencies.

Agreed, do you want to add that in the docs? Help with documentation is always appreciated. :-) Although in a separate PR, please.

> I'll do the `.webroot_create`, but do tell me if you think it's cleaner chopping the dir creation off the roles instead.

No, the application roles configure all the other aspects of the UNIX accounts (shell, GECOS, additional group memberships) and this functionality needs to be included there for completeness sake.

tasosalvas:
Ok, makes sense. I'll be happy to add the setting to the roles that need it if you get me the list from Gitlab CI. 

I'll do a doc PR later yeah (once I got it working here ;) ), I've been reading it for the past month anyway so I guess I should get comfy.

tasosalvas:
## The grand `item.webroot_create` tour
So, I'm `ack`ing the roles folder for `"nginx"` and setting `webroot_create: False` to any roles with `root`, `group`, or `owner` set, since those would have been implicitly skipping webroot creation while using the old `debops.nginx`. (No instances of `.group` or `.owner` were actually encountered.)

At a quick glance, I suspect there are other roles (i.e. those using nginx as a proxy) that would benefit from `webroot_create: False` as well, but that's beyond the scope of these changes -- I'm just guessing these roles (such as `debops.apt_cacher_ng`) allowed the webroot to be created and then never bothered with it again. Otherwise, wouldn't they have had the idempotency issues we saw with `debops.dokuwiki`?

I think it might make sense for `debops.nginx` to automatically skip webroot creation for proxy configurations, but I'm probably unqualified to make that call.

I have also noted a couple of roles that implemented their servers in a non-standard way. Personally my only issue was that I couldn't just `$ ack "__nginx_dependent_servers"` -- Not my call how important it is for those to be uniform within the codebase, and I didn't review the roles themselves.


### Roles containing `"nginx"` left unchanged
(excluding any that only refereced nginx in comments, such as `debops.apache`)
* `debops-contrib.homeassistant`: Proxy
* `debops.apt_cacher_ng`: Proxy
* `debops.docker_gen`: Doesn't seem to define a server.
* `debops.docker_registry`: Auth, custom proxy.
* `debops.etherpad`: No root defined, bespoke server with custom user. (bug?)
* `debops.gitusers`: Reads local.nginx.www and local.nginx.user but defines no server.
* `debops.hashicorp`: Reads local.nginx.www but defines no server.
* `debops.kibana`: Proxy, auth
* `debops.minio`: Proxies defined in `templates/lookup/minio__env_nginx_dependent_servers.j2`
* `debops.monit`: No idea. Just jacks onto an existing nginx process?
* `debops.netbox`: No root defined, has custom user.
* `debops.pki`: Reads `ansible.local.acme` and `.acme_root`, defines no server.
* `debops.rabbitmq_management`: Proxy
* `debops.rstudio_server`: No root defined, has custom user.


### Edited Contrib roles
Keeping those in their own commit, in case I was not supposed to mess with them.
* `debops-contrib.foodsoft`
* `debops-contrib.volkszaehler`


### Edited Debops roles
* `debops.debops_api`
* `debops.dokuwiki` Including it here for completeness' sake.
* `debops.gitlab` Also defines a `gitlab__nginx_pages_server` proxy I left unchanged.
* `debops.icinga_web`
* `debops.librenms`
* `debops.mailman`
* `debops.mosquitto`
* `debops.owncloud`
* `debops.phpipam`
* `debops.phpmyadmin` Sets `phpmyadmin_nginx_server` in `vars/main.yml` through `meta/main.yml`.
* `debops.preseed`: Sets `preseed__nginx__servers`.
* `debops.rails_deploy`: Sets `rails_deploy_nginx_server` through `meta/main.yml`.
* `debops.reprepro`: Sets `reprepro_nginx_server_http` and `reprepro_nginx_server_https` through `meta/main.yml`.
* `debops.roundcube`
* `debops.sks`: Sets `sks_nginx_frontend` and `sks_nginx_ssl_frontend` in `vars/main.yml` through `meta/main.yml`. The first has `root` set, so it gets the correction. The second I can't imagine working with non-default variables.

I thought `vars/main.yml` was supposed to be the end user's business and roles should do theirs in `defaults/main.yml` am I wrong?

Among the edited roles `debops.gitlab` and `debops.sks` probably need you to have a look.

---

Heh, that wasn't that bad. Seriously beats just reading the doc while learning the tool. I hope my notes are of some use.

drybjed:
@tasosalvas Thanks for investigating and fixing the idempotency issues! The different takes on `nginx` configuration are mostly because different roles were developed over the years and then left unchanged, for example some still with configuration in `vars/main.yml` which cannot be overridden via inventory. I want to clean up as much of these old stuff as possible before the `v2.0.0` release in January 2020, so this roundup will be very handy.

Skipping webroot creation if the server is configured as a proxy might be a good idea. I'm not sure how it affects the security of `nginx` and the application - if `root` directory specified in the configuration file doesn't exist, does `nginx` look for files elsewhere or returns 404? But that can be implemented in a separate PR.

PR #1124, opened 2019-10-28
[debops.grafana] Add debops.grafana role

LeonardoAndrei:
This new role is a dashboard and graph composer, which runs as a web application and can be used as a part of a monitoring system.


ansible/playbooks/service/grafana.yml
```diff
@@ -0,0 +1,38 @@
...
+      when: grafana__nginx_enabled|bool
+
+    - role: debops.grafana
+      tags: [ 'role::grafana', 'skip::grafana' ]
+
+    - role: debops.ferm
+      tags: [ 'role::ferm', 'skip::ferm' ]
+      ferm__dependent_rules:
+        - '{{ grafana__ferm__dependent_rules }}'
+        - '{{ nginx__ferm__dependent_rules if (grafana__nginx_enabled|bool) else [] }}'
```
drybjed: The `debops.ferm` dependency should be put before the `debops.nginx` role, that way the firewall will be opened for the service ports right away.

ansible/roles/debops.grafana/defaults/main/main.yml
```diff
@@ -0,0 +1,193 @@
...
+# The DNS domain used by other variables in the ``debops.grafana`` role.
+grafana__domain: '{{ ansible_local.core.domain
+                    if (ansible_local|d() and ansible_local.core|d() and
+                        ansible_local.core.domain|d())
+                    else ansible_domain }}'
+
+                                                                   # ]]]
+                                                                   # ]]]
+# Grafana configuration [[[
+# --------------------------------
```
drybjed: You should make sure that the header underlines are flush with the `[[[` characters. If the header lines are too short, the reStructuredText parser will error out.

ansible/roles/debops.grafana/defaults/main.yml
```diff
@@ -0,0 +1,193 @@
...
+# .. envvar:: grafana__base_packages [[[
+#
+# List of APT packages which are required by the Grafana role.
+grafana__base_packages: [ 'grafana' ]
+
+                                                                   # ]]]
+# .. envvar:: grafana__packages [[[
+#
+# List of APT packages to install on all hosts in Grafana inventory.
+grafana__packages: [ 'software-properties-common', 'apt-transport-https' ]
```
drybjed: The `grafana__packages` list should be empty by default, the packages defined here would have to be moved to the `grafana__base_packages` list. This one exists for the users to add additional ones via Ansible inventory, which would overridde the default value.

The `apt-transport-https` APT package is included in APT by default since Debian Buster, and you shouldn't have to install it separately. IMO it can be dropped from here.

The `software-properties-common` APT package is required mostly for PPA support on Ubuntu, which should have it installed by default. On Debian it's not needed, I would remove this one as well, leaving an empty list.

ansible/roles/debops.grafana/defaults/main.yml
```diff
@@ -0,0 +1,193 @@
...
+# - ``system``: install Grafana using current OS packages, possibly from the
+#               backports repository.
+#
+# - ``upstream``: install Grafana from the upstream APT repository on
+#                 Launchpad (PPA, usable on Debian as well).
+#
+# - ``bootstrap``: install APT packages required for Grafana and build a local
+#                  ``.deb`` package using the upstream GitHub repository.
+#
+grafana__deploy_type: "upstream"
```
drybjed: Looks like you based this variable on `debops.ansible` role. :-)

Currently Grafana is not present in any Debian repositories. Ubuntu has it in the `xenial` archive, but not anywhere else. It seems that the OS distributions don't plan to package Grafana anytime soon, therefore I would remove the `system` option from here.

The `bootstrap` method is used in `debops.ansible` to allow installation of a newer version for testing and development purposes - I would mark it as an exception to the rule since DebOps is based on Ansible. In the Grafana case though, I wouldn't want to build it from source, since upstream provides APT repository which makes upgrades of new releases painless. So, I would drop the `bootstrap` installation type as well.

This leaves the `upstream` as the only type and makes the variable redundant. It can be safely removed to make conditions in tasks easier.

ansible/roles/debops.grafana/defaults/main.yml
```diff
@@ -0,0 +1,193 @@
...
+
+                                                                   # ]]]
+# .. envvar:: grafana__allow [[[
+#
+# List of IP addresses or CIDR subnets which will be allowed to connect to the
+# Grafana in :command:`ip(6)tables` and TCP wrappers. If it's empty, remote
+# connections are not allowed.
+grafana__allow: [ '0.0.0.0/0' ]
```
drybjed: Currently the value here is contradictory - it will allow access over IPv4 to the service from anywhere, but not over IPv6. And if the user will want to open the access over IPv6, he/she will have to copy over the `0.0.0.0/0` value as well because if not, it will not be present in overriden variable in the inventory and that will flip the IPv4/IPv6 access.

You should decide if access to Grafana should be public or private by default. Since this is a web service with authentication, I think that having it accessible by default through the firewall is OK. In that case, I would make the `grafana__allow` list empty, and change the `accept_any` parameter in `debops.ferm` configuration to `True` - this will allow access from anywhere by default and if users want to limit access to their network, they still can.

ansible/roles/debops.grafana/defaults/main.yml
```diff
@@ -0,0 +1,193 @@
...
+# .. envvar:: grafana__nginx__dependent_servers [[[
+#
+# Server configuration for the :ref:`debops.nginx` Ansible role.
+grafana__nginx__dependent_servers:
+  - name: ['{{ grafana__fqdn }}']
+    by_role: 'debops.grafana'
+    filename: 'debops.grafana'
+    location:
+      '/': |
+        proxy_pass http://grafana;
```
drybjed: Try using the `proxy` server type instead, with `proxy_pass` parameter that points to the `grafana` upstream.

ansible/roles/debops.grafana/defaults/main.yml
```diff
@@ -0,0 +1,193 @@
...
+  - name: 'grafana'
+    server: [ 'localhost' ]
+    port: '{{ grafana__port }}'
+
+                                                                   # ]]]
+# .. envvar:: grafana__nginx__dependent_servers [[[
+#
+# Server configuration for the :ref:`debops.nginx` Ansible role.
+grafana__nginx__dependent_servers:
+  - name: ['{{ grafana__fqdn }}']
```
drybjed: The `name` parameter should handle a string just fine, you don't have to define it as a list.

ansible/roles/debops.grafana/handlers/main.yml
```diff
@@ -0,0 +1,12 @@
+---
+
+- name: Restart grafana
+  service:
+    name: 'grafana-server'
+    state: 'restarted'
+    enabled: yes
```
drybjed: Setting `enabled: True` here might have issues on installations, where `grafana.service` was intentionally disabled by system administrator. The service will probably be enabled by default on installation, so you don't need this parameter here.

ansible/roles/debops.grafana/tasks/main.yml
```diff
@@ -0,0 +1,41 @@
...
+  args:
+    creates: '/etc/grafana/grafana.ini.dpkg-divert'
+
+- name: Configure main grafana config file
+  template:
+    src: 'grafana.ini.j2'
+    dest: '/etc/grafana/grafana.ini'
+    owner: 'root'
+    group: 'adm'
+    mode: '0644'
```
drybjed: Kind of weird group selection here... Is the configuration file really owned by `root:adm`? Also, are there any passwords in that configuration file? In this case you should use stricter mode.

drybjed (changes requested):
Nice work. :-) There are a few changes I would like to see though, see the comments below. I hope that the default configuration will be a bit more sane.

Do you plan to add [LDAP support](https://grafana.com/docs/auth/ldap/) and integrate it with the `debops.ldap` environment?

alinalexandru:
@drybjed We want to have ldap integration. For now maybe we can just add support to configure that file. (same way as grafana.ini). 
As we don't have experience with debops.ldap role, we decided to not include the support in first version

ansible/playbooks/service/grafana.yml
LeonardoAndrei: moved

ansible/roles/debops.grafana/defaults/main/main.yml
LeonardoAndrei: solved

ansible/roles/debops.grafana/defaults/main.yml
LeonardoAndrei: deleted

ansible/roles/debops.grafana/defaults/main.yml
LeonardoAndrei: also deleted

ansible/roles/debops.grafana/defaults/main.yml
LeonardoAndrei: modified in both places

ansible/roles/debops.grafana/defaults/main.yml
LeonardoAndrei: used 
type: proxy
proxy_pass: 'http://grafana/'

ansible/roles/debops.grafana/defaults/main.yml
LeonardoAndrei: changed

ansible/roles/debops.grafana/handlers/main.yml
LeonardoAndrei: deleted enabled entry

ansible/roles/debops.grafana/tasks/main.yml
LeonardoAndrei: changed group to grafana and permissions to 0640

ansible/roles/grafana/defaults/main.yml
```diff
@@ -0,0 +1,182 @@
...
+# .. include:: ../../../../includes/global.rst
+
+
+# Domain name configuration [[[
+# -----------------------------
+
+# .. envvar:: grafana__hostname [[[
+#
+# The hostname part of the Grafana installation
+grafana__hostname: 'grafana'
```
drybjed, line 19: Why not use `{{ ansible_hostname }}` instead? Actually, this variable is not used anywhere besides `grafana__fqdn`, I think that it can be safely removed.

ansible/roles/grafana/defaults/main.yml
```diff
@@ -0,0 +1,182 @@
...
+#
+# The Grafana fingerprint.
+grafana__upstream_apt_id: '4E40 DDF6 D76E 284A 4A67  80E4 8C8C 34C5 2409 8CB6'
+
+                                                                   # ]]]
+# .. envvar:: grafana__upstream_apt_repository [[[
+#
+# The APT repository URI of the upstream Grafana repository.
+grafana__upstream_apt_repository: 'deb https://packages.grafana.com/oss/deb stable main'
+
```
drybjed, line 59: Can you remove the empty line here? That way the variables are evenly spaced from each other. This is commonly done at the end of the sections.

ansible/roles/grafana/meta/main.yml
```diff
@@ -0,0 +1,29 @@
+---
+
+dependencies:
+
+  - role: ansible_plugins
```
drybjed, line 5: You should move this to the task list, use the `import_role` module.

ansible/roles/grafana/templates/etc/ansible/facts.d/grafana.fact.j2
```diff
@@ -0,0 +1,23 @@
...
+def cmd_exists(cmd):
+    return any(
+        os.access(os.path.join(path, cmd), os.X_OK)
+        for path in os.environ["PATH"].split(os.pathsep)
+    )
+
+
+output = {}
+
+output['installed'] = cmd_exists('grafana-server')
```
drybjed, line 21: You can combine the above lines as:
```
output = {'installed': cmd_exists('grafana-server')}
```

ansible/roles/grafana/COPYRIGHT
```diff
@@ -0,0 +1,18 @@
+debops.grafana - Install Grafana on a Debian/Ubuntu host using Ansible
+
+Copyright (C) 2019 Innobyte Bechea Leonardo https://www.innobyte.com/
+Copyright (C) 2019 Innobyte Alin Alexandru https://www.innobyte.com/
```
drybjed, line 4: Can you change the copyright attributions to something [like this](https://github.com/debops/debops/blob/master/ansible/roles/tinyproxy/COPYRIGHT#L3-L5) instead? That way the script that builds the manual pages can correctly extract the authorship information from the copyright files.

ansible/playbooks/service/grafana.yml
```diff
@@ -0,0 +1,38 @@
...
+
+    - role: etc_services
+      tags: [ 'role::etc_services', 'skip::etc_services' ]
+      etc_services__dependent_list:
+        - '{{ grafana__etc_services__dependent_list }}'
+
+    - role: keyring
+      tags: [ 'role::keyring', 'skip::keyring', 'role::grafana' ]
+      keyring__dependent_apt_keys:
+        - '{{ grafana__keyring__dependent_apt_keys }}'
```
drybjed: You should reorder the `keyring` role to be the first one in the playbook, that way on any issues with the keyservers Ansible can stop early.

ansible/playbooks/service/grafana.yml
```diff
@@ -0,0 +1,38 @@
...
+
+    - role: ferm
+      tags: [ 'role::ferm', 'skip::ferm' ]
+      ferm__dependent_rules:
+        - '{{ grafana__ferm__dependent_rules }}'
+        - '{{ nginx__ferm__dependent_rules if (grafana__nginx_enabled|bool) else [] }}'
+
+    - role: nginx
```
drybjed, line 29: The `nginx` role has the `python` role as a dependency (I think it changed some time after you made the PR). Can you update the playbook and add it as well?

ansible/playbooks/service/grafana.yml
```diff
@@ -0,0 +1,38 @@
+---
+
+- name: Manage Grafana
+  hosts: [ 'debops_service_grafana', 'debops_grafana' ]
```
drybjed, line 4: Don't use the `debops_grafana` group scheme, they were removed before the `v2.0.0` release.

docs/ansible/roles/grafana/index.rst
```diff
@@ -0,0 +1,24 @@
...
+==============
+
+The ``debops.grafana`` role can be used to install Grafana on a Debian or
+Ubuntu host. Grafana will be installed from upstream APT repository.
+
+.. toctree::
+   :maxdepth: 2
+
+   getting-started
+   defaults/main
```
drybjed, line 13: There's no `defaults-detailed` document that explains the `grafana__*_configuration` variables. Also, the documentation was rearranged a bit to add support for manual pages, perhaps you could update it here as well?

drybjed (changes requested):
Thanks for the role, just reviewed it. Sorry that it took so long that a bunch of changes in the project were introduced. Perhaps you could refresh the PR when you have some time, then it can be tested and merged.

ansible/playbooks/service/grafana.yml
LeonardoAndrei: removed

ansible/playbooks/service/grafana.yml
LeonardoAndrei: added python dep

ansible/playbooks/service/grafana.yml
LeonardoAndrei: moved on top

ansible/roles/grafana/COPYRIGHT
LeonardoAndrei: changed, as you requested

ansible/roles/grafana/templates/etc/ansible/facts.d/grafana.fact.j2
LeonardoAndrei: modified the 2 lines

ansible/roles/grafana/meta/main.yml
LeonardoAndrei: moved in tasks as import_role

ansible/roles/grafana/defaults/main.yml
LeonardoAndrei: removed

ansible/roles/grafana/defaults/main.yml
LeonardoAndrei: done. removed the variable

alinalexandru:
@pedroluislopez  added LDAP support

reixd:
Any news on this role?

alinalexandru:
Probably @drybjed have to review it. 
@realloc have to tested the PR? Do you have any feedback?

TuxCoder:
@LeonardoAndrei Thank nice role, grafana was also on my TODO list...

could merge successfully to my working branch and deploy to my test system.

also auto installation of the upstream works like a charm :)

Ldap config createation works also, but didn't test, because have a different ldap setup ...
Currently try to switch to Oauth2/OIDC that works with some config tweaks.

please add root_url wenh nginx is enabled:
otherwise the server url in emails, oauth, .. has wrong port/proto
```yaml
grafana__configuration:
  - name: 'server'
    options:
      - name: root_url 
        value: 'https://{{ grafana__fqdn }}'
        state: '{{ "present" if grafana__nginx_enabled else "absent" }}'
```


Probably more a future request, make upstream repository optional over a flag,
What I know currently you want to use the upstream version, but it is not stable as in "debian stable".
So if grafana has a usable version in the next debian relelease the upstream flag can be disable to have a "debian stable" version to ensure no api changes.

ypid:
LGTM, thanks. I made a few improvements in https://github.com/innobyte/debops/pull/9. @TuxCoder I included the `root_url` by default now.

alinalexandru:
Approved @ypid improvements.
I will have to do a rebase

ypid:
Thanks @alinalexandru 

> I will have to do a rebase

This branch merges beautifully into master without merge conflicts. No real need for a rebase. It would destroy git commit signatures made with OpenPGP.

But it would not be the end of the world if you had to. I can just implicitly sign it again when I next commit something which includes your work.

docs/ansible/roles/grafana/getting-started.rst
```diff
@@ -0,0 +1,47 @@
...
+.. Copyright (C) 2019 DebOps <https://debops.org/>
+.. SPDX-License-Identifier: GPL-3.0-or-later
+
+Getting started
+===============
+
+.. contents::
+   :local:
+
+
```
tasosalvas, line 7: A __Default Configuration__ section should give a brief overview of the state the application will be in if deployed with defaults, especially:

* choices that were made that might differ from its out-of-the-box defaults
* things the end user must pay attention to before deploying to production

julien-lecomte:
I've tested this role and it works great, some notes:
- port is 3001 event if nginx is not installed. It rather should be 3000
- analytics.reporting_enabled should rather be turned off by default.

PR #1125, opened 2019-10-29
[debops.postldap] New role for Virtual Mail management with LDAP backend

reixd:
@drybjed moved the LDAP and vmail changes to a new role. Please take a look :)

There are some improvements needed so do not merge yet.


reixd:
This is the initial role development. It works with the "old" Postfix mail scheme.
I will take a look to other schemes and make the transition accordingly.

ansible/roles/debops.postfix_virtual_ldap/meta/main.yml
```diff
@@ -0,0 +1,35 @@
+---
+
+dependencies:
+  - role: debops.ansible_plugins
+  - role: debops.secret
+  - role: debops.ldap
+  - role: debops.postfix
```
drybjed: You shouldn't include the `debops.ldap` and `debops.postfix` roles as hard dependencies. They are already present in the role playbook; adding them here only makes them run twice.

docs/ansible/roles/debops.postfix_virtual_ldap/copyright.rst
```diff
@@ -0,0 +1,4 @@
+Copyright
+=========
+
+.. literalinclude:: ../../../../ansible/roles/debops.postfix_virtual_ldap/COPYRIGHT
```
drybjed: This section should be included in the `index.rst` file, see newer roles like `debops.apt_mark` as an example.

docs/ansible/roles/debops.postfix_virtual_ldap/introduction.rst
```diff
@@ -0,0 +1,23 @@
+Introduction
+============
+
+.. include:: ../../../includes/global.rst
+
+The ``debops.postfix_virtual_ldap`` Ansible role can be used to configure and
+manage a `Postfix`_ Virtual Mail server with an LDAP backend.
+It allows configuration of Postfix using Ansible inventory variables.
```
drybjed: This section should be moved to the `index.rst` file.

docs/ansible/roles/debops.postfix_virtual_ldap/introduction.rst
```diff
@@ -0,0 +1,23 @@
...
+
+
+Installation
+~~~~~~~~~~~~
+
+This role requires at least Ansible ``v2.7.0``. To install it, run:
+
+.. code-block:: console
+
+   ansible-galaxy install debops.postfix_virtual_ldap
```
drybjed: The installation section can be removed, the example installation command will not work at all with DebOps being an Ansible Galaxy Collection.

ansible/roles/debops.postfix_virtual_ldap/defaults/main.yml
```diff
@@ -0,0 +1,685 @@
...
+# debops.postfix_virtual_ldap default variables [[[
+# =================================================
+
+# .. contents:: Sections
+#    :local:
+#
+# .. include:: ../../../../includes/global.rst
+
+# APT packages [[[
+# -------------------------
```
drybjed: The section underlines should be flush with the `[[[` characters, it looks more readable this way.

ansible/playbooks/service/postfix_virtual_ldap.yml
```diff
@@ -0,0 +1,49 @@
...
+        - '{{ etc_aliases__secret__directories }}'
+        - '{{ postfix__secret__directories }}'
+
+    - role: debops.ferm
+      tags: [ 'role::ferm', 'skip::ferm' ]
+      ferm__dependent_rules:
+        - '{{ postfix__ferm__dependent_rules }}'
+
+    - role: debops.etc_aliases
+      tags: [ 'role::etc_aliases', 'skip::etc_aliases' ]
```
drybjed: The `debops.postfix_virtual_ldap` role requires the `debops.postfix` role to work. Since this playbook will be included later in the `site.yml` playbook than the `postfix.yml` playbook, we can assume that the base Postfix service, including `/etc/aliases` and the firewall are already configured.

Withh this assumption, you don't need to include the `debops.ferm` and `debops.etc_aliases` roles in the `debops.postfix_virtual_ldap` role playbook, because these services will not be affected by it. This makes the playbook execution shorter and more focused.

ansible/roles/debops.postfix_virtual_ldap/defaults/main.yml
```diff
@@ -0,0 +1,685 @@
...
+
+# .. envvar:: postfix_virtual_ldap__packages [[[
+#
+# List of additional APT packages needed for the LDAP support.
+postfix_virtual_ldap__packages: '{{ [ "postfix-ldap" ] if postfix_virtual_ldap__ldap_enabled else [] }}'
```
drybjed: The `debops.postfix_virtual_ldap` role doesn't do much apart from setting up the virtual UNIX account. To indicate that the packages are installed by the `debops.postfix` role, you should rename this variable to `postfix_virtual_ldap__postfix__dependent_packages` and put it at the end of the `defaults/main.yml` file.

ansible/roles/debops.postfix_virtual_ldap/defaults/main.yml
```diff
@@ -0,0 +1,685 @@
...
+# ---------------------------
+
+# .. envvar:: postfix_virtual_ldap__pki [[[
+#
+# Enable or disable support for TLS in Postfix, managed by the :ref:`debops.pki`
+# Ansible role.
+postfix_virtual_ldap__pki: '{{ (True
+                               if (ansible_local|d() and ansible_local.pki|d() and
+                                   ansible_local.pki.enabled|d() | bool)
+                               else False) | bool }}'
```
drybjed: The PKI support is configured in the `debops.postfix` role, therefore it's not needed here. The default LDAP connection will most likely not require a specific CA certificate, and when users want to use that functionality, they can specify the CA certificate file directly via the Ansible inventory.

ansible/roles/debops.postfix_virtual_ldap/defaults/main.yml
```diff
@@ -0,0 +1,685 @@
...
+# or make the directory world writable.
+postfix_virtual_ldap__mailbox_base: '/var/vmail'
+                                                                   # ]]]
+# .. envvar:: postfix_virtual_ldap__system_users [[[
+#
+# Postfix Virtual Users
+postfix_virtual_ldap__system_users:
+  - name: '{{ postfix_virtual_ldap__vmail_posix_user }}'
+    group: '{{ postfix_virtual_ldap__vmail_posix_group }}'
+    home: '{{ postfix_virtual_ldap__mailbox_base }}'
```
drybjed: I suppose that having a way to create multiple UNIX accounts could be handy to support separate `virtual` "realms" protected by each being in its own UNIX account... But the rest of the role does not support that configuration. Perhaps changing this to just one UNIX account with `postfix_virtual_ldap__user`, `postfix_virtual_ldap__group`, `postfix_vritual_ldap__home` etc. variables could simplify things a bit?

ansible/roles/debops.postfix_virtual_ldap/defaults/main.yml
```diff
@@ -0,0 +1,685 @@
...
+# to other local or remote address.
+# The default option, when using Virtual Mail with LDAP, will first go over the
+# alisaes set by :ref:`debops.etc_aliases` and if no match is found query
+# the LDAP for the email address.
+#
+# Specify zero or more `type:name` lookup tables, separated by whitespace or comma.
+# Tables will be searched in the specified order until a match is found.
+# Note: these lookups are recursive.
+postfix_virtual_ldap__virtual_alias_maps:
+  - '$alias_maps'
```
drybjed: The `$alias_maps` Postfix option defines alias maps for local delivery. I think that including them in `virtual_alias_maps` will break things. You should remove them from here.

ansible/roles/debops.postfix_virtual_ldap/defaults/main.yml
```diff
@@ -0,0 +1,685 @@
...
+# Postfix 'main.cf' configuration [[[
+# -----------------------------------
+
+# These variables define the contents of the :file:`/etc/postfix/main.cf`
+# configuration file. See :ref:`postfix__ref_maincf` for more details.
+
+# .. envvar:: postfix_virtual_ldap__virtual_maincf [[[
+#
+# List of options needed by `postfix` for the Virtual Mail configuration
+postfix_virtual_ldap__virtual_maincf:
```
drybjed: You should rename this variable to `postfix_virtual_ldap__postfix__dependent_maincf` since it will be passed via the role dependent variables to the `debops.postfix` role.

ansible/roles/debops.postfix_virtual_ldap/defaults/main.yml
```diff
@@ -0,0 +1,685 @@
...
+  # Postfix is final destination for the specified list of virtual alias domains,
+  # that is, domains for which all addresses are aliased to addresses in other local or remote domains
+  - name: 'virtual_alias_domains'
+    value: 'ldap:/etc/postfix/ldap_virtual_alias_domains.cf'
+    section: 'virtual'
+    state: '{{ "absent"
+            if (postfix_virtual_ldap__ldap_enabled|bool and
+                postfix__version is version_compare("2.0", ">="))
+            else "absent" }}'
```
drybjed: With this condition, this parameter will never be included in the `/etc/postfix/main.cf`. Is this what you want?

ansible/roles/debops.postfix_virtual_ldap/defaults/main.yml
```diff
@@ -0,0 +1,685 @@
...
+    state: '{{ "present"
+            if (postfix_virtual_ldap__ldap_enabled|bool and
+                postfix__version is version_compare("2.0", ">="))
+            else "absent" }}'
+
+  # The minimum user ID value that the virtual delivery agent accepts as a result from $virtual_uid_maps table lookup
+  - name: 'virtual_minimum_uid'
+    value: '{{ postfix_virtual_ldap__vmail_posix_uidnumber }}'
+    section: 'virtual'
+    state: 'absent'
```
drybjed: This configuration will remove the `virtual_minimum_uid` option from `/etc/postfix/main.cf` but it probably isn't there in the first place. I think that this can be safely removed.

ansible/roles/debops.postfix_virtual_ldap/defaults/main.yml
```diff
@@ -0,0 +1,685 @@
...
+    value: 'ldap:/etc/postfix/ldap_virtual_senders.cf'
+    section: 'base'
+    state: '{{ "present"
+            if (postfix_virtual_ldap__ldap_enabled|bool)
+            else "absent" }}'
+                                                                   # ]]]
+# .. envvar:: postfix_virtual_ldap__postfix__lookup_header [[[
+#
+# The header of the lookup table containing the settings to connect to the LDAP server
+postfix_virtual_ldap__postfix__lookup_header:
```
drybjed: This variable is not directly passed to the `debops.postfix` role via role dependent variables, therefore it shouldn't use two `__` separators in the name. How about, `postfix_virtual_ldap__lookuo_connection_header`?

Also, that's an interesting solution for having standardized header for all `ldap_*` lookup tables. Since `debops.postfix` role does not have a dedicated solution yet, it can stay for now. :-)

ansible/roles/debops.postfix_virtual_ldap/defaults/main.yml
```diff
@@ -0,0 +1,685 @@
...
+# is set to e.g. @foobar.com
+postfix_virtual_ldap__postfix__lookup_virtual_senders:
+  search_base: '{{ postfix_virtual_ldap__ldap_people_dn|join(",") }}'
+  query_filter: '(&(mailAlias=%s){{ postfix_virtual_ldap__ldap_user_filter }})'
+  result_attribute: 'uid, mail'
+                                                                   # ]]]
+# .. envvar:: postfix_virtual_ldap__postfix__lookup_tables [[[
+#
+# The Virtual Mail lookup tables for postfix
+postfix_virtual_ldap__postfix__lookup_tables:
```
drybjed: This variable should be renamed to `postfix_virtual_ldap__postfix__dependent_lookup_tables`.

ansible/roles/debops.postfix_virtual_ldap/defaults/main.yml
```diff
@@ -0,0 +1,685 @@
...
+# Secrets [[[
+# -------------------------
+
+# .. envvar:: postfix_virtual_ldap__no_log [[[
+#
+# This variable can be used in the Ansible tasks to set the ``no_log``
+# parameter. Changing it to ``False`` can help with debugging issues with the
+# roles, but otherwise it should be set to ``True`` to ensure that Ansible
+# doesn't leak sensitive data in its logs.
+postfix_virtual_ldap__no_log: True
```
drybjed: Although this variable might be useful for debugging, it might clutter the configuration of the `debpos.postfix` and `debops.ldap` roles and make it less readable. I would consider removing it and specifying the `item.no_log: True` instead where appropriate. Changing the logging option for a specific lookup table will probably be more frequent than enabling/disabling logging for everything at once.

ansible/roles/debops.postfix_virtual_ldap/defaults/main.yml
```diff
@@ -0,0 +1,685 @@
...
+# .. envvar:: postfix_virtual_ldap__postfix__lookup_tables [[[
+#
+# The Virtual Mail lookup tables for postfix
+postfix_virtual_ldap__postfix__lookup_tables:
+  - name: 'ldap_virtual_aliases.cf'
+    state: 'present'
+    comment: |
+      The virtual_alias_maps setting is used to find the target mailbox
+      (configured in the attribute mail, aka the mailbox or maildrop).
+    group: 'postfix'
```
drybjed: Should the lookup tables have the `postfix` group? I think that Postfix opens the lookup tables as `root` before dropping the privileges, so the access from the `postfix` UNIX account might be less secure.

ansible/roles/debops.postfix_virtual_ldap/defaults/main.yml
```diff
@@ -0,0 +1,685 @@
...
+    mode: '0640'
+    no_log: '{{ postfix_virtual_ldap__no_log|d(True) }}'
+    config: '{{ postfix_virtual_ldap__postfix__lookup_header |
+                 combine(postfix_virtual_ldap__postfix__lookup_virtual_aliases) }}'
+  - name: 'ldap_virtual_recipients.cf'
```
drybjed: Each entry should be separated by a blank line, this makes the list of YAML dictionaries a bit more readable.

ansible/roles/debops.postfix_virtual_ldap/defaults/main.yml
```diff
@@ -0,0 +1,685 @@
...
+    no_log: '{{ postfix_virtual_ldap__no_log|d(True) }}'
+    config: '{{ postfix_virtual_ldap__postfix__lookup_header |
+                 combine(postfix_virtual_ldap__postfix__lookup_virtual_senders) }}'
+                                                                   # ]]]
+                                                                   # ]]]
+# LDAP [[[
+# ----------------------
+
+# LDAP authentication [[[
+# ^^^^^^^^^^^^^^^^^^^^^^
```
drybjed: I think that these sections should be on the same level as the `LDAP` section, you don't need to put them on yet another level. Just have two or more sections related to LDAP configuration.

ansible/roles/debops.postfix_virtual_ldap/defaults/main.yml
```diff
@@ -0,0 +1,685 @@
...
+# The Distinguished Name of the LDAP object which contains the user accounts
+# used by Postfix.
+postfix_virtual_ldap__ldap_people_dn: '{{ [ postfix_virtual_ldap__ldap_people_rdn ]
+                                          + postfix_virtual_ldap__ldap_base_dn }}'
+
+                                                                   # ]]]
+# .. envvar:: postfix_virtual_ldap__ldap_uri [[[
+#
+# List of LDAP URIs that point to the directory servers which should be used by
+# nextcloudFIXME.
```
drybjed: Hello, Nextcloud. :-)

ansible/roles/debops.postfix_virtual_ldap/defaults/main.yml
```diff
@@ -0,0 +1,685 @@
...
+                                    else [""]) | first }}'
+
+                                                                   # ]]]
+# .. envvar:: postfix_virtual_ldap__ldap_private_groups [[[
+#
+# When this variable is enabled, the :ref:`debops.ldap` role will create
+# a separate LDAP objects that manage the Postfix groups as subtree of the
+# DokiWiki service LDAP object. If you set this parameter to ``False``, the
+# role will use the global ``ou=Groups,dc=example,dc=org`` subtree instead.
+postfix_virtual_ldap__ldap_private_groups: True
```
drybjed: Hello, DokuWiki. :-)

I'm not sure if managing multiple separate mail domains using single LDAP infrastructure is that common, especially that you probably want to have just a single set of `ou=People` objects, which would imply that users share accounts on multiple domains, so having just a single set of mail domains for the entire LDAP tree would suffice. This would also make the configuration a bit simpler... Thoughts?

ansible/roles/debops.postfix_virtual_ldap/defaults/main.yml
```diff
@@ -0,0 +1,685 @@
...
+#
+# The LDAP configuration options defined by default.
+postfix_virtual_ldap__ldap_default_config: []
+
+                                                                   # ]]]
+# .. envvar:: postfix_virtual_ldap__ldap_config [[[
+#
+# List of custom LDAP configuration options defined for all hosts
+# in the Ansible inventory.
+postfix_virtual_ldap__ldap_config: []
```
drybjed: You most likely don't need the `postfix_virtual_ldap__*_ldap_config` variables.

ansible/roles/debops.postfix_virtual_ldap/defaults/main.yml
```diff
@@ -0,0 +1,685 @@
...
+# the role tasks and templates.
+postfix_virtual_ldap__ldap_combined_config: '{{ postfix_virtual_ldap__ldap_default_config
+                                              + postfix_virtual_ldap__ldap_config
+                                              + postfix_virtual_ldap__group_ldap_config
+                                              + postfix_virtual_ldap__host_ldap_config }}'
+                                                                   # ]]]
+# .. envvar:: postfix_virtual_ldap__ldap_dependent_tasks [[[
+#
+# Configuration for the :ref:`debops.ldap` Ansible role.
+postfix_virtual_ldap__ldap_dependent_tasks:
```
drybjed: This variable should be named `postfix_virtual_ldap__ldap__dependent_tasks`.

docs/ansible/roles/debops.postfix_virtual_ldap/getting-started.rst
```diff
@@ -0,0 +1,61 @@
...
+.. code-block:: ini
+
+   [debops_service_ldap]
+   hostname_of_ldap_server
+
+   [debops_service_slapd]
+   hostname_of_ldap_server
+
+   [debops_service_postfix_virtual_ldap]
+   hostname_of_mail_server
```
drybjed: The Postfix host should be included in the `[debops_service_postfix]` Ansible inventory group to apply the base `debops.postfix` configuration including `/etc/aliases` and firewall.

drybjed (changes requested):
Very nice work, thanks. Although, when I look at it, using `postfix_virtual_ldap__*` prefix everywhere just makes variable names unreadable for me. For example what would you say about `postfix_virtual_ldap__virtual_alias_maps`? To me just just mishmash of similar words...

How about renaming the role to `debops.postldap`? The LDAP support most likely implies virtual mail in this case. And variable names would be a lot shorter, and `postldap__virtual_*` and `postldap__ldap_*` variable prefixes make a bit more sense that way. This would also be in line with `debops.postconf`, `debops.postscreen`, `debops.postwhite` roles. The `debops.postgresql` role throws a wrench in the works, but I think that we can forgive that since it's a completely separate project. I have written my comments with the current role name, but consider these to be renamed accordingly as well.

Thanks for not including the LDAP schema changes, they definitely can be done separately.

ansible/roles/debops.postfix_virtual_ldap/meta/main.yml
reixd: removed

docs/ansible/roles/debops.postfix_virtual_ldap/introduction.rst
reixd: done

docs/ansible/roles/debops.postfix_virtual_ldap/introduction.rst
reixd: Done

ansible/roles/debops.postfix_virtual_ldap/defaults/main.yml
reixd: Done

ansible/playbooks/service/postfix_virtual_ldap.yml
reixd: Removed as requested

ansible/roles/debops.postfix_virtual_ldap/defaults/main.yml
reixd: Renamed

ansible/roles/debops.postfix_virtual_ldap/defaults/main.yml
reixd: Cleanup not needed PKI stuff. Set tls_ca_cert_dir to default folder, in order to check the TLS cert of the LDAP server for a valid cert.

ansible/roles/debops.postfix_virtual_ldap/defaults/main.yml
reixd: Simplify posix user/group management and config

ansible/roles/debops.postfix_virtual_ldap/defaults/main.yml
reixd: Actually it solved things. 
Without this line emails to any email/alias in `/etc/aliasies` will be dropped or sent to a catch-all address.
Therefore the `$alias_maps` is required here on the first position, in order for postfix to find a match first in an alias and when not found in the LDAP mails or aliases.

reixd: Does anyone else, familiar with postfix, support this config?

ansible/roles/debops.postfix_virtual_ldap/defaults/main.yml
reixd: Renamed

ansible/roles/debops.postfix_virtual_ldap/defaults/main.yml
reixd: typo fixed

ansible/roles/debops.postfix_virtual_ldap/defaults/main.yml
reixd: This is currently a workaround. I have problem injecting the correct int value to the main.cf config. Somehow I get an OID value, where it should be an INT. Really weird.

ansible/roles/debops.postfix_virtual_ldap/defaults/main.yml
reixd: renamed var, It was a copy-pasta typo

> Also, that's an interesting solution for having standardized header for all `ldap_*` lookup tables. Since `debops.postfix` role does not have a dedicated solution yet, it can stay for now. :-)

;-)  I do not like code which repeats and is error prone.
If there is an easier method, please give me hint :)

ansible/roles/debops.postfix_virtual_ldap/defaults/main.yml
reixd: I disagree.  I think it is useful to specify this type of no_log condition ad-hoc while debugging, whitout modifying the role in the filesystem.
Example: 
`debops service/postfix_virtual_ldap  -e postfix_virtual_ldap__no_log=False -v`

ansible/roles/debops.postfix_virtual_ldap/defaults/main.yml
reixd: Initially it was root:root with 0600. But then postfix complained it couldn't read the virtual config file. At the end I need to use this settings.

ansible/roles/debops.postfix_virtual_ldap/defaults/main.yml
reixd: readability increased

ansible/roles/debops.postfix_virtual_ldap/defaults/main.yml
reixd: Renamed

ansible/roles/debops.postfix_virtual_ldap/defaults/main.yml
reixd: it doesn't bother here. But if you insist I will remove the vars.

reixd: I think this type of config helps to keep the roles consistent with the host, groups, all inventory config

ansible/roles/debops.postfix_virtual_ldap/defaults/main.yml
reixd: Renamed

ansible/roles/debops.postfix_virtual_ldap/defaults/main.yml
reixd: fixed

ansible/roles/debops.postfix_virtual_ldap/defaults/main.yml
reixd: I think it helps if for example you are hosting email for different domains, not subdomains but completely different domains. This works for the use case of and "email provider" outside of and organisation.

ansible/roles/debops.postfix_virtual_ldap/defaults/main.yml
reixd: I like to have this hierarchy of sections to keep the topic

reixd:
PUH! A lot of small changes. Pushed, let's get to round two :)

ansible/roles/debops.postfix_virtual_ldap/defaults/main.yml
drybjed: Very well, let's leave it for now then. In my opinion in production code there shouldn't be any debugging options at all to avoid accidental secret exposure, but I realize that makes actual debugging much harder on larger scales. The next best thing would be to implement some way of global, generalized debugging flag but with Ansible design the only way to get reliable global variables is either to stick them in inventory or to add a new role that's included in all other roles similarly to how `debops.secret` works. Maybe that would be the way to go - finally make a `debops.globals` role that's included everywhere via `import_role` and be done with it.

ansible/roles/debops.postfix_virtual_ldap/defaults/main.yml
drybjed: Perhaps [this thread on ServerFault](https://serverfault.com/questions/644306/) can shine some light on this. Perhaps you could show an example from `/etc/aliases` that resulted in mail being dropped when `$alias_maps` was not included in `$virtual_alias_maps`? I could then test this in my development environment, would be very interesting to know this behaviour.

ansible/roles/debops.postfix_virtual_ldap/defaults/main.yml
drybjed: It might be not an OID but an IPv4 address from a weird subnet, like `1.0.4.3` or something. Can you share the configuration you are trying to apply? I can then check it on my end and see what happens.

ansible/roles/debops.postfix_virtual_ldap/defaults/main.yml
drybjed: Well, OK then. I'll have to see the rendered documentation to properly judge it. :-)

ansible/roles/debops.postfix_virtual_ldap/defaults/main.yml
drybjed: In that case I would rename the variable to `postfix_virtual_ldap__ldap_private_subtree` to be more precise about the intended purpose, assuming that the "host" subtree would contain more Postfix-related objects than just some groups.

But then again, I suppose that's the intended scalability model. For example, in an environment you might want to deploy a few Postfix instances for round-robin load balancing. Would you expect all instances to use the same set of LDAP objects (horizontal scaling, I think), or each Postfix host to use separate set of users/mailboxes/domains/aliases (vertical scaling)? With how the DebOps project directories are focused on a single domain and DNS MX records support horizontal scaling, I would make use of the global set of LDAP objects rather than private subtree as the default, `False` here. The variable of course can stay so that vertical scaling can easily be enabled.

ansible/roles/debops.postfix_virtual_ldap/defaults/main.yml
drybjed: I currently don't see `postfix_virtual_ldap__ldap_combined_config` used anywhere, do you have any plans to use it?

ansible/roles/debops.postfix_virtual_ldap/defaults/main.yml
reixd: Set new default, updated docs

ansible/roles/debops.postfix_virtual_ldap/defaults/main.yml
reixd: When using different mail servers and domains changing the LDAP task this way is useful. But no, currently not used

ansible/roles/debops.postfix_virtual_ldap/defaults/main.yml
reixd: As of [Confused about alias_maps and virtual_alias_maps](https://serverfault.com/a/644393):

> Parameter virtual_alias_maps
>> Always invoked first time before any other address classes. It doesn't care whether the recipient domain was listed in mydestination, virtual_mailbox_domains or other places. It will override the address/alias defined in other places.

So this means, that if  we have users in LDAP with the following attributes:
- Alice:
  - `mail` alice.dow@example.net
  - `mailAlias: admin@example.net`

- Bob:
  - `mail` bob.jones@example.net
  - `mailAlias: @example.net`
  - `mailAlias: bob@anotherdomain.com`
  - `mailAlias: webmaster@anotherdomain.com`

 And this `/etc/aliases`:
```
root:                                admin@example.net
webmaster:                     admin@example.net
foo:                                  carl@othermailprovider.com
```

Using my proposed configuration, the following recipients will be routed as follows:
- webmaster --> Alias:admin@example.net --> Mailbox:alice.dow@example.net
- webmaster@example.com --> Alias:admin@example.net --> Mailbox:alice.dow@example.net
- alice@example.net -->CatchAll-Alias:@example.net --> Mailbox:bob.jones@example.net
- a@example.net -->CatchAll-Alias:@example.net --> Mailbox:bob.jones@example.net
- webmaster@anotherdomain.com -> Mailbox:bob.jones@example.net
- foo --> MailBox: carl@othermailprovider.com

Without explicitly listing  `$alias_maps` in `virtual_alias_maps` , `virtual_alias_maps` will just ask the LDAP server for `mail` and `mailAlias` without querying the local aliases database.

ansible/roles/debops.postfix_virtual_ldap/defaults/main.yml
drybjed: Actually you might have come with an optimal solution. I'm currently implementing it as a separate `item.connection` key which can be used to define the connection parameters, otherwise not much will change. I'll try to commit it today so that you can implement it in the role.

drybjed:
@reixd, you might want to check the changes in https://github.com/debops/debops/pull/1127 for better lookup table support.

ansible/roles/debops.postfix_virtual_ldap/defaults/main.yml
reixd: When the `virtual_minimum_uid` option is set to `state: present` I get the following result:

```
# --------------------- Virtual mail configuration ---------------------

virtual_mailbox_domains = ldap:/etc/postfix/ldap_virtual_domains.cf
virtual_alias_maps = $alias_maps,ldap:/etc/postfix/ldap_virtual_aliases.cf
virtual_mailbox_base = /var/vmail
virtual_mailbox_maps = ldap:/etc/postfix/ldap_virtual_recipients.cf
virtual_uid_maps = static:997
virtual_gid_maps = static:992
virtual_minimum_uid = 0.0.3.229    #<<--- WTF!?
```

The string there should be `997`, but I am unable to get this value correctly. Filtering `postfix_virtual_ldap__vmail_posix_uidnumber | int` to an INTEGER does not help either.

reixd: Nevertheless this config option can be actually removed:

- We have only one UID & GID storing the emails to `/var/vmail`, this is the `vmail` posix system user
- Current postfix virtual mail config does not allows to "virtual users" to access the mailDir directly, because the MailDir is not in the home folder of these users and the users may not be POSIX users but just Virtual Mail Users.
- So, without direct access to the filesystem (e.g. over SSH), users cannot read the MailDir directly, thus no need to specify different UIDs/GIDs.
- Mail reading will be handled entirely by Dovecot over IMAP. And dovecot will access the MailDir folder as the `vmail` user.

I hope it is clear....

ansible/playbooks/service/postldap.yml
```diff
@@ -0,0 +1,41 @@
...
+  become: True
+
+  environment: '{{ inventory__environment | d({})
+                   | combine(inventory__group_environment | d({}))
+                   | combine(inventory__host_environment  | d({})) }}'
+
+  roles:
+
+    - role: debops.etc_aliases/env
+      tags: [ 'role::etc_aliases', 'role::secret', 'role::postfix' ]
```
drybjed: Since the `debops.etc_aliases` role has been removed from the playbook, `debops.etc_aliases/env` subrole can also be removed.

ansible/roles/debops.postldap/COPYRIGHT
```diff
@@ -0,0 +1,18 @@
+debops.postldap - Configure Postfix SMTP service to use Virtual Mail with LDAP support
+
+Copyright (C) 2019 Rainer 'rei' Schuth <devel@reixd.net>
+Copyright (C) 2014-2017 DebOps https://debops.org/
```
drybjed: You should change the year to `2019` since that's when this role has been introduced to the project.

ansible/roles/debops.postldap/defaults/main.yml
```diff
@@ -0,0 +1,627 @@
...
+#
+# Virtual Mail POSIX username
+# For the accesses to the mailbox directories a separate user `vmail` (`Virtual Mail`) is created,
+# under which the accesses of Postfix, Dovecot and other components of the mail server should take place.
+# On the one hand this prevents mail server components from accessing sensitive system directories,
+# on the other hand it protects the mailboxes from external access.
+# Only vmail (and root) are allowed to access the mailboxes.
+postldap__vmail_posix_user: 'vmail'
+                                                                   # ]]]
+# .. envvar:: postldap__vmail_posix_uidnumber [[[
```
drybjed, line 67: You should add one empty line after the variable and before the `# ]]]` mark to make the variable a bit more readable.

On the next set of variables, you should remove the `#` before the variable for better code readability.

ansible/roles/debops.postldap/defaults/main.yml
```diff
@@ -0,0 +1,627 @@
...
+  bind: 'yes'
+  bind_dn: '{{ postldap__ldap_binddn }}'
+  bind_pw: '{{ postldap__ldap_bindpw }}'
+  scope: 'sub'
+                                                                   # ]]]
+# .. envvar:: postldap__postfix__lookup_virtual_aliases [[[
+#
+# The virtual_alias_maps setting is used to find the target mailbox
+# (configured in the attribute mail, aka the mailbox or maildrop).
+postldap__postfix__lookup_virtual_aliases:
```
drybjed: The `postldap__postfix__lookup*` variables should be renamed to `postldap__postfix_lookup*` because they are not directly passed via the role dependent variables and are not directly used in the playbook.

ansible/roles/debops.postldap/defaults/main.yml
```diff
@@ -0,0 +1,627 @@
...
+                                    if (ansible_local|d() and ansible_local.ldap|d() and
+                                        ansible_local.ldap.uri|d())
+                                    else [""] }}'
+
+                                                                   # ]]]
+# .. envvar:: postldap__ldap_private_subtree [[[
+#
+# When this variable is enabled, the :ref:`debops.ldap` role will create
+# separate LDAP objects that manage the Postfix groups as subtree of the
+# DokiWiki service LDAP object. If you set this parameter to ``False``, the
```
drybjed: You should change this to "Postfix service LDAP object". It's a different role, after all. :-)

ansible/roles/debops.postldap/defaults/main.yml
```diff
@@ -0,0 +1,627 @@
...
+# ^^^^^^^^^^^^^^^^^
+
+# .. envvar:: postldap__ldap_user_filter [[[
+#
+# The LDAP filter used to look up user accounts in the directory.
+postldap__ldap_user_filter: '(&
+                                           (objectClass=inetOrgPerson)
+                                           (objectClass=postfixUser)
+                                         )
+                                         (mailEnabled=TRUE)
```
drybjed: I think that without the custom mail schema which is currently planned but not available yet, and with `postfixUser` object class and `mailEnabled` attribute defined as required, no user accounts will be found by the LDAP filter, even with the required `authorizedService` values. Perhaps you could drop the custom object class and attribute from the search filter for now, at least until the proper mail schema is developed?

ansible/roles/debops.postldap/defaults/main.yml
```diff
@@ -0,0 +1,627 @@
...
+                                                                   # ]]]
+# .. envvar:: postldap__ldap_quota_attribute [[[
+#
+# Set the LDAP attribute value to be read by Postfix in order to get the user quota.
+postldap__ldap_quota_attribute: 'mailQuota'
+                                                                   # ]]]
+# .. envvar:: postldap__ldap_quota_default [[[
+#
+# User default LDAP quota. Use human-readable values, e.g. "2 GB".
+postldap__ldap_quota_default: '10 GB'
```
drybjed: Postfix doesn't support quota in a standard installation, there are patches for that but it looks like Debian `postfix` package does not carry one. There is probably a better solution offered by Dovecot - a [policy service](https://blog.sys4.de/postfix-dovecot-mailbox-quota-en.html) which can be queried by Postfix. It might be a better solution in a distributed environment.

Since the quota variables are not used elsewhere in the PR, and they won't be used due to above reasons, I would remove them from the PR entirely. The `mailQuota` attribute might come in handy with Dovecot, but that's not relevant for this PR.

ansible/roles/debops.postldap/defaults/main.yml
```diff
@@ -0,0 +1,627 @@
...
+# .. envvar:: postldap__ldap_combined_config [[[
+#
+# The variable that combines default and user LDAP configuration and is used in
+# the role tasks and templates.
+postldap__ldap_combined_config: '{{ postldap__ldap_default_config
+                                              + postldap__ldap_config
+                                              + postldap__group_ldap_config
+                                              + postldap__host_ldap_config }}'
+                                                                   # ]]]
+# .. envvar:: postldap__ldap__dependent_tasks [[[
```
drybjed, line 587: The dependent variables should be put in their own section at the end of the `defaults/main.yml` file. I would also consider moving other dependent variables here for consistency.

ansible/roles/debops.postldap/tasks/main.yml
```diff
@@ -0,0 +1,41 @@
...
+  getent:
+    database: 'group'
+    split: ':'
+    key: '{{ postldap__vmail_posix_group }}'
+
+- name: Get vmail passwd info
+  getent:
+    database: 'passwd'
+    split: ':'
+    key: '{{ postldap__vmail_posix_user }}'
```
drybjed: I think that having the UID/GID gathered at runtime might not be idempotent, if the playbook runs only with the `role::postfix` tag, but I noticed that you added it to the `debops.postldap` role as well, which should cover this situation. Still, I think that getting the UID/GID via an Ansible local fact and using the facts to define them in the default variables instead of `getent_group` and `getent_passwd` variables might be a bit better approach.

And the role should include the fact script anyway, to indicate that it has been applied on the host. It could be just a basic script without the UID/GID gathering if you don't want to implement it at this point, that can come later.

docs/ansible/roles/debops.postldap/getting-started.rst
```diff
@@ -0,0 +1,61 @@
...
+Default configuration
+---------------------
+
+The ``debops.postldap`` role configures a Postfix SMTP server with
+support for a ``virtual user mail system``, i.e. where the senders and
+recipients do not correspond to the Linux system users.
+Hence it is possible to host emails for other domains.
+The users, email alias and domains will be managed with LDAP.
+Local mail is enabled by default, support for mail aliases is provided by
+the ``debops.etc_aliases`` Ansible role and the LDAP user attribute
```
drybjed: You should use an internal reference here:
```
:ref:`debops.etc_aliases`
```

docs/ansible/roles/debops.postldap/guides.rst
```diff
@@ -0,0 +1,34 @@
...
+It is also possible to configure accounts with `wildcard` (catch-all)
+email addresses. The default configuration uses first the aliases set by
+:ref:`debops.etc_aliases` and then queries the LDAP server, if no match was found.
+
+See also :ref:`debops.dovecot` and :ref:`debops.roundcube` for an IMAP server
+and Email-Webclient correspondingly.
+
+.. code-block:: yaml
+
+  Work in progress...
```
drybjed: You can either expand this document before the PR is merged, or remove the "work in progress" note.

drybjed (changes requested):
@reixd Thanks for renaming the role to `debops.postldap`, I think that the variables look much better now. There are a few places I would like to see changes, check the comments below.

After the role rename, some of the value indentation in many variables is off. Perhaps you could make a pass on them and update the indentation?

ansible/roles/debops.postldap/defaults/main.yml
reixd: > On the next set of variables, you should remove the `#` before the variable for better code readability.

I do not quite understand your comment here. Can you give me a hint, a line number or an example?

drybjed: Sure. For example, you have two variables:

```yaml
# .. envvar:: postldap__vmail_posix_uidnumber [[[
#
# Virtual Mail POSIX uidNumber
#
postldap__vmail_posix_uidnumber: '{{ getent_passwd[postldap__vmail_posix_user][1]|int }}'
                                                                   # ]]]
# .. envvar:: postldap__vmail_posix_group [[[
#
# Virtual Mail POSIX group
#
postldap__vmail_posix_group: 'vmail'
                                                                   # ]]]
```

They should instead look like this:

```yaml
# .. envvar:: postldap__vmail_posix_uidnumber [[[
#
# Virtual Mail POSIX uidNumber.
postldap__vmail_posix_uidnumber: '{{ getent_passwd[postldap__vmail_posix_user][1]|int }}'

                                                                   # ]]]
# .. envvar:: postldap__vmail_posix_group [[[
#
# Virtual Mail POSIX group.
postldap__vmail_posix_group: 'vmail'

                                                                   # ]]]
```

This is probaly more readable?

ansible/roles/debops.postldap/defaults/main.yml
reixd: IMHO: Moving the dependent variables to a section "Configuration for other Ansible roles" at the end, breaks the logic behind the organisation of the variables. 
I tested your proposition by moving all dependent variables to the end, but then configuring the role while reading the docs is weird. Sections are split, it seems incoherent to me.

Although I acknowledge my proposal has a different structure than other roles, this role modifies postfix heavily, thus building the configuration docs in "chronological" order (what needs to be done first) helps for understanding.
But if you want it changed, its your call.

drybjed: I see. Readability and understanding of complex roles is important, probably even more so than the rules about variable order, and some roles can have exceptions to the rules. If you think that the current order better explains how the role works, let's keep it. :-)

ansible/roles/debops.postldap/defaults/main.yml
reixd: Yeah, thank you! WILCO

ansible/roles/debops.postldap/tasks/main.yml
reixd: > I think that having the UID/GID gathered at runtime might not be idempotent, 

you do not want idempotency here, you want the correct values at runtime.

> Still, I think that getting the UID/GID via an Ansible local fact and using the facts to define them in the default variables instead of `getent_group` and `getent_passwd` variables might be a bit better approach.

One second after writing the UID/GID values in the fact script, this values may be old and have changed. There is no guarantee they will be kept up to date and do not change. Thus getting the values just when you need them from the official POSIX/Linux user store is a good way to have up-to-date data.

> And the role should include the fact script anyway, to indicate that it has been applied on the host. It could be just a basic script without the UID/GID gathering 

What do you mean, just a simple fact script, so "we" know debops.postldap was executed before?

reixd:
@drybjed There is still work to be done: SASL authentication for users sending mail. So do not merge yet.

> "SASL –> Provides authentication for the SMTP server (since users are only allowed to send mail after they authenticated). User and password verification is done via LDAP."

There are two possibilities here: 
- Cyrus SASL (`libsasl2-2 sasl2-bin`)
- or, Dovecot, which has an SASL provider

Using dovecot is easier, less dependencies, less LDAP service users, lesser complexity, however as postfix needs to work standalone without dovecot, I will set SASL auth with Cyrus up.

ansible/roles/debops.postldap/tasks/main.yml
drybjed: > One second after writing the UID/GID values in the fact script, this values may be old and have changed. There is no guarantee they will be kept up to date and do not change. Thus getting the values just when you need them from the official POSIX/Linux user store is a good way to have up-to-date data.

I could say the same about the task-based method. After all, each task is in reality a separate Python script that gets sent to the remote host for execution, and returns with the response. If we are talking about second-based resolution, depending on the network and I/O latency the UID/GID could be changed even between first and second `getent` task.

> > And the role should include the fact script anyway, to indicate that it has been applied on the host. It could be just a basic script without the UID/GID gathering
> 
> What do you mean, just a simple fact script, so "we" know debops.postldap was executed before?

Yes, that's one of the reasons. The other is that DebOps roles are purposefully kept as separate as possible, and Ansible facts guarantee loose coupling between various roles. Not to mention that other roles like `debops.dovecot` could use the `postldap.fact` script facts for their own purposes, for example to configure UID/GID access if it's ever needed somewhere. In that case there will be no need to create another set of two `getent` tasks, because the facts are available to all roles.

drybjed:
> There are two possibilities here:
> 
>     * Cyrus SASL (`libsasl2-2 sasl2-bin`)

This is already done in the `debops.saslauthd` role. I think that LDAP integration should work out of the box when you enable it, but I haven't tested LDAP-based users yet. Check the role documentation.

>     * or, Dovecot, which has an SASL provider
> 
> Using dovecot is easier, less dependencies, less LDAP service users, lesser complexity, however as postfix needs to work standalone without dovecot, I will set SASL auth with Cyrus up.

I think that work in Dovecot can be done in a separate PR. This one has lots of various changes already, splitting the LDAP integration in separate PRs focused on different parts of the project will make it easier to digest and integrate with other potential changes down the road.

ansible/roles/debops.postldap/defaults/main.yml
```diff
@@ -508,16 +508,11 @@ postldap__ldap_start_tls: '{{ ansible_local.ldap.start_tls
...
-                                         (|
-                                           (authorizedService=postfix)
-                                           (authorizedService=mail)
-                                           (authorizedService=\*)
-                                         )'
+postldap__ldap_user_filter: '(|
+                               (authorizedService=postfix)
+                               (authorizedService=mail)
+                               (authorizedService=\*)
+                             )'
```
drybjed, line 535: I think that now the user filter is too broad and would let almost any LDAP object be treated as a mail account. Check the filter [from the `debops.saslauthd` role](https://github.com/debops/debops/blob/master/ansible/roles/debops.saslauthd/defaults/main.yml#L294-L313):
```yaml
      - name: 'ldap_filter'
        value: '(|
                  (&
                    (objectClass=inetOrgPerson)
                    (uid=%U)
                    (|
                      (authorizedService=smtpd)
                      (authorizedService=\2a)
                    )
                  )
                  (&
                    (objectClass=account)
                    (uid=%U)
                    (host=%r)
                    (|
                      (authorizedService=smtpd)
                      (authorizedService=\2a)
                    )
                  )
                )'
```
This filter is very narrow, but supports both People and Service accounts, with different attributes. But this filter is focused on mail submission authentication, and i don't think this would be a good fit for mail accounts because services that send mail don't necessarily have to have real mailboxes.

I would probably replace the `(objectClass=account)` part with something focused on functional accounts, think "marketing", "finances", "support", etc. Basicaly department-based accounts which might be shared between a few users. I think that a good way to support them would be to include support for [shared mailboxes](https://vogasec.wordpress.com/2012/07/01/ubuntu-postfix-dovecot-shared-mailboxes/) in Dovecot. But that's out of scope of this PR.

Also, changing the `authorizedService=postfix` to `authorizedService=smtpd` would make the attribute values more agnostic; other SMTP servers could support the same filter.

docs/ansible/roles/debops.postldap/guides.rst
```diff
@@ -28,7 +28,99 @@ email addresses. The default configuration uses first the aliases set by
...
 and Email-Webclient correspondingly.
 
+The following example shows a real-world™ setup on the Hetzner Cloud. It consists of two servers,
+one ``controller`` and a ``mail-server``. The LDAP directory is hosted also in the ``controller``.
+``mail-server`` has access to LDAP over an internal network (10.10.10.0/24) attached directly to the VMs.
+This setup has no internal DNS server (no split-DNS), thus internal IPs are mapped to DNS entries in the form ``$service.hetzner.mydomain.net``.
+See `secret__ldap_fqdn`.
+
+.. code-block:: yaml
+  :caption: ansible/inventory/group_vars/hetzner/ldap.yml
```
drybjed, line 37: This. Changes. Everything.

Edit: Ehh, not so much. I expected a bit better from the ReadTheDocs theme we are using. Perhaps this could somehow be improved from within DebOps documentation via CSS injection? Need to research more.

docs/ansible/roles/debops.postldap/guides.rst
```diff
@@ -28,7 +28,99 @@ email addresses. The default configuration uses first the aliases set by
...
+  ldap__enabled: True
+
+  ldap__domain: 'mydomain.net'
+
+  # Domain used for LDAP base DN and to select default LDAP server.
+  secret__ldap_domain: 'mydomain.net'
+
+  # LDAP server hostname / internal IP address which holds the database.
+  # ldap_* modules will connect to it natively, so it should be available at least from the Ansible Controller.
+  secret__ldap_fqdn: 'ldap.hetzner.mydomain.net'
```
drybjed: The `secret__ldap_*` variables are not relevant anymore and will be removed from the role before the `v2.0.0` DebOps release. The `debops.ldap` role has taken over their duties.

reixd: So remove them from this PR?

ansible/roles/debops.postldap/defaults/main.yml
reixd: It is a chicken egg problem.... Will change as proposed

docs/ansible/roles/debops.postldap/guides.rst
```diff
@@ -28,7 +28,99 @@ email addresses. The default configuration uses first the aliases set by
...
+  slapd__group_allow:
+    # IPv4/v6 localhost
+    - '127.0.0.1'
+    - '::1'
+    - '{{ ansible_default_ipv4.address
+          if (ansible_default_ipv4|d() and ansible_default_ipv4.address|d() )
+          else "127.0.0.1" }}'
+    - '{{ ansible_default_ipv6.address
+          if (ansible_default_ipv6|d() and ansible_default_ipv6.address|d() )
+          else "::1" }}'
```
drybjed: The above 4 entries shouldn't be needed. The `debops.ferm` firewall already accepts all traffic on `localhost` explicitly.

I'm not sure why would you want to talk to the `slapd` service over external IP addresses from the same host. But if you found out that this is needed for some reason, I think that a better solution would be to put the `ansible_default_ipv*` entries in the `debops.slapd` role, probably in a separate variable that gets activated when the public access to the OpenLDAP service is disabled. Just need a good use case for this.

docs/ansible/roles/debops.postldap/guides.rst
```diff
@@ -28,7 +28,99 @@ email addresses. The default configuration uses first the aliases set by
...
+  # basic Postfix SMTP server with configuration similar to the "Internet Site"
+  # MTP service listens for connections on port 25 from all hosts.
+  # Mail relay is authorized from localhost, other hosts are deferred.
+
+  postfix__domain: 'mydomain.net'
+
+  postfix__pki_realm: 'mail.mydomain.net'
+
+  # Enable debugging
+  postldap__no_log: False
```
drybjed: I feel that this should be removed from the example, because users will copy it over to their inventory and forget to disable it, making their installations insecure. You could mention that in order to help with debugging issues there's this variable to make the hidden configuration visible, but don't include it directly in the example - just add an `:envvar:` reference in the documentation.

docs/ansible/roles/debops.postldap/guides.rst
```diff
@@ -28,7 +28,99 @@ email addresses. The default configuration uses first the aliases set by
...
+
+  postldap__ldap_private_subtree: True
```
drybjed: This variable probably needs a bit of explanation why it is here and what will happen when it gets included in the inventory.

reixd:
> This is already done in the debops.saslauthd role

There are so many roles, I forget what can be used... ;-)

>I  think that work in Dovecot can be done in a separate PR.

This is part of my Virtual Mail Setup. Already working on it. Of course a separate PR. Will upload soon for for initial review. Just need to solve the sasl auth stuff.
But my PoC works xD, just need to make everything prettier

docs/ansible/roles/debops.postldap/guides.rst
```diff
@@ -28,7 +28,99 @@ email addresses. The default configuration uses first the aliases set by
...
+  postldap__ldap_user_filter: '(&
+                                 (objectClass=inetOrgPerson)
+                                 (objectClass=postfixUser)
+                               )
+                               (mailEnabled=TRUE)
+                               (|
+                                 (authorizedService=postfix)
+                                 (authorizedService=mail)
+                                 (authorizedService=\*)
+                               )'
```
drybjed: This included in the example begs the question why this filter wasn't included by default in the role... Check my other comment about the user filter - I think that with the improved default this particular example could be removed from the guide.

ansible/roles/debops.postfix/defaults/main.yml
```diff
@@ -900,7 +900,9 @@ postfix__original_mastercf:
     private: False
     chroot: True
     command: 'smtpd'
-    state: 'comment'
+    state: '{{ "present"  
+               if (postfix__pki|bool)
+               else "comment" }}'
```
drybjed: This is a wrong place to make this change. This variable holds the original `/etc/postfix/master.cf` contents, so that they can be changed by other DebOps roles, in particular `debops.postconf`. In there, the `smtps` aka `submissions` service will be enabled automatically when `auth` capability is enabled as well. You should revert this change.

docs/ansible/roles/debops.postldap/guides.rst
drybjed: Yes.

ansible/roles/debops.postldap/defaults/main.yml
drybjed: This is not a chicken-egg problem, you don't initialize mail accounts in LDAP via Postfix. Besides, DebOps being in a monorepo lets us create a single definition of an e-mail account and keep it coherent across multiple Ansible roles. Nothing chikeny about that, I'm afraid. :-)

ansible/roles/debops.postldap/defaults/main.yml
```diff
@@ -19,6 +19,7 @@
 # List of additional APT packages needed for the LDAP support.
 postldap__postfix__dependent_packages: '{{ [ "postfix-ldap" ]
                                            if postldap__ldap_enabled else [] }}'
+
                                                                    # ]]]
                                                                    # ]]]
```
drybjed, line 23: The section endings are exception to the rule - since you already have two lines here, the spacing between variables is fine, and you shouldn't add additional blank space. Now it will look "off" in the other direction. Sorry for pedantry. :-)

drybjed:
Hundredth comment in a PR, woo!

reixd:
This is actually the second PR. So we have more comments... :D

docs/ansible/roles/debops.postldap/guides.rst
reixd: > I'm not sure why would you want to talk to the `slapd` service over external IP addresses from the same host. But if you found out that this is needed for some reason,

The ext IPs rules are needed to access the SLAPd over SSH-Port-Forwarding. So I am able to use ApacheDirectoryStudio to modify LDAP attributes.

> I think that a better solution would be to put the `ansible_default_ipv*` entries in the `debops.slapd` role, probably in a separate variable that gets activated when the public access to the OpenLDAP service is disabled. Just need a good use case for this.

See above. But I do not think it is a good reason to be a default. Instead write down this solution in a Knowledge Base.

@drybjed What do prefer: keep or remove ?

docs/ansible/roles/debops.postldap/guides.rst
reixd: True, my bad. I just copy-pasted everything :D

docs/ansible/roles/debops.postldap/guides.rst
reixd: Removed subtree var. If the admin want this setting --> RTFM.

ansible/roles/debops.postldap/defaults/main.yml
reixd: I was actually referring to the "Problem" of the unfinished LDAP schema.

docs/ansible/roles/debops.postldap/guides.rst
reixd: What do you have in mind?

I thought of a "simple" tutorial/example on how postfix and other services can be configured (copy-pasted) in order to have a working solution in no time.

docs/ansible/roles/debops.postldap/guides.rst
reixd: Removed for the sake of simplicity, however the question with the SSH PF still remains

docs/ansible/roles/debops.postldap/guides.rst
drybjed: The default configuration in roles is supposed to give you a working solution - DebOps is designed for use in production, you are supposed to be able to set up new hosts, apply the playbook and go, all of the role default variables are there so that you can tweak stuff from a typical use case to a more specific one. If that's not the case with some roles that are supposed to configure stuff automatically, it needs to be fixed in the roles, not provided as some kind of additional tutorial with required configuration changes.

But in this specific case I was referring to the `:caption:` option in reStructuredText. I would like to be able to label various inventory examples with specific paths, but when I tried out `:caption:` with the current RTD theme it was kinda meh - the path was centered in kind of away from the file snippet. It could look better, IMO.

docs/ansible/roles/debops.postldap/guides.rst
drybjed: Ah, SSH port forwarding... I see. I just connect to the OpenLDAP server directly, over TLS. You could open the firewall if you have static IP address. For dynamic IP address, Tinc VPN could be a good solution, so that you have access to the internal network.

Wtih SSH port forwarding, I think that since you forward the TCP port to `localhost`, the firewall should work fine without additional rules. If you show what `ssh` command you use to create the tunnel, or other needed configuration, I could test it locally to see if you really needed to open the host's external IP addresses.

reixd: Yeah, tinc is in the todo list...

reixd:
How do I need to do, to get posfix with smtps ans sasl and LDAP going? I added the corresponding host to the ansible group. It installed, all green. but it's not working.
Ideas?

drybjed:
Make sure that PKI and LDAP is set up on the host. Add the host to groups:
```
[debops_service_postfix]
[debops_service_postconf]
[debops_service_saslauthd]
```
Run the command:
```
debops service/postfix service/saslauthd service/postconf -l host
```
This should do the trick.

reixd:
> This should do the trick.

It helped after adding the "missing" postconf ansible group to the mailserver. It is not quite clear from the current doc, that to have a working mail server setup you need to install at least 3 roles (postfix, postconf, saslauth).

ansible/roles/debops.postldap/tasks/main.yml
reixd: So, added a facts script with the data from user module.
Does this works for you?

ansible/roles/debops.postldap/tasks/main.yml
```diff
@@ -0,0 +1,54 @@
...
+    owner: '{{ postldap__vmail_posix_user }}'
+    group: '{{ postldap__vmail_posix_group }}'
+    mode: '0750'
+  tags: [ 'role::postldap:user' ]
+
+- name: Make sure Ansible local facts directory exists
+  file:
+    dest: '/etc/ansible/facts.d'
+    state: 'directory'
+    owner: 'root'
```
drybjed, line 39: You can drop the `owner` and `group` parameters from the `file`/`template` tasks if both are `root`.

ansible/roles/debops.postldap/templates/etc/ansible/facts.d/postldap.fact.j2
```diff
@@ -0,0 +1,12 @@
...
+cat << EOF
+{
+    "installed": true,
+    "vmail_posix_user" : "{{ postldap__vmail_posix_user }}",
+    "vmail_posix_uidnumber" : {{ postldap__register_vmail_posix_user.uid|int }},
+    "vmail_posix_group" : "{{ postldap__vmail_posix_group }}",
+    "vmail_posix_gidnumber" : "{{ postldap__register_vmail_posix_user.group|int }}",
+    "mailbox_base" : "{{ postldap__mailbox_base }}"
+}
+EOF
```
drybjed, line 12: Good first try, but you should use Python instead of Bash for the fact scripts. There are lots of good examples in various roles.

Optimally the facts should get the information themselves, not rely on Jinja templating - that should only be for data that cannot be inferred from the host itself, for example if particular option is defined in the Ansible role. In this particular case, you could check if some Dovecot binary is present to define if Dovecot is installed, store the user and group name in the fact script via Jinja and JSON loading, and get their corresponding UID and GID from the system user and password database via the `pwd` and `grp` Python modules. I'm not sure about mailbox base - do we go with the home directory of the `vmail` user, or perhaps a subdirectory? I guess that because this directory is created by the role itself, it should be stored in the script via Jinja and not inferred somehow from the host since I wouldn't know from where get that information.

drybjed:
@reixd Good work, keep it up. The fact script needs to be rewritten but otherwise I think that the PR is ready for testing, I'll try to do that in the next few days.

drybjed:
I tested the `debops.postldap` role and sending mail to the virtual users works! Great job, @reixd! I plan to work on the fact script in the evening, and maybe some slight changes here and there, I'll let you review the changes before merging.

One thing I noticed is that relying on the `mail` attribute as the mailbox target gets complicated when you have more than one value. For example with two I got the maildir in `/var/vmail/example.org/user/mailbox/,/other.example.org/user/mailbox/` directory. Confusing. :-) We need to develop that Mail LDAP schema soon, hopefully it will be ready before DebOps v2.0.0 release.

drybjed:
@reixd I added some improvements to your proposed role in https://github.com/debops/debops/pull/1161, let me know what you think about the changes. I'll be away for the weekend, if you think the PR is OK it will be merged on Monday.

PR #1130, opened 2019-11-04
[debops.ferm] Prefix conntrack vars w/ "tracking_"

imrejonk:
I was scratching my head a bit too long trying to customise the
iptables target in a connection tracking rule. I would add an
`item.invalid_target` variable as written down in the `debops.ferm`
documentation, but this didn't seem to change anything. Turns out that
the related template file only looks at `item.tracking_invalid_target`.
This variable was not documented anywhere. The same goes for
`item.tracking_active_target` and `item.tracking_module`.

These changes rename the `item.{active_target|invalid_target|module}`
variables by prefixing them with "tracking_". The documentation gets
updated as well to reflect this.


drybjed (approved):
Sounds good, thanks. I'll add the upgrade notes.

PR #1131, opened 2019-11-05
Allow to add additional MIME Types to dokuwiki

reixd:


ansible/roles/debops.dokuwiki/defaults/main.yml
```diff
@@ -381,6 +381,37 @@ dokuwiki__protected_conf_php: |
...
+# .. code-block:: yaml
+#
+#      # Extra allowed MIME Types
+#      txt     text/plain
+#      tex     text/plain
+#      conf    text/plain
+#      xml     text/xml
+#      csv     text/csv
+#      svg     image/svg+xml
+#      epub    application/epub+zip
```
drybjed: I think that these are reasonable choices for a default set, perhaps you could move them to the variable instead of providing them as an example?

drybjed (approved):
Looks good, thanks. I'll backport this change to the stable releases as well.

drybjed:
Unfortunately, there are conflicts during backporting, so that won't happen. Sorry.

PR #1133, opened 2019-11-07
[debops.rsyslog] Fix setting perimssions on /var/log

gaudenz:
The command to update directory and file permissions only changed the
permissions on /var/log if the group was not correct. If the group was
correct but the permissions were wrong, the permissions were not fixed.
As the group for /var/log is already fixed when creating the the system
group, the permissions stayed wrong.

This is fixed by adding an independent check for the permissions and
fixing them if they don't match.


drybjed (approved):
Thanks. This change will only affect Ubuntu hosts, so I hope it works fine there. I'll backport the patch to the stable releases.

PR #1138, opened 2019-11-10
[debops.dovecot] Add Virtual Mail capabilities with LDAP backend

reixd:
This PR allows dovecot to provide Virtual Mail together with Postfix using an LDAP backend.


ansible/roles/debops.dovecot/defaults/main.yml
```diff
@@ -58,10 +59,13 @@ dovecot_deny_users: [ 'root' ]
...
-# Mailbox location. For maildir set something like ``maildir:~/Maildir``.
+# Mailbox location. For mbox set something like ``mbox:~/mail:INBOX=/var/mail/%u``.
 # For more information about the supported format, check `Dovecot Mail
-# Location <http://wiki2.dovecot.org/MailLocation>`_
-dovecot_mail_location: 'mbox:~/mail:INBOX=/var/mail/%u'
+# Location <https://doc.dovecot.org/configuration_manual/mail_location/>`_
+dovecot_mail_location: 'maildir:{{ "/var/vmail/%d/%n/mailbox"
+                                  if (dovecot__ldap_enabled|d(False))
+                                  else
+                                  "~/Maildir" }}'
```
ganto, line 73: What's the reason for changing the default from mbox to maildir? I personally also use maildir, but that's a breaking change that is not topic related.

ansible/roles/debops.dovecot/defaults/main.yml
```diff
@@ -206,32 +210,64 @@ dovecot_passwdfile_mailbox_path: '%u'
 # .. envvar:: dovecot_pki [[[
 #
 # Enable or disable support for TLS in Dovecot (using debops.pki)
-dovecot_pki: True
+dovecot_pki: '{{ (True
+                                if (ansible_local|d() and ansible_local.pki|d() and
```
ganto: Something is weird with the indentation here

ansible/roles/debops.dovecot/defaults/main.yml
```diff
@@ -206,32 +210,64 @@ dovecot_passwdfile_mailbox_path: '%u'
...
+                                    ansible_local.pki.enabled|d() | bool)
+                                else False) | bool }}'
 
                                                                    # ]]]
 # .. envvar:: dovecot_pki_path [[[
 #
 # Base PKI directory
-dovecot_pki_path: '{% if (ansible_local is defined and ansible_local.pki is defined) %}{{ ansible_local.pki.base_path }}{% else %}/etc/pki{% endif %}'
+dovecot_pki_path: '{{ ansible_local.pki.path
+                                         if (ansible_local|d() and ansible_local.pki|d() and
```
ganto: Indentation

ansible/roles/debops.dovecot/defaults/main.yml
```diff
@@ -241,12 +277,12 @@ dovecot_pki_key: 'default.key'
...
 
                                                                    # ]]]
-# .. envvar:: dovecot_ssl_protocols [[[
+# .. envvar:: dovecot_ssl_min_protocol [[[
 #
 # SSL ciphers to use. On new distros you only specify the minimum ssl
 # protocol version Dovecot accepts, defaulting to TLSv1.
 # On old distros disable SSLv2,3 and allow TLSv1.0 or better.
-dovecot_ssl_protocols: '{{ "!SSLv2 !SSLv3"
+dovecot_ssl_min_protocol: '{{ "!SSLv2 !SSLv3"
```
ganto, line 290: If you rename the variable also use the proper prefixing with two __ as you did for the new variables.

ansible/roles/debops.dovecot/defaults/main.yml
```diff
@@ -281,14 +318,65 @@ dovecot_ssl_cipher_list_better_cypto: 'EDH+CAMELLIA:EDH+aRSA:EECDH+aRSA+AESGCM:E
...
+# .. envvar:: dovecot_imap_default_config_map [[[
+#
+# Default IMAP configuration properties. See :ref:`dovecot_imap_config_map`
+# for more details.
+dovecot_imap_default_config_map:
+  login-service:
+    inet_listener:
+      imap:
+        # Use only localhost when PKI is true, otherwise listen on all IPs
+        address: '{{ "127.0.0.1, [::0]" if (dovecot_pki|d(True)) else "0.0.0.0, [::]" }}'
```
ganto: Looks reasonable but breaks compatibility. Just a note.

ansible/roles/debops.dovecot/defaults/main.yml
```diff
@@ -281,14 +318,65 @@ dovecot_ssl_cipher_list_better_cypto: 'EDH+CAMELLIA:EDH+aRSA:EECDH+aRSA+AESGCM:E
...
+dovecot_imap_listeners: '{{ [ "imap", "imaps" ]
+                            if ( dovecot_pki|d(True) )
+                            else [ "imap" ] }}'
+
+                                                                   # ]]]
+# .. envvar:: dovecot_imap_default_config_map [[[
+#
+# Default IMAP configuration properties. See :ref:`dovecot_imap_config_map`
+# for more details.
+dovecot_imap_default_config_map:
```
ganto, line 373: Can use `dovecot__` prefix for new variable

ansible/roles/debops.dovecot/defaults/main.yml
```diff
@@ -297,12 +385,21 @@ dovecot_imap_listeners: [ 'imap', 'imaps' ]
 # for more details.
 dovecot_imap_config_map: {}
 
+                                                                   # ]]]
+# .. envvar:: dovecot_imap_combined_config_map [[[
+#
+# Combined IMAP configuration properties.
```
ganto, line 399: Is it necessary to add this to the defaults? Are you using own custom variables to combine?

ansible/roles/debops.dovecot/templates/etc/dovecot/conf.d/auth-ldap.conf.ext.j2
```diff
@@ -0,0 +1,15 @@
+# {{ ansible_managed }}
+
+# Virtual Mail user and group
+mail_uid = {{ dovecot__vmail_posix_user }}
+mail_gid = {{  dovecot__vmail_posix_group }}
```
ganto:
```suggestion
mail_gid = {{ dovecot__vmail_posix_group }}
```

ganto (commented):
Thanks a lot for all your effort. Generally looks very well. Maybe it would be nice if you could fix all the indentation issues. I guess you were using tabs instead of spaces?

@drybjed : This PR contains some breaking changes. If we merge this for a new major release, we might as well also adjust the variable naming with the double-underline prefix and maybe drop support for wheezy/jessie/precise/trusty? What do you think?

reixd:
> Thanks a lot for all your effort. Generally looks very well. 
Thank you for also reviewing the PR and not just drybjed :)

> Maybe it would be nice if you could fix all the indentation issues. I guess you were using tabs instead of spaces?

Shure. Actually I only work with spaces. Dunno why this indentation issues happened.


> we might as well also adjust the variable naming with the double-underline prefix and maybe drop support for wheezy/jessie/precise/trusty?

+1

ansible/roles/debops.dovecot/defaults/main.yml
reixd: I wanted to be as close as possible as current dovecot config defaults. Looking at their docs and configs, I chosed to use the MailDir sane default.

ansible/roles/debops.dovecot/defaults/main.yml
reixd: Actually the new variables with two underlines (__) are actually wrong. As this role does not have the new variable name scheme with two underlines, I needed to stick to this "old format".

ansible/roles/debops.dovecot/defaults/main.yml
reixd: It's a sane secure default, avoid the transmission of clear text passwords over the Interweb, specially when using a default Roundcube setup on another webserver.

Using the same argumentation as for nginx: nginx forwards almost everything from 80 to 443. Therefore we could to something similar with mail.

ansible/roles/debops.dovecot/defaults/main.yml
reixd: See https://github.com/debops/debops/pull/1138#discussion_r345620339

drybjed:
@ganto Thanks for reviewing! Sure, renaming all of the variables in a proper commit sounds good to me, of course that won't be backported to the stable releases.

I'm still planning to rewrite the entire role to utilize `parse_kv_items` / `parse_kv_config` filters so that the Dovecot configuration can be controlled better, but I haven't got motivation for that particular role yet. And I'm currently preparing for the next stable release, and the next one I would like to use to clean up old code and rename all roles for better Ansible Galaxy Collections support, so that won't happen for at least February.

ansible/roles/debops.dovecot/defaults/main.yml
reixd: Yes it helps.

This is part of my inventory config for dovecot:

```
dovecot_imap_config_map:

  login-service:
    service_count: 1
    process_min_avail: 1

  service:
    process_limit: 512

  protocol:
    mail_max_userip_connections: 15
```

reixd:
> Sure, renaming all of the variables in a proper commit sounds good to me, of course that won't be backported to the stable releases.

So what now? 

- Should we review this role and make it work with stable and master branches,
- or in addition also rename the vars to the new scheme and drop backporting? 

What do you prefer?

ansible/roles/debops.dovecot/defaults/main.yml
drybjed: If we are going with variable renames, I think that might a good moment to change the default to Maildir as well. The `mbox` format is old and I'm not sure about scalability, Maildir seems to be more sane. I wonder how [`dbox` storage format](https://wiki2.dovecot.org/MailboxFormat/dbox) would fare in this case and how the conversion process goes for these changes.

ansible/roles/debops.dovecot/defaults/main.yml
drybjed: I think that this is backwards. If we have PKI available, IMAP should listen for all connections because IMAP clients exist - we don't have just Roundcube after all. If we don't have PKI available, switching to `localhost` only is reasonable, to be secure.

Your comparsion is right for the wrong reasons - `nginx` provides "public" services which are exepcted to be reachable from anywhere by default, of course over TLS. IMAP in my view is the same - if we provide a mail storage service, offering access via IMAP by default, over TLS, is a reasonable policy, making the Dovecot service "public" as well.

reixd: So keep :)

drybjed:
@reixd I think that renaming all variables would push the merge of this role to the next release, after DebOps `v1.2.0` is released in late November. I'd like to use the remaining days to fix bugs and stabilize the `master` branch before the next release, not introduce major changes.

I'll try to test the mail-LDAP PRs until the end of the week, depending on available time. If you think that it's fine to push this PR to the next release then it could be polished better with time. Your choice.

reixd:
> I'll try to test the mail-LDAP PRs until the end of the week, depending on available time. If you think that it's fine to push this PR to the next release then it could be polished better with time. Your choice.

I would prefer to merge this PRs in the next days. I will be very busy next month and won't be able to make fast or any changes to the PR.
But I do not need a backporting to stable releases. Master is fine.

reixd:
> I'll try to test the mail-LDAP PRs until the end of the week, depending on available time. 

FYI, this is a sample LDAP ldiff to test my changes. It uses a different object class name `postfixUser`, so that a change to `mailUser``can happen without problems ;)
```
# Postfix attribute definitions
# -------------------------------

attributetype ( 1.3.6.1.4.1.4203.666.1.202
                NAME 'mailEnabled'
                DESC 'TRUE to enable, FALSE to disable mail account'
                EQUALITY booleanMatch
                SYNTAX 1.3.6.1.4.1.1466.115.121.1.7 SINGLE-VALUE )

attributetype ( 1.3.6.1.4.1.4203.666.1.200
                NAME 'mailAlias'
                DESC 'Postfix mail alias attribute'
                EQUALITY caseIgnoreMatch
                SUBSTR caseIgnoreSubstringsMatch
                SYNTAX 1.3.6.1.4.1.1466.115.121.1.15{1024} )


# Postfix object definitions
# ----------------------------

objectclass ( 1.3.6.1.4.1.4203.666.1.100
              NAME 'postfixUser'
              DESC 'Postfix mail user class'
              SUP top
              AUXILIARY
              MAY  ( mailEnabled $ mailAlias ) )
```

ansible/roles/debops.dovecot/defaults/main.yml
drybjed: @reixd What? No, reverse the polarity - put the localhost IP addresses where the `0.0.0.0` and `[::]` are and vice versa, then it will be fine.

reixd: What I want to achieve is:

```
If (PKI is active) then
  IMAP Listen in v4,v6 localhost addresses;
  IMAPS Listen in all v4,v6 addresses;
Else
  IMAP Listen in all v4,v6 addresses;
  IMAPS is deactivated;
```

drybjed: IMAP over STARTTLS is also a thing. So in this case only enabling or disabling IMAPS makes sense depending on availability of PKI. You should keep IMAP open in both cases.

reixd: So you want this?
IMAP listen to Internet connections if pki enaed, otherwise keep to localhost?
```
address: '{{ "0.0.0.0, [::]" if (dovecot_pki|d(True)) else "127.0.0.1, [::0]" }}'
```

drybjed: Yes, that sounds reasonable. You could also check the firewall rules and enable/disable them accordingly.

reixd:
I think all reviews are implemented. Did I miss something?

For the sake of simplicity the ferm rules format has been updated to the XXI century :D

PR #1140, opened 2019-11-11
[debops.slapd] LDAP ACL reorganization

### LINKED ISSUE #1110: [LDAP] Planned major changes in the OpenLDAP ACLs
I would like to change the type of some objects created by default in an OpenLDAP installation from `groupOfNames` to `organizationalRole`. This involves modifications in the default OpenLDAP ACLs which might break installations if people are unaware of the changes needed.

The change comes from the fact that the `groupOfNames` object class requires at least one `member` attribute to be present, and the `organizationalRole` does not. This results in some of the objects in the LDAP directory, for example `cn=LDAP Replicators,ou=System Groups,dc=example,dc=org` to require at least one `member`, even if there are no replicators configured yet. And when the last `member` of a given group is deleted, the entry is not updated by the `refint` overlay because removing the `member` attribute would break the `groupOfNames` object consistency.

In comparsion, the `organizationalRole` object class does not enforce any `roleOccupant` attributes (equivalent to the `member` attribute) to be present, so we can have "roles" defined without anybody that performs them and no issues with directory consistency on removals. Additionally, `organizationalRole` has much more diverse set of attributes that can describe a role than `groupOfNames` which might be a plus.

Changing the object class will require modifications to the OpenLDAP ACLs to base the lookups on the new object attributes, changes in the `ldap/init-directory.yml` playbook which will create the new objects as well as additional indexes at least for `roleOccupant` and more attributes in the `refint` overlay configuration. Because this change is very intrusive and the LDAP directory structure will have to be carefully modified, I'll schedule it for the DebOps `v2.0.0` release. The migration process will be carefully planned and described in the upgrade notes - I think that first running the `ldap/init-directory` playbook and then applying the `debops.slapd` role LDAP tasks to change the ACLs should be enough, but that will have to be tested first.

Perhaps at the same time the objects in the `ou=System Groups,dc=example,dc=org` subtree could be moved to `ou=Groups,dc=example,dc=org`. After using the current set of ACLs and directory layout for a while I see some issues with using two separate sets of groups in various applications. This will probably require more rules to protect certain groups like `cn=LDAP Administrators`, but that might be beneficial in the long run.

drybjed:
I found a way to make migration of the ACLs to the new layout described in https://github.com/debops/debops/issues/1110 as easy as possible (just the UNIX groups need to be moved by hand), so I decided to go ahead with the change early, as well as backport the changes to the stable releases for consistency.

Fixes #1110

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -129,6 +129,15 @@ LDAP
 - The Access Control List rules can now be tested using the :man:`slapacl(8)`
   command via a generated :ref:`test suite script <slapd__ref_acl_tests>`.
 
+- The default ACL rules have been overhauled to add support for the
+  ``ou=Roles,dc=example,dc=org`` subtree and use of the ``organizationalRole``
+  LDAP objects for authorization. The old set of rules is still active to
+  ensure that the existing environments work as expected.
+
+  If you use a modified ACL configuration, you should include the new rules as
+  well to ensure that changes in the :ref:`debops.ldap` support are working
+  correctly.
+
 :ref:`debops.unbound` role
 ''''''''''''''''''''''''''
 
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -138,6 +138,11 @@ LDAP
   well to ensure that changes in the :ref:`debops.ldap` support are working
   correctly.
 
+- You can now hide specific LDAP objects from unprivileged users by adding them
+  to a special ``cn=Hidden Objects,ou=Groups,dc=example,dc=org`` LDAP group.
+  The required ACL rule will be enabled by default; the objects used to control
+  visibility will be created by the :file:`ldap/init-directory.yml` playbook.
+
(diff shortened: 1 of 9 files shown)


ypid:
I might actually review all your LDAP work more closely for my LPIC-2. For now just one thing, why 2FA with SMS when we have TOTP ;-) ?

drybjed:
There's already `debops.smstools` role which I would like to adapt to support the `mobile` attribute from LDAP. Also, SMS is easier to understand for casual users than TOTP. This is not just for us, remember.

drybjed:
@ypid Heading for certification? Let me know if you find some issues or improvements.

ypid:
> Also, SMS is easier to understand for casual users than TOTP.

I guess :-/

> Heading for certification? Let me know if you find some issues or improvements.

As always :) Yes, LPIC seems nice. Third in the series would then be LPIC-3 303: Security, some day I hope :)

PR #1144, opened 2019-11-12
[debops.roundcube] Update Roundcube to version 1.4.0.

reixd:
This PR updates the current role to support the new v1.4.0 including the elastic responsible skin.


ansible/roles/debops.roundcube/templates/srv/www/sites/config.inc.php.j2
```diff
@@ -4,8 +4,12 @@
  +-----------------------------------------------------------------------+
  | Local configuration for the Roundcube Webmail installation.           |
  |                                                                       |
+ | This is a sample configuration file only containing the minimum       |
+ | setup required for a functional installation. Copy more options       |
+ | from defaults.inc.php to this file to override the defaults.          |
```
ganto: I don't think that's a good idea to re-add these comments here. They don't really make sense, as this is not a sample configuration nor should the user be adviced to copy anything manually into this file.

ansible/roles/debops.roundcube/templates/srv/www/sites/config.inc.php.j2
```diff
@@ -19,19 +23,34 @@
...
 {% if roundcube__database_map[roundcube__database].dbtype == 'mysql' %}
 $config['db_dsnw'] = 'mysql://{{ roundcube__database_map[roundcube__database].dbuser }}:{{ roundcube__database_map[roundcube__database].dbpass }}@{{ roundcube__database_map[roundcube__database].dbhost }}/{{ roundcube__database_map[roundcube__database].dbname }}';
 {% endif %}
 
-// The mail host chosen to perform the log-in.
+// ----------------------------------
+// LOGGING/DEBUGGING
+// ----------------------------------
+// log driver:  'syslog', 'stdout' or 'file'.
+$config['log_driver'] = 'syslog';
```
ganto: Would be nice to have this as a role variable, no?

ansible/roles/debops.roundcube/templates/srv/www/sites/config.inc.php.j2
```diff
@@ -19,19 +23,34 @@
...
+// ----------------------------------
+
 // Database connection string (DSN) for read+write operations
 {#// Format (compatible with PEAR MDB2): db_provider://user:password@host/database
-// Currently supported db_providers: mysql, pgsql, sqlite, mssql or sqlsrv
+// Currently supported db_providers: mysql, pgsql, sqlite, mssql, sqlsrv, oracle
 // For examples see http://pear.php.net/manual/en/package.database.mdb2.intro-dsn.php
-// NOTE: for SQLite use absolute path: 'sqlite:////full/path/to/sqlite.db?mode=0646' #}
+// NOTE: for SQLite use absolute path (Linux): 'sqlite:////full/path/to/sqlite.db?mode=0646'
+//       or (Windows): 'sqlite:///C:/full/path/to/sqlite.db' #}
```
ganto: No need to add windows example. Confusing...

ansible/roles/debops.roundcube/defaults/main.yml
```diff
@@ -245,21 +245,50 @@ roundcube__database_schema: '{{ roundcube__git_checkout + "/SQL/mysql.initial.sq
 roundcube__domain: [ 'roundcube.{{ ansible_domain }}' ]
 
                                                                    # ]]]
-# .. envvar:: roundcube__default_host [[[
+# .. envvar:: roundcube__imap_server [[[
 #
-# Mail host chosen to perform the log-in.
-roundcube__default_host: 'localhost'
+# The IMAP host chosen to perform the log-in.
+# This variable is also called `default_server` by Roundcube  .
```
ganto, line 251: It's called `default_host` in the Roundcube config, no?

ansible/roles/debops.roundcube/defaults/main.yml
```diff
@@ -245,21 +245,50 @@ roundcube__database_schema: '{{ roundcube__git_checkout + "/SQL/mysql.initial.sq
...
+# Leave blank to show a textbox at login, give a list of hosts
+# to display a pulldown menu or set one host as string.
+# To use SSL/TLS connection, enter hostname with prefix ssl:// or tls://
+# Supported replacement variables:
+# %n - hostname ($_SERVER['SERVER_NAME'])
+# %t - hostname without the first part
+# %d - domain (http hostname $_SERVER['HTTP_HOST'] without the first part)
+# %s - domain name after the '@' from e-mail address provided at login screen
+# For example %n = mail.domain.tld, %t = domain.tld
+roundcube__imap_server: 'localhost'
```
ganto, line 262: Why you rename the variable? I don't really like this breaking change for no obvious benefit.

ansible/roles/debops.roundcube/tasks/main.yml
```diff
@@ -51,6 +51,10 @@
   tags: [ 'role::roundcube:deployment' ]
 
 
+# ---- Skins/Themes ----
+- include: configure_skins.yml
```
ganto: Change to `include_tasks`

ansible/roles/debops.roundcube/tasks/configure_skins.yml
```diff
@@ -0,0 +1,30 @@
...
+      file:
+        path: '{{ roundcube__git_checkout }}/skins/{{ roundcube__skin_folder }}/{{ item.dest }}'
+        owner: '{{ roundcube__user }}'
+        group: '{{ roundcube__group }}'
+        mode: '0644'
+      loop: '{{ roundcube__fact_skin_elastic_css_files }}'
+
+  when: roundcube__skin_folder is defined
+        and roundcube__skin_folder == 'elastic'
+        and roundcube__git_checkout is defined
```
ganto, line 29: I think it's easier to read the code if you move the `when` condition to the `include` and remove the block here. The playbook output is also less noisy if the condition is false.

ansible/roles/debops.roundcube/defaults/main.yml
```diff
@@ -317,6 +352,31 @@ roundcube__default_plugins: [ 'archive', 'filesystem_attachments', 'jqueryui', '
...
+                                                                   # ]]]
+# .. envvar:: roundcube__skin_logo [[[
+#
+# Logo image replacement. Specifies location of the image as:
+#
+# - URL relative to the document root of this Roundcube installation
+# - full URL with http:// or https:// prefix
+# - URL relative to the current skin folder (when starts with a '/')
+#
+roundcube__skin_logo: 'null'
```
ganto, line 388: What does `null` mean?

ansible/roles/debops.roundcube/templates/srv/www/sites/config.inc.php.j2
```diff
@@ -79,18 +109,29 @@ $config['product_name'] = 'Roundcube Webmail';
...
+// ----------------------------------
+// USER PREFERENCES
+// ----------------------------------
+
 // skin name: folder from skins/
-$config['skin'] = 'larry';
-{% endif %}
+$config['skin'] = '{{ roundcube__skin_folder }}';
+
+// Logo image replacement. Specifies location of the image as:
```
ganto, line 125: No need to add all these comments. They are redundant with the documentation in the variable defaults. 

I can't see why I originally added them to the template, but they should at least be excluded with `{# ...  #}` to avoid that they end up in the generated config file.

ganto (changes requested):
Very nice update and extension. Thanks a lot for your effort. However I feel the breaking changes are not really necessary.

Somehow I missed the release of Roundcube 1.4.0. Looks like I need to give it a try.

ansible/roles/debops.roundcube/templates/srv/www/sites/config.inc.php.j2
reixd: My purpose was to have a template very similar to the oficial `config.inc.php`. Including comments.
Which actually help while developing the template...

What does says the benevolent dictator @drybjed ?

ansible/roles/debops.roundcube/templates/srv/www/sites/config.inc.php.j2
reixd: True, wilco

ansible/roles/debops.roundcube/defaults/main.yml
reixd: Yes it is. But it is kinda confusing. 
I am trying to make it easier to configure...

Any more opinions?

ansible/roles/debops.roundcube/defaults/main.yml
reixd: See https://github.com/debops/debops/pull/1144#discussion_r345631451

ansible/roles/debops.roundcube/tasks/main.yml
reixd: Good catch!
(https://docs.ansible.com/ansible/latest/user_guide/playbooks_reuse_includes.html)

ansible/roles/debops.roundcube/tasks/configure_skins.yml
reixd: My intention was to have a skin task file, so all special skins configurations in the future land here.

reixd: Any other opinions?

ansible/roles/debops.roundcube/defaults/main.yml
reixd: See https://github.com/roundcube/roundcubemail/blob/master/config/defaults.inc.php#L424

reixd:
Travis says: [NO](https://travis-ci.org/debops/debops/jobs/611271205?utm_medium=notification&utm_source=github_status)
locally I get all green

ansible/roles/debops.roundcube/templates/srv/www/sites/config.inc.php.j2
drybjed: I think that if these comments were already gone  from the template included in the role, adding them back doesn't have any useful porpose, so this should go.

cultcom:
Why do you prefer installing nodejs on a webserver in favor of using the official tarball ?

reixd:
> Why do you prefer installing nodejs on a webserver in favor of using the official tarball ?

Can you please be more specific with your question?

`debops.roundcube` installs roundcube from the official git repo, thus the necessity to compile the (new) CSS files.

cultcom:
> `debops.roundcube` installs roundcube from the official git repo, thus the necessity to compile the (new) CSS files.

yes, I am questioning this strategy. I think the complexity decreases when the role simply downloads the official tarball and drops dependencies to nodejs and git etc.

I just wanted to hear other opinions about that.

drybjed:
My view on that is that I want to be protected from third-party projects like Roundcube suddenly vanishing and me being unable to recreate the installation due to the lack of sources. If we have source code cloned somewhere, we can always in such case publish it in a reachable `git` repository so that role can grab that cloned version and install Roundcube again. This is the same strategy as Debian's to include all source code in the repositories, without binary-only packages or vendored code. Any code compilation on the production server, if it's needed, cannot be avoided. An alternative would be to install the application from upstream APT repository, and when that's not available, the user would have to make their own `.deb` package and put it in a local APT repository.

I'm fine with having binary-only installation from upstream (that would be the functional result in this case), as long as installation "from source" is still available as an option. Then this can be selected conditionally by the user via Ansible inventory and both sides should be satisfied.

Don't forget about ahtneticating the source code or binary installs using GPG keys.

PR #1148, opened 2019-11-12
[debops.nginx] Fix erroneous paths on docs

Erethon:
I checked the `etc/nginx/sites-available/default.conf.j2` template and as far as I can tell this seems to be a typo or a relic of the past, as `/srv/www/sites/` is used everywhere by default.


drybjed (approved):
Thanks for fixing the documentation! :+1:

PR #1152, opened 2019-11-15
[debops.gitlab] Check if deploy.html page file exists before upgrade

reixd:


drybjed:
Hmm, this change seems a bit weird to me... You first create the `public/` directory and then you check if the `public/index.html` page exists. But the `public/` directory is in the GitLab `git` repository itself, and the `public/index.html` page is not commited in the repository, so it won't be present.

Then, you deploy the `deploy.html` page if the `public/index.html` file exists, but it won't exist so the `deploy.html` page will not be deployed as `index.html`... So the second task will not be executed, unless you somehow create `public/index.html` file by something else... Did GitLab change the upgrade mechanism recently of which I might not be aware of, or something else happened?

I think that you added this code to solve an issue, but I'm not sure what could that be. Care to elaborate and explain the reasoning?

reixd:
> Did GitLab change the upgrade mechanism recently of which I might not be aware of, or something else happened?

Apparently... 

> I think that you added this code to solve an issue, but I'm not sure what could that be. Care to elaborate and explain the reasoning?

Without this changes I was unable to install gitlab 12.2 in a vanilla system (debops bootstrap ; debops) with almost default config and registry.

reixd:
@drybjed Does my last patch work for you?

PR #1156, opened 2019-11-17
bionic features dovecot 2.2.33

romor001:
Dovecot < 2.3 needs ssl_protocols instead of ssl_min_protocol in config.


drybjed (approved):
Nice work, thanks. :+1:

PR #1160, opened 2019-11-20
[debops.varnish] Add debops.varnish role (Work in progress)

LeonardoAndrei:
The goal is to create the possibility that the user can change the default template used.
It would be nice if the user could create his own config file based on the default one.

An idea would be using simultaneously parse_kv_items and template_src option.

If anyone has any suggestions, please share them with us.


alinalexandru:
With the latest changes I was able to use it in several environments.

PR #1161, opened 2019-11-21
[debops.postldap] New, improved role for Virtual Mail management with LDAP backend

drybjed:
Based on #1125


reixd:
Good work! +1

PR #1163, opened 2019-11-29
[debops.haproxy] Add debops.haproxy role

LeonardoAndrei:


ansible/roles/haproxy/defaults/main.yml
```diff
@@ -0,0 +1,205 @@
...
+# Copyright (C) 2020 DebOps <https://debops.org/>
+# SPDX-License-Identifier: GPL-3.0-or-later
+
+# debops.haproxy default variables
+# ================================
+
+# .. contents:: Sections
+#    :local:
+#
+# .. include:: ../../../../includes/global.rst
```
tasosalvas, line 15: This `global.rst` is something we're trying to move away from.

If you're not using any of the refs contain in there you can just remove it. Otherwise, just inline the refs you need.

docs/ansible/roles/haproxy/getting-started.rst
```diff
@@ -0,0 +1,42 @@
+Getting started
+===============
+
+.. contents::
+   :local:
+
+
```
tasosalvas, line 7: A __Default Configuration__ section should give a brief overview of the state the application will be in if deployed with defaults, especially:
* choices that were made that might differ from its out-of-the-box defaults
* things the end user _must_ pay attention to before deploying to production

PR #1167, opened 2019-11-30
Allow options to be specified for the nfs_server bind mount

thomasfinstad:
This change will allow the bind mount to be customized with options while still being backwards compatible.

This came up as a need for me when trying to use the bindmount on a box with zfs, when the box reboots the bindmount will try to run before zfs is done mounting as zfs is a more complex system and mount later in the boot process. Reference of the experienced problem: https://github.com/zfsonlinux/zfs/issues/971#issuecomment-325165200

### DIFF
--- a/docs/ansible/roles/debops.nfs_server/defaults-detailed.rst
+++ b/docs/ansible/roles/debops.nfs_server/defaults-detailed.rst
@@ -134,7 +134,7 @@ Anyone can access it, barring any firewall configuration:
         src: '/usr'
         options:
 
-          - x-systemd.requires=zfs-mount.service
+          - 'x-systemd.requires=zfs-mount.service'
 
        options: [ 'ro', 'no_subtree_check', 'async' ]
        acl: '*'
--- a/docs/ansible/roles/debops.nfs_server/defaults-detailed.rst
+++ b/docs/ansible/roles/debops.nfs_server/defaults-detailed.rst
@@ -43,6 +43,14 @@ Each list entry is a YAML dictionary with specific parameters:
   bind-mounted to the specified ``path``. This is useful when you want to
   export directories that are outside of the NFS4 root pseudo-filesystem.
 
+  ``src``
+    A string acting the same way as if you assigned the value directly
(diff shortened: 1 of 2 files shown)


docs/ansible/roles/debops.nfs_server/defaults-detailed.rst
```diff
@@ -112,3 +121,20 @@ networks, with different set of parameters:
...
+
+.. code-block:: yaml
+
+   nfs_server__exports:
+     - path: '/srv/nfs/usr'
+       bind:
+        src: '/usr'
+        options:
+
+          - x-systemd.requires=zfs-mount.service
```
drybjed: Please quote the value in the example, just to be on the safe side.

drybjed:
That's a very interesting approach, thanks! :-) Update the example and I think it will be good for merging.

thomasfinstad:
@drybjed Should be good to go now.

docs/ansible/roles/debops.nfs_server/defaults-detailed.rst
```diff
@@ -112,3 +121,20 @@ networks, with different set of parameters:
...
+root filesystem, but only after the ZFS service has started.
+Anyone can access it, barring any firewall configuration:
+
+.. code-block:: yaml
+
+   nfs_server__exports:
+     - path: '/srv/nfs/usr'
+       bind:
+        src: '/usr'
+        options:
```
drybjed: The `src` and `options` parameters need to be indented with 2 spaces, not 1. Otherwise, YAML will not be parsed correctly.

thomasfinstad: Did not have any such problem myself, will fix.

thomasfinstad: Done

drybjed (approved):

thomasfinstad:
@drybjed thanks for the guidance. :smile:

PR #1178, opened 2019-12-09
PostgreSQL replication

gaudenz:


drybjed:
Nicely done. Looks like the changes can be backported to the stable releases as well, the code doesn't seem to impact existing installations.

Could you change the examples to use the YAML code blocks? With something like:
```
.. code-block:: yaml
```
at the beginning. That way the syntax highlighting will work and documentation will look at bit better.

PR #1184, opened 2019-12-13
[debops.gitlab] Added variables for gitlab gravatar configuration

exciler:
As discussed in IRC yesterday, I added configuration variabled for gitlab gravatar service/feature.
This PR makes all three gravatar configuration entries as documented at https://docs.gitlab.com/ee/administration/libravatar.html configurable via ansible variables.

The default behaviour is not changed as the gitlab__gravatar_enabled variable defaults to True. Both URL variables have a default setting of '' which excludes them from the gitlab.yml file so that the gitlab defaults apply.

As it was mentioned in IRC yesterday: My commit can easily be applied to the current stable-1.2 branch if you should consider backporting this change.


drybjed (approved):
Looks good, thanks. :-)

PR #1188, opened 2019-12-18
Fixes and improvements

ypid:


drybjed:
Looks good, I think that all of these can be backported to the stable releases as well.

Could make a similar change from the `bootstrap.yml` playbook in the `bootstrap-ldap.yml` playbook as well? Both of these can be used in initial provisioning, so their case will be essentially the same with CD-based installation.

ypid:
Good idea with the bootstrap-ldap.yml. Done. Not sure why I overlooked that. Note that I have not tested the ldap one end to end.

I guess backporting is ok. You could leave the debops-contrib.firejail one out as this is the only feature here.

CHANGELOG.rst
```diff
@@ -623,6 +623,11 @@ General
   documentation have been fixed and now point to the correct source files on
   GitHub.
 
+- Reorder :file:`bootstrap.yml` Ansible playbook to also work for systems freshly
+  installed from CD. :ref:`debops.apt` needs to be run early to regenerate
+  :file:`/etc/apt/sources.list` which might still contain a now not functional
+  CD entry.
+
```
drybjed: Looks like this entry is way, way down the Changelog file, in a completely wrong release. It's late so I'll merge the PR tommorow; if you want you could move this entry to the correct place, I'd say in the Changed section. For that you probably woul have to rebase the commits on latest `master` as well. If not, I can do that tommorow.

CHANGELOG.rst
```diff
@@ -637,6 +642,10 @@ General
 - The ``dmz`` firewall configuration will use the ``dport`` parameter instead
   of ``port``, otherwise filtering rules will not work as expected.
 
+- The ``dmz`` firewall configuration will now not interpret the port as part of
+  a IPv6 address anymore. We now protect the IPv6 address by surrounding it by
+  ``[]``.
+
```
drybjed, line 657: This entry also needs to be moved to the correct place in the Changelog.

ypid:
Thanks for your careful review! Fixed.

drybjed (approved):

PR #1189, opened 2019-12-20
Fixes hashicorp consul detection

trly:
Fixes detection of consul installation in debops.unbound and debops.dnsmasq

### DIFF
--- a/ansible/roles/debops.dnsmasq/defaults/main.yml
+++ b/ansible/roles/debops.dnsmasq/defaults/main.yml
@@ -272,8 +272,10 @@ dnsmasq__default_configuration:
     raw: |
       server = /consul/127.0.0.1#8600
     state: '{{ "present"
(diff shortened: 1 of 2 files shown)


drybjed:
I don't think that's a good idea. The `debops.hashicorp` role only installs `consul` for you, you would still have to have a `consul` role which would presumably create the fact that the `debops.dnsmasq` and `debops.unbound` roles look for.

Besides, with the new `debops.golang` role I think that having `consul` installed via the `debops.hashicorp` role will be come obsolete. A new `debops.consul` role will most likely use the `debops.golang` role to install the binary, similarly to how `debops.minio` and `debops.lxd` do things now. Then your solution will have to be reverted for the roles to work correctly.

trly:
That makes perfect sense. I get why those conditionals are structured the way they are now.

Thanks!

PR #1193, opened 2019-12-27
[debops.roundcube] Add PostgreSQL support

imrejonk:
Thanks a lot for your recent `debops.roundcube` improvements, @drybjed! I'm really happy with the LDAP address book support. Let's see if I can contribute something as well :)

These changes add PostgreSQL support to the `debops.roundcube` role in
a backwards-compatible way. The configuration is similar to the
pre-existing MySQL support, with two noticeable changes:

1. The name of the default PostgreSQL database equals the name of the
   PostgreSQL role instead of 'roundcubemail'. This is because of the
   'samerole' requirement in pg_hba.conf.

2. The role now uses a database type to schema file mapping in order to
   import the correct initial database schema. The new mapping can be
   found in roundcube__database_schema_map.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -65,6 +65,9 @@ New DebOps roles
   configuration of Sieve filter scripts. The role will use the DNS SRV resource
   records to find the Sieve service host and port to use.
 
+- The role can now use PostgreSQL as a database backend. The database server
+  can be managed with the :ref:`debops.postgresql_server` role.
+
 Changed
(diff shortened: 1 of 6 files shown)


drybjed (approved):
Nicely done, thanks! The Roundcube role could be further improved by adding the database autodetection, but for that I think the db code would have to be redesigned entirely, which can be done later. Improving the db server roles, especially with proper facts, wouldn't hurt either. It will be done someday. :-)

PR #1196, opened 2019-12-31
[debops-contrib.dropbear-initramfs] Add settings to set DROPBEAR_OPTIONS

SuperSandro2000:


drybjed (approved):
Sounds good. Although I would put the various option detection in the template itself and give an option to override the detection with a variable if needed. But this can be done later, when the role is moved to the main "namespace"; this PR will serve as a reminder to keep the functionality.

SuperSandro2000:
I think it is easier this way as I didn't add all the options but only the ones that are generally useful and increase security. You can also already override the final options by just setting them (at least I hope that, I didn't explicitly test this).

ypid (commented):
LGTM. I also tested it end to end – The system stills boots with this PR (full disk encryption, unlocked with FDEunlock) :)

PR #1198, opened 2020-01-01
[debops.docker_server] Add option wether to install python virtualenv

SuperSandro2000:
I don't want to install the virtual python as I use the debian packages for docker-compose.


drybjed (approved):
Sounds good to me. The virtualenv could be disabled on OS-only installs, but then the playbooks that rely on it would be broken with non-existent interpreters. But that could be mitigated by the role as well. I'll look into it when I have some time.

SuperSandro2000:
Which things do we miss? `docker-compose`? Then we just need to set `docker_server__packages` to `docker-compose` if this option is activated. This would also install the `docker` python package.

PR #1200, opened 2020-01-02
[debops.docker_server] Replace deprecated graph option with data-root

SuperSandro2000:
I was debugging why my dockerd was not starting and found the following warning and fixed it.
```
level=warning msg="The \"graph\" config file option is deprecated. Please use \"data-root\" instead."
```

I am not sure what a proper migration path is. Suggestions are welcome!


drybjed (approved):
Sounds good, thanks. I wonder what was the process behind naming that option in the first place...

PR #1201, opened 2020-01-02
[debops-contrib.dropbear-initramfs] Add changelog for #1196

SuperSandro2000:
I am not sure on the wording. Improvements are welcome.

Ref: #1196

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -27,6 +27,11 @@ New DebOps roles
 - The :ref:`debops.lxd` role brings support for LXD on Debian hosts by building
   the Go binaries from source, without Snap installation.
 
+debops-contrib.dropbear-initramfs role
+''''''''''''''''''''''''''''''''''''''
+
+- Add support to set `DROPBEAR_OPTIONS`.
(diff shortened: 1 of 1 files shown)


CHANGELOG.rst
```diff
@@ -27,6 +27,11 @@ New DebOps roles
 - The :ref:`debops.lxd` role brings support for LXD on Debian hosts by building
   the Go binaries from source, without Snap installation.
 
+debops-contrib.dropbear-initramfs role
+''''''''''''''''''''''''''''''''''''''
+
+- Add support to set `DROPBEAR_OPTIONS`.
```
ypid, line 33: I think a DebOps user who does not review all commit diffs will not know what `DROPBEAR_OPTIONS` is. That is an implementation detail I would not include in the changelog. Maybe something like:

```RST
Add support to set dropbear daemon command line options. Refer to :envvar:`dropbear_initramfs__dropbear_options` for details.
```

ypid (changes requested):
Please see my suggestion.

PR #1204, opened 2020-01-02
[debops-contrib.firejail] Support python3

komachi:
Currently it fails with python3, but it's easy to fix.

Docs: https://docs.ansible.com/ansible/latest/user_guide/playbooks_python_version.html#dictionary-views


drybjed (approved):
Sounds good, thanks.

PR #1206, opened 2020-01-02
[debops.docker_server] Install docker-compose system package if virtualenv is deactivated

SuperSandro2000:
Follow up to #1198


drybjed:
The `docker-compose` APT package is installed automatically [when upstream install is disabled](https://github.com/debops/debops/blob/master/ansible/roles/debops.docker_server/defaults/main.yml#L127), therefore adding it separately is not needed. Also, `docker_server__packages` list is reserved for the usage via Ansible inventory, you shouldn't add anything to it in the defaults - otherwise that will be overwritten by the inventory.

SuperSandro2000:
Oh, I overlooked that.

PR #1208, opened 2020-01-03
Fix Postfix SMTP auth

reixd:
I found the missing config in Dovecot in order for Postfix to authenticate an user which is trying to send an email over SMTP.

See `/etc/dovecot/conf.d/10-master.conf.dpkg-divert`

The missing snippet is:
```
service auth {
  # Postfix smtp-auth
  unix_listener /var/spool/postfix/private/auth {
    group = postfix
    mode = 0660
    user = postfix
  }
}
```

Although I updated the debops.dovecot role to include this config (analogous to LMTP), ansible is not creating the correct template. @drybjed can you please take a look?

### DIFF
--- a/ansible/roles/debops.dovecot/defaults/main.yml
+++ b/ansible/roles/debops.dovecot/defaults/main.yml
@@ -502,14 +502,14 @@ dovecot_postfix_transport: 'lmtp:unix:private/dovecot-lmtp'
                                                                    # ]]]
 # .. envvar:: dovecot_auth_listeners [[[
 #
-# List of LMTP ``inet_listeners`` or ``unix_listeners`` which will be enabled.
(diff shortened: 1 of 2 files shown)


ansible/roles/debops.dovecot/defaults/main.yml
```diff
@@ -499,6 +499,27 @@ dovecot_lmtp_config_map:
...
+# Custom AUTH service configuration properties. See :ref:`dovecot_auth_config_map`
+# for more details.
+dovecot_auth_config_map:
+  service:
+    # Postfix smtp-auth socket.
+    unix_listener:
+      /var/spool/postfix/private/auth:
+        user: 'postfix'
+        group: 'postfix'
+        mode: 0600
```
SuperSandro2000, line 521: from IRC: missing quotes

PR #1214, opened 2020-01-08
Update default values of debops.gitlab_runner role to improve interaction with the debops.gitlab role

reixd:


ansible/roles/debops.gitlab_runner/defaults/main.yml
```diff
@@ -126,7 +126,7 @@ gitlab_runner__concurrent: '{{ ansible_processor_vcpus
 # .. envvar:: gitlab_runner__domain [[[
 #
 # The default domain used in different places of the role.
-gitlab_runner__domain: '{{ ansible_domain }}'
+gitlab_runner__domain: '{{ gitlab_domain|d(ansible_domain) }}'
```
drybjed: That will work only if the user has `gitlab_domain` set in the inventory. If you expected that this value can be set through the `debops.gitlab` role, it will work only if the `debops.gitlab` role is in the same playbook as `debops.gitlab_runner` role which is not the case in DebOps.

A better solution would be to expose the GitLab domain and FQDN via the Ansible local facts defined by the `debops.gitlab` role, then the `debops.gitlab_runner` role could use them in the above way to configure itself... However since the Runner(s) would probably be configured and used on separate hosts from the main GitLab instance for security and performance, that probably won't work. So instead, perhaps the role could check the DNS SRV resource records, say `_gitlab._tcp.example.org` to find the domain of the GitLab host and configure the Runners that way. Of course this requires setting up that DNS record first, but might be a useful feature.

ansible/roles/debops.gitlab_runner/defaults/main.yml
```diff
@@ -139,7 +139,7 @@ gitlab_runner__fqdn: '{{ ansible_fqdn }}'
 #
 # Fully Qualified Domain Name of the GitLab CI host which will control the
 # runner operation.
-gitlab_runner__api_fqdn: 'code.{{ gitlab_runner__domain }}'
+gitlab_runner__api_fqdn: '{{ gitlab__fqdn|d("code." + gitlab_runner__domain) }}'
```
drybjed: This is the same case as with `gitlab_domain` and won't work as you expect in the default configuration.

docs/ansible/roles/debops.gitlab_runner/getting-started.rst
```diff
@@ -40,7 +40,7 @@ For example storing the token in the :file:`ansible/secret/credentials` folder:
 .. code-block:: console
 
    gitlab_runner__token: '{{ lookup("password", secret
-                           + "/credentials/" + ansible_fqdn
+                           + "/credentials/" + gitlab_runner__api_fqdn
```
drybjed: This is actually a good catch, instead of putting token in multiple directories based on the Runner FQDN, put it in a directory based on the GitLab instance. Although the documentation should probably be updated to mention that as well. :+1:

ansible/roles/debops.gitlab_runner/defaults/main.yml
reixd: Based on your works for `_imap._tcp` and so on, updated the defaults accordingly.
Please check on your side if it still works ;)

CHANGELOG.rst
```diff
@@ -263,6 +263,16 @@ General
...
+:ref:`debops.gitlab_runner` role
+''''''''''''''''''''''''''''''''
+
+- Solve 'Build occasionally get "No such file or directory" accessing build directory' error.
+  See also https://gitlab.com/gitlab-org/gitlab-runner/issues/1379 .
+  Occasionally the runner stops with an error due to the conflicts created by `~/.bash_logout`
+  wipping the screen/terminal clean, thus confusing the runner. Therefore updated the
+  `debops.gitlab_runner` user creation to no add an user skeleton and just an empty home dir.
+  Additionally added task to clean up manually any `~/.bash_logout` files from current installations.
+
```
drybjed: This section should definitely be lower since the roles should be ordered alphabetically. Also check if it's in the "unreleased" section.

CHANGELOG.rst
```diff
@@ -472,6 +482,11 @@ Changed
 Updates of upstream application versions
 ''''''''''''''''''''''''''''''''''''''''
 
+- In the :ref:`debops.gitlab_runner` role, the example showing how to store the
+  `gitlab_runner__token` in the inventory has been updated. The example now shows
+  how to store the token inside the Gitlab server credentials
+  under :file:`ansible/secret/credentials/`.
+
```
drybjed: I think that this should be moved higher in the file, currently it's located in the `v1.2.0` release section.

ansible/roles/gitlab_runner/tasks/main.yml
```diff
@@ -25,6 +25,12 @@
     state: 'present'
     generate_ssh_key: '{{ gitlab_runner__ssh_generate | bool }}'
     ssh_key_bits: '{{ gitlab_runner__ssh_generate_bits }}'
+    skeleton: null
+
+- name: "Remove '{{ gitlab_runner__home }}/.bash_logout' to avoid conflicts with the shell runner"
+  file:
+    state: 'absent'
+    path: "{{ gitlab_runner__home }}/.bash_logout"
```
drybjed, line 33: Nicely done. :+1:

ansible/roles/gitlab/tasks/main.yml
```diff
@@ -112,4 +112,4 @@
     dest: '/etc/ansible/facts.d/gitlab.fact'
     owner: 'root'
     group: 'root'
-    mode: '0644'
+    mode: '0755'
```
drybjed: The `gitlab.fact` file is not a script at the moment, so this is unnecessary and might even break Ansible which will try to execute it blindly, I suspect.

drybjed (changes requested):
Nice work, although there are some modifications you should do. See the comments.

ansible/roles/gitlab_runner/tasks/main.yml
reixd: :)

drybjed (approved):
Thanks for the fixes. I'll squash the commits and fix small syntax issues on the way, since you did not sign these commits.

PR #1215, opened 2020-01-09
Rename DebOps roles, improve Ansible Collection support

drybjed:


SuperSandro2000:
Any reason why this is needed? Just want to understand the reasoning behind it.

drybjed:
Sure, although this requires a bit of explanation.

For the longest time, Ansible supported role distribution on Ansible Galaxy using the format:
```
<role_owner>.<role_name>
```
This notation worked in the playbooks and role dependencies. When the role repositories were imported from GitHub, Galaxy automatically converted the estabilished `ansible-<role>` repository name to `<role>`, and prefixed with the `<role_owner>` by itself. In DebOps, to allow use of roles from the repository and from Galaxy in the playbooks, the roles were named in that same format.

Then, Ansible Collections were created. But now, Ansible team decided that they would want to allow multiple collections for each "publisher" with multiple roles in each, so the format of the role name in the playbooks was changed to:
```
<collection_owner>.<collection_name>.<role_name>
```
Ansible additionally has code that can resolve the `<role_name>` to its exact Collection automatically, using the `collections` playbook keyword. This works on the playbook level, and in roles via the `import_role`/`include_role` modules, but cruically, Ansible Collections do not support the old `dependencies` definition in the `meta/main.yml` files - this is reserved for single roles, not Collections. When a Collection with dependencies defined like this is imported, the import fails.

This creates a problem for DebOps. Because roles are named in the `<role_owner>.<role_name>` format, the role dependencies used the same format, which broke full support for Ansible Collections. But if the role dependencies where changed to use only `<role_name>` format, that would be compatible with Ansible Collections (so, via the `import_role` module), but without renaming the role directories themselves, use of DebOps roles/playbooks from outside of the Ansible Galaxy installation would break.

To resolve that, I decided to drop the `debops.` prefix from the role names, which allows use of the `<role_name>` format in role dependencies and playbooks provided in DebOps. This also allows the roles installed via Ansible Collections to correctly resolve their dependencies in the custom playbooks that define the `debops.debops` collection, like this:
```yaml
---
- name: Playbook
  hosts: 'all'
  collections:
    - 'debops.debops'
```
I plan to add this configuration in the DebOps playbooks, although this breaks the project usage with Ansible 2.7.7 currently present in Debian Buster. Ansible 2.9.x is already available in Debian Bullseye, so I hope that soon a backported version will be available for Buster. Because of that I'll probably go ahead and add the `collections` keyword in the playbooks before DebOps v2.0.0 is released. At the same time I'll probably switch from using the `roles:` keyword to `import_role` tasks on the playbook level, which allows use of the `tasks_from` parameter - that in turn could replace the current mechanism of `<role_name>/env` "subroles" to prepare the environment used throughout DebOps.

Only the "live" code was changed, in the documentation the roles are still referred to as `debops.<role>`. This might be updated in the future, although I wonder if use of the same name for a role and the software that it manages (think `debops.nginx` vs `nginx`) would lead to confusion or not, if it's not explicitly mentioned what we are talking about. I guess we will see.

SuperSandro2000:
Thanks for the long text!

Also apt pinning could be an option to backport ansible or pip install.

I also think this is a bigger change that should be under 2.0.0.

drybjed:
I plan to release DebOps v2.0.0 near the end of January, so this was a good moment to do the rename - there's still some time to catch any bugs that show up, and test things like backporting changes to older releases.

In case you haven't seen it, this was mentioned in the [DebOps v1.2.0 release announcement](https://lists.debops.org/pipermail/debops-users/2019-December/000247.html).

As for Ansible - sure, users can install Ansible 2.9 on Debian Buster via `pip3` or build the `.deb` package from GitHub sources, but I like to keep to the release available in Debian Stable as long as possible for convenience. But looks like it's time to move on, especially that Ansible 2.9 has been released some time ago.

SuperSandro2000:
> build the .deb package from GitHub sources

if it is the same as the one on the ppa it is py2 only and this is kinda what you want.

Apt pinning is probably the best option everyone should take.

Also mailing lists are so '90. You should write such important notes into the release notes and nit just link them. I see myself not reading this and getting frustrated and pushing the PR back to later.

drybjed:
> Also mailing lists are so '90.

I like mailing lists - the one thing I check constantly is various mail accounts here and there. With mailing lists, interesting messages come to me, I don't have to look on a bunch of different websites and forums to keep up with threads. Also, messages on the mailing list are signed by my GPG key (I'm using `mutt` to send them), so the signatures can be easily verified. It probably could be done easily in a forum post, but that would be more hassle with putting the text of the message in a text editor, signing it manually, etc. which is pretty much how I write e-mails.

Perhaps when Mailman3 role with a better web interface is written, this would be easier to stomach - Mailman3 supports online posting and has better web UI.

> You should write such important notes into the release notes and nit just link them. I see myself not reading this and getting frustrated and pushing the PR back to later.

Currently there are three places worth keeping track of with various levels of importance in DebOps:

- The `CHANGELOG.rst` file in the root of the DebOps monorepo is used to write notes about visible changes in the probject that can be of general interest for the users - new roles, large scale changes in existing roles, modifications in role behaviour that might require user attention, etc.

- The upgrade notes in `docs/news/upgrades.rst` file in the DebOps monorepo contain detailed instructions on what has to be changed either in the Ansible inventory or on the remote hosts to make the roles work correctly. Sometimes role default variables are renamed, or their meaning/usage changes, or there are changes on the remote hosts that might require manual intervention by the user - this is the place you will find them listed from release to release. Such changes will not be backported to stable releases.

- The release notes published on the project mailing list when a new stable DebOps release is created. These are meant to give a high level overview of what changed in the project from release to release and what will be the focus for the next development cycle. These release notes are totally optional and shouldn't be a required reading for DebOps users.

Is that a good way to manage this kind of project? I don't know, I'm just trying things out as I go. One thing I keep in mind is to provide a well known "interface" for the Debian users which might be used to the project mailing lists and bug tracker. Having a forum requires forum administrators, moderators, anit-spam measures... A mailing list seems to me to be an easier thing to manage.

As for writing the notes on the release notes page on GitHub - I did that once, but at the time I just copied the Changelog entries, which were written in reStructuredText, to the release notes which supported Markdown, and they looked horrible. Perhaps with a bit of tweaking I can add the notes I post on the mailing list on the release notes page as well; I'll keep that in mind when DebOps v2.0.0 is released near the end of January, and we will see how it looks.

SuperSandro2000:
Now I am confused. Should I read all three things or just the upgrade notes? Also I kinda want to know what else changed so I also read the changelog. In my opinion they could be combined and update notes can just be in the first lines of release.

drybjed:
I suppose that I'm complicating this a bit. In short:

- Read the Changelog
- Keep an eye on the upgrade notes, you might see variables that you set in the inventory that need to be updated. Changelog mentions these cases as well, with the link to upgrade notes.
- Release notes on the mailing list are optional, in case you want to know what I'll be focusing on next.

ypid:
I just noticed that this change breaks all external links to our roles. @drybjed Can you maybe configure some clever Nginx directive in the reverse proxy that redirects to the new URL?

Keep up the good work!

drybjed:
@ypid Good idea, done.

PR #1221, opened 2020-01-15
Allow to enable or disable the Gitlab backup procedure

reixd:
For example in order to save space under `/var/backups`.

### DIFF
--- a/ansible/roles/gitlab/tasks/configure_gitlab_ce.yml
+++ b/ansible/roles/gitlab/tasks/configure_gitlab_ce.yml
@@ -271,6 +271,9 @@
   cron:
     name: 'Create GitLab backup'
     special_time: '{{ gitlab_backup_frequency }}'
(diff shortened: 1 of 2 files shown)


ansible/roles/gitlab/tasks/configure_gitlab_ce.yml
```diff
@@ -271,6 +271,9 @@
   cron:
     name: 'Create GitLab backup'
     special_time: '{{ gitlab_backup_frequency }}'
+    disabled: '{{ False
+                  if (gitlab_backup_enabled|d(True))
```
drybjed: The `|d(True)` here is meaningless because the `gitlab_backup_enabled` variable is a default variable which will always be defined in the role context. I would change this to `|bool` instead to check for the boolean value.

drybjed (approved):
Nicely done, thanks.

PR #1225, opened 2020-01-16
[rsyslog] Fix systemd-tmpfiles permissions for /var/log

gaudenz:
The systemd package ships a systemd-tmpfiles configuration which resets
the permissions of /var/log on every boot. This adds a configuration
override to prevent this if rsyslog__unprivileged is set on Debian.

### DIFF
--- a/ansible/roles/rsyslog/tasks/main.yml
+++ b/ansible/roles/rsyslog/tasks/main.yml
@@ -58,6 +58,15 @@
   changed_when: rsyslog__register_file_permissions.stdout|d()
   notify: [ 'Restart rsyslogd' ]
 
+- name: Create systemd-tmpfiles override
+  copy:
+    dest: '/etc/tmpfiles.d/rsyslog-var-log.conf'
+    owner: 'root'
+    group: 'root'
+    mode: '0755'
+    content: 'z {{ rsyslog__home }} 0775 root {{ rsyslog__group }} -'
+  when: ansible_service_mgr == "systemd" and rsyslog__unprivileged|bool and ansible_distribution == "Debian"
+
 - name: Divert main rsyslog configuration
   command: dpkg-divert --quiet --local --divert /etc/rsyslog.conf.dpkg-divert --rename /etc/rsyslog.conf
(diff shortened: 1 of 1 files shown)


SuperSandro2000:
Cant you mask or disable the offending systemd file?

gaudenz:
The offending setting is in `/usr/lib/tmpfiles.d/var.conf`. This is a generic file from the systemd package which configures various files and directories in `/var`. So we can't just mask the whole file. I don't know a better way to override the permission setting just for this particular directory.

ansible/roles/rsyslog/tasks/main.yml
```diff
@@ -58,6 +58,15 @@
...
   notify: [ 'Restart rsyslogd' ]
 
+- name: Create systemd-tmpfiles override
+  copy:
+    dest: '/etc/tmpfiles.d/rsyslog-var-log.conf'
+    owner: 'root'
+    group: 'root'
+    mode: '0755'
+    content: 'z {{ rsyslog__home }} 0775 root {{ rsyslog__group }} -'
+  when: rsyslog__unprivileged|bool and ansible_distribution == "Debian"
```
drybjed: Could you add a condition to check if the `systemd` is used on the host? Otherwise `/etc/tmpfiles.d/` directory might not be present and the role run will fail. For example:
```
when: ansible_service_mgr == "systemd" and ...
```

gaudenz: Done as suggested. Did not know that non systemd systems are supported by debops.

drybjed: Thanks! Well, perhaps not fully supported, but I'm not enforcing `systemd`-only policy here. If there are users that want to use DebOps with Devuan, for example, and they provide PRs to improve support for it, I don't mind as long as it doesn't interfere with officially supported OSes (Debian, Ubuntu).

gaudenz:
I have now pushed the change which I actually already wanted to push on Friday. Sorry for that.

PR #1226, opened 2020-01-16
[icinga_web] Update Icinga Director to v1.7.2

imrejonk:
The Icinga Director version installed by the :ref:`debops.icinga_web`
role has been updated to the v1.7.2 release. Notable changes in v1.7.x
are new German and Japanese translations, side-by-side sync previews, a
new background daemon to replace the job runner and new module
dependencies. Other Icinga Web modules have also been updated to their
latest versions.

The `icinga2-director-jobs.service` systemd service has been replaced
with `icinga-director.service`. This service manages a new daemon that
is required for Icinga Director v1.7.0+.


drybjed (approved):
Awesome work, thanks! I tested the changes on an existing installation, upgrade seems to work fine and the database migration can be done afterwards via the UI. :+1:

PR #1230, opened 2020-01-18
Minor fixes

ypid:


ansible/roles/etckeeper/defaults/main.yml
```diff
@@ -176,6 +176,9 @@ etckeeper__default_gitignore:
   - name: 'zfs-zpool-cache'
     ignore: 'zfs/zpool.cache'
 
+  - name: 'etesync-secret'
+    ignore: 'etesync-server/secret.txt'
```
drybjed, line 180: That's a bit random, isn't it? It's the first time I hear about [EteSync](https://www.etesync.com), it's not even available in Debian. I don't see an RFP or ITP messages about it either. There's no efforts to create a `.deb` package in its GitHub repositories that I could see on a quick glance, looking for an Ansible playbook/role for it also doesn't show any results, nothing on Ansible Galaxy...

The `etckeeper` role is designed to handle such cases via Ansible inventory perfectly fine, so why add this particular case to the role defaults?

ansible/roles/python/templates/etc/ansible/facts.d/python.fact.j2
```diff
@@ -29,4 +31,29 @@ output = loads("""{{ ({'enabled':  python__enabled | bool,
...
+        output['version3'] = subprocess.check_output([
+            'python3', '-c',
+            textwrap.dedent("""
+            import platform
+            print(platform.python_version(), end="")
+            """)
+        ])
+    except subprocess.CalledProcessError:
+        pass
+
```
drybjed, line 54: It's a good idea to add the available Python version to the facts, however it's kind of redundand since Ansible does this in its "internal" facts as well, although only for the interpreter. However, why did you go with such an elaborate way to do this? Why not just go with parsing the output of `python3 --version` and `python2 --version` instead? I'm curious because this script will be executed frequently during Ansible runs, so I'm not sure which method would be faster.

drybjed (commented):
Thanks for your improvements! However there are some questions I would like you to answer first. See the comments below.

ansible/roles/etckeeper/defaults/main.yml
ypid: I guess because I have a DebOps role 80 % done for a EteSync server and adding it to etckeeper ignores appeared as one usual preparation work for me :)

drybjed: Very well then, it can stay. :-)

ansible/roles/python/templates/etc/ansible/facts.d/python.fact.j2
ypid: Because `platform.python_version` is the API driven way to do it and I expect it to be more stable/reliable. I did some testing and it does not make much of a difference but I changed it anyway.

One of those version checks could be switched to Ansible "internal" facts. But not both. For my use case, it did not work obviously thus I added it to the local facts. I still have Python2 as `interpreter_python` but I want to run EteSync (for which this Python version check is for) under Python3.

drybjed (approved):

PR #1231, opened 2020-01-19
[icinga] Also listen on IPv6 addresses

imrejonk:
The Icinga REST API included with the Icinga 2 version in Debian Buster
listens only on IPv4 addresses by default. This commit adds a role
variable to configure the listen address, which defaults to '::' (all
IPv4 and IPv6 addresses).


drybjed (approved):
Sounds good, thanks!

PR #1232, opened 2020-01-19
[docker_registry] Support simple authentication schema w/o token auth

ypid:
This might be a bit controversial. Let me know what you think.


drybjed (approved):
Sounds good to me. Since this is disabled by default, I'll backport the change to stable releases as well.

PR #1234, opened 2020-01-20
[owncloud] Upgrade to 10.3 version

pedroluislopez:
Upgrade owncloud to 10.3 version

### DIFF
--- a/ansible/roles/owncloud/tasks/setup_owncloud.yml
+++ b/ansible/roles/owncloud/tasks/setup_owncloud.yml
@@ -153,7 +153,7 @@
 
 # Configure cron job for background tasks [[[
 
-- name: Setup cron service
+- name: Setup cron service (nextcloud)
   cron:
     name: 'ownCloud Background Jobs'
     minute: '{{ owncloud__cron_minute }}'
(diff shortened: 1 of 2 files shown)


ansible/roles/owncloud/tasks/setup_owncloud.yml
```diff
@@ -158,8 +158,8 @@
     name: 'ownCloud Background Jobs'
     minute: '{{ owncloud__cron_minute }}'
     user: '{{ owncloud__app_user }}'
-    job: 'test -x /usr/bin/php && test -e {{ (owncloud__deploy_path + "/cron.php") | quote }}
-          && /usr/bin/php -f {{ (owncloud__deploy_path + "/cron.php") | quote }}'
+    job: 'test -x /usr/bin/php && test -e {{ (owncloud__deploy_path + "/occ") | quote }}
+          && /usr/bin/php {{ (owncloud__deploy_path + "/occ") | quote }} system:cron'
     cron_file: 'owncloud'
```
ypid, line 173: This does not work for Nextcloud. Can you make this change conditional based on if Nextcloud or ownCloud is used?

Ref: https://docs.nextcloud.com/server/17/admin_manual/configuration_server/background_jobs_configuration.html

ypid (changes requested):
Please do not break Nextcloud support :)

ansible/roles/owncloud/tasks/setup_owncloud.yml
pedroluislopez: Thank you. Done

ypid (approved):
LGTM. The ownCloud variant has not received much attention over the last years. Out of curiosity, why do you use ownCloud instead of Nextcloud? The plan was to drop ownCloud support some day.

pedroluislopez:
Because i have a old instance of owncloud and now I don't have time to migrate to nextcloud. But i need to maintain this and i needes this upgrade.

Maybe in the future i will migrate to nextcloud.

drybjed (approved):
Nicely done, thanks.

PR #1236, opened 2020-01-20
Improve Chrony support

ypid:


ansible/roles/ntp/templates/etc/chrony/chrony.conf.j2
```diff
@@ -21,6 +21,9 @@ bindaddress {{ address }}
 # Only listen on localhost for command connections
 bindcmdaddress 127.0.0.1
 
+# Do not listen on udp control port on loopback. chronyc can still talk to chronyd via Unix sockets.
+cmdport 0
+
```
SuperSandro2000, line 26: Can we add an option to turn that back on, please?

ypid: I couldn't think of a use case where this would be needed. Even for Monitoring, you don’t need it. What do you have in mind?

SuperSandro2000: Monitoring over network? Maybe healthchecks over tcp.

ypid: For performance reasons, I assume proper monitoring systems would do that locally (CheckMK does) and then ship all results. Checks of the service itself work anyway because NTP is a network service. Only additional metrics need to be pulled from this port. And even when I now make it configurable, the ferm Firewall also does not allow this access. I would leave it as is and just see if there is a need. It is not difficult to add that and as I mentioned, I am not breaking anything. The scenario you have in mind was never supported in the first place.

drybjed: @SuperSandro2000 What's your decision on this?

SuperSandro2000: @drybjed It wasn't my intention to block this PR with that but just a thought which we might want to consider. If we don't break anything I guess this is fine.

drybjed (approved):
Looks good, thanks. I'll backport the changes to the stable releases as well.

ansible/roles/ntp/templates/etc/chrony/chrony.conf.j2
DouglasHeriot: FYI a use for this setting is using [telegraf](https://github.com/influxdata/telegraf) to monitor chrony. When `telegraf` runs as its own user, it cannot read from the `chrony` UNIX socket. (chrony has no option to modify its permissions - it is only for root and the *user* chrony, no the *group* chrony)

It should be safe in many circumstances to allow `chrony` to listen for commands on the localhost address. It only allows read-only commands. From the [`chronyc` man page](https://chrony.tuxfamily.org/doc/3.5/chronyc.html):
> Only the following monitoring commands, which do not affect the behaviour of chronyd, are allowed from the network

So right now I'm looking into either:
* [Adding sudo support into telegraf](https://github.com/influxdata/telegraf/pull/7669) when running `chronyc`
* Asking the chrony project if they would add a setting to allow the *group* `chrony` to write to the control socket
* Or @drybjed would you accept a pull request that added a variable to set `cmdport` to something other than 0?

drybjed: @DouglasHeriot Sure, you can create a pull reuquest that exposes this parameter as a variable.

PR #1237, opened 2020-01-21
[role.yadm] install the upstream version for xenial

### LINKED ISSUE #1235: [debops.yadm] Detect the installed yadm version
I discovered that Ubuntu 16 comes with yadm 1.03 and the bootstrap option has been added in 1.07.
https://github.com/TheLocehiliosan/yadm/releases/tag/1.07
I propose: 
- either to add xenial into the `yadm__upstream_enabled`. This is the easiest one.
- either to detect the version and compare with 1.07 (or 1.12.0 as it's in `yadm__upstream_version`). If it's lower use upstream, otherwise use the package from OS. The advantage here is that it doesn't depend on the OS version.

@drybjed What do you think is the best way?

alinalexandru:
Fix #1235

### DIFF
--- a/ansible/roles/yadm/defaults/main.yml
+++ b/ansible/roles/yadm/defaults/main.yml
@@ -115,7 +115,7 @@ yadm__host_packages: []
 # :command:`git` repository.
 yadm__upstream_enabled: '{{ True
                             if (ansible_distribution_release in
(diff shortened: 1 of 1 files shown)


drybjed (approved):
Thanks for the fix. :-)

Issue #1233, opened 2020-01-19
RFC: Support subpath/subdir hosting for additional security


ypid:
With the last three services that I deployed (searx, Vaultwarden and EteSync) I started thinking/worrying because I was required to expose those services on the open internet and with that basically everybody could talk to their API. The services are easy to find because they obviously use HTTPS (think about CT logs). Now this exposes quite a bit of attack surface. For other services, I would have reduced that by client certificates, firewalling or VPN. All of this was not feasible here so I came up with misusing the "sub dir/path deployment schema" (ref: https://docs.nextcloud.com/server/15/admin_manual/installation/nginx.html#nextcloud-in-a-subdir-of-nginx) for this.

The idea is to host services on https://etesync.example.org/JZ27jIoxLKxB6mu09eirMPc instead of just https://etesync.example.org/. It depends on the application of how much this is distracting for the user but for Vaultwarden and EteSync it works fine because during normal usage, the URLs are not visible to the end user.

~~Also for other services like Nextcloud this could be done. Maybe with a shorter "PSK" because there it will be visible to the end user.~~ After looking into it I consider it not worth the hassle because for Nextcloud, I would need to break the `/.well-known` URIs. Without breaking those URIs, the access control would be useless. Ref: https://docs.nextcloud.com/server/21/admin_manual/installation/nginx.html#nextcloud-in-a-subdir-of-the-nginx-webroot

~~Maybe even https://etesync.example.org/JZ27jIoxLKxB6mu09eirMPc_psk could be used to give a hint what it is about.~~ Decided against this. The PSK acronym probably does not mean much to the average user.

The reason I think we could do this is that nowadays, bookmarks and browser auto complete is so common that what in the old days used to be done by humans is now offloaded to machines thus the URL does not have to be remembered for my taste. It is still in the estimation of the admin to weight up security vs. usability.

Implementation would be done by Nginx so it is very easy to audit. The application only gets the request if the requester knows the PSK.

Example implementation:

```YAML
# .. envvar:: etesync__http_subpath [[[
#
# Deploy EteSync on a random subpath that acts as a protection of the web app
# from people not knowing this PSK.
etesync__http_subpath: '{{ lookup("password", secret + "/etesync/" +
                           inventory_hostname + "/config/subpath chars=ascii_letters,digits length=23") }}'

                                                                   # ]]]

# .. envvar:: etesync__nginx__dependent_servers [[[
#
# Server configuration for the :ref:`debops.nginx` Ansible role.
[...]

PR #1242, opened 2020-01-25
Add 'etesync' Ansible role

ypid:
The first new DebOps role of the decade :)

Project: https://www.etesync.com/

WIP: @drybjed I am not sure if the integration test is ok. I have not yet played with the current testing approach enough.

Implements: #1233 
Updates: #1230

Reasons to update #1230:

* The etckeeper role should not get bloated by each and every application that could possibly write secret files to /etc.
* Users might not use the DebOps etckeeper role and instead only use the application roles and would then not have secret files excluded from Etckeeper if they still use it.

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -511,6 +511,15 @@ stages:
     JANE_DIFF_PATTERN: '.*/roles/etherpad/.*'
     JANE_LOG_PATTERN: '\[debops\.etherpad\]'
 
+'etesync role':
+  <<: *test_role_1st_deps
+  variables:
+    JANE_TEST_FACT: 'etesync.fact'
+    JANE_TEST_PLAY: '${DEBOPS_PLAYBOOKS}/service/etesync.yml'
+    JANE_INVENTORY_GROUPS: 'debops_service_etesync'
+    JANE_DIFF_PATTERN: '.*/roles/etesync/.*'
+    JANE_LOG_PATTERN: '\[debops\.etesync\]'
+
 
 # --- f --- [[[2
 
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -512,7 +512,7 @@ stages:
     JANE_LOG_PATTERN: '\[debops\.etherpad\]'
 
 'etesync role':
-  <<: *test_role_1st_deps
+  <<: *test_role_2nd_deps
   variables:
     JANE_TEST_FACT: 'etesync.fact'
     JANE_TEST_PLAY: '${DEBOPS_PLAYBOOKS}/service/etesync.yml'
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -27,7 +27,11 @@ New DebOps roles
 - The :ref:`debops.lxd` role brings support for LXD on Debian hosts by building
   the Go binaries from source, without Snap installation.
 
-- The :ref:`debops.etesync` role allows to setup a EteSync server.
+- The :ref:`debops.etesync` role allows to setup a EteSync__ server.
+  EteSync is a cross-platform project to provide secure, end-to-end encrypted,
+  and privacy respecting sync for your contacts, calendars and tasks.
+
+.. __: https://www.etesync.com/
 
 General
 '''''''
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -27,6 +27,8 @@ New DebOps roles
 - The :ref:`debops.lxd` role brings support for LXD on Debian hosts by building
   the Go binaries from source, without Snap installation.
 
+- The :ref:`debops.etesync` role allows to setup a EteSync server.
+
 General
 '''''''
 
--- a/ansible/roles/etckeeper/defaults/main.yml
+++ b/ansible/roles/etckeeper/defaults/main.yml
@@ -176,9 +176,6 @@ etckeeper__default_gitignore:
   - name: 'zfs-zpool-cache'
     ignore: 'zfs/zpool.cache'
 
-  - name: 'etesync-secret'
-    ignore: 'etesync-server/secret.txt'
-
                                                                    # ]]]
 # .. envvar:: etckeeper__gitignore [[[
 #
--- a/ansible/roles/etesync/defaults/main.yml
+++ b/ansible/roles/etesync/defaults/main.yml
@@ -1,0 +1,456 @@
+---
+# .. vim: foldmarker=[[[,]]]:foldmethod=marker
+
+# .. _etesync__ref_defaults:
+
+# debops.etesync default variables [[[
+# ===================================
+
+# .. contents:: Sections
+#    :local:
+#
+# .. include:: ../../../../includes/global.rst
+
+
+# Domain name configuration [[[
+# -----------------------------
+
+# .. envvar:: etesync__domain [[[
+#
+# The DNS domain used by other variables in the ``debops.etesync`` role.
+etesync__domain: '{{ ansible_local.core.domain
+                     if (ansible_local|d() and ansible_local.core|d() and
+                         ansible_local.core.domain|d())
+                     else ansible_domain }}'
+
+                                                                   # ]]]
+# .. envvar:: etesync__fqdn [[[
+#
+# String of the Fully Qualified domain names on which the EteSync application
+# will be available, used by the webserver.
+etesync__fqdn: 'etesync.{{ etesync__domain }}'
+
+                                                                   # ]]]
+                                                                   # ]]]
+# APT packages [[[
+# ----------------
+
+# .. envvar:: etesync__base_packages [[[
+#
+# List of APT packages which are required by the EteSync server.
+etesync__base_packages:
+  - 'git'
+                                                                   # ]]]
+                                                                   #
+# .. envvar:: etesync__packages [[[
+#
+# List of additional APT packages to install with EteSync.
+etesync__packages: []
+                                                                   # ]]]
+                                                                   # ]]]
+# Application user, group, home [[[
+# ---------------------------------
+
+# .. envvar:: etesync__user [[[
+#
+# Name of the UNIX system account used to manage EteSync.
+etesync__user: 'etesync'
+
+                                                                   # ]]]
+# .. envvar:: etesync__group [[[
+#
+# Name of the UNIX primary group used to manage EteSync.
+etesync__group: 'etesync'
+
+                                                                   # ]]]
+# .. envvar:: etesync__gecos [[[
+#
+# Contents of the GECOS field set for the EteSync account.
+etesync__gecos: 'EteSync'
+
+                                                                   # ]]]
+# .. envvar:: etesync__shell [[[
+#
+# The default shell set on the EteSync account.
+etesync__shell: '/usr/sbin/nologin'
+                                                                   # ]]]
+                                                                   # ]]]
+# Directory paths [[[
+# -------------------
+
+# .. envvar:: etesync__home [[[
+#
+# The EteSync account home directory.
+etesync__home: '{{ (ansible_local.root.home
+                   if (ansible_local|d() and ansible_local.root|d() and
+                       ansible_local.root.home|d())
+                   else "/var/local") + "/" + etesync__user }}'
+
+                                                                   # ]]]
+# .. envvar:: etesync__etc [[[
+#
+# Directory where the role stores EteSync configuration.
+etesync__etc: '/etc/etesync-server'
+
+                                                                   # ]]]
+# .. envvar:: etesync__src [[[
+#
+# Directory where the role stores EteSync version control sources.
+etesync__src: '{{ (ansible_local.root.src
+                  if (ansible_local|d() and ansible_local.root|d() and
+                      ansible_local.root.src|d())
+                  else "/usr/local/src") + "/" + etesync__user }}'
+
+                                                                   # ]]]
+# .. envvar:: etesync__lib [[[
+#
+# Directory where the EteSync server directory is located.
+etesync__lib: '{{ (ansible_local.root.lib
+                  if (ansible_local|d() and ansible_local.root|d() and
+                      ansible_local.root.lib|d())
+                  else "/usr/local/lib") + "/" + etesync__user }}'
+
+                                                                   # ]]]
+# .. envvar:: etesync__data [[[
+#
+# Directory where EteSync data is stored.
+etesync__data: '{{ (ansible_local.root.data
+                   if (ansible_local|d() and ansible_local.root|d() and
+                       ansible_local.root.data|d())
+                   else "/srv") + "/" + etesync__user }}'
+                                                                   # ]]]
+                                                                   # ]]]
+# Application sources and deployment [[[
+# --------------------------------------
+
+# .. envvar:: etesync__git_gpg_key_id [[[
+#
+# The GPG ID of the key used for signing EteSync releases.
+etesync__git_gpg_key_id: '9E21 F091 FC39 5F36 6A47  43E2 D2E5 84C3 7C47 7933'
+
+                                                                   # ]]]
+# .. envvar:: etesync__git_repo [[[
+#
+# The URI of the EteSync :command:`git` source repository.
+etesync__git_repo: 'https://github.com/etesync/server.git'
+
+                                                                   # ]]]
+# .. envvar:: etesync__git_version [[[
+#
+# The :command:`git` branch or tag which will be installed.
+# git commit hash lock to release 0.2.1.
+# Note that this hash locking is not very effective because the main
+# implementation of EteSync is in additional Python packages.
+etesync__git_version: 'cf0dc8e6a81f0b87bfba6d952ce9085d21257c18'
+
+                                                                   # ]]]
+# .. envvar:: etesync__git_dest [[[
+#
+# Path where the :command:`git` source bare repository will be stored.
+etesync__git_dest: '{{ etesync__src + "/" + etesync__git_repo.split("://")[1] }}'
+
+                                                                   # ]]]
+# .. envvar:: etesync__git_checkout [[[
+#
+# Path where EteSync sources will be checked out (installation path).
+etesync__git_checkout: '{{ etesync__lib + "/app" }}'
+                                                                   # ]]]
+                                                                   # ]]]
+# Python virtualenv configuration [[[
+# -----------------------------------
+
+# .. envvar:: etesync__virtualenv [[[
+#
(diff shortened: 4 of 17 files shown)


ansible/roles/etesync/defaults/main.yml
```diff
@@ -0,0 +1,456 @@
...
+#
+# Configuration for the :ref:`debops.python` Ansible role.
+etesync__python__dependent_packages3:
+
+  - 'python3-setproctitle'
+
+  - 'python3-dev'
+
+  ## Django>=2.1.7,<2.1.999 is required which is currently only in Debian sid.
+  # - 'python3-django'
```
SuperSandro2000, line 397: Apt pinning?

ypid: Please refer to https://docs.debops.org/en/master/dep/dep-0003.html Debian sid is not enabled on production servers.

SuperSandro2000: oh yeah. Also DEP-11 is probably ambiguous with debian appstream.

CHANGELOG.rst
```diff
@@ -27,6 +27,8 @@ New DebOps roles
 - The :ref:`debops.lxd` role brings support for LXD on Debian hosts by building
   the Go binaries from source, without Snap installation.
 
+- The :ref:`debops.etesync` role allows to setup a EteSync server.
```
drybjed: Perhaps you could add a sentence or two about what EteSync server can be used for because I'm not sure if it's essential or not... :-) An anonymous link to the website here wouldn't hurt either - this will be a first place new users will hear about it.

.gitlab-ci.yml
```diff
@@ -511,6 +511,15 @@ stages:
     JANE_DIFF_PATTERN: '.*/roles/etherpad/.*'
     JANE_LOG_PATTERN: '\[debops\.etherpad\]'
 
+'etesync role':
+  <<: *test_role_1st_deps
```
drybjed: The `nginx` role already is on the first level of dependencies, so this should be on the second level, `test_role_2nd_deps`.

ansible/roles/etckeeper/defaults/main.yml
```diff
@@ -176,9 +176,6 @@ etckeeper__default_gitignore:
   - name: 'zfs-zpool-cache'
     ignore: 'zfs/zpool.cache'
 
-  - name: 'etesync-secret'
-    ignore: 'etesync-server/secret.txt'
```
drybjed, line 180: At the moment this change is bundled with a change in the `etesync` role. This will be hard to backport cleanly to older DebOps releases. But I suppose that I can just do a separate revert of the https://github.com/debops/debops/commit/c6b19c45fd18523186a6ac3fac5b1da8481839f2 commits.

ansible/roles/etesync/defaults/main.yml
```diff
@@ -0,0 +1,456 @@
...
+etesync__git_repo: 'https://github.com/etesync/server.git'
+
+                                                                   # ]]]
+# .. envvar:: etesync__git_version [[[
+#
+# The :command:`git` branch or tag which will be installed.
+# git commit hash lock to release 0.2.1.
+# Note that this hash locking is not very effective because the main
+# implementation of EteSync is in additional Python packages.
+etesync__git_version: 'cf0dc8e6a81f0b87bfba6d952ce9085d21257c18'
```
drybjed: Perhaps you should contact upstream author and ask him to sign tags instead of only random commits? That would make a separate task to verify the signature unnecessary and would let us use git tags instead of specific commit here.

ansible/roles/etesync/defaults/main.yml
drybjed: Does that mean that EteSync cannot be used on Debian Buster without Sid package installed? Or do you install Django in the virtualenv instead?

~~Edit (ypid): I somehow cannot comment on this on GitHub so I edit instead. For now virtualenv is used.~~

ansible/roles/etesync/defaults/main.yml
```diff
@@ -0,0 +1,456 @@
...
+  - user: '{{ etesync__user }}'
+    group: '{{ etesync__group }}'
+    home: '{{ etesync__home }}'
+    id: '{{ etesync__git_gpg_key_id }}'
+
+                                                                   # ]]]
+# .. envvar:: etesync__python__dependent_packages3 [[[
+#
+# Configuration for the :ref:`debops.python` Ansible role.
+etesync__python__dependent_packages3:
```
drybjed, line 390: Can EteSync be used with Python 2.7 environment?

ansible/roles/etesync/tasks/main.yml
```diff
@@ -0,0 +1,204 @@
...
+
+- name: Update Ansible facts if they were modified
+  action: setup
+  when: etesync__register_facts is changed
+
+- name: Print URL entry point if EteSync is deployed on a subpath
+  debug:
+    msg: 'URL: {{ etesync__url }}'
+  when: etesync__http_psk_subpath != ""
+  tags: [ 'role::etesync:show_url' ]
```
drybjed: I don't like to have `debug` modules in the production code, this implies that Ansible output is meant to be read and interactive which is not strictly true. We need to find a better, general purpose solution for this. Perhaps a mail to `root` on first execution would suffice? This would also ensure that the message with the details can be archived if necessary. The SMTP configuration should make sure that it will be sent to the correct person.

docs/ansible/roles/etesync/getting-started.rst
```diff
@@ -0,0 +1,48 @@
...
+
+.. only:: html
+
+   .. contents::
+      :local:
+
+Example inventory
+-----------------
+
+To configure Docker on a given remote host, it needs to be added to the
```
drybjed: Docker? What?

docs/ansible/roles/etesync/getting-started.rst
```diff
@@ -0,0 +1,48 @@
...
+.. code-block:: none
+
+   [debops_service_etesync]
+   hostname
+
+
+Example playbook
+----------------
+
+Here's an example playbook that can be used to manage Docker:
```
drybjed: And Docker again here. :)

drybjed (changes requested):
EteSync seems like an interesting solution for the CalDAV/CardDAV conundrum. I wonder if there are any blockers for it to be added to Debian archive, at least the server part.

There are some questions or issues I'd like to see your answers to, before merging this PR. See the comments below.

ansible/roles/etesync/defaults/main.yml
ypid: Sid packages are not required obviously. I just tried to install as much of the dependencies from stable but then had to give up. I use virtualenv and pip for now.

ansible/roles/etckeeper/defaults/main.yml
ypid: Should be fine to revert as you have not made a 1.2.x release yet including it. Then I guess you support this switch in how we use the etckeeper role and my reasoning?

ansible/roles/etesync/defaults/main.yml
ypid: Upstream really cares about security in my eyes so they already sign tags (ref: https://github.com/etesync/server/tags). When you look at the role, you will notice that I of course also verify this signature.

It is just that over at https://github.com/hashbang/os, we really like the concept of "hash locking" particulare versions even if they are already signed. In Android, it is not unseen that signed tags are changed (not even a compromise, just regular behavior) … That is something where hash locking helps. I don’t see the big downside of having that additional security. When I hash lock some upstream project that usually also indicates that I have done at least some degree of research and review of it and that is best signaled with exactly referring to what version I checked.

ansible/roles/etesync/defaults/main.yml
ypid: I don’t care about Python2 as it is EOL. Python 3 is ~~installed~~ available everywhere and there it works. Do we still need to care?

> Or do you install Django in the virtualenv instead?

Yes, I have to.

.gitlab-ci.yml
ypid: Thanks, fixed.

CHANGELOG.rst
ypid: Sounds good. I am not creative so I just copied the intro from the upstream website one more time :-/.

ansible/roles/etesync/tasks/main.yml
ypid: Very good idea, thanks. I don’t operate (with DebOps/Ansible) at such a scale but I can definitely see this come in handy. Done.

ypid (commented):
> EteSync seems like an interesting solution for the CalDAV/CardDAV conundrum. I wonder if there are any blockers for it to be added to Debian archive, at least the server part.

From having reviewed (a bit) the software myself, I could not find any blockers. Most if not all dependencies are already in Debian. I guess the only issue is that nobody has stepped up to package it yet.

> There are some questions or issues I'd like to see your answers to, before merging this PR. See the comments below.

Thanks for your careful (as always) review!

ansible/roles/etesync/templates/etc/etesync-server/etesync-server.ini.j2
```diff
@@ -0,0 +1,16 @@
...
+[allowed_hosts]
+{% for host in ([ etesync__config_allowed_hosts ]
+				if etesync__config_allowed_hosts is string
+				else etesync__config_allowed_hosts) %}
+allowed_host{{ loop.index }} = {{ host }}
+{% endfor %}
+
+[database]
+engine = django.db.backends.sqlite3
+name = {{ etesync__data }}/db.sqlite3
```
drybjed, line 16: I'm suprised that the role does not support PostgreSQL as an alternative to SQLite. I suppose that this can be added later and backported to a stable release.

ypid: Jup. The only reasons I did not add PostgreSQL support is that for my small deployment I decided to go with SQLite. If someone needs PostgreSQL, it can be added.

ansible/roles/etesync/tasks/main.yml
```diff
@@ -0,0 +1,209 @@
...
+    source {{ etesync__virtualenv }}/bin/activate
+    echo "from django.contrib.auth.models import User;
+          User.objects.create_superuser('${SUPERUSER_NAME}', '${SUPERUSER_EMAIL}', '${SUPERUSER_PASSWORD}')" | ./manage.py shell
+  args:
+    chdir: '{{ etesync__git_checkout }}'
+    executable: 'bash'
+  become: True
+  become_user: '{{ etesync__user }}'
+  when: (not etesync__register_installed.stat.exists|bool and
+         etesync__register_migration is not changed)
```
drybjed: The above task which is supposed to create a superuser account did not get executed, therefore there's no admin account or password. I suspect that the `when:` conditions are wrong, because the they depend on the database being not changed, which isn't the case on first install. Perhaps you should change the condition. I wonder if the Django command is idempotent or can it be made so if not.

drybjed (changes requested):
I tried out the EteSync role in a development environment. Application installed fine but `gunicorn` didn't start properly because it's not present on Buster managed by DebOps by default  (Python3-only), but `gunicorn3` is. This is a bug in the `gunicorn` role, I'll fix that shortly.

When I went to the EteSync site I was greeted with message:
```
It works!
Please refer to the README to complete the final steps if you haven't done so already.
```
Is that expected? I thought that was the webserver at first, but it seems that this is indeed the EteSync application. I figured that going to the `/admin` page would give me a login form, but due to superuser account issues mentioned elsewhere I didn't had proper credentials to test it further.

Perhaps you could mention this in the Getting Started section of the role documentation so that it is expected?

ansible/roles/etesync/tasks/main.yml
ypid: Good catch, fixed. I copied most of the Django setup from the NetBox role :)

ypid:
> This is a bug in the gunicorn role, I'll fix that shortly.

Thanks!

> Is that expected?
> Perhaps you could mention this in the Getting Started section of the role documentation so that it is expected?

It is expected. I added a paragraph to the docs about the feature set. I know upstream docs are not comprehensive but I don’t want to document in a role that setups up the service how to use the service itself.

ansible/roles/etckeeper/defaults/main.yml
ypid: @drybjed Ping

drybjed: I remember about this. I'll revert the commit when the `etesync` role is merged.

ansible/roles/etesync/tasks/main.yml
```diff
@@ -0,0 +1,203 @@
...
+    source {{ etesync__virtualenv }}/bin/activate
+    echo "from django.contrib.auth.models import User;
+          User.objects.create_superuser('${SUPERUSER_NAME}', '${SUPERUSER_EMAIL}', '${SUPERUSER_PASSWORD}')" | ./manage.py shell
+  args:
+    chdir: '{{ etesync__git_checkout }}'
+    executable: 'bash'
+  become: True
+  become_user: '{{ etesync__user }}'
+  when: ("etesync" not in ansible_local)
+  no_log: True
```
drybjed, line 172: After the change the task is executed, but (on Debian Buster) it errors out with:
```
fatal: [etesync]: FAILED! => {"changed": true, "cmd": "## -o nounset can be enabled with virtualenv >= 16.2.\n## Ref: https://github.com/pypa/virtualenv/pull/922\nset -o pipefail -o errexit;\nsource /usr/local/lib/etesync/virtualenv/bin/activate\necho \"from django.contrib.auth.models import User;\n      User.objects.create_superuser('${SUPERUSER_NAME}', '${SUPERUSER_EMAIL}', '${SUPERUSER_PASSWORD}')\" | ./manage.py shell\n", "delta": "0:00:00.182531", "end": "2020-01-29 22:54:04.625757", "msg": "non-zero return code", "rc": 1, "start": "2020-01-29 22:54:04.443226", "stderr": "Traceback (most recent call last):\n  File \"./manage.py\", line 22, in <module>\n    execute_from_command_line(sys.argv)\n  File \"/usr/local/lib/etesync/virtualenv/lib/python3.7/site-packages/django/core/management/__init__.py\", line 381, in execute_from_command_line\n    utility.execute()\n  File \"/usr/local/lib/etesync/virtualenv/lib/python3.7/site-packages/django/core/management/__init__.py\", line 375, in execute\n    self.fetch_command(subcommand).run_from_argv(self.argv)\n  File \"/usr/local/lib/etesync/virtualenv/lib/python3.7/site-packages/django/core/management/base.py\", line 316, in run_from_argv\n    self.execute(*args, **cmd_options)\n  File \"/usr/local/lib/etesync/virtualenv/lib/python3.7/site-packages/django/core/management/base.py\", line 353, in execute\n    output = self.handle(*args, **options)\n  File \"/usr/local/lib/etesync/virtualenv/lib/python3.7/site-packages/django/core/management/commands/shell.py\", line 92, in handle\n    exec(sys.stdin.read())\n  File \"<string>\", line 2\n    User.objects.create_superuser('drybjed', 'drybjed', 'fE0hheO78pC16Ths.U2Z')\n    ^\nIndentationError: unexpected indent", "stderr_lines": ["Traceback (most recent call last):", "  File \"./manage.py\", line 22, in <module>", "    execute_from_command_line(sys.argv)", "  File \"/usr/local/lib/etesync/virtualenv/lib/python3.7/site-packages/django/core/management/__init__.py\", line 381, in execute_from_command_line", "    utility.execute()", "  File \"/usr/local/lib/etesync/virtualenv/lib/python3.7/site-packages/django/core/management/__init__.py\", line 375, in execute", "    self.fetch_command(subcommand).run_from_argv(self.argv)", "  File \"/usr/local/lib/etesync/virtualenv/lib/python3.7/site-packages/django/core/management/base.py\", line 316, in run_from_argv", "    self.execute(*args, **cmd_options)", "  File \"/usr/local/lib/etesync/virtualenv/lib/python3.7/site-packages/django/core/management/base.py\", line 353, in execute", "    output = self.handle(*args, **options)", "  File \"/usr/local/lib/etesync/virtualenv/lib/python3.7/site-packages/django/core/management/commands/shell.py\", line 92, in handle", "    exec(sys.stdin.read())", "  File \"<string>\", line 2", "    User.objects.create_superuser('drybjed', 'drybjed', 'fE0hheO78pC16Ths.U2Z')", "    ^", "IndentationError: unexpected indent"], "stdout": "", "stdout_lines": []}
```
I think that making it closer to the similar task in the `netbox` role by changing the content to not use the `: |` syntax will fix this, but it's getting late here so I'm heading to bed. I'll try and fix it tommorow if @ypid won't do it first. :-)

ansible/roles/etesync/templates/etc/etesync-server/etesync-server.ini.j2
SuperSandro2000: I would like to request postgres as my backup solution can't really deal with sqlite.

drybjed: In that case I suggest that the PR will be merged after `v2.0.0` release which is imminent (perhaps today). I'll revert the `etckeeper` change as well.

ansible/roles/etesync/tasks/main.yml
ypid: Thanks for the review and the hint. Fixed it.

ansible/roles/etesync/templates/etc/etesync-server/etesync-server.ini.j2
ypid: @SuperSandro2000 Feel free to add PostgreSQL support.

drybjed (approved):
@ypid Thanks for the new role. Since we are at the beginning of the new DebOps development cycle, I think that the current PR can be emrged and the PostgreSQL support can be added later.

There's a web client in the works, do you know how far along it is from a stable release? It might be useful to have a role for it as well, or add support to the `etesync` role itself in the future.

ypid:
You are welcome, thanks for maintaining :+1: The web client is useful and I have tested it a bit. That would be interesting to add to DebOps. When that is done, also [Signed Pages](https://github.com/tasn/webext-signed-pages) should be supported, not sure yet how that can be implemented. I have some interesting in that. But this is not a priority for me meaning that when current demand does not change, I will not setup the web client at all. PostgreSQL I will only do when I run into limits with SQLite in my small deployment.

PR #1244, opened 2020-01-27
Add 'mailservice' LDAP schema

### LINKED ISSUE #1128: [debops.slapd] Discussion about Virtual Mail LDAP scheme
After reading the links on [this comment](https://github.com/debops/debops/pull/1108#discussion_r337470001) from #1108, here are my propositions for a postfix+dovecot Virtual Mail User LDAP scheme.

- The Virtual Mail system uses only "virtual users", even if they are real POSIX users. So these users do not have the MailDir directly in their Home dir. Therefore there is no need to specify the user UID and GID for the Mailbox
- Using the attributes specified [here](https://github.com/tleuxner/ldap-virtualMail/blob/master/virtualMail.ldif) as an example:
  - `mailHomeDirectory`: is not needed, `postfix`+`postldap` take care of this by storing all emails under `/var/vmail`
  - `mailAlias`: it's a keeper. This is useful.
  - `mailDrop`: this is the "real" email where emails sent to this address or aliases is finally stored. While it is the correct term, also used officially by Postfix, I think we should just `mail`. Because `mail` is just easier to remember and understand.
  - Combining `mail` and `mailAlias` attributes to denote the email addresses for an user is easy to manage and understand by the LDAP admin. No need to check extra docs and so on.
  - `mailUidNumber`, `mailGidNumber`: see my argument above
  - `mailEnabled`: keep. Useful to "disable" an email account, without deleting `authorizedService` attributes. Although I acknowledge it's kind of a duplicate with `authorizedService`
  - `mailQuota`: keep. self explaining.
  - `mailGroupACL`: keep. Allows to enforce ACL rules with dovecot [Example](https://github.com/tleuxner/ldap-virtualMail/blob/75bf753ed97b54b1c53f189b10979a13d415a162/etc/dovecot/dovecot-ldap-userdb.conf#L9)
  - `mailExpungeTrash`: keep: autoexpunge 
  - `mailForward`: keep
- Other attributes are not really needed. Even the [Postfix Book](https://github.com/variablenix/ldap-mail-schema/blob/master/postfix-book.schema) does not have exotic attributes.
- If there is a need to use groups or mailing list, I think it is better to use Mailman or an additional schema.

More sources:
- https://kb.leuxner.net/article/mail-server-using-ldap-backend-w-postfix-dovecot/
- http://acidx.net/wordpress/2014/06/installing-a-mailserver-with-postfix-dovecot-sasl-ldap-roundcube/

drybjed:
Fixes #1128

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -103,6 +103,14 @@ General
 - The role can now use PostgreSQL as a database backend. The database server
   can be managed with the :ref:`debops.postgresql_server` role.
 
+:ref:`debops.slapd` role
+''''''''''''''''''''''''
+
+- The :ref:`mailservice <slapd__ref_mailservice>` LDAP schema has been added to
+  the :ref:`debops.slapd` role. It provides a set of object classes and
+  attributes useful for defining e-mail recipients and simple mail distribution
+  lists in the LDAP directory.
+
 Changed
 ~~~~~~~
 
--- a/ansible/roles/slapd/defaults/main.yml
+++ b/ansible/roles/slapd/defaults/main.yml
@@ -630,6 +630,8 @@ slapd__default_tasks:
         - 'host eq,sub'
         - 'gid eq'
         - 'mail eq,sub'
+        - 'mailAddress eq,sub'
+        - 'mailAlternateAddress eq,sub'
 
   - name: 'Enable the monitor database'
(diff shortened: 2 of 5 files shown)


reixd:
+1 approved from my side.
Thank you, this schema is more generic than my proposal.

PR #1245, opened 2020-01-29
[postgresql_server] Support more efficient backukp compression

gaudenz:
Add configuration options to configure autopostgresqlbackup to compress
the backups directly with pg_dump instead of first dumping an
uncompressed file and then compressing it.

This uses the autopostgresqlbackup configuration option COMMCOMP which
adds the --compress option to pg_dump. Unfortunately this is labeled as
"communication compression" in the original file, but it has nothing to
do with compressing the communication. See the pg_dump manpage for
details about the --compress option.

### DIFF
--- a/ansible/roles/postgresql_server/defaults/main.yml
+++ b/ansible/roles/postgresql_server/defaults/main.yml
@@ -563,7 +563,9 @@ postgresql_server__auto_backup_encryption_suffix: '.enc'
                                                                    # ]]]
 # .. envvar:: postgresql_server__auto_backup_compression [[[
 #
-# Specify compression method to use for snapshots.
+# Specify compression method to use for snapshots. The special value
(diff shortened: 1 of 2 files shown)


drybjed (commented):
Sounds good. Do you think that the `pg_dump` mode can be set as default? Will that change anything on existing installations?

gaudenz:
I think the end result should be the same as with the current defaults. I'm a bit reluctant to set this as the default as it looks a bit like the original author of autopostgresqlbackup did not intend to use the variable this way. It looks like he misunderstood what the `--compress` option does.
On the other hand the current options are suboptimal for large databases as you need lots of free space just to do the uncompressed dump and compress it afterwards.

drybjed (approved):
Very well then. I'll merge the current PR so the change can be easily tested. If there are no issues in production when `pg_dump` mode is enabled, we can switch to it as the default at a later date.

PR #1256, opened 2020-02-03
Redesign DebOps Collections, fix small issues

drybjed:


ypid:
> The DebOps Collection published on Ansible Galaxy has been split into multiple Collections *due to the number of Ansible roles present in DebOps.*

Can you go into more detail about the reasoning? `debops.roles01`, `debops.roles02` and `debops.roles03` looks a bit random?

drybjed:
Sure. With the changes to the `v2.0.0` release, I was unable to import the DebOps collection as a whole unit to Ansible Galaxy repository. The issue [was reported to the development team](https://github.com/ansible/galaxy/issues/2202), they determined that the current import process limits might be the problem.

I also noticed that the DebOps collection tarball has the 1,5 MB size. Since the official limit for an Ansible Collection tarball is 2 MB, I decided that this is as good moment as any to make the split - sooner or later we would have to do this anyway even if Galaxy successfully imported the tarball due to the size limits.

The new set of Collections is the `debops.debops` as the "main" Collection which contains various playbooks, Ansible plugins, custom modules and the DebOps roles which are used in the common and bootstrap playbooks. Other collections just contain various roles and are marked as a dependency of the `debops.debops` collection which will pull the additional collections automatically during installation. I cannot define a reverse dependency in the `debops.rolesXY` collections on the `debops.debops` collection because Galaxy will not let me import a collection with dependencies that don't exist. Currently I import the collections in reverse order.

The contents of the collections (what roles go into which one) are determined by several factors - how many times a given role is used in the playbooks, how many times local facts of a given role are referenced by other roles, if a given role contains dependencies, and so on. The final list of roles per collection is computed during collection creation via the `lib/ansible-galaxy/make-collection` script. Because of that some roles might be moved around between collections in the future, but this shouldn't be an issue as long as the `debops.debops` collection is installed. The playbooks should also include all DebOps collections for full compatibility.

ypid:
Thanks very much :) The moving of roles in collection might not be an issue for proper DebOps users, but I image it could be an issue for users of individual roles?

drybjed:
The problem is that many roles in the auxillary collections depend on custom DebOps plugins like `parse_kv_items`, etc. which are included in the `debops.debops` collection. Without it, roles will be broken, so it has to be installed. And when it is installed, it will pull all of the other collections with it.

ypid:
Fair enough then.

PR #1257, opened 2020-02-04
[resources] Add support for the `*_time`; [cryptsetup docs] Add second boot disk to FDE with same passphrase; Minor stuff

ypid:


ansible/roles/resources/defaults/main.yml
```diff
@@ -29,6 +29,15 @@ resources__enabled: True
...
 resources__src: '{{ inventory_dir | realpath + "/../resources/" }}'
                                                                    # ]]]
+# .. envvar:: resources__time_format [[[
+#
+# Time format to be used for the `Ansible file module`_.
+# This is used for example when the ``access_time`` is specified.
+# It is set here to a human readable profile of `ISO 8601`_ because the Ansible
+# default format is hardly human readable
+# (without separators like ``-`` and ``:``).
+resources__time_format: '%Y-%m-%dT%H:%M:%S'
```
drybjed, line 40: I'm not sure what the time format is used for. Does that indicate how Ansible should interpret the string provided by the user? In that case, perhaps you should add an example or two of the usage with a date specified in the desired format, in the variable documentation.

ypid: > Does that indicate how Ansible should interpret the string provided by the user?

That is correct. And the Ansible default format is totally unreasonable and not very human readable.

I checked the variable docs but you decided to refer to the Ansible module docs which I find very reasonable. Therefore I documented it in the `defaults/main.yml`. I now added a example to it as well.

drybjed: No, I meant a proper example in the variable documentation in `docs/ansible/roles/resources/defaults-detailed.rst`, sorry for the misunderstanding. The small example near the variable definition can stay.

ypid: Done that. Should be ready for merging.

drybjed:
@ypid: Almost there. :-) Got the error from the docs on Travis-CI:
```
/home/travis/build/debops/debops/docs/ansible/roles/cryptsetup/defaults-detailed.rst:599:Duplicate explicit target name: "cryptsetup__ref_devices_chaining_multiple_ciphers".
```

ypid:
Fixed :)

drybjed (approved):
Looks good, thanks. I'll backport selected commits to older DebOps releases.

PR #1265, opened 2020-02-13
Postgresql custom log directory

hvisage:
needs to check/create them too ;)


ansible/roles/postgresql_server/tasks/main.yml
```diff
@@ -46,6 +46,17 @@
...
   when: postgresql_server__data_directory != '/var/lib/postgresql'
 
+
+- name: Make sure that custom PostgreSQL Log directory exists
+  file:
+    path: '{{ postgresql_server__log_directory }}'
+    state: 'directory'
+    owner: '{{ postgresql_server__user }}'
+    group: '{{ postgresql_server__group }}'
+    mode: '0755'
```
drybjed: Perhaps the directory more should be more strict (`0750`), I'm not sure if PostgreSQL emits confidential information in its logs...

ansible/roles/postgresql_server/tasks/manage_clusters.yml
```diff
@@ -65,6 +65,19 @@
...
+  file:
+    name: "{{item.log_directory}}"
+    mode: 0755
+    owner: '{{ item.user | d(postgresql_server__user) }}'
+    group: '{{ item.group | d(postgresql_server__group) }}'
+    state: directory
+    recurse: yes
+  with_flattened:
+    - '{{ postgresql_server__clusters }}'
+  when: item.log_directory is abs
```
drybjed: What? `abs`?

hvisage: Absolute path.

hvisage: I recall the problem was that when the postgresql directory is absolute, then these are needed, otherwise, it "fits" into the system nicely.

Specifically I had `/pgdb` and `/pglogs` separate ZFS filesystems/zpools

ansible/roles/postgresql_server/tasks/main.yml
SuperSandro2000: At least in my configuration it logs failed inserts and I use pretty much the defaults.

hvisage: Failure messages?
I might need to add the test for empty/undefined `postgresql_server__log_directory` as in my setup(s) I always have a log_directory defined

SuperSandro2000: > Failure messages?

Yes but this might be unrelated to this as I don't use this role to manage my postgres. Just wanted to give insights that this could happen.

hvisage: That is why I'd like to see what the error message(s) is that you are getting

drybjed: @hvisage I think that you changed the mode in the incorrect task. :-) Remember to update the mode in the other file as well.

hvisage: :blush: fixed the log directory permissions.

BUT: when you have the data directory with a 0755, then I'd make the case, you are already lax enough that a 0750 on the log directory isn't buying you any extra security, thus the need for the data to be as or stricter than the log directory.

SuperSandro2000:
````
[40] STATEMENT:  INSERT INTO devices (user_id, device_id, display_name, hidden) VALUES('*******************', '**********', NULL, false)
````

I replaced the sensitive values with stars. It logged the full email address and device id.

Comment (hvisage) [reply ansible/roles/postgresql_server/tasks/main.yml]: @SuperSandro2000 that is  unrelated to my PR which was ONLY adding the needed for a custom *LOG DIRECTORY* ;)
Perhaps a separate PR to handle that issue please ?

Comment (SuperSandro2000) [reply ansible/roles/postgresql_server/tasks/main.yml]: Well, no as this container is totally unrelated to this project. I just wanted to answer the question is postgres can log sensitive data. Yes, it can.

File: ansible/roles/postgresql_server/tasks/manage_clusters.yml [hunk 3]
```diff
@@ -65,6 +65,19 @@
   loop_control:
     label: '{{ item.item }}'
 
+- name: Log directory path
+  file:
+    name: "{{item.log_directory}}"
+    mode: 0750
```
drybjed: The octal value should be quoted to avoid issues with Jinja interpreting it and converting to decimal value.

ansible/roles/postgresql_server/tasks/main.yml
```diff
@@ -43,9 +43,20 @@
...
 
+
+- name: Make sure that custom PostgreSQL Log directory exists
+  file:
+    path: '{{ postgresql_server__log_directory }}'
+    state: 'directory'
+    owner: '{{ postgresql_server__user }}'
+    group: '{{ postgresql_server__group }}'
+    mode: '0750'
+  when: postgresql_server__log_directory != '/var/lib/postgresql'
```
drybjed, line 57: The role does not have `postgresql_server__log_directory` variable defined by default. In the generated `postgresql.conf` file the `log_directory` variable has the value `pg_log`. The default logs are stored in the `/var/log/postgresql/` directory, per cluster I think. Any idea what value should be the default? `/var/lib/postgresql` seems to be incorrect.

ansible/roles/postgresql_server/tasks/manage_clusters.yml
hvisage: done.

ansible/roles/postgresql_server/tasks/main.yml
hvisage: from the pg_createcluster manpages: http://manpages.ubuntu.com/manpages/trusty/man8/pg_createcluster.8.html
```
  -l path, --logfile=path
           Explicitly set the path for the postmaster(1) server log file. Defaults to
           /var/log/postgresql/postgresql-version-cluster.log.
```

hvisage: Thus when you do specify it for pg_createcluster, you actually have to specify with the full path, ie. `/directory/filename.log`, though that gets "overriden" by the `postgres.conf` log file setting(s), and then becomes the value that `pg_lsclusters` reports (afaik), though there is a systemd time that will use that `/directory/filename.log' for the initial log file(s) before the postmaster started to read the `postgres.conf` file.

drybjed: I'm sorry that it takes so long to merge this, but something else bugs me. If you already create the directory by setting it as `item.log_directory` below, and you need to specify an absolute path, why additionally create some kind of global log directory? I think that the above task is not really needed. Thoughts?

hvisage:
> On 20 Feb 2020, at 22:24 , Maciej Delmanowski <notifications@github.com> wrote:
> 
> @drybjed commented on this pull request.
> 
> In ansible/roles/postgresql_server/tasks/main.yml:
> 
> >    when: postgresql_server__data_directory != '/var/lib/postgresql’

This, is the *DATA* directory, ie. /zTank01/database

> +- name: Make sure that custom PostgreSQL Log directory exists
> +  file:
> +    path: '{{ postgresql_server__log_directory }}'
> +    state: 'directory'
> +    owner: '{{ postgresql_server__user }}'
> +    group: '{{ postgresql_server__group }}'
> +    mode: '0750'
> +  when: postgresql_server__log_directory != '/var/lib/postgresql’

and this is for the *log* directory, ie. /zLogs07/logspace

> I'm sorry that it takes so long to merge this, but something else bugs me. If you already create the directory by setting it as item.log_directory below, and you need to specify an absolute path, why additionally create some kind of global log directory? I think that the above task is not really needed. Thoughts?

My logs and databases aren’t necessarily in the same “space” as using ZFS, they have different compression, back end and sync settings.

And when the database clusters log with different names in the log filename, then it’s not a problem throwing them inside the same global directory… easier for things pgBadger, but also for the base etc. if I do want it in separate directories.

drybjed:
I see. In that case, I'll add the `postgresql_server__log_directory` default variable and set it as empty, as well as change the condition of the task - that way the task will not be applied if the custom log directory "root" is not defined. This should keep the current behaviour where by default PostgreSQL on Debian logs to `/var/log/postgresql/` directory.

PR #1266, opened 2020-02-16
Add 'dpkg_cleanup' role, remove cleanup script from 'nullmailer' role

drybjed:


ypid:
I have just seen this work pop up in my `git pull && git verify-commit HEAD && git diff HEAD@{1}..HEAD` combo and have not done any testing yet but the design looks beautiful from what I have seen. Nice that it is now easy for other roles to also use it :+1:

drybjed:
Thanks! I was looking into a way to add support for custom commands to have a more extensible options but `eval` seemed risky to me to use for this, so I left that out for now. Perhaps this will be sufficient most of the time.

ypid:
> Perhaps this will be sufficient most of the time.

I would guess so. `eval` can scare people :) But the `eval` would effectively be an inline raw code block. Not really a break of our security model but can definitely cause a safety issue. If we can avoid it we should :)

PR #1267, opened 2020-02-18
[debops.telegraf] Add telegraf role

LeonardoAndrei:


scibi:
@drybjed  In Elastic stack there is separate role `elastic_co`. Maybe in Influx stack there should be a similar role?

Long time ago I started to write `influxdata` role. It's here: https://github.com/scibi/ansible-influxdata. It's almost identical to `elastic_co` and it's not using `keyring` role because it didn't exist 2 years ago ;)

drybjed:
@scibi Currently 4 potential roles (telegraf, influxdb, chronograf, capacitor), so sure. Your role is almost ready it looks like. Want to update it to use `keyring` and make a PR?

alinalexandru:
@scibi We wanted initially to create a module for managing the influxdata repository,
I don't remember why we have drop the support.
Thanks for your PR.
After it's merged, we can update our roles to depend on your role.

drybjed:
Another Telegraf role implementation was merged into DebOps via https://github.com/debops/debops/pull/1828. Closing this issue.

PR #1275, opened 2020-02-22
Updates of copyright notices and license information in DebOps

drybjed:
This patch adds the SPDX license identifiers in files which were
detected by the 'reuse' tool and have existing copyright details.

I decided to review the copyright and licensing in DebOps and make it compliant with the [REUSE Specification](https://reuse.software/spec/). This hopefully will make the project easier to include in Debian archive in the future. The code is checked using the [`reuse`](https://pypi.org/project/reuse/) tool which eventually will be added to the Travis-CI test suite.

Review will take some time, but it's a one time process and after that it will be easy to stay compliant.


LICENSES/MIT.txt
```diff
@@ -0,0 +1,19 @@
+MIT License Copyright (c) <year> <copyright holders>
```
SuperSandro2000: I have no clue what you could do with that line but leaving the default in seems strange.

ansible/roles/owncloud/templates/srv/www/sites/themes/debops-template/defaults.php.j2
```diff
@@ -13,7 +13,7 @@ Raw: https://raw.githubusercontent.com/owncloud/core/master/themes/example/defau
  * @author Björn Schießle <schiessle@owncloud.com>
  * @author Jan-Christoph Borchardt, http://jancborchardt.net
  * @copyright Copyright (c) 2015, ownCloud, Inc.
- * @license AGPL-3.0-only
```
SuperSandro2000, line 16: I think this is some magic php comment and should maybe stay.

docs/_templates/breadcrumbs.html
```diff
@@ -1,3 +1,7 @@
+{# Copyright (C) 2016      Robin Schneider <ypid@riseup.net> #}
```
SuperSandro2000, line 1: Does this highlight mean anything or is it just some strange github thing?

LICENSES/MIT.txt
drybjed: As I, not a lawyer, understand it: the files in the `LICENSES/` directory are just for reference, because you MUST provide the license with the project. They shouldn't be changed, because quick check with `sha256sum` allows to verify if the license was changed or not

 Also, if I replaced the example text here with say, current year and my name and email, that probably would mean that suddenly DebOps is licensed under MIT license, right? Without changing the details here, it probably means that the MIT license applies only to the files that explicitly refer to it.

Also, the [answer](https://reuse.software/faq/#edit-license) from the REUSE FAQ about this:

> *Should I edit my license files?*

> You should never edit license files. When you use an existing license, you should always copy it verbatim.
> There are many reasons for why you should not alter license texts, but if you do alter the texts, you should use a different SPDX identifier for this license.

ansible/roles/owncloud/templates/srv/www/sites/themes/debops-template/defaults.php.j2
drybjed: Sounds good, I'll bring it back then.

docs/_templates/breadcrumbs.html
drybjed: It's just the syntax highliter being confused, probably it thinks these are some kind of CSS references.

LICENSES/MIT.txt
SuperSandro2000: I think the text you linked is about the legal text in the license which talks about your rights not the header. So lines 3 to 19 should stay the same. I think the best choice would be to drop the year and holders completely from this file.

Also sha sums are irrelevant cause as soon as you change any whitespace it is different which does not change the license what so ever.

drybjed: I removed the `Copyright (c) <year> <copyright holders>` string from the file and the `reuse` script doesn't scream about any issues, so I suppose that's fine.

ypid (dismissed):
Looks like a great idea. I reviewed the PR and it looks good to me. I did not know reuse before. Thanks for maintaining!

.reuse/dep5
```diff
@@ -0,0 +1,90 @@
...
+Files: ansible/roles/dokuwiki/files/srv/www/dokuwiki/farm/animal/*
+Copyright: 2004-2010 Andreas Gohr <andi@splitbrain.org>
+License: GPL-2.0-only
+
+Files: ansible/roles/owncloud/files/usr/local/src/owncloud/keyring/0x*
+Copyright: 2015-2016 Maciej Delmanowski <drybjed@gmail.com>
+Copyright: 2015      Hartmut Goebel <h.goebel@crazy-compilers.com>
+Copyright: 2015-2019 Robin Schneider <ypid@riseup.net>
+Copyright: 2015-2019 DebOps <https://debops.org/>
+License: GPL-3.0-or-later
```
ypid: Not sure if GPL is the best license for OpenPGP keys. Maybe put it under Public Domain?

ypid:
Out of interest. How are you adding those headers? Is it somewhat tool-assisted based on the git history?

drybjed:
I want this PR to be a "demarcation point" between basically chaotic license and copyright information in DebOps in the past, and known good state in the future. So, no automation - all 3462 files are/were checked by hand by me and the copyright and license details written separately (I don't automate GPG signatures either, so yeah - all these commits were signed by me personally).

I wanted to capture the current state of the repository as-is, but at the same time I didn't want to do it for like 3-4 weeks combing through various `git` repositories of the project. Because of that mot all copyright information is accurate - to save time I looked through `git` history for most of the non-role related files in available repositories and I tried to capture rough years where a given author touched these files, but not all is there. For the role files, I copied the details from the `COPYRIGHT` files of each role treating all files under a given subdirectory with the same copyright and license data. In the future I hope that more people will help with filling up the missing copyright credits.

.reuse/dep5
drybjed: There is no "Public Domain" license in SPDX list [due to reasons](https://wiki.spdx.org/view/Legal_Team/Decisions/Dealing_with_Public_Domain_within_SPDX_Files). In my opinion these keys should be removed entirely from the repository - the `keyring` role can deal with GPG keys not published on the keyserver network by adding them to the local keyring directory. The role can mention that these GPG keys need to be downloaded separately. But since for now I wanted to capture the existing state which can be changed later, I went with the same license as the role itself.

I'm not sure what license should be applied to GPG keys... What do you think about [CC0](https://creativecommons.org/share-your-work/public-domain/cc0/) instead?

ypid: CC0 is also fine. Actually, when I said Public Domain, I already included CC0 in my thought :)

ypid:
Thanks for the details and your commitment! Sounds good.

.reuse/dep5
drybjed: OK, will do.

drybjed:
Final summary from the `reuse` tool:
```
# SUMMARY

* Bad licenses:
* Deprecated licenses:
* Licenses without file extension:
* Missing licenses:
* Unused licenses:
* Used licenses: AGPL-3.0-only, Apache-2.0, BSD-4-Clause, CC-BY-SA-4.0, CC0-1.0, GPL-2.0-only, GPL-2.0-or-later, GPL-3.0-or-later, MIT
* Read errors: 0
* Files with copyright information: 3462 / 3462
* Files with license information: 3462 / 3462

Congratulations! Your project is compliant with version 3.0 of the REUSE Specification :-)
```

ypid:
Well done!

Can you talk a bit about why you changed from GPL-3.0-only to GPL-3.0-or-later? Do you fear that it would be more difficult to switch to a new version later (as in the Linux Kernel for example)?

drybjed:
It's hard to tell if there ever will be a new version of the GPL license. The current one seems to cover most of the "loopholes", and the AGPL is an even more strict version for the SaaS world, I don't think that another iteration is necessary - but who knows... Perhaps the recent emergence of Commons Clause and similar efforts to fight with big players taking over projects and profiting fom them without giving back (something AGPL was probably meant to tackle) will result in another attempt at a new GPL license.

As for switching to the `-or-later` version within the DebOps project - I tried to make the status quo explicit... But now when I look at the license texts closely it seems that I made a mistake with giving all of the roles the `later` treatment. Since the scripts were done first, and they were licensed with `-or-later` version of the license, I must have assumed that the roles were licensed the same and I din't pay enough attention to the text of the license in the `COPYRIGHT`files which don't explicitly state this.

Hmmm, what to do what to do... I'm not really opposed at setting the `GPL-3.0-or-later` version to be the default. if we assume that a potential new version of the GPL license will be an improvement over the current 3.0 one, having an option to switch to it might be useful in the future. This also aligns in my opinion with DebOps being a long term project - future-proofing it might be handy.

ypid:
My personal (and at work) policy is to go with `3.0-only` because I would like to review new versions of the GPL first. But I am not sure how well this scales thus I am open to have `3.0-or-later` in DebOps. The issue is that such license upgrade (GPL-2.0-only to GPL-3.0) might require approval from all copyright owners. This is impossible for projects like the Linux Kernel.

Nextcloud is `AGPL-3.0-or-later` (ref: https://github.com/nextcloud/server/blob/master/package.json) and I trust them with making a responsible choice as they had a bit of a learning curve :)

drybjed:
I see. Do you think that the current state with `GPL-3.0-or-later` everywhere is fine, or should I go over the files again, revert to `GPL-3.0-only` where `-or-later` is not explicitly permitted, after that add missing copyright holders to files... And then we can switch the license over when all contributors to a given section of the code are fine with it? I think that as long as the source is not releases as tarballs/Ansible Collections/Python packages, we have a "window of opportunity" to straighten out any SPDX identifiers that were wrongly selected, as long as the changes are documented in the commit history.

Also, any thoughts about adopting [Developer Certificate of Origin](https://developercertificate.org/) for DebOps? That would entail [adding `Signed-Off-By:` string](https://stackoverflow.com/a/1962112) in the commit text. This can be done using `git commit -s` command, but would that add additional hurdle for contributing to DebOps? Or is there no benefit at this point?

ypid:
I fully agree with your assesment of the situation. It only counts when it is included in a release. The safer thing to do would be to revert to `GPL-3.0-only`.

> Developer Certificate of Origin

I will have to check that. I have not looked into this before.

drybjed:
Very well then, round two it is.

PR #1277, opened 2020-02-23
[ferm] Fix incorrect removal of avahi rule

### LINKED ISSUE #1183: [debops.ferm] role keeps removing debops.avahi firewall rule
The avahi playbook runs the debops.ferm role with these `ferm__dependent_rules`:
```
  - name: 'avahi'
    type: 'accept'
    dport: 'mdns'
    saddr: '{{ avahi__allow }}'
    protocol: 'udp'
    accept_any: True
    rule_state: '{{ "present" if (avahi__enabled|bool) else "absent" }}'
```
Without any variables overriding the role defaults, this renders `/etc/ferm/rules.d/100_rule_avahi.conf` as follows:
```
# This file is managed remotely, all changes will be lost

domain (ip ip6) table filter chain INPUT {
    protocol udp dport mdns {
        ACCEPT;
    }
}
```
Which seems fine. However, the rule gets removed when running the debops.ferm role again, but without the `ferm__dependent_rules` variable as shown above:
```
TASK [debops.ferm : Remove firewall rules] **************************************************************
--- before
+++ after
@@ -1,4 +1,4 @@
 {
     "path": "/etc/ferm/rules.d/100_rule_avahi.conf",
-    "state": "file"
+    "state": "absent"
 }

changed: [pancake.gwl.amsw.nl] => (item={'key': 'avahi', 'value': {'accept_any': True, 'dport': 'mdns', 'name': 'avahi', 'protocol': 'udp', 'rule_state': 'absent', 'rules': [{'accept_any': True, 'dport': 'mdns', 'protocol': 'udp', 'saddr': [], 'type': 'accept'}], 'saddr': [], 'type': 'accept'}})
[...]
```
The `debops.ferm` role should not remove earlier defined dependent rules. It is unclear to me why this only seems to happen with the `debops.avahi` rule.

## System information

- DebOps v1.2.0
- Python 3.7
- Debian 10

imrejonk:
The `ferm` role kept removing the dependent rule set by the `avahi`
role on IPv6-enabled systems. This happended because the default avahi
rule in the `ferm` role only detected the `mdns4_minimal` NSS service,
while it should detect the `mdns_minimal` NSS service as well.

Closes #1183

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -108,6 +108,12 @@ Removed
 Fixed
 ~~~~~
 
+:ref:`debops.ferm` role
+'''''''''''''''''''''''
+
+- Fixed incorrect removal of the ferm rule set by :ref:`debops.avahi` on
+  IPv6-enabled systems.
+
 :ref:`debops.netbase` role
 ''''''''''''''''''''''''''
 
--- a/ansible/roles/ferm/defaults/main.yml
+++ b/ansible/roles/ferm/defaults/main.yml
@@ -533,8 +533,10 @@ ferm__default_rules:
     rule_state: '{{ "present"
                     if ((ansible_local|d() and ansible_local.nsswitch|d() and
                          ansible_local.nsswitch.conf|d() and
-                         "mdns4_minimal" in q("flattened",
-                                              ansible_local.nsswitch.conf.hosts|d([]))) and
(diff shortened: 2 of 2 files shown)


drybjed (approved):
Thanks for finding the bug and fixing it! :+1:

PR #1279, opened 2020-02-24
Fix prosody pki hook. 

reixd:
Set correct prosody config file name 
Replace prosody check call

### DIFF
--- a/ansible/roles/prosody/templates/etc/pki/hooks/prosody.j2
+++ b/ansible/roles/prosody/templates/etc/pki/hooks/prosody.j2
@@ -6,7 +6,7 @@
 
 set -o nounset -o pipefail -o errexit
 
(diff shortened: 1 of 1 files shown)


reixd:
The usage of ```prosodyctl check``` is not ideal here, because if the user has a DNS missconfiguration the call to ```prosodyctl check``` can fail, thus not reloading the daemon and thus not changing the TLS cert used.

Anyway, it works for now.

Any comments on this?

drybjed:
Looks good to me.

If the user has a DNS misconfiguration, I think that the correct way would be to fix that misconfiguration instead of worrying that Prosody doesn't work - this is just a symptom of the problem with DNS. It's always DNS. :-)

drybjed (approved):

PR #1282, opened 2020-02-29
Add [debops.influxdb_server] and [debops.influxdb] roles

pedroluislopez:
- Split influxdb server and influxdb (client) roles.
- Integration with debops influxdata role.
- Debops PKI support
- Use parse_kv_items for all configuration
- Auto influxdb backup support

Mixed with https://github.com/innobyte/debops/tree/role-influxdb.


drybjed:
@pedroluislopez Thanks for working on this! I'll check it out after copyright and license PR is done, in a day or two.

@alinalexandru Care to make a review as well?

pedroluislopez:
@drybjed some documentation is missing. I will add it tomorrow.

alinalexandru:
Yes, i will have a look at it.
Maybe now that we are more involved, we can add the pki hook as well.

ansible/roles/influxdb_server/defaults/main.yml
```diff
@@ -0,0 +1,382 @@
...
+#
+# Length of automatically generated user accounts, saved in the ``secret/``
+# directory. See :ref:`debops.secret` role for more details about passwords.
+influxdb_server__password_length: '48'
+
+                                                                   # ]]]
+# .. envvar:: influxdb_server__root_password [[[
+#
+# Password for default root admin user.
+influxdb_server__root_password: '{{ lookup("password", secret + "/influxdb/" + ansible_fqdn +
```
alinalexandru, line 277: As I know the recommendation  where is to use ansible_hostname. Right @drybjed ?

docs/ansible/role-index.rst
```diff
@@ -118,6 +118,7 @@ Databases
 - :ref:`debops.redis_server`
 - :ref:`debops.redis_sentinel`
 - :ref:`debops.slapd`
+- :ref:`debops.influxdb`
 - ``debops.phpmyadmin``
```
alinalexandru, line 127: influxdb_server needs to be added as well.

alinalexandru (changes requested):

docs/ansible/role-index.rst
pedroluislopez: Yes. Today I will upload documentation for debops.influxdb_server role.

ansible/roles/influxdb_server/defaults/main.yml
drybjed: No, that would be `inventory_hostname` instead.

pedroluislopez: It's a problem.

If I use `inventory_hostname`, it generate the password in `secret/influxdb/influxdb-ams3-01/credentials/root/password` path.

But `influxdb` role use this configuration for search root password:

```password: '{{ lookup("password",
                  secret + "/influxdb/" + influxdb__delegate_to +
                  "/credentials/root/password " +
                  "length=" + influxdb__password_length) }}'

# .. envvar:: influxdb__server [[[
#
# FQDN hostname of the InfluxDB server. If a local InfluxDB server is detected, it
# will override this variable automatically. Only one server at a time is
# currently supported per host.
influxdb__server: ''

# .. envvar:: influxdb__delegate_to [[[
#
# When the InfluxDB server is used remotely, Ansible needs to run tasks on the
# correct host. This variable controls the task delegation to the correct
# database server.
#
# If the InfluxDB server is run locally, this should point to the inventory name
# of the current host, NOT ``localhost`` because that would delegate the tasks
# to the Ansible Controller.
influxdb__delegate_to: '{{ influxdb__server
                           if (influxdb__server|d() and
                               influxdb__server != "localhost")
                           else ansible_fqdn }}'
```
For remote connections `influxdb__server` is a FQDN, like `influxdb__server: 'influxdb-ams3-01.neuromobile.io'`, other roles can use `influxdb__server` inventory var for configure connections to InfluxDB database, example:

```sensu_server__pipe_handlers:

  - name: 'influx-db'
    command: 'sensu-influxdb-handler -d sensu'
    env_vars:
      INFLUXDB_ADDR: 'https://{{ influxdb__server }}:8086'
      INFLUXDB_USER: 'sensu'
      INFLUXDB_PASS: '{{ lookup("password",
                         secret + "/influxdb/" + influxdb__server +
                         "/credentials/sensu/password") }}'
```

And `influxdb__delegate_to` will be configured with `influxdb__server` value.

So, it will search the root password in a incorrect path: `secret/influxdb/influxdb-ams3-01.neuromobile.io/credentials/root/password`.

pedroluislopez: In other database roles use local accounts, but influxdb use a HTTP API to connect and needs found root password in the correct path.

drybjed: I see, so the DNS name is required in that case. That explains things. :-) you should mention that in the documentation just in case.

In this special case (REST API) I'm not sure if delegating the task to the InfluxDB server host has any sense, since the task will be ran over HTTP, right? So `delegate_to` functionality can probably be removed.

scibi: I would keep `delegate_to`. It can be useful when you have some firewalls between Ansible controller and the host. I have some environments where I use `ssh` with several jumphosts to access servers managed with DebOps and I'm not able to access InfluxDB port directly.

pedroluislopez: Upgraded documentation

ansible/roles/influxdb_server/tasks/main.yml
```diff
@@ -0,0 +1,122 @@
...
+             and ansible_local.influxdb_server.installed|d(false)|bool)
+
+- name: Create default admin user on first install
+  influxdb_user:
+    user_name: 'root'
+    user_password: '{{ influxdb_server__root_password }}'
+    state: 'present'
+    admin: 'yes'
+  when: not (ansible_local|d() and ansible_local.influxdb_server|d()
+             and ansible_local.influxdb_server.installed|d(false)|bool)
```
scibi, line 37: I've tried to run this role on a test machine in a tricky environment I manage. In that environment you have to use HTTP proxy to access anything outside local network. So I have:

```yaml
inventory__group_environment:
  http_proxy: 'http://proxy.address.local:8080'
  https_proxy: 'http://proxy.address.local:8080'
```

And at this point the role breaks because i tries to use http proxy to access the server.

scibi: Afer some testing I found one possible solution:

```yaml
- name: Create default admin user on first install
  influxdb_user:
    user_name: 'root'
    user_password: '{{ influxdb_server__root_password }}'
    state: 'present'
    admin: 'yes'
    proxies: 
      "http": 
      "https": 
  when: not (ansible_local|d() and ansible_local.influxdb_server|d()
             and ansible_local.influxdb_server.installed|d(false)|bool)
```

pedroluislopez: Using `delegate_to` maybe is also required disable proxies in `influxdb` role, tasks in `manage_contents.yml`, right?

An other option is clear http and https proxies environment in the task, like this:

```yaml
- name: Create default admin user on first install
  influxdb_user:
    user_name: 'root'
    user_password: '{{ influxdb_server__root_password }}'
    state: 'present'
    admin: 'yes'
  when: not (ansible_local|d() and ansible_local.influxdb_server|d()
             and ansible_local.influxdb_server.installed|d(false)|bool)```
  environment: 
      http_proxy: ''
      https_proxy: ''
```

But I'm not sure about environment precedence playboook/task. Can you test it?

pedroluislopez: It works also for tasks in `manage_contents.yml` in `influxdb` role:

```yaml
- name: Create default admin user on first install
  influxdb_user:
    user_name: 'root'
    user_password: '{{ influxdb_server__root_password }}'
    state: 'present'
    admin: 'yes'
  when: not (ansible_local|d() and ansible_local.influxdb_server|d()
             and ansible_local.influxdb_server.installed|d(false)|bool)```
  environment: 
      http_proxy: ''
      https_proxy: ''
```

But, with this solution never you will can use proxies for `influxdb_user` and `influxdb_database` tasks.

@drybjed @scibi , what do you think?

pedroluislopez: Also we could deactivate environment proxies by configuration. Example:

```yaml
influxdb__disable_http_proxies: True
```

```yaml
- name: Create databases
  influxdb_database:
    database_name: '{{ item.database | d(item.name) }}'
    state: 'present'
    hostname: '{{ influxdb__server }}'
    port: '{{ influxdb__port }}'
    ssl: '{{ influxdb__pki }}'
    password: '{{ lookup("password",
                  secret + "/influxdb/" + influxdb__delegate_to +
                  "/credentials/root/password " +
                  "length=" + influxdb__password_length) }}'
  with_flattened: '{{ influxdb__databases + influxdb__dependent_databases|d([]) }}'
  delegate_to: '{{ influxdb__delegate_to }}'
  when: ((item.database|d(False) or item.name|d(False)) and
         (item.state is undefined or item.state != 'absent'))
  environment:
      http_proxy: '{{ "" if influxdb__disable_http_proxies|d(True) else omit }}'
      https_proxy: '{{ "" if influxdb__disable_http_proxies|d(True) else omit }}'
```

scibi: Both `influxdb_database` and `influxdb_user` support `proxies` option. I thing it's better to use this option than environment variables.

I'm wondering if `influxdb__disable_http_proxies` is really necessary. Why do you use both `delegate_to` and `hostname` when managing users and databases? If you always delegate to machine running InfluxDB you could safely omit `hostname` option so it will always use `localhost`. And in this case you can always disable proxies.

drybjed: Since we are talking REST API here, host FQDN comes into play - the X.509 certificates used for the endpoint will have a specific CN which needs to be the same as the endpoint address we're using, otherwise the TLS connection will be invalid.

I think that delegation should be decoupled from the hostname and by default not enabled at all. Using the `hostname` we can specify the API endpoint address which will be reached from the remote host we are managing. Optionally like in the @scibi's case, we can use `delegate_to` to perform the actual task on a different host, which can actually reach the one specified in the `hostname`. How about this?

pedroluislopez: It's a HTTP API. With SSL it dosen't work with `localhost`, it can't validate certificate names.

For task `Create default admin user on first install` in `influxdb_server` role it isn't a problem, because SSL is not configured at these moment, but it's a problem for tasks in `influxdb` role with a SSL server configured.

pedroluislopez: Anyway, I think I can always disable http proxies. Because it configure users and databases from the same host. But DNS name have to be configured with a IP in the host, vía DNS server or `/etc/hosts`.

pedroluislopez: @drybjed 

So, i can disable http proxies for task `Create default admin user on first install` in `influxdb_server` role.

And drop delegation in `influxdb` role. Anyway, the host where you configure users and databases has to able to access InfluxDB server via DNS name.

Right?

drybjed: Yes, I think that the admin user task doesn't need to be proxied, since we are talking to the server directly anyway over SSH.

The `infludb` role can use `delegate_to` like in the case in the @scibi's environment, but I would set it as `{{ omit }}` by default. This should make it "disabled" in normal usage, and just on special case it can be enabled through the inventory.

scibi: I've run a test with a current codebase. In all tasks in `manage_contents.yml` I added:
```yaml
proxies:
  http:
  https:
```

And it works.

@drybjed what's the benefit of making `delegate_to` optional?

scibi: I think that there is another issue with making direct calls to the API from Ansible Controller:

In my case Ansible Controller is my pc. I manage many environments. Sometimes they are created just to test something. All of them use different PKI and I don't add root certificates to trusted CA. When I have to expose some service to users I use external certificates or use Let's Encrypt. I didn't test this but I guess that when `delegate_to` is omitted it will fail to connect because root CA is not trusted.

pedroluislopez: About proxies. For admin task we can disable proxies, it access via localhost. But I don't know if it's a good idea disable proxies and don't allow user to configure proxies if he needs these in `manage_contents.yml` tasks. I think that we can to allow user configure proxies in `influxdb__users` and `influxdb__databases` inventory vars.

About CA's. If `delegate_to` is disabled, the calls to the API are done from the host where runs `influxdb` playbook. Only with `influxdb__delegate_to: 'localhost'` the calls to the API are done from the Ansible controller. Anyway, InfluxDB ansible modules don't allow configure root CA, so it needs a PR to ansible modules. But otherwise, we can allow disable `validate_certs`.

A flexible proposal:

[influxdb_server]
```yaml
- name: Create default admin user on first install
  influxdb_user:
    user_name: 'root'
    user_password: '{{ influxdb_server__root_password }}'
    state: 'present'
    admin: 'yes'
    proxies:
      http:
      https:
  when: not (ansible_local|d() and ansible_local.influxdb_server|d()
             and ansible_local.influxdb_server.installed|d(false)|bool)
```

[influxdb]
```yaml
- name: Create user accounts
  influxdb_user:
    user_name: '{{ item.user | d(item.name) }}'
    user_password: '{{ item.password | d(lookup("password",
                       secret + "/influxdb/" + influxdb__delegate_to +
                       "/credentials/" + item.user | d(item.name) + "/password " +
                       "length=" + influxdb__password_length)) }}'
    grants: '{{ item.grants | d(omit) }}'
    admin: '{{ item.admin | d(False) }}'
    state: 'present'
    hostname: '{{ influxdb__server }}'
    port: '{{ influxdb__port }}'
    ssl: '{{ influxdb__pki }}'
    password: '{{ lookup("password",
                  secret + "/influxdb/" + influxdb__delegate_to +
                  "/credentials/root/password " +
                  "length=" + influxdb__password_length) }}'
    proxies: '{{ item.proxies | d(omit) }}'
    validate_certs: '{{ item.validate_certs | d(True) }}'
  with_flattened: '{{ influxdb__users + influxdb__dependent_users }}'
  delegate_to: '{{ influxdb__delegate_to }}'
  when: ((item.user|d(False) or item.name|d(False)) and
         (item.state is undefined or item.state != "absent"))
```

scibi: Maybe something like this?

```yaml
proxies: '{{ item.proxies | d(influxdb__proxies | d(omit)) }}'
```

This way it will be easier to write and use roles like `telegraf` which depend on `influxdb`.

pedroluislopez: I have made some changes about `proxies` and `delegate_to`. Can you review it?

scibi: Hi!

I've set `influxdb__proxies` in my inventory

```yaml
influxdb__proxies:
  http:
  https:
```

And role works great both with and without `influxdb__delegate_to` set.

scibi:
I've been testing both roles in another environment. First I've run `debops` without adding the host to the `debops_service_influxdb_server` and `debops_service_influxdb`. At first try it aborted because of #1295. I applied the patch and rerun `debops`. It finished without any errors. Then I run `influxdb_server` role. It finished cleanly. And then I run `influxdb` role and it aborted:

```
TASK [influxdb : Create databases] ********************************************************************************************************************************************************************************
failed: [influxdb.mydomain.io -> influxdb.mydomain.io] (item={u'name': u'telegraf'}) => {"ansible_loop_var": "item", "changed": false, "item": {"name": "telegraf"}, "msg": "HTTPSConnectionPool(host='influxdb.mydomain.io', port=8086): Max retries exceeded with url: /query?q=SHOW+DATABASES&db=telegraf (Caused by SSLError(SSLError(1, u'[SSL: WRONG_VERSION_NUMBER] wrong version number (_ssl.c:727)'),))"}
```

After some debugging I found that there is no TLS configured in `/etc/influxdb/influxdb.conf`. After some more digging I found that pki is not configured on this machine.

`/etc/ansible/facts.d/pki.fact` looks like this:
```
output = loads('''{
    "acme": "false",
    "base_path": "/etc/pki/realms",
    "ca": "CA.crt",
    "ca_realm": "domain",
    "crt": "default.crt",
    "enabled": "false",
    "hooks": "/etc/pki/hooks",
    "internal": "false",
    "key": "default.key",
    "known_realms": [],
    "path": "/etc/pki/realms",
    "pem": "default.pem",
    "realm": "domain",
    "trusted": "trusted.crt"
}''')
```

As you can see boolean values are returned as strings. So `influxdb__pki` is set to True. Some debug code:

```
TASK [influxdb : influxdb__pki] ***********************************************************************************************************************************************************************************
ok: [influxdb.mydomain.io] => {
    "msg": true
}
```

And the client role tries to use TLS when it's not configured and that's the reason of this error from the beginning.

What more interesting after running `pki` once again and no configuration changes PKI was enabled on the host and `/etc/ansible/facts.d/pki.fact` looks like this:

```
output = loads('''{
    "acme": "false",
    "base_path": "/etc/pki/realms",
    "ca": "CA.crt",
    "ca_realm": "domain",
    "crt": "default.crt",
    "enabled": "true",
    "hooks": "/etc/pki/hooks",
    "internal": "true",
    "key": "default.key",
    "known_realms": [
        "domain"
    ],
    "path": "/etc/pki/realms",
    "pem": "default.pem",
    "realm": "domain",
    "trusted": "trusted.crt"
}''')
```

Sill using strings instead of boolean values. Running `influxdb_server` role again enabled TLS in InfluxDB config and `influxdb` role worked great.

I will run some more test later: on fresh machine without te need to rerun `debops` because of error and then using current master branch to setup everything except InfluxDB.

scibi:
I've run few more tests and boolean values are saved as a text even in newest master (bug in `pki` role) and `influxdb` role doesn't work when `ansible_local.pki.enabled` is a text "false".

ansible/roles/influxdb/defaults/main.yml
```diff
@@ -0,0 +1,118 @@
...
+                                                                   # ]]]
+# SSL configuration [[[
+# ---------------------
+
+# .. envvar:: influxdb__pki [[[
+#
+# Enable or disable support for SSL in InfluxDB (using :ref:`debops.pki`).
+influxdb__pki: '{{ (True
+                   if (ansible_local|d() and ansible_local.pki|d() and
+                       ansible_local.pki.enabled|d())
```
scibi: It should be:
```
ansible_local.pki.enabled|d()|bool
```

ansible/roles/influxdb_server/defaults/main.yml
```diff
@@ -0,0 +1,382 @@
...
+      - ca-certs: '"{{ influxdb_server__pki_path + "/" + influxdb_server__pki_realm +
+                       "/" + influxdb_server__pki_ca }}"'
+
+                                                                   # ]]]
+# .. envvar:: influxdb_server__pki [[[
+#
+# Enable or disable support for SSL in InfluxDB (using :ref:`debops.pki`).
+influxdb_server__pki: '{{ (True
+                          if (ansible_local|d() and ansible_local.pki|d() and
+                              ansible_local.pki.enabled|d() and
```
scibi: It should be

```
ansible_local.pki.enabled|d()|bool and
```

ansible/roles/influxdb/defaults/main.yml
pedroluislopez: Fixed!

ansible/roles/influxdb_server/defaults/main.yml
pedroluislopez: Fixed!

pedroluislopez:
I have included "create retention policies" feature.

drybjed (approved):
Nice work with meshing two InfluxDB implementations together, @pedroluislopez :-)

PR #1285, opened 2020-03-02
Add 'clean-git' target in the Makefile

drybjed:
The 'clean-git' target will remove files and directories in the current
'git' worktree which are ignored by 'git'. This will be used to remove
any extra files for Ansible Collection generation.


nqb:
Nice !

PR #1293, opened 2020-03-06
Clarify GPL 3.0 license terms

drybjed:
Updates: #1275


SuperSandro2000:
Why not GPL-3.0-or-later?

drybjed:
I wanted to "capture" the current state of the licensing in the project. Since roles [were licensed with GPL 3.0 only license](https://github.com/debops/debops/blob/master/ansible/roles/core/COPYRIGHT), that's what should be defined via SPDX identifiers before we can change it.

You can see how "GPL 3.0 or later" license looks like [in the `bin/debops` file](https://github.com/debops/debops/blob/master/bin/debops#L13-L14).

ypid:
Looks good to me, thanks! Also thanks for fixing the license of `get-reachable-apt-proxy` and debops-api that I overlooked in my review before :+1:

SuperSandro2000:
We should really try to use GPL-3.0-or-later to avoid relicensing issues in the future.

drybjed:
That would be my intention as well. However before that happens, I plan to get as much as the copyright information (who modified files) as possible into the copyright sections.

@ypid, do you know if changing from GPL-3.0-only to GPL-3.0-or-later would require a permission from the copyright holder? i suspect that it would, I'm not sure what to ask Google about it.

SuperSandro2000:
I am pretty sure it does require permission.

ypid:
> I am pretty sure it does require permission.

Me too. With GPL-2.0-only I know this for a fact because that is used for the Linux Kernel. Switching or including GPL-3.0 would require permission from all copyright holders. This is also confirmed by https://www.gnu.org/licenses/gpl-faq.en.html#AllCompatibility.

For our case, the GPL FAQ seems to suggest that we could put new releases under GPL-3.0-or-later but the or-later would not have any effect. I guess only with permission it takes effect.

PR #1295, opened 2020-03-10
[nullmailer] Don't dig empty DNS domains

drybjed:
If the 'ansible_domain' variable is empty (no DNS domain configured on
a host), don't use the 'dig' lookup plugin to find the '_smtp._tcp' SRV
resource record. This fixes the "dns.resolver unhandled exception A DNS
label is empty." error during Ansible execution.


reixd:
I think it is time to make a dns resolve wrapper plugin for DebOps. It could help to hide the dig commands from the config.

drybjed:
I'm not exactly sure what you have in mind, perhaps you could expand the suggestion a bit?

PR #1297, opened 2020-03-14
[owncloud] No parameter validate in module file.

paulinsche:
ansible file module has no parameter validate. Looks like a cut&paste error from block above.


drybjed (approved):
Thanks for the fix! :-)

PR #1301, opened 2020-03-17
[nginx] Enable TLSv1.3 for nginx 1.13.0 and up

imrejonk:
The nginx web server supports TLSv1.3 since version 1.13.0. These
changes enable TLSv1.3 by default when nginx 1.13.0+ is used.

Note that TLSv1.3 only works in nginx with OpenSSL 1.1.1. OpenSSL also
needs to be build with TLSv1.3 support (which is the case in Debian
stable).


drybjed (approved):
Nicely done, thanks. :+1:

PR #1302, opened 2020-03-17
[nginx] Set HSTS header regardless of status code

imrejonk:
The current Strict-Transport-Security header in the default.conf
template is only added when the response code equals 200, 201 (1.3.10),
204, 206, 301, 302, 303, 304, 307 (1.1.16, 1.0.13), or 308 (1.13.0).

For added security, this change ensures that the HSTS header is always
set, regardless of the response code. This is done by adding the
'always' parameter to the HSTS add_header directive.

I believe that this change is compliant with RFC 6797.

Ref: http://nginx.org/en/docs/http/ngx_http_headers_module.html#add_header

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -182,6 +182,12 @@ General
   in case that the DNS infrastructure is not configured. This avoids 60s
   timeouts during Ansible fact gathering in such cases.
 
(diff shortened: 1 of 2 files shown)


drybjed (approved):
Sounds good to me.

PR #1303, opened 2020-03-19
[gunicorn] Add secret role

thiras:
I've tried to integrate the secret role into gunicorn playbook. I've taken minio playbook as en example since it's relatively newly written.

### DIFF
--- a/ansible/playbooks/service/gunicorn.yml
+++ b/ansible/playbooks/service/gunicorn.yml
@@ -13,12 +13,6 @@
                    | combine(inventory__group_environment | d({}))
                    | combine(inventory__host_environment  | d({})) }}'
 
-  pre_tasks:
-
-    - import_role:
-        name: 'gunicorn'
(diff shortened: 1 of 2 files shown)


thiras:
I'm not sure if both pre role and role are required but nevertheless it worked for my case.

thiras:
I've noticed that the playbook runs twice with a secret role as pre_task. So I've removed it.

drybjed:
You should add the `secret` role in the task list instead, via the `import_role` Ansible module - see [the example in the `minio` role](https://github.com/debops/debops/blob/master/ansible/roles/minio/tasks/main.yml#L9-L10). That way users don't have to remember about including the `secret` role in all playbooks that use the `gunicorn` role.

drybjed (approved):
The change looks OK, not sure why Travis-CI does not want to test this PR now. I'll cherry pick the final commit to have a cleaner history. Thanks!

PR #1305, opened 2020-03-21
[postfix] only restrict submission port to sasl_authenticated

TuxCoder:
I got an issue on a users client, that it get denied because of an invalid host name in the helo message on the submission port, as the client is authenticated it should ignore this error(IMHO).
used client with the bug: debian - claws mail 3.17.3

this comments in the additional `mua_*` variables and set them with the proper values
and set in the main.cf the corresponding variables.

### DIFF
--- a/ansible/roles/postfix/defaults/main.yml
+++ b/ansible/roles/postfix/defaults/main.yml
@@ -712,22 +712,22 @@ postfix__restrictions_maincf:
     weight: 70
     value:
 
-       - name: 'mua_sender_restrictions'
-         weight: -300
-       - name: 'mua_client_restrictions,'
-         weight: -200
-       - name: 'mua_helo_restrictions'
-         weight: -100
+      - name: 'mua_sender_restrictions'
+        weight: -300
+      - name: 'mua_client_restrictions,'
+        weight: -200
+      - name: 'mua_helo_restrictions'
+        weight: -100
 
   - name: 'mua_client_restrictions'
     section: 'restrictions'
     weight: 80
     value:
 
-       - name: 'permit_sasl_authenticated'
-         weight: -200
-       - name: 'reject'
-         weight: -100
+      - name: 'permit_sasl_authenticated'
+        weight: -200
+      - name: 'reject'
+        weight: -100
 
 
   - name: 'mua_sender_restrictions'
@@ -735,19 +735,18 @@ postfix__restrictions_maincf:
     weight: 90
     value:
 
-       - name: 'permit_sasl_authenticated'
-         weight: -200
-       - name: 'reject'
-         weight: -100
+      - name: 'permit_sasl_authenticated'
+        weight: -200
+      - name: 'reject'
+        weight: -100
 
   - name: 'mua_helo_restrictions'
     section: 'restrictions'
     weight: 100
     value:
 
-       - name: 'permit'
-         weight: -100
-
+      - name: 'permit'
+        weight: -100
 
                                                                    # ]]]
 # .. envvar:: postfix__maincf [[[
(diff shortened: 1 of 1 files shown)


ansible/roles/postfix/defaults/main.yml
```diff
@@ -866,15 +908,12 @@ postfix__original_mastercf:
...
         value: '$mua_client_restrictions'
-        state: 'comment'
 
       - name: 'smtpd_helo_restrictions'
         value: '$mua_helo_restrictions'
-        state: 'comment'
 
       - name: 'smtpd_sender_restrictions'
         value: '$mua_sender_restrictions'
-        state: 'comment'
```
drybjed: Instead of modifying the `postfix__original_mastercf` variable which contains the values that come with the Debian `postfix` package, you should override them in the `postfix__default_mastercf` variable. This way the changes applied by DebOps should be easier to spot. You can see how it's done in the `postfix__default_maincf` variable for comparsion.

ansible/roles/postfix/defaults/main.yml
```diff
@@ -703,6 +703,48 @@ postfix__restrictions_maincf:
...
+
+  - name: 'mua_helo_restrictions'
+    section: 'restrictions'
+    weight: 100
+    value:
+
+       - name: 'permit'
+         weight: -100
+
+
```
drybjed, line 750: One empty line should be removed from here to maintain whitespacing.

drybjed (changes requested):
Good idea, makes sense to me. The Travis-CI detected some indentation issues, can you check them? Also, you shoul change the original values in a separate variable instead of directly in the original variable; see the comment for explanation.

TuxCoder:
Thanks for the feedback, going to look into it these days.
Yeah forgot about the testing system, was so a small change and still some styling issue :)

TuxCoder:
I will run some tests, but looks now ok, a bit unsure if `status: present` is enough.

As debian stable supports docker again, local testing is now not so complicated. :)
Will also deploy on a machine again, will report in 1 or 2 days.
I had some problems with the backup and need to wait until that is resynced.

ansible/roles/postfix/defaults/main.yml
TuxCoder: have refactor that

TuxCoder:
had a bit of a mess with my git,... so had to push force a rebased version on the current master branch.
From my view it looks now good, do you have any improvements?

TuxCoder:
I recognized after some testing, that I have a config collision with: `mua_sender_restrictions = reject_authenticated_sender_login_mismatch` in the `postconf` role.

Probably do not set `mua_sender_restrictions` or move all to `postconf` ?

drybjed:
The `mua_*_restrictions` should most likely be defined in the `postfix` role initially, then the `postconf` role can modify them further if needed.

I'm not sure if the current PR contents are correct, but I know what you are trying to do. I'll test the config in my development environment and add tweaks where necessary.

TuxCoder:
I think I applied now the current master with this patch on my mailserver,
My last deployed environment (postfix/postconf/postwhite) was a bit older and some stuff in debops.postfix/postconf changed.
Sorry for not testing with the current version....


The result is that this two get overwritten by debops.postconf:
Also with this patch, but the values are ok
```
smtpd_helo_restrictions=
smtpd_sender_restrictions=reject_authenticated_sender_login_mismatch
```
The only value that not gets overwritten is `smtpd_client_restrictions` and that is by default empty/or gets set by this patch to:
`smtpd_client_restrictions=$mua_client_restrictions`


IMHO should this be ok values and look like to work, I haven't tested with this specific client that causes the bug, will try to do in the following weeks.


[Note] because it's not so easy to see in the PR the seted `mua_*` variables
```
mua_client_restrictions = permit_sasl_authenticated, reject
mua_sender_restrictions =
    permit_sasl_authenticated,
    reject

mua_helo_restrictions = permit

```

So updating to postconf solved the problem.

**Edited:**
Clarifying behavior

So Probably can be closed or have a "better" default behavior with the `postfix` role alone.

drybjed:
@TuxCoder does this mean that you don't need this patch anymore? Since the current default seems to be a bit more clear to me and closer to the "vanilla" Postfix setup. Adding `mua_*_restrictions` might be confusing so I would like to avoid this for now.

TuxCoder:
@drybjed  yes, updating to the newer `postconf` role solved the problem (at least I think).
Otherwise I can open a new request.

Tanks for helping!

PR #1307, opened 2020-03-23
[pki] Allow adding contact details to ACME account

imrejonk:
These changes allow specifying contact URLs
(e.g. 'mailto:webmaster@example.com') for acme-tiny to use while
registering a new ACME account. Let's Encrypt for example uses this
information to notify you about expiring certificates and emergency
revocation.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -50,6 +50,13 @@ New DebOps roles
 
   .. __: https://influxdata.com/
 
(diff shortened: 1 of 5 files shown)


drybjed (approved):
Sounds good, thanks. I would proably try and handle the `mailto:` prefix in a list via `regex_replace` filter, but this can be done when the `pki` role is eventually refactored.

PR #1308, opened 2020-03-25
[debops.python] Install Python modules

thiagotalma:


drybjed (changes requested):
The idea sounds good, although system-wide Python module installation should be discouraged, and virtualenv/unprivileged installs should be preferred. But for the DebOps `python` role system-wide install is probably best, the other modes are more complicated to implement properly from the `root` account.

However, the `python` role supports both Python 2.7 and Python 3.x, and you need to account for this in the variables and tasks. Use two sets for module lists, one for Py2, and another for Py3 version. And, you need to add another task that will use `pip3` to install the requested modules. Don't forget to enable/disable them conditionally depending on wether each Python version is enabled.

thiagotalma:
I agree with you. We should not encourage this type of installation.

Issue #953, opened 2019-07-30
[FEATURE] add support for external components to debops.prosody


sjthespian:
The debops.prosody role supports components, but it looks like it is missing support for external components (https://prosody.im/doc/components#adding_an_external_component). It needs to provide the connection parameters for external components like:
```
Component "msn.example.org"
         component_secret = "mysecretcomponentpassword"
```

I don't see any way to do this with the existing role. This patch seems to work, but given the problem I mention below I can't test it...
```
diff --git a/ansible/roles/debops.prosody/templates/etc/prosody/prosody.cfg.lua.j2 b/ansible/roles/debops.prosody/templates/etc/prosody/prosody.cfg.lua.j2
index 2edcfe51d..4deb8fd8e 100644
--- a/ansible/roles/debops.prosody/templates/etc/prosody/prosody.cfg.lua.j2
+++ b/ansible/roles/debops.prosody/templates/etc/prosody/prosody.cfg.lua.j2
@@ -47,4 +47,7 @@ VirtualHost "{{ item.name }}"

 {% for _component in prosody__combined_config_components %}
 Component "{{ _component.domain }}" {{ _component.params }}
+    {% if _component.external_params is defined %}
+    {{ lua_print(_component.external_params) }}
+    {% endif %}
 {% endfor %}
```

I am also running into issues with sql connect parameters, I cannot come up with a valid syntax in the `prosody__host_config_global` block that doesn't fubar the config.  For example:
```
  sql:
  - 'driver: "MySQL", database: "prosody", username: "prosody", password: "{{ lookup("password", secret + "/credentials/prosody/sql/password") }}", host: "localhost"'
```
ends up being quoted in the generated file, but without quotes in the yaml the parser fails. The above generates:
```
sql =   {
    "driver: "MySQL", database: "prosody", username: "prosody", password: "xxxx", host: "localhost"";
  }
```
but it needs to look like:
```
sql = { driver = "MySQL", database = "prosody", username = "prosody", password = "xxxx", host = "localhost" }
```

drybjed:
Try writing the configuration like this:
```yaml
prosody__host_config_global:
  sql:
    driver: "MySQL"
    database: "prosody"
    username: "prosody"
    password: "xxxx"
    host: "localhost"
```
As for the support for component configuration, I'm not sure if that's implemented at the moment. The `debops.prosody` role was written by @TuxCoder, perhaps he will be able to shed some light on this.

TuxCoder:
[...]

PR #1312, opened 2020-04-05
debops.prosody: split up prosody config

reixd:
As discussed in #953 I propose to split the prosody config in three components:

-  global config
- virtual hosts config
- components config

So, we have the main config /etc/prosody/prosody.cfg.lua. There are all global config parameters needed.
Then every VirtualHost and Component has it own .lua file under /etc/prosody/conf.avail with the corresponding symlinks to /etc/prosody/conf.d.

New introduced variables analogous to the nginx server config, allow to specify config parameters for
each virtual host and component.

The PR is a work in progress.


ansible/roles/prosody/templates/etc/prosody/conf.avail/VirtualHost.j2
```diff
@@ -0,0 +1,31 @@
...
+-- {{ ansible_managed }}
+{% import 'templates/import/debops__tpl_macros.j2' as debops__tpl_macros with context %}
+{% import 'templates/import/prosody_macros.j2' as prosody_macros without context %}
+-- Virtual Host '{{ item.name }}' Configuration File
+
+
+{% set prosody__tpl_pki_realm = item.pki_realm %}{#| d((debops__tpl_macros.get_realm_yaml_list(item.name, prosody__pki_realm) | from_yaml)[0]) %#}
+{% set prosody__tpl_pki_realm_path = prosdoy__pki_realm_path + "/" + prosody__tpl_pki_realm %}
+VirtualHost "{{ item.name }}"
+  enabled = {{ item.enabled }}
```
TuxCoder: enabled should have default true

TuxCoder (commented):
after looking a bit longer into the code,
I think `debops__tpl_macros.merge_dict(current_dict, to_merge_dict, dict_key='name')` is exactly what you are searching for to merge dicts.

ansible/roles/prosody/defaults/main.yml
```diff
@@ -283,89 +283,165 @@ prosody__combined_config_global: '{{ prosody__default_config_global | combine(
...
+#           muc_log: 'sql'
+#     params:
+#       - key1: value1
+#       - key2: value2
+
 
+# .. envvar:: prosody__default_components [[[
+#
+# Variable which contains default components used by prosody
+prosody__default_components:
```
TuxCoder, line 411: Good job, I also think the domain has to be unique between all components.
That's a pretty easy structure and it fits the needs!

ansible/roles/prosody/defaults/main.yml
```diff
@@ -283,89 +283,165 @@ prosody__combined_config_global: '{{ prosody__default_config_global | combine(
...
-#  XEP-0045
-prosody__config_muc:
-  - domain: 'conference.{{ prosody__domain }}'
-    params: '"muc"'
+# Variable which contains virtual hosts used by prosody,, combined from all other
+# variables.
+prosody__combined_virtual_hosts: '{{ prosody__default_virtual_hosts
+                                    | combine(prosody__virtual_hosts)
+                                    | combine(prosody__host_virtual_hosts)
+                                    | combine(prosody__group_virtual_hosts) }}'
```
TuxCoder: The ordering from the combined is important, https://docs.ansible.com/ansible/latest/user_guide/playbooks_filters.html#combining-hashes-dictionaries

The first get overwritten by the following.
So would suggest, in the order `default`, `all`,`group`,`host`

ansible/roles/prosody/defaults/main.yml
```diff
@@ -283,89 +283,165 @@ prosody__combined_config_global: '{{ prosody__default_config_global | combine(
...
-prosody__combined_config_components: '{{ prosody__default_config_components
-                                       + prosody__config_components
-                                       + prosody__group_config_components
-                                       + prosody__host_config_components  }}'
+# Variable which contains components used by prosody,, combined from all other
+# variables.
+prosody__combined_components: '{{ prosody__default_components
+                                  | combine(prosody__components)
+                                  | combine(prosody__host_components)
+                                  | combine(prosody__group_components) }}'
```
TuxCoder: same as for `prosody__combined_virtual_hosts`

TuxCoder (commented):
Some small comments.

Overall looks good, is it ready for testing
or you want to do some changes first?

ansible/roles/prosody/defaults/main.yml
reixd: good catch.

drybjed:
Nice work on the PR! Although I would like to see the configuration converted to the `parse_kv_items` format, since using dictionaries directly makes overriding the changes conditionally harder, IMO. Also, you should update the copyright notices and add your own lines where appropriate. :-)

Let me know when the PR is ready for merge.

reixd:
@drybjed please check the new changes as you asked for.

reixd:
@TuxCoder regarding the XMPP Compliance (https://compliance.conversations.im/) I think it would fit better to make a new PR, if changes in this direction are needed.
This PR was all about new "backend infrastructure".

reixd:
@drybjed any news on this?

PR #1313, opened 2020-04-08
mosquitto: Ignore meta variables for bridge file generation

pokerazor:
To set the linux file owner group and permission mode, one can set the variables 'mode' and 'group' see (https://github.com/debops/debops/blob/master/ansible/roles/mosquitto/tasks/main.yml#L172), but those variables need also to be ignored when generating the configfile from the template, otherwise the resulting configfile will be invalid.


drybjed (approved):
Sounds like a good idea, I hope that these parameters are not Mosquitto configuration options. Thanks.

PR #1314, opened 2020-04-08
[nullmailer] Make nullmailer-smtpd dual stacked

imrejonk:
With these changes, the Nullmailer SMTPD service will listen on both 127.0.0.1 and ::1.
It used to only listen on ::1. This is implemented by adding a second
xinetd.conf file so that xinetd can be configured to bind to another IP
address.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -123,6 +123,13 @@ General
 
 - TLSv1.3 is now enabled by default for nginx version 1.13.0 and up.
 
+:ref:`debops.nullmailer` role
+'''''''''''''''''''''''''''''
+
+- The Nullmailer smtpd service can now listen on both IPv4 and IPv6 addresses.
+  It listens on both loopback addresses by default, where it used to only
+  listen on the IPv6 loopback address.
+
 :ref:`debops.owncloud` role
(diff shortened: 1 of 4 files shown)


drybjed (approved):
Nicely done, thanks! I'll backport it to the older releases as well.

imrejonk:
Hi @drybjed, it looks like this change is not present in DebOps 2.0.4, 1.2.4, 1.1.10 1.0.18. Did you forget to backport it?

drybjed:
Seems like it, thanks for noticing. I'll do the backport right away, but tagged releases will be a bit later.

drybjed:
@imrejonk I remember now why I didn't backport these changes - the `dpkg_divert` role that is used in `nullmailer` was added in `stable-2.1` branch so the patch couldn't be backported cleanly. The current policy states that everything should go through `master` first and changes are backported if they can be cleanly applied, so I don't think that this change should go to previous releases.

imrejonk:
Ah, of course! I remember having to do something with `dpkg_divert` as well when working with the `nullmailer` role. That indeed won't make a clean backport. Thanks for clarifying.

PR #1317, opened 2020-04-15
[cron] Ensure enironment name & value is a list

tumd:
... and select the first item in the list.

Fixes the error;
`The task includes an option with an undefined variable. The error was: dict_keys object has no element 0`
That I get using Ansible 2.9.6 and on Python 3.7.6. Not sure when or what causes the issue.


drybjed (approved):
Nicely done, thanks! I'll backport it as well.

PR #1318, opened 2020-04-18
[debops.gitlab] Bump Gitlab to latest stable v12.9

reixd:
As discussed in the IRC, using a newer golang version (>=1.14) from buster-backports I was able to install the latest Gitlab stable version.

### DIFF
--- a/ansible/roles/gitlab/defaults/main.yml
+++ b/ansible/roles/gitlab/defaults/main.yml
@@ -1130,6 +1130,7 @@ gitlab__etc_services__dependent_list:
 gitlab__apt_preferences__dependent_list:
 
   - package: 'git git-*'
+    state: 'present'
     backports: [ 'jessie' ]
     reason:  'Meet version requirement of GitLab 8.17 (Git version >= 2.7.3) on Debian Jessie.'
     by_role: 'debops.gitlab'
--- a/ansible/roles/gitlab/defaults/main.yml
+++ b/ansible/roles/gitlab/defaults/main.yml
@@ -1134,6 +1134,12 @@ gitlab__apt_preferences__dependent_list:
     reason:  'Meet version requirement of GitLab 8.17 (Git version >= 2.7.3) on Debian Jessie.'
     by_role: 'debops.gitlab'
 
+  - package: 'golang-go golang-src'
+    state: 'present'
+    backports: [ 'buster' ]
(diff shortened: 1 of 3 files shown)


reixd:
Uhm, apparently Travis does not like to install the `latest` packages: https://travis-ci.org/github/reixd/debops/jobs/676743466#L519
In some ways it is right.

@drybjed any idea how to guarantee, that the `golang-go` pkg is updated to the latest backports version on existing installations?
Upgrade notes and Changelog is a way, however I think, error prone. And if `golang` is in the wrong version, the whole installation breaks and you need to restore the Server/VM from backup.

drybjed:
This would be the wrong way to do it. The modifications to the `golang` packages should be done in the `golang` role; namely you want to add `buster` in the [list of backports to enable](https://github.com/debops/debops/blob/master/ansible/roles/golang/defaults/main/dependent.yml#L48). I would probaly add `bionic` if that is available, not sure.

With this, installation of GitLab 2.19 on a fresh host/container should work fine. As for the existing installations, manual `apt install golang` should upgrade the packages to the backported versions, and then the playbook run should theoretically go as expected. i haven't tested that yet.

Of course all this needs to be mentioned in the Changelog and the upgrade notes. I plan to release new DebOps versions in the next week, perhaps we could wait with these changes before the DebOps v2.1.0 is released?

reixd:
> This would be the wrong way to do it. The modifications to the `golang` packages should be done in the `golang` role; 

This would mean, that `golang-go` would be always installed from backports in `Buster` (and `bionic`)?
If this is wished, then I will change the code accordingly.


> As for the existing installations, manual `apt install golang` should upgrade the packages to the backported versions, and then the playbook run should theoretically go as expected. i haven't tested that yet.

In theory true, however, maybe we should add a dependency version check to the gitlab role and throw an error. Otherwise forgetting to upgrade `golang-go` will destroy the Gitlab installation.


> Of course all this needs to be mentioned in the Changelog and the upgrade notes. 
Of course

> I plan to release new DebOps versions in the next week, perhaps we could wait with these changes before the DebOps v2.1.0 is released?

This are kind of breaking changes, so yes. The changes can wait.

drybjed:
I did the change in a slightly different way via the https://github.com/debops/debops/pull/1332 PR. The `golang` APT packages need to be upgraded system-wide, so the preference has to be done in the `golang` role, not specifically the `gitlab` role. Otherwise you might end up with a system that has Go applications compiled using different Go runtimes.

reixd:
Works for me

PR #1326, opened 2020-04-29
Enforce role doc order

tasosalvas:
# Standardizing the order of role documentation
I've listed details on the specific order in the commit itself, it's pretty much what we discussed on irc.
I've checked how the changed order affects manpages, too. Nothing to report, really.

I'm not sure it's a good idea that some content is excluded from `man`.
Especially the default vars are pretty much the core of a role's doc - but also generally there's an assumption of completeness.

## Misc other stuff
* Pruned role `upgrade.rst`s that remained from the monorepo move.
* Changed section headers to bold text in `owncloud` and `cryptsetup`'s `man_description.rst`s,
  as their inclusion was breaking the html's menu hierarchy. (just click on them on the role index)
* Added a section in `contributing-docs.rst` documenting role doc structure.

## Saved the orphans
The word "orphan:" renders when documents with the `:orphan:` header are included.
The issue appeared in `owncloud`, `cryptsetup` and `apache`, both on the manpages and html. (i.e. see owncloud/getting-started)

I moved `cryptsetup`'s refs inside the file that used them.
`apache` and `owncloud` are actually sharing the refs between multiple documents,
so I hacked around it by adding `:start-line:` in the includes
(`apache` was actually already doing this in one place).

## golang is such a rebel
All its documentation lives in the role as `defaults/main/environment`,
`defaults/main/packages` and `defaults/main/dependent`. It reads nicely,
but apart from `dependent` it doesn't really map to `defaults/main` and `variable-details`.

Its manpage ends up only containing `getting-started`. I didn't touch any of it.


drybjed (dismissed):
Nice work, thanks! I don't plan to backport this since number of changed files is so large.

drybjed:
As for the `golang` documentation being a mess - usually when I modify or make a new role, I like to try out some new idea before applying it to all the other relevant roles. In this particular case I wanted to split the `defaults/*.yml` files in the role, and in the documentation combine the details of relevant defaults per-file to have them in the same place, so to speak. The result was kinda OKay-ish in the HTML documentation and a horrible disaster in the manual pages (that was before manpages were a thing though), so I don't plan to do this to any more roles. And the `golang` docs should probably be changed to move details back to one file.

tasosalvas:
Yeah I don't think backporting this is important, even if the number of files didn't make it fussy. It's only doc changes and a single comment in `owncloud/defaults/main.yml`.

I never called `golang`'s doc a mess :). It's just different, it reads pretty ok.

I do think that in a project with such diverse goals, every bit of standardization we can arrange helps us preserve the mental bandwitdth to deal with higher-level stuff like boxes and stacks.

I have never messed with Go, and I'm not very confident with my understanding of the structure of ansible/debops roles yet. Plus it looked a bit harder at 2AM. :)

Regarding the attempted categorization, `environment` seems to be the only one that wouldn't really map to many other roles. `dependent` and `packages` are pretty universal. I do think, however, that from a usage viewpoint, having all top-level role variables in a single document one can mindlessly `Ctrl+F` in is superior.

Having a look at it today, I think I could handle dumping all documents into a `/main.yml` it if you want me to, since only the sphinx includes in those files seem to be path-dependent.

tasosalvas:
Hey,

So I had a go (no pun intended :p) at rearranging the role's defaults.
Being the only irregular thing in the role doc, I just couldn't let it ... go. (ok intended this time xD)

Overall I think things ended up fitting nicely in the common role structure:
* `environment`, `packages` and `dependent.rst` are all `main.yml` now
* `packages.rst` included an `includes/defaults-packages.rst` this became `usage-examples.rst` and `defaults-detailed.rst`.

The manpage now pretty much contains the same stuff it did before, just labeled and included in the same way as the other roles.

Also added `fsfe-reuse` to the testing documentation.

And this time I didn't spot anything else so I think I'm happy with this PR if you are.

drybjed (approved):
Nice work, thanks. There are some issues with the whitespace, I'll fix that before merging.

tasosalvas:
Happy to help. Tell me what needs fixing, if you want, so I can save you the trouble, make sure my next stuff is up to the project's standards, and maybe document it as well.

PR #1335, opened 2020-05-06
[ownCloud] Updates ownCloud to 10.4 and fixes the apt repo url

Gomez:
This patch updates ownCloud to 10.4 and fixes the apt repo url

### DIFF
--- a/ansible/roles/owncloud/defaults/main.yml
+++ b/ansible/roles/owncloud/defaults/main.yml
@@ -435,7 +435,7 @@ owncloud__variant_name_map:
 #
 # For ownCloud refer to the `ownCloud Maintenance and Release Schedule <https://github.com/owncloud/core/wiki/Maintenance-and-Release-Schedule>`_
 # and the `package index <https://download.owncloud.org/download/repositories/>`_ for more details.
-owncloud__release: '{{ "10.3"
+owncloud__release: '{{ "10.4"
                        if (owncloud__variant == "owncloud")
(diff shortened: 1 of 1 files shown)


ansible/roles/owncloud/defaults/main.yml
```diff
@@ -499,7 +497,7 @@ owncloud__src_remote_dir: '{{
 # .. envvar:: owncloud__apt_repo_source
 #
 # APT ``sources.list`` URL of the ownCloud ``.deb`` repository.
-owncloud__apt_repo_source: '{{ "deb http://" + owncloud__apt_repo_base + "/" +
+owncloud__apt_repo_source: '{{ "deb https://" + owncloud__apt_repo_base + "/" +
```
SuperSandro2000, line 500: We have ``apt-transport-https``installed right?

Gomez: Not on my clean buster, but the apt version is 1.8.x and can handel https. Should i change to http, for older debian?

drybjed: The `apt` role [will install it by default on older OS releases](https://github.com/debops/debops/blob/master/ansible/roles/apt/defaults/main.yml#L105-L109), on newer ones it is not needed since APT handles `https://` URLs natively. I think you're good here.

drybjed (approved):
Looks good, thanks!

Issue #1338, opened 2020-05-09
debops-defaults failed if external command 'view' doesn't exist


jeansorgemoel:
On my configuration, when I launch debops-defaults, command crash.
Probably, I have bad configuration, but this crash is very strange.

<pre>
$ debops-defaults 
Traceback (most recent call last):
  File "/usr/local/bin/debops-defaults", line 133, in <module>
    main(args.role)
  File "/usr/local/bin/debops-defaults", line 105, in main
    require_commands(cmd_args[0])
  File "/usr/local/lib/python3.7/dist-packages/debops/cmds/__init__.py", line 83, in require_commands
    error_msg("%s: command not found" % name)
  File "/usr/local/lib/python3.7/dist-packages/debops/cmds/__init__.py", line 67, in error_msg
    print(SCRIPT_NAME+':', severity+':', message)
TypeError: a bytes-like object is required, not 'str'
</pre>

When I replace print to write on file, I have this message :
debops-defaults:Error:view: command not found

We can do that to solve issue:
<pre>
--- old/debops/cmds/__init__.py	2020-05-09 14:02:11.136689203 +0000
+++ new/debops/cmds/__init__.py	2020-05-09 14:02:50.203355905 +0000
@@ -64,7 +64,10 @@
     """
     Display error message and exit
     """
+    old = sys.stdout
+    sys.stdout = sys.__stdout__
     print(SCRIPT_NAME+':', severity+':', message)
+    sys.stdout = old
     if severity == "Error":
         raise SystemExit(1)
</pre>

PR #1340, opened 2020-05-09
Fix issue #1338 - debops-defaults failed if command 'view' not exist

jeansorgemoel:
Fix issue #1338 - debops-defaults failed if external command 'view' doesn't exist


drybjed (approved):
Thanks for fixing the problem with displaying the error message. :-)

PR #1344, opened 2020-05-14
[bin/debops-init] fix project paths in .gitignore and .gitattributes

### LINKED ISSUE #1343: [bin/debops-init] .gitignore paths that should be absolute are relative
Heh, unless I'm blanking out on syntax or I'm missing some advanced use-case.
I did check my configs and the relevant [git-scm](https://git-scm.com/docs/gitignore) page, though.
Maybe it's something we all fix in our projects and never talk about? ;)

So, these are the debops project related entries:
```ini
\
debops
ansible.cfg
ansible/secret
secret

# Remove the line below to store encrypted secrets in
# the git repository
.encfs.secret

roles
playbooks
```

In my projects I have to prepend a `/` to all of these. (and every time I ponder on whether the mysterious backslash on the first line actually does something). :)

Git-scm, btw, says:
>If there is a separator at the beginning or middle (or both) of the pattern, then the pattern is relative to the directory level of the particular `.gitignore` file itself. Otherwise the pattern may also match at any level below the `.gitignore` level.

My main issue was that `roles` and `playbooks` in `/ansible/` were being ignored. The paragraph also implies that out of all these entries, `ansible/secret` is actually treated as absolute ("relative to the .gitignore dir").

`.gitattributes` could also do with a couple of slashes.

Anyway, I can go fix it, just thought I'd ask here first in case I missed something.

tasosalvas:
* Fixes #1343
* Adds a few comments and a link to the relevant doc

Heh I understood what the `\` before the first line change of the `r"""` was trying to do.
Sadly `r` strings only respect escaping their closing quote marks.


tasosalvas:
Lol sorry I did it in nano for some inexplicable reason

drybjed (approved):
Thanks for fixing the issues. I'll backport this to the stable releases as well.

PR #1345, opened 2020-05-14
[journald] Don't enable persistent logs by default

drybjed:


drybjed:
The tests seem to pass fine, not sure why Travis does not update the status. Merging.

PR #1347, opened 2020-05-19
[mailman3] Add mailman3 role

imrejonk:
The `debops.mailman3` role manages Mailman Suite installations, which consist of the Mailman Core backend, Postorius web frontend and Hyperkitty list archiver. The role manages the database with debconf, integrates with NGINX and Postfix and configures LDAP authentication for superusers.

I opted for writing a new role for Mailman 3 instead of updating the existing `debops.mailman` role, because of the many backwards-incompatible changes. The Debian Mailman Team made the same choice.

We're using this role at CipherMail to manage https://lists.ciphermail.com/ with a PostgreSQL database. Support for MariaDB and SQLite 3 has been tested in a Vagrant box.

### DIFF
--- a/ansible/roles/mailman3/templates/etc/ansible/facts.d/mailman3.fact.j2
+++ b/ansible/roles/mailman3/templates/etc/ansible/facts.d/mailman3.fact.j2
@@ -1,0 +1,20 @@
+#!{{ ansible_python['executable'] }}
+
+# Copyright (C) 2020 CipherMail B.V. <https://www.ciphermail.com/>
+# Copyright (C) 2020 DebOps <https://debops.org/>
+# SPDX-License-Identifier: GPL-3.0-only
+
+# {{ ansible_managed }}
+
+from json import dumps
+import apt
+
+cache = apt.Cache()
+output = {}
+
+try:
+    output['installed'] = cache['mailman3'].is_installed
+except KeyError:
+    output['installed'] = False
+
+print(dumps(output, sort_keys=True, indent=4))
--- a/ansible/roles/mailman3/templates/etc/ansible/facts.d/mailman3.fact.j2
+++ b/ansible/roles/mailman3/templates/etc/ansible/facts.d/mailman3.fact.j2
@@ -1,20 +1,22 @@
 #!{{ ansible_python['executable'] }}
 
-# Copyright (C) 2020 CipherMail B.V. <https://www.ciphermail.com/>
-# Copyright (C) 2020 DebOps <https://debops.org/>
+# Copyright (C) 2018 Maciej Delmanowski <drybjed@gmail.com>
+# Copyright (C) 2018 DebOps <https://debops.org/>
 # SPDX-License-Identifier: GPL-3.0-only
 
 # {{ ansible_managed }}
 
 from json import dumps
-import apt
+import os
 
-cache = apt.Cache()
-output = {}
 
-try:
-    output['installed'] = cache['mailman3'].is_installed
-except KeyError:
-    output['installed'] = False
+def cmd_exists(cmd):
+    return any(
+        os.access(os.path.join(path, cmd), os.X_OK)
+        for path in os.environ["PATH"].split(os.pathsep)
+    )
+
+
+output = {'installed': cmd_exists('mailman')}
 
 print(dumps(output, sort_keys=True, indent=4))
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -953,6 +953,15 @@ stages:
     JANE_DIFF_PATTERN: '.*/roles/mailman/.*'
     JANE_LOG_PATTERN: '\[debops\.mailman\]'
 
+'mailman3 role':
+  <<: *test_role_3rd_deps
+  variables:
+    JANE_TEST_FACT: 'mailman3.fact'
+    JANE_TEST_PLAY: '${DEBOPS_PLAYBOOKS}/service/mailman3.yml'
+    JANE_INVENTORY_GROUPS: 'debops_service_mailman3'
+    JANE_DIFF_PATTERN: '.*/roles/mailman3/.*'
+    JANE_LOG_PATTERN: '\[debops\.mailman3\]'
+
 'mariadb role':
   <<: *test_role_1st_deps
   variables:
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -54,6 +54,11 @@ New DebOps roles
   used to install the InfluxDB time series database service and manage its
   databases and users, respectively.
 
+- The :ref:`debops.mailman3` role manages Mailman Suite installations, which
+  consist of the Mailman Core backend, Postorius web frontend and Hyperkitty
+  list archiver. The role manages the database with debconf, integrates with
+  NGINX and Postfix and configures LDAP authentication for superusers.
+
 :ref:`debops.pki` role
 ''''''''''''''''''''''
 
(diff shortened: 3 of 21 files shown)


drybjed:
Dang, and I just finally made my own rewrite of the `mailman` role work with Mailman3... Well, let's see what you cam up with. :-)

As for naming it `mailman3`... DebOps initially started PHP support with a `php5` role. When Debian switched to PHP 7.x, that name was not relevant anymore so i rewrote the role to support new package layout and named it `php` instead. Since then I'm wary of naming roles with numerical suffixes.

As I understand, Mailman 2.x is in maintenance mode in Debian and might be removed at any point in the future. The same seems to be true on the upstream side. That's why I decided to rewrite the `mailman` role instead of adding a new one. What do you think?

drybjed:
After reviewing... LDAP - cool; systemd change - COOOL, now I know why `mailman3` fails on startup all the time. I might grab that, if you don't mind, with attribution of course.

Our roles are similar, obviously the subject is the same so that's a given. I went a different route though - generating the configuration files from scratch via `parse_kv_items` filter which lets you add and remove options conditionally. I'll try to finish mine before the weekend, then you will be able to compare the PRs.

imrejonk:
Oof, didn't you get my "I'm working on debops.mailman3" e-mail on May 6th?

I thought it'd be nice if the old `debops.mailman` would stay available in later DebOps releases, so that users of that role would have less trouble updating to a later DebOps version. `debops.mailman` would be in maintenance mode as well.

About the systemd change: of course, take it, that's what the GPL is for :)
Mailman 3 failing to start without a database was actually a surprise to me, I had imagined that it would just keep retrying or something. Note that my workaround would not work for external databases.

Using `parse_kv_items` to manage configuration options sounds like a good idea, I imagine my templates are a bit static compared to yours.

ansible/roles/mailman3/templates/etc/ansible/facts.d/mailman3.fact.j2
```diff
@@ -0,0 +1,20 @@
...
+# {{ ansible_managed }}
+
+from json import dumps
+import apt
+
+cache = apt.Cache()
+output = {}
+
+try:
+    output['installed'] = cache['mailman3'].is_installed
```
drybjed: Do you know which method of checking if package is installed (either via APT cache or via presence in `$PATH`) is faster and/or less taxing on the system in question? Keep in mind that this script will be executed each time Ansible gets the facts which happens frequently in DebOps.

drybjed:
> Oof, didn't you get my "I'm working on debops.mailman3" e-mail on May 6th?

I can't find it from a cursory search. But I had some issue with my mail server and VPS provider recently so it might have been dropped on the way. Where did you send it? You could try resending it.

Anyway, I worked on the `mailman` role rewrite for some time now. Perhaps we need some better way to coordinate such things, maybe a service that tracks roles and who works or wants to work on them. Food for thought.

> I thought it'd be nice if the old `debops.mailman` would stay available in later DebOps releases, so that users of that role would have less trouble updating to a later DebOps version. `debops.mailman` would be in maintenance mode as well.

Well, I guess you have a point there, I can rename the role. Mailman maintainers probably don't plan to switch to Mailman 4 any time soon. The old `mailman` role still works in GitLab CI and can stay for the forseeable future.

Scratch that - it looks like [the `mailman` package has been removed from Debian Unstable](https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=959435) on May 2nd, so I would assume that it won't be present in Debian Bullseye. So I think that it's a good moment to switch to the Mailman 3 version but keep the old `mailman` role name - I want to avoid numerical roles as much as possible.

ansible/roles/mailman3/templates/etc/ansible/facts.d/mailman3.fact.j2
imrejonk: Let's go with the `$PATH` approach:
```
imre@lists2:/etc/ansible/facts.d$ time ./mailman3-apt.fact
{
    "installed": true
}

real	0m0,364s
user	0m0,327s
sys	0m0,035s
imre@lists2:/etc/ansible/facts.d$ time ./mailman3-path.fact
{
    "installed": true
}

real	0m0,047s
user	0m0,030s
sys	0m0,017s
```

imrejonk:
> I can't find it from a cursory search. But I had some issue with my mail server and VPS provider recently so it might have been dropped on the way. Where did you send it? You could try resending it.

I sent it to your Gmail address. Just resent it, can you check if it arrived?

> Anyway, I worked on the `mailman` role rewrite for some time now. Perhaps we need some better way to coordinate such things, maybe a service that tracks roles and who works or wants to work on them. Food for thought.

Well we have https://github.com/debops/debops/projects/5. I did check there as well when starting on `debops.mailman3`, but saw the Mailman 3 update in 'Planned things' (where it still is). Maybe you can create a project board with write access for contributors? Not sure if GitHub allows that though, I have yet to actively use that feature.

> Well, I guess you have a point there, I can rename the role. Mailman maintainers probably don't plan to switch to Mailman 4 any time soon. The old `mailman` role still works in GitLab CI and can stay for the forseeable future.
> 
> Scratch that - it looks like [the `mailman` package has been removed from Debian Unstable](https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=959435) on May 2nd, so I would assume that it won't be present in Debian Bullseye. So I think that it's a good moment to switch to the Mailman 3 version but keep the old `mailman` role name - I want to avoid numerical roles as much as possible.

Fair enough, but can we make the `debops.mailman` role, when updated for Mailman 3, backwards compatible? Or do we tell anyone who wants to keep using the old role that it's now unsupported and that they are best off grabbing the old role from Git history if they want to update to future DebOps 2.1?

drybjed:
> > I can't find it from a cursory search. But I had some issue with my mail server and VPS provider recently so it might have been dropped on the way. Where did you send it? You could try resending it.
> I sent it to your Gmail address. Just resent it, can you check if it arrived?

It did, indeed. Not sure why GMail classified it as Spam though, I fixed that.

> Well we have https://github.com/debops/debops/projects/5. I did check there as well when starting on `debops.mailman3`, but saw the Mailman 3 update in 'Planned things' (where it still is). Maybe you can create a project board with write access for contributors? Not sure if GitHub allows that though, I have yet to actively use that feature.

There is an option to create team-based project boards, but it's buried so far inside the GitHub interface that I would probably not use it after a while. I guess if it was in browser bookmarks... Still, it's not really as discoverable by other users as the project boards in the git repositories themselves.

I changed the project you linked to be "team-based" with people having their own columns - want a column for yourself? I guess you can ping me on IRC if you want to move/edit things, let's see how that goes. And if more people are interested, we can move to the real "team project board".

> > Well, I guess you have a point there, I can rename the role. Mailman maintainers probably don't plan to switch to Mailman 4 any time soon. The old `mailman` role still works in GitLab CI and can stay for the forseeable future.
> > Scratch that - it looks like [the `mailman` package has been removed from Debian Unstable](https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=959435) on May 2nd, so I would assume that it won't be present in Debian Bullseye. So I think that it's a good moment to switch to the Mailman 3 version but keep the old `mailman` role name - I want to avoid numerical roles as much as possible.
> 
> Fair enough, but can we make the `debops.mailman` role, when updated for Mailman 3, backwards compatible? Or do we tell anyone who wants to keep using the old role that it's now unsupported and that they are best off grabbing the old role from Git history if they want to update to future DebOps 2.1?

I don't think that i can be backwards compatible, the two applications are too different.

I would add exrensive migration guide from Mailman 2.x to Mailman 3 in the documentation. This would also help people that don't use DebOps but have old Mailman installations and would like to switch to the new one.

drybjed:
BTW, here's my current progress: https://github.com/debops/debops/compare/master...drybjed:redesign-mailman

PR #1349, opened 2020-05-20
Replace include with import_playbook in project directories user guide

lewer:
The doc for playbook overriding mentions `include: playbook.yml` but it seems outdated, now it should be done with `import_playbook`.


drybjed (approved):
Thanks for keeping the documentation updated.

PR #1350, opened 2020-05-20
Relax mount device fstype check

lewer:
As item.fstype is set to d('auto') in the task, it is not required


drybjed (approved):
Sounds good, thanks.

PR #1353, opened 2020-05-22
Allow custom default inventory in .debops.cfg

lewer:
Currently debops generates ansible.cfg with default inventory set automatically. A user may need to define another default inventory (for instance it there are multiple inventory files in ansible/inventory and only one of them has to be targeted by default)

### DIFF
--- a/bin/debops
+++ b/bin/debops
@@ -238,7 +238,7 @@ def main(cmd_args):
         if play:
             play_list.append(play)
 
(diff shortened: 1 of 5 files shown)


drybjed:
The PR sounds good to me. @lewer, do you think it is finished, or are you still working out the kinks?

lewer:
@drybjed OK for me

drybjed (approved):

PR #1355, opened 2020-05-23
Fix issue 1354 - [debops.core] Save local facts - uuid issue

jeansorgemoel:


drybjed:
I'm not sure what causes the issue you are trying to solve with this PR, but this change will write the entire `core__register_uuid` variable contents in the fact script, resulting in something similar to:
```
output = loads('''"{'cmd': 'set -o nounset -o pipefail -o errexit && dmidecode --string system-uuid | tail -n 1 || true', 'stdout': '', 'stderr': '/sys/firmware/dmi/tables/smbios_entry_point: permission denied\\n/dev/mem: no such file or directory', 'rc': 0, 'start': '2020-06-03 00:10:04.150820', 'end': '2020-06-03 00:10:04.153580', 'delta': '0:00:00.002760', 'changed': false, 'stdout_lines': [], 'stderr_lines': ['/sys/firmware/dmi/tables/smbios_entry_point: permission denied', '/dev/mem: no such file or directory'], 'failed': false}"''')
```
So that will not be helpful. But, I'll probably remove this fact soon, it's currently not used in DebOps and Ansible facts already have `ansible_machine_id` variable which can perform a similar function.

PR #1367, opened 2020-06-04
[mailman] Redesign the role, switch to Mailman 3.x

drybjed:
Fixes #1347


imrejonk:
Cool! We'll be trying out your role on our infrastructure when I get back from vacation, in two weeks or so.

PR #1370, opened 2020-06-09
Include debops logstash role

pedroluislopez:
Features:
- Default pipeline Beats -> Filter -> Elasticsearch.
- PKI support for incoming request from beats.
- Memory tunning
- Support multiple pipelines: https://www.elastic.co/guide/en/logstash/6.8/multiple-pipelines.html

Missing:
- Documentation

### DIFF
--- a/ansible/roles/logstash/tasks/main.yml
+++ b/ansible/roles/logstash/tasks/main.yml
@@ -1,0 +1,149 @@
+---
+# Copyright (C) 2020 Pedro Luis Lopez <pedroluis.lopezsanchez@gmail.com>
+# Copyright (C) 2020 DebOps <https://debops.org/>
+# SPDX-License-Identifier: GPL-3.0-only
+
+- import_role:
+    name: 'secret'
+
+- name: Check if the dependent config file exists
+  stat:
+    path: '{{ secret + "/logstash/dependent_config/" + inventory_hostname + "/config.json" }}'
+  register: logstash__register_dependent_config_file
+  become: False
+  delegate_to: 'localhost'
+  when: (ansible_local|d() and ansible_local.logstash|d() and
+         ansible_local.logstash.installed|d())
+  tags: [ 'role::logstash:config' ]
+
+- name: Load the dependent configuration from Ansible Controller
+  slurp:
+    src: '{{ secret + "/logstash/dependent_config/" + inventory_hostname + "/config.json" }}'
+  register: logstash__register_dependent_config
+  become: False
+  delegate_to: 'localhost'
+  when: (ansible_local|d() and ansible_local.logstash|d() and
+         ansible_local.logstash.installed|d() and
+         logstash__register_dependent_config_file.stat.exists|bool)
+  tags: [ 'role::logstash:config' ]
+
+- name: Add Logstash server user to specified groups
+  user:
+    name: '{{ logstash__user }}'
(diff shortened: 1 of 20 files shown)


ansible/roles/logstash/tasks/main.yml
```diff
@@ -0,0 +1,149 @@
...
+- name: Add Logstash server user to specified groups
+  user:
+    name: '{{ logstash__user }}'
+    groups: '{{ logstash__append_groups | join(",") | default(omit) }}'
+    append: True
+    createhome: False
+  when: logstash__pki|bool
+
+- name: Divert original configuration files
+  command: dpkg-divert --quiet --local --divert {{ item }}.dpkg-divert
```
ypid: DebOps has the module `dpkg_divert` now ;-)

ypid:
Hi @pedroluislopez 

From a quick look, this role looks very solid. Well done!

Just a kinda personal note on Logstash. That thing is a beast. Worst thing in my opinion is the high startup time and CPU usage. When I would start out from scratch I would choose anything but Logstash.

And to get Logstash in line somehow, I would not template the conf.d files (input/filter/output) using Ansible. I would write them by hand without adding yet another abstraction. This allows you to test the config (shameless self plug: https://github.com/geberit/logstash-config-integration-testing). Sometimes this abstraction makes sense when you can for example use it to do some clever all/group/host merging so thats why you see it in DebOps together with the custom Jinja filters. I am just not convinced that Logstash conf.d (which is a custom format) is worth it. Configuration can (should?) be the same on all Logstashes.

Another option is to have the Elastic Stack push the Logstash conf.d to connected Logstash instances.

And don’t get me wrong, I enjoy the other parts of the Elastic Stack (ES, Kibana, Beats).

ypid:
Just to document what I said just now in the first DebOps video meeting. I looked into Logstash alternatives and even discussed this with a few entities including Elastic support at length. @pedroluislopez Have you looked at [Elasticsearch ingest pipelines](https://www.elastic.co/guide/en/elasticsearch/reference/current/ingest.html) which can do some of the things that Logstash can do?

Also, if I may spoiler something, @drybjed is working on a filebeat role :) Note that Filebeat is actually more than a log shipping agent by now as it ships with [quite a few modules/log parsers](https://www.elastic.co/guide/en/beats/filebeat/current/filebeat-modules.html) that do the job people used Logstash previously for. This is also based on ingest pipelines.

ypid:
> Just a kinda personal note on Logstash. That thing is a beast. Worst thing in my opinion is the high startup time and CPU usage. When I would start out from scratch I would choose anything but Logstash.

Logstash 7.9 has greatly improved the situation. But I only tested it yet in my test environment. https://www.elastic.co/blog/whats-new-logstash-7-9-0 So maybe Logstash support is not the worst thing that could happen to DebOps after all :wink: I have not ultimately decided how to proceed (Logstash or something else).

ypid:
I deep dived into https://github.com/timberio/vector a few weeks ago and have something running with it. Vector is already very well designed. It already crushes Logstash in terms of performance (startup and throughput), some features are missing but it is already usable and I would say it has a great future ahead.

I am not yet sure if I really want to vendor look myself in with ingest pipelines.

ypid:
My posts here good a bit off-topic. I created #1679 for this where the architecture can be discussed.

PR #1372, opened 2020-06-11
The 'core' role reorganization

drybjed:


ypid (commented):
I reviewed the diff, LGTM.

Just a note as as German speaking person. "hier" means "here". That confused me a bit. So there is not association to hierarchy because my German association/interpretation as "here" is way stronger.

How about naming it `fhs` (Filesystem Hierarchy Standard)?

drybjed:
I was not aware of that, but that's probably a common problem in many languages.

Thers's no `fhs` Debian package, so I guess the name is free to use. I guess in the context it will be fine to use this name, with maybe an explanation in the documentation that this implements the `/usrlocal/` etc. hierarchy. I'll rename the role before the release, after my PC is not busy testing all the roles in the GitLab CI pipeline.

ypid:
Thanks :)

> I was not aware of that, but that's probably a common problem in many languages.

True.

SuperSandro2000:
What means hier from English to German? I couldn't find any translation and it is so distracting.

ypid:
"hier" is a German word which means "here" in English. Also find it distracting because you speak German? (I don’t quite understand what you mean with your comment).

The role was initally supposed to be called "hier" (as in **hier**archy).

SuperSandro2000:
> The role was initally supposed to be called "here" (as in hierarchy).

That is what I was looking for.

PR #1373, opened 2020-06-11
Rename 'hier' role to 'fhs'

drybjed:
The 'hier' name has an unfortunate meaning of "here" in German language,
therefore it might create issues with German speakers. The 'fhs' name
seems to be a better fit for what the role is meant to do.


ypid:
LGTM. Thanks for renaming :)

Updates: #1372

PR #1374, opened 2020-06-11
Minor fixes and improvments

ypid:
Fixes: #1373

### DIFF
--- a/ansible/roles/python/templates/etc/pip.conf.j2
+++ b/ansible/roles/python/templates/etc/pip.conf.j2
@@ -1,0 +1,7 @@
+# {{ ansible_managed }}
+
+# Copyright (C) 2020 Robin Schneider <ypid@riseup.net>
+# SPDX-License-Identifier: GPL-3.0-only
+
+[global]
+disable-pip-version-check = {{ not python__pip_version_check }}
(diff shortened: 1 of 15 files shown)


ansible/roles/python/templates/etc/pip.conf.j2
```diff
@@ -0,0 +1,7 @@
+# {{ ansible_managed }}
+
+# Copyright (C) 2020 Robin Schneider <ypid@riseup.net>
+# SPDX-License-Identifier: GPL-3.0-only
```
drybjed: When possible, the copyright notice should be for the template and not the final config - in other words, put it in `{# #}` jinja comment section. Exceptions to this are bash/python scripts since they need to pass `shellcheck` and `pycodestyle` checks.

docs/ansible/roles/fhs/man_seealso.rst
```diff
@@ -5,7 +5,7 @@
 See Also
 ========
 
-- Manual pages: :man:`hier(7)`
+- Manual pages: :man:`fhs(7)`
```
drybjed: There is no `fhs(7)` manpage. The `hier(7)` is correct here.

ansible/roles/python/templates/etc/pip.conf.j2
ypid: Makes sense. Just for this reason I checked some other recently added role to follow the same style. Seems I picked the wrong role model file. Updated both.

docs/ansible/roles/fhs/man_seealso.rst
ypid: And I learned something new :) Now I know how you picked the name. So the name "hier" has some background. I removed this commit from the PR. No reason to have it in the eternal DebOps git history ;-)

drybjed (approved):
Looks good, thanks.

PR #1375, opened 2020-06-11
fix potential logic error for interfaces managed by systemd

sysbot:
the way i understand this code is that `systemd_interface_instances` track all
the interfaces that are managed by systemd. so if `iface_name` is not in
`systemd_interface_instances` then restart networking and sync.

Where correctly it should be if `iface_name` is in `systemd_interface_instances`
then state is correct and continue with restarting networking and sync.

This resulting in errors like these when I attempt to use this script, with
added debug lines. Let me know if i'm off or this was a red herring since I'm
new to this code base.

```
$ sudo /usr/local/lib/ifupdown-reconfigure-interfaces
Detected interfaces to reconfigure: bond0,enp1s0f0,enp1s0f1
Detected interfaces to reconfigure: bond0,enp1s0f0,enp1s0f1
Found active systemd ifup@ instances: bond0,enp1s0f0,enp1s0f1
Found active systemd iface@ instances: bond0,enp1s0f0,enp1s0f1,
Found active systemd iface@ instances: bond0,enp1s0f0,enp1s0f1,
Bringing down 'enp1s0f1' interface
inspect enp1s0f1, and active systemd iface@ instances: bond0,enp1s0f0,enp1s0f1,,
Error: Script was working on 'enp1s0f1' network interface when it lost knowledge about the network interface state. The '/etc/network/interfaces.d/' might be desynchronized. Exiting to avoid loss of connectivity, investigate the issue.

```

thanks,

### DIFF
--- a/ansible/roles/ifupdown/files/script/ifupdown-reconfigure-interfaces
+++ b/ansible/roles/ifupdown/files/script/ifupdown-reconfigure-interfaces
@@ -287,7 +287,7 @@ if [ -d "${interface_request_path}" ] ; then
                     if ifquery --state "${iface_name}" > /dev/null 2>&1 ; then
 
                         # The interface in question was definitely not set up by 'ifup@.service'
-                        if ! containsElement "${iface_name}" "${systemd_interface_instances[@]:-}" ; then
+                        if containsElement "${iface_name}" "${systemd_interface_instances[@]:-}" ; then
 
                             log_message -m "The '${iface_name}' interface is still active, shutting down networking service"
                             stop_networking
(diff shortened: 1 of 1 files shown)


drybjed:
The problem with `ifupdown` and `systemd` is that on some systems, network is configured using a global `networking` services, while on later systems (if I remember correctly since Stretch), it is done via separate per-interface `ifup@.service` instance. The script checks if the latter is true (the desired outcome), and if not it will assume that the old `networking` setup is used. In that case whe drop all interfaces and re-configure them using separate `ifup@` and `iface@` service instances.

Since you are working with bonding, I think that the problem could be that the script tried to bring down one of the Ethernet interfaces in the bond, and you should probably bring down the bond itself instead. Did you check the system state when that error occured, especially what network interfaces are up, if they have IP addresses and so on? Is this error reproductible in your environment?

drybjed (approved):
Thanks. I'll backport the fix to the stable releases as well.

PR #1380, opened 2020-06-12
Fixing a few bugs (Ansible facts with Python2, influxdb_server handlers)

ypid:
I am still working on my Grafana/InfluxDB setup but I consider the two bug fixes as serious enough to open a PR right away.


drybjed:
The `coding` magic line would probably have to be applied to all Python scripts included in DebOps, not just the fact scripts, right? Also, perhaps using `encoding` instead of `coding` would sound better.

Apparently [PEP 3120](https://www.python.org/dev/peps/pep-3120/) mandates use of UTF-8 as the default encoding in Python 3. But since we are supporting the project use in Python 2.7 environments, I suppose that declaring the proper encoding still has value.

ypid:
> The coding magic line would probably have to be applied to all Python scripts included in DebOps, not just the fact scripts, right? Also, perhaps using encoding instead of coding would sound better.

I tested this, also works. I am in favor of this as I also think that it sounds better. I just cannot really find that being used elsewhere. Everyone is using the `coding` keyword. I would just stick to the default. It is a workaround for legacy Python2 anyway.

> But since we are supporting the project use in Python 2.7 environments, I suppose that declaring the proper encoding still has value.

I guess so. I will try to make that happen.

ypid:
> I guess so. I will try to make that happen.

Done that. I did some checking to ensure I did not miss any (see commit messages).

If you want me to update the header line (use `encoding`) I can now do that easily.

ypid:
I also searched for `ag 'encoding: utf-8'`. Nothing yet. I don’t have a strong opinion on this, but I guess even we use the kinda standard `coding` word.

drybjed (approved):
Changes look good, thanks.

ypid:
Thanks for merging.

PR #1382, opened 2020-06-16
Improve influxdb roles; Use Ansible 2.8 syntax shortcut; A few other fixes

### LINKED ISSUE #1371: Jinja Undefined values improved in Ansible 2.8
Hi @drybjed

You are aware of https://docs.ansible.com/ansible/latest/porting_guides/porting_guide_2.8.html#jinja-undefined-values ?

This finally allows us to change this:

```YAML
cryptsetup__keyfile_remote_location: '{{ (ansible_local.root.var
                                         if (ansible_local|d() and ansible_local.root|d() and
                                             ansible_local.root.var|d())
                                         else "/var/local") + "/keyfiles" }}'
```

Into this:

```YAML
cryptsetup__keyfile_remote_location: '{{ (ansible_local.root.var
                                         if ansible_local.root.var|d()
                                         else "/var/local") + "/keyfiles" }}'
```

ypid:
Closes: #1371

I reviewed and tested this with all the use cases I have. But I don’t use all roles so I have not tested everything. The regex I came up with does quite a bit of validation so it does not blindly replace stuff. The one known issue is that opening/closing parenthesis `(`, `)` are maybe not removed together, leaving one there which is a syntax error. Please test this a bit with your use cases before you merge. Other than this, I hope I fixed more bugs than I introduce with this.


ansible/roles/apt_cacher_ng/defaults/main.yml
```diff
@@ -201,12 +201,8 @@ apt_cacher_ng__max_download_speed_kib: ''
 # One mirror per line.
 # Set to an empty string to let the package scripts from Apt-Cacher NG decide
 # which upstream mirror to use.
-apt_cacher_ng__upstream_mirror_debian: '{{ ansible_local.apt.default_sources_map.Debian[0]
-                                           if (ansible_local|d() and ansible_local.apt|d() and
-                                               ansible_local.apt.default_sources_map|d() and
-                                               ansible_local.apt.default_sources_map.Debian|d() and
-                                               ansible_local.apt.default_sources_map.Debian[0]|d())
-                                           else "http://deb.debian.org/debian" }}'
+apt_cacher_ng__upstream_mirror_debian: '{{ ansible_local.apt.default_sources_map.Dsdfebian[0]
```
drybjed: Spelling mistake (`Dsdfebian`).

ansible/roles/dnsmasq/tasks/main.yml
```diff
@@ -29,8 +29,7 @@
     owner: 'root'
     group: 'root'
     mode: '0755'
-    unsafe_writes: '{{ True if (core__unsafe_writes|d(True if (ansible_local|d() and ansible_local.core|d()
-                       and ansible_local.core.unsafe_writes|d() | bool) else False) | bool) else omit }}'
+    unsafe_writes: '{{ True if (core__unsafe_writes|d(ansible_local.core.unsafe_writes|d()) | bool) else omit }}'
```
drybjed, line 32: Well... I'm not sure why I didn't notice this usage before (the default variable is referenced first instead of the fact) but this is essentially redundant.

Actually the use of the `ansible_local.core.*` facts in various roles could probably be replaced to references to variables defined in the inventory directly. The information flow would also be better, instead of getting the value from the remote host via facts, it could be taken from the inventory. The reason I didn't do this is that there wasn't a good variable namespace to reserve for this. But I think that after renaming the `debops` role to `controller`, the `debops__` namespace could be reserved for global variables, like `debops__unsafe_writes` and similar. Then we could drop most of the stuff in the `core` role facts and rely on the users to define the variables in the inventory. Then the above code could be changed to:
```yaml
unsafe_writes: '{{ True if debops__unsafe_writes|bool else omit }}'
```
As for the current cases in the file - I guess you can leave them as they are for now.

ypid: Sounds good. `debops__` works for me. Then I will leave this as is for now.

ansible/roles/apt_cacher_ng/defaults/main.yml
ypid: That was me testing if `d()` also works with array access and then forgetting to change it back before committing (and failing to self review my own commit changes). Thanks for your careful review!

drybjed (approved):
All files reviewed. Nice job, thanks!

ypid:
Thanks for reviewing this so quickly :+1: 

Also thanks for fixing the remaining issues. For reference, https://github.com/debops/debops/pull/1386, https://github.com/debops/debops/pull/1385 and https://github.com/debops/debops/pull/1384.

drybjed:
I wish that I had more scrutiny before checking this PR, it added hours of testing and fixing bugs, and since I'm currently using my workstation for the CI pipeline, I cannot use it during the full test due to the load which takes hours. Oh well, new release will probably be over the weekend after I fix all the bugs...

ypid:
I warned you ;-) :

> Please test this a bit with your use cases before you merge.

And sorry for the work it caused. At least we now have much more clean `ansible_local` usage.

PR #1386, opened 2020-06-18
Additional fixes after fact changes

drybjed:


drybjed:
Not sure why Travis-CI doesn't report a successful test, but it should be fine. Merging.

ansible/roles/php/tasks/main_env.yml
```diff
@@ -51,7 +51,8 @@
   set_fact:
     php__version: '{{ ansible_local.php.version|d(php__register_version.stdout) }}'
     php__long_version: '{{ ansible_local.php.long_version
-                           if (ansible_local.php.long_version|d() != "(none)")
+                           if (ansible_local|d() and ansible_local.php|d() and
+                               ansible_local.php.long_version|d() != "(none)")
```
ypid, line 55: Does the Ansible 2.8 shortcut not work here?

drybjed: Apparently, you can see the error [in the CI pipeline test](https://gitlab.com/debops/debops/-/jobs/601197359). I'm not sure what caused this, it also shows up with the `debug` module, so it's probably a quirk of Jinja.

ypid: Wired. Also does not make sense to me. Ok then.

PR #1389, opened 2020-06-20
[pki] Fixed symbolic link paths for openssl selfsigned root cert

jstruebel:
Added basename command to symbolic linking of the selfsigned root certificate in the openssl function that mirrors the same line in the gnutls function. This fixes the error "failed to create symbolic link 'selfsigned/root.pem': No such file or directory" from the pki-realm command when creating a realm that uses selfsigned certificate and the openssl library.


drybjed (approved):
Thanks! I tried to recreate the issue in my environment, but the role refuses to cooperate. So I hope that's the correct fix. :-)

I'll backport this to older releases as well.

jstruebel:
It should be, it's working in my environment with the fix. I had a custom realm with acme=true, internal=false, and library=openssl when I discovered it.

PR #1394, opened 2020-06-22
Expand ~ in XDG_* env var

lewer:
If not, fish shell will get crazy and create '~' empty directory everywhere


drybjed (approved):
Thanks. I'm not sure why Travis-CI doesn't want to test it, but the patch seems fine. I'll backport it to previous releases as well.

PR #1404, opened 2020-07-01
Add variable to set chrony cmdport, which was disabled in #1236

DouglasHeriot:
In #1236 chrony's `cmdport` setting was set to `0` which disabled localhost UDP monitoring commands.

An example use where that doesn't work, it using [telegraf](https://github.com/influxdata/telegraf) to monitor chrony. When `telegraf` runs as its own user, it cannot read from the `chrony` UNIX socket. (chrony has no option to modify its permissions - it is only for root and the *user* chrony, no the *group* chrony)

It should be safe in many circumstances to allow `chrony` to listen for commands on the localhost address. It only allows read-only commands. From the [`chronyc` man page](https://chrony.tuxfamily.org/doc/3.5/chronyc.html):
> Only the following monitoring commands, which do not affect the behaviour of chronyd, are allowed from the network

I'm also working on [adding sudo support into telegraf](https://github.com/influxdata/telegraf/pull/7669) when running `chronyc`, but that's going to take a bit longer than simply changing the `cmdport` setting.

### DIFF
--- a/ansible/roles/ntp/defaults/main.yml
+++ b/ansible/roles/ntp/defaults/main.yml
@@ -164,6 +164,17 @@ ntp__purge_packages:
 ntp__openntpd_options: '-f /etc/openntpd/ntpd.conf -s'
                                                                    # ]]]
                                                                    # ]]]
+# Chrony configuration [[[
+# --------------------------
+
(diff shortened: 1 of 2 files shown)


drybjed (approved):

PR #1405, opened 2020-07-01
[postfix] Use long form in master.cf

imrejonk:
Postfix main.cf configuration overrides are now written to master.cf in 'long form'. This allows specifying parameter values that contain whitespace, for example:

`-o { smtpd_sender_restrictions = check_sasl_access ldap:${config_directory}/ldap_known_sender_relays.cf, reject_authenticated_sender_login_mismatch }`

Ref: https://manpages.debian.org/master.5.html

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -33,6 +33,12 @@ General
   variables. All role variables have been renamed along with the role inventory
   group, you will have to update your inventory.
 
+:ref:`debops.postfix` role
+''''''''''''''''''''''''''
+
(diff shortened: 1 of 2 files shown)


drybjed (approved):
Sounds like a good idea, thanks. This feature is supported since Postfix 3.0. Postfix 3.1 was introduced in Debian Stretch which is Oldstable by this point. Making it conditional based on version could lead to broken configuration, so let's go with the long form by default; older installations should probably be upgraded.

I'll backport this change to the stable DebOps releases as well.

PR #1409, opened 2020-07-03
[rsnapshot] Fix dry run error

imrejonk:
The 'Scan SSH fingerprints of the configured hosts' task fails because
it needs a list of already scanned host fingerprints, which is not
populated on dry runs. This change makes the 'Get list of already
scanned host fingerprints' task execute during dry runs as well.


drybjed (approved):
Looks good, thanks. I'll backport it to the stable releases as well.

PR #1411, opened 2020-07-04
[owncloud] Fix dry run errors

imrejonk:
I identified the following issues with the tasks in tarball.yml when performing a dry run:

- 'Create source directories' fails, because `owncloud__register_full_version` is not set during dry run.
  Fix: force the task setting `owncloud__register_full_version` to run during dry run. This will trigger a 'changed' notice in the 'Create source directories' task when a new version is available.

- 'Download application files needed for verification' is always changed in dry run mode.
  Fix: skip this task in dry run mode. This should not have a negative impact because these files only need to be downloaded when actually updating the ownCloud version, which does not happen in dry run mode.

- 'Read checksum from file' will fail when there is a new version available, because the verification files have not been downloaded.
  Fix: skip this task in dry run mode.

For a related PR (but for the rsnapshot role), see https://github.com/debops/debops/pull/1409

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -42,6 +42,12 @@ General
 - Fixed an issue where the :command:`debops` scripts did not expand the
   :file:`~/` prefix of the file and directory paths in user home directories.
 
+:ref:`debops.owncloud` role
+''''''''''''''''''''''''''''
+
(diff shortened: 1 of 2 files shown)


drybjed (approved):
Looks good, thanks. :-) I'll fix the conflict in the Changelog and backport the fixes to the stable releases.

PR #1413, opened 2020-07-06
[debops.postgresql] Drop role/database if requested and fix issues with when clauses

lewer:
This PR is meant to fulfil the following goals:

- currently postgresql role doesn't take into account `state: 'ignore'` in `postgresql__*_databases` items nor `postgresql__*_roles` nor `postgresql__*_extensions`
- tasks should fail instead of being skipped when a required key is not defined in `postgresql__*` dictionaries
- databases and roles can be dropped with `state: 'absent'` in `postgresql__*_roles` and `postgresql__*_databases`


drybjed (approved):
Looks good, thanks. I tested the changes a bit, seems to work fine. I'll backport the PR to the stable releases as well.

Issue #1398, opened 2020-06-23
tinc interfaces doesnt become bridge member


jankowa:
Tinc systemd service seems not to be waiting for the bridge to come ready if it's configured as a bridge port.

To reproduce on Debian Buster: Setting up a default tinc mesh described here with:

```
tinc__host_networks:
  'mesh0':
    bridge: 'br2'

ifupdown__interface_layout: 'static'
ifupdown__host_interfaces:
  'br2':
    type: 'bridge'
    inet: 'static'
    nat: True
    addresses:
      - '10.0.0.10/24'
```

works fine on setup (the only problem is that ifupdown role always looses connection and has to be done twice). But after reboot the tap-mesh0 interface isn't attched to br2.

`service tinc@mesh0 restart`

helps - and tinc is working again like expected.

```
:~# brctl show
bridge name	bridge id		STP enabled	interfaces
```

After
`  service tinc@mesh0 restart`

```
bridge name	bridge id		STP enabled	interfaces
br2		8000.deb203db0cd1	no		tap-mesh0
```

drybjed:
I think that I experienced this issue as well on one or two machines but I didn't have time to pinpoint the reasons for it... Looks like you found that.

I wonder if a solution to this could be to add `Requires=` or `Wants=` or some other service attributes in the `tinc` service configuration, most likely via `tinc@mesh0.service.d/override.conf` file or something similar. Perhaps you could try adding something like this and checking if that helps?

jankowa:
This works for me:

systemctl edit tinc@mesh0.service

```
[Unit]
Requires=sys-subsystem-net-devices-br2.device
After=sys-subsystem-net-devices-br2.device
```

I'll think how to integrate this best to the tinc role.

jankowa:
> This works for me:
> 
> systemctl edit [tinc@mesh0.service](mailto:tinc@mesh0.service)
> 
> ```
> [Unit]
> Requires=sys-subsystem-net-devices-br2.device
> After=sys-subsystem-net-devices-br2.device
> ```

Yes, but not in any cases. If tinc is configured with link_type: 'static' it doesn't.

It's true that the interface is added to the bridge - but we don't have network connectivity - even if tinc log says every thing is ok.

In the journal log I get.

`br2: received packet on tap-mesh0 with own address as source address `

This was the moment I inspected the mac-address more carefully and I realized that both of my tinc host got the same mac-address. Bingo. And the mac adress is generated in tinc-up:

```
[...]

PR #1417, opened 2020-07-10
tinc interfaces doesnt become bridge member #1398

jankowa:
I made the change for the systemd unit. But since I'm not very familiar with systemd I'm still unsure how parameters of the tinc.service are inherited by tinc@.service.

In tinc.service there is 

Wants=network-online.target

But neither this is sufficient nor it does make a difference if we put this in tinc.service.j2
```
After=local-fs.target network-pre.target apparmor.service systemd-sysctl.service systemd-modules-load.service networking.service ifup-allow-all-auto.service ifup-allow-boot.service
```
so it doesn't inherit here anything from tinc.service.


drybjed (approved):
Sounds good. I'll backport the change to the stable releases.

drybjed:
Merged in https://github.com/debops/debops/commit/b51f681f0997a3c02ea8b8083ac7694556aceeb0, closing.

PR #1418, opened 2020-07-10
[icinga] WIP: add option for generating config/feature files on delegated host

lewer:
When director is not used, we might want to upload a host's configuration on a master/satellite rather than on the host itself.

### DIFF
--- a/ansible/roles/icinga/handlers/main.yml
+++ b/ansible/roles/icinga/handlers/main.yml
@@ -6,11 +6,26 @@
 - name: Check icinga2 configuration and restart
   command: icinga2 daemon -C
   notify: [ 'Restart icinga2' ]
+  delegate_to: '{{ item }}'
+  with_items: |
+    {% set changed_hosts = [] %}
+    {% for c in icinga__configuration_result.results +
+                icinga__features_result.results %}
+    {%   if c.changed %}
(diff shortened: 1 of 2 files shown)


drybjed:
A question about this idea... Since the per-host configuration is done in the Ansible inventory, why not just define it in the master/satellite host variables directly? I think that delegating might complicate things a bit too much in such case.

drybjed:
Here are my current thoughts about the whole thing... Your approach of using `item.delegate_to` for the tasks is wrong. I don't see the Icinga configuration split piecemeal between different masters - instead, each icinga node has a specific master host - is that how you set up your environment?

This means that there should be `icinga__delegate_to` variable, by default targeting `inventory_hostname`, which is uses in the configuration task. Using this variable you can tell via the inventory where the configuration should be generated.

There should be a separate `Check and restart remote Icinga` set of handlers, each one defined with the `delegate_to` keyword, but only "firing" when `icinga__delegate_to` is not equal to `inventory_hostname`. I think that will be the cleanest way to do this. I'm not sure what the effect will be with multiple masters or when multiple hosts fire the handler at once - I would think that Ansible should run it once according to the handler rules, but that would have to be tested.

@lewer, can you check if something like this works for you?

lewer:
The lack of proper variable treatment in handlers makes it too tricky to achieve... I close the PR.

PR #1419, opened 2020-07-12
[apt] Install gnupg before downloading PGP keys

imrejonk:
I noticed this while running the bootstrap playbook: the apt role, when
configured to do so through inventory variables, tries to add APT keys
from a PGP keyserver without installing gnupg2 first. This results in
the following error:

`Failed to find required executable gpg in paths: /usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin`

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -46,7 +46,7 @@ General
 ''''''''''''''''''''''
 
 - Fixed an issue where the role would attempt adding APT keys from a PGP
-  keyserver without installing gnupg2 first.
+  keyserver without installing gnupg first.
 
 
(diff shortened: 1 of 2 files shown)


drybjed:
The `gnupg2` package is a transitional package since Debian Stretch. I think that you want the `gnupg` package instead.

imrejonk:
@drybjed you're right, thanks for correcting me!

drybjed (approved):
Looks good, thanks. I'm not sure why Travis isn't picking this up... I'll fix the merge conflict and backport the changes.

PR #1423, opened 2020-07-19
[ntp] Don't try to disable systemd-timesyncd if not available

### LINKED ISSUE #1422: ntp: "Could not find the requested service systemd-timesyncd" when using chrony
In my inventory I define `ntp__daemon: chrony` 

Now the `debops.ntp` role will fail when trying to "disable" systemd-timesyncd on a Debian Buster host:
```
TASK [debops.ntp : Manage systemd-timesyncd state] ***********************************************************************************
fatal: [host.example.com]: FAILED! => {"changed": false, "msg": "Could not find the requested service systemd-timesyncd: host"}
```

There is no `systemd-timsyncd.service` available on this host.

ganto:
Fixes #1422

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -174,6 +174,13 @@ LDAP
 - Fixed multiple issues with adding and updating hosts to the LDAP directory
   when these hosts were configured for network bonding.
 
+:ref:`debops.ntp` role
+''''''''''''''''''''''
+
+- Don't try to disable or stop the ``systemd-timesyncd`` service when using an
+  alternative NTP service implementation and ``systemd-timesyncd`` is not
+  available.
+
 :ref:`debops.owncloud` role
 ''''''''''''''''''''''''''''
 
--- a/ansible/roles/ntp/tasks/install.yml
+++ b/ansible/roles/ntp/tasks/install.yml
@@ -11,12 +11,18 @@
   register: ntp__register_apt_install
   until: ntp__register_apt_install is succeeded
 
+- name: Query available systemd services
+  service_facts:
+  when: ansible_service_mgr == "systemd"
+
 - name: Manage systemd-timesyncd state
   service:
     name: 'systemd-timesyncd'
     state: '{{ "started" if (ntp__daemon in [ "systemd-timesyncd" ]) else "stopped" }}'
     enabled: '{{ True if (ntp__daemon in [ "systemd-timesyncd" ]) else False }}'
-  when: ansible_service_mgr == "systemd"
+  when:
+    - ansible_service_mgr == "systemd"
+    - ("systemd-timesyncd.service" in ansible_facts.services)
 
 # NTPdate does not need to be configured separately as the script `/usr/sbin/ntpdate-debian`
 # will pick up the configuration from `/etc/ntp.conf` when it exists.


nqb:
Just for my knowledge, in which case systemd-timesyncd is not available ?

ganto:
No idea... on the host I preseeded myself, there is no systemd-timesyncd.

drybjed:
The condition is redundant with a check for `systemd` init. In my environment which uses unprivileged LXC containers, `systemd-timesyncd.service` unit is still present in the container even though the service is not running. The service file comes with the `systemd` package so it will probably be defined everywhere that `systemd` is used.

@ganto, can you tell a bit more details about your environment? What do you use to deploy it, is `systemd` used there?

ganto:
> can you tell a bit more details about your environment? What do you use to deploy it, is systemd used there?

It's a KVM virtual machine that I installed in March 2019 via a minimal `ganto.libvirt-machine` Ansible role that is using `virt-install` with a locally mirrored Debian installer (at that time it was still Sarge) and a generated preseed.cfg (I completely forgot about this role, so I never uploaded it to github) but it looks more or less like [this](https://github.com/inofix/common-playbooks/blob/master/roles/virt-install/tasks/main.yml).

From then on it was managed via DebOps, starting with 0.8.1 and then 1.0.x until I found recently that 1.0.18 or maybe already earlier (I skipped a few releases) was failing to execute the 'ntp' role. I have backports enabled, so currently it runs `systemd-245.6-1~bpo10+1`. The `systemd-timesyncd` package is not installed. I'm explicitly using `chrony` for time keeping.

drybjed:
Oh, that explains it. Since `systemd` 245, the `timesyncd` service has been split into [its own Debian package](https://packages.debian.org/sid/main/systemd-timesyncd). This can be seen in the [package Changelog](https://metadata.ftp-master.debian.org/changelogs//main/s/systemd/systemd_245.7-1~bpo10+1_changelog):

>   * Ship systemd-timesyncd in a separate package.
    The new systemd-timesyncd package conflicting with other NTP-related
    packages resolves the problems arising when running systemd-timesyncd
    and other NTP servers on the same system.
    (LP: #1849156, Closes: #805927, #947936)

In that case the support in DebOps needs to be updated before Bullseye is released.

I plan to split the NTP support from the `ntp` roles into separate ones (`chrony`, `timesyncd`, `openntpd`) and make the `ntp` role a ["wrapper role"](https://lists.debops.org/hyperkitty/list/debops-devel@lists.debops.org/thread/LRP4XPQATH6FNZA3TMY6A7JHQW2CGIS7/) that will let you choose between different sevice. However this will take some time.

In any case, this PR that you are proposing will not work with hosts that don't use `systemd` 245+, which includes Debian Buster. There's no Ansible local fact that provides version of the init system, so we cannot use a condition with that. Basing the condition on an OS release will also not work because you are using `systemd` from Backports. I need to think about this a bit.

ganto:
> In any case, this PR that you are proposing will not work with hosts that don't use systemd 245+, which includes Debian Buster. 

Why is that?

drybjed:
Scratch that - you're right, since the condition checks if the service exists or not, it should work fine. I guess we can go with that then for the time being. Do you want to rebase the PR on the current `master` branch and fix merge conflicts? I can do it otherwise.

tasosalvas (approved):

drybjed (approved):
Looks good, thanks.

PR #1427, opened 2020-07-22
[docs] Add project roadmap

drybjed:


nqb:
>  Final goal: DebOps is used to set up a data center in an extraterrestrial colony 

:heart:

tasosalvas:
Hey that's not a *roadmap*, it is a *Goals*

docs/meta/roadmap.rst
```diff
@@ -0,0 +1,205 @@
...
+It's not a pipe dream. Debian has already made it into space, due to `being
+used as the operating system on various laptops used on the International
+Space Station`__. Humanity already made it to the Earth's Moon in the past, so
+we will be getting there at some point in the future; other planets like Mars
+will also be visited. Eventually a new human colony will be estabilished,
+either on a planetary/lunar surface, on an asteroid or within an artifical space
+station. In such case, at least one local data center will have to be set up to
+faciliate computing tasks optimally.
+
+Obviously the first couple of data centers set up in space won't be based on
```
ypid, line 188: This sentence bugs me. I would like to be more optimistic here, I mean we are talking about the end goal. What else would you deploy first if not Free Software? What else would you trust enough to keep you alive? I only accept HAL 9000 or Free Software :) Oh wait, HAL 9000 would have not let us humans alive.

drybjed: Life support system, traffic control for local airspace, temperature control... These systems are better handled by a real time OS such as WxWorks. Will it be Free Software? Hopefully. But I don't think that will be Linux, unless RTOS support is improved.

We are talking about stuff done much later, meant for normal non-critical systems.

ypid: Then we are on the same page. I was also thinking about life support systems, but the sentence allows for a different, more proprietary interpretation was well. I guess we can just be more specific. I proposed something in https://github.com/debops/debops/pull/1432/files?short_path=cd789e6#diff-cd789e669a2579f3de19462cabe39f08

PR #1429, opened 2020-07-22
[docs] Rename 'roadmap' to 'goals', more precise

drybjed:


drybjed:
This patch only changes the documentation. The problem on Travis-CI is caused by never version of `shellcheck`  (0.7.0) which detects new issues. I'll merge this PR to get it in the tree and then fix the shellcheck issues so that syntax checks work again.

PR #1431, opened 2020-07-22
[netbox] Update to Netbox 2.8

ypid:
There is one issue preventing this from getting merged. The initial setup does not create the first admin user at task "Perform database installation" anymore. The DB users table remains empty. Because I just wanted to get started with Netbox, I did it manually for now so I don’t have a solution yet. But I have not debugged deep into this to be honest. My hope is that someone else who is as exited as me about the latest Netbox versions can pick up this PR, also, I guess plugin installation could be improved by the role (mostly pip).


tasosalvas (dismissed):
Clear in terms of doc and style, do flag me if some doc-related concern pops up

CHANGELOG.rst
```diff
@@ -61,6 +61,11 @@ General
   :ref:`debops.global_handlers` role to allow for easier cross-role handler
   notification. The role has been imported in roles that rely on the handlers.
 
+:ref:`debops.netbox` role
+'''''''''''''''''''''''''
+
+- Update to Netbox 2.8. The newly added plugin support can be configured from DebOps.
```
drybjed: The update notice should be moved to the "Updates of upstream application versions" section in the Changelog. It's currently present if it hasn't been there before. Fixing this will also fix the merge conflict.

ansible/roles/netbox/defaults/main.yml
```diff
@@ -135,7 +135,7 @@ netbox__git_repo: 'https://github.com/netbox-community/netbox.git'
 # .. envvar:: netbox__git_version [[[
 #
 # The :command:`git` branch or tag which will be installed.
-netbox__git_version: 'v2.6.3'
+netbox__git_version: 'v2.8.7'
```
drybjed: You should also update the `meta/watch-netbox` watch file.

Since this PR was made, Netbox was updated to `v2.8.9`, perhaps you could use this instead. Or, we can wait for merging this PR for `v2.9.0` release, although I'm not sure when it will be released. I'm fine with switching to `v2.8.9` for now.

ansible/roles/netbox/defaults/main.yml
```diff
@@ -583,24 +579,31 @@ netbox__config_media_root: '{{ netbox__data + "/media" }}'
 # Absolute path where Netbox reports (Python modules) are stored.
 netbox__config_reports_root: '{{ netbox__data + "/reports" }}'
                                                                    # ]]]
+# .. envvar:: netbox__config_plugins [[[
+#
+# List of Netbox plugins.
+netbox__config_plugins: []
+                                                                   # ]]]
```
drybjed, line 588: You could add an empty line before `# ]]]` marker to have the variables better stand out in the defaults file.

ansible/roles/netbox/defaults/main.yml
```diff
@@ -583,24 +579,31 @@ netbox__config_media_root: '{{ netbox__data + "/media" }}'
...
+# .. envvar:: netbox__config_plugins_config [[[
+#
+# Netbox plugins configuration settings.
+netbox__config_plugins_config: {}
+                                                                   # ]]]
```
drybjed, line 594: Any plans to add an example in the documentation for how the plugins can be configured? I guess since there's no plugins yet that I know of, you could use the example from the official documentation.

ansible/roles/netbox/tasks/main.yml
```diff
@@ -195,6 +196,7 @@
     SUPERUSER_NAME: '{{ netbox__superuser_name }}'
     SUPERUSER_EMAIL: '{{ netbox__superuser_email }}'
     SUPERUSER_PASSWORD: '{{ netbox__superuser_password }}'
+    ## TODO: Does not work with 2.8 anymore.
```
drybjed, line 199: You could switch the superuser account creation to the [method used in the `mailman` role](https://github.com/debops/debops/blob/master/ansible/roles/mailman/tasks/main.yml#L114-L119). The password will not be generated anymore, but the application should send an e-mail to the superuser with the password reset link. This is how Mailman 3 works, I'm not sure how Netbox will behave, but since it's Django this is probably built-in.

ansible/roles/netbox/templates/usr/local/lib/netbox/configuration.py.j2
```diff
@@ -167,6 +175,13 @@ NAPALM_ARGS = {{ netbox__config_napalm_args | to_nice_json }}
 # Determine how many objects to display per page within a list. (Default: 50)
 PAGINATE_COUNT = {{ netbox__config_paginate_count }}
 
+# Enable installed plugins. Add the name of each plugin to the list.
+PLUGINS = {{ netbox__config_plugins | to_nice_json }}
+
+# Plugins configuration settings. These settings are used by various plugins that the user may have installed.
+# Each key in the dictionary is the name of an installed plugin and its value is a dictionary of settings.
+PLUGINS_CONFIG = {{ netbox__config_plugins_config | to_nice_json }}
+
```
drybjed, line 184: Thanks for updating the config. I think that in the future we should switch to the templating solution done in the `mailman` role, that way configuration options can be done conditionally, based on the NetBox version. But that can be dealt with in another pull request.

docs/ansible/roles/netbox/getting-started.rst
```diff
@@ -15,8 +15,8 @@ General deployment notes
 ------------------------
 
 The application will be by default deployed both on ``ipam.{{ ansible_domain
-}}`` as well ass ``dcim.{{ ansible_domain }}`` DNS domains, for convenience.
```
drybjed, line 18: Whoops. :-) I think that my keyboard is the culprit - I noticed occasional repeats of keys which are unintended. I'm currently using [SteelSeries Merc Stealth](https://steelseries.com/blog/steelseries-discontinued-gaming-keyboards-130) but it seems that the keyboard is starting to show its age... Any suggestions for a replacement?

drybjed (changes requested):
Thanks for the update. I tried it out in a development environment with production data, and it works as expected. In the comments I added some suggestions for improvements.

docs/ansible/roles/netbox/getting-started.rst
tasosalvas: There's that company named "Microsoft", their cheapest keyboards are pretty good. Something to do with wanting their logo on as many desks as possible... I don't know why, I think they also sell mice.

Mine doubles as an ashtray as well.

ypid: > Whoops. :-)

Happens :) I don’t have a recommendation.

ansible/roles/netbox/defaults/main.yml
ypid: I guess I know. I just did not put so much effort into this because I wrote software that can lint/fix this.

> Never send a human to do a machine's job.

Ref: https://github.com/debops/debops/issues/1460

ansible/roles/netbox/defaults/main.yml
ypid: Makes sense. Examples are always nice.

Flags: @tasosalvas

ansible/roles/netbox/tasks/main.yml
ypid: What would you prefer? Email or the way it was before? For my setup and strategy, the way it was before is preferred for the following reasons:

* I don’t like all this mail reset/set stuff that is currently done by Email.
* I would ultimately like to have DebOps automatically generate and write passwords into my password manager for me to just login. (I should probably say that for my self-hosting, I decided against LDAP. This decision shifts my prios a bit of course.). (You can see first steps into this direction with the `etesync__url` variable.).

ansible/roles/netbox/templates/usr/local/lib/netbox/configuration.py.j2
ypid: Sounds like a plan. I agree. TBD later.

ansible/roles/netbox/tasks/main.yml
drybjed: I had to deal with a similar problem while updating the `mailman` role. The "manual" method similar to this one didn't work there as well, and the official `./manage.py createsuperuser` command is an interactive command, not suitable for Ansible unless we deal with it somehow via `expect` - I tried to do that but I couldn't manage to do it properly. In the end, the `django_manage` module and just letting the user handle it via password reset email is what I went with.

It's not really elegant and requires a working mail setup, but for a LDAP-based environment which I'm slowly aiming DebOps towards it seems to be sufficient. In the NetBox case, you can define separate groups in LDAP for users, staff and administrators so when you add your account to the admin group, password reset is not needed.

If you manage to somehow make superuser account with autogenerated password via Ansible and it's not a horrible hack, let me know. :)

ansible/roles/netbox/defaults/main.yml
ypid: Sounds good to me. Seems 2.8.9 fixed a bug I faced so I am motivated :) Let me test this for a few days, I will come back to the PR.

drybjed: @ypid, the NetBox v2.9.1 has been released, do you want to try it out and update the PR for the current version?

ypid: And again, v2.9.x fixes a few issues I faced. I just upgraded :) Lets see when I can finish the PR.

drybjed (approved):
I tested the clean installation (worked fine) and upgrade from the `v2.6.3` version which was the previously supported one. In the latter case a direct upgrade to `v2.9.1` failed during database migration, but an upgrade to `v2.8.9` version first worked fine.

Nice job with the changes. Since they are extensive, I will not backport them to the stable DebOps releases.

I think that the `netbox_topology_views` plugin might be useful enough to be installed by default. But we can add that later.

ypid:
Thanks very much. So the Django user creation works (again)? One backlog entry that solved itself? I don’t have that often :wink:

drybjed:
Hmmm, shoot. I forgot about that. No, it didn't solve itself... But I can adapt the method from `mailman` for now, since there's no better alternative.

Well, the PR was not backported to stable releases, so at least that's a saving grace for me. :)

PR #1434, opened 2020-07-24
[roundcube] Discard cleandb.sh stdout

imrejonk:
The 'Roundcube daily database housekeeping' cronjob has output similar
to this:
```
0 records deleted from 'contacts'
0 records deleted from 'contactgroups'
0 records deleted from 'identities'
```
The output is sent by email to the roundcube user, whose mailbox is
usually forwarded to the postmaster or system administrator. The
receiver of these emails may find them annoying.

This change discards the output by redirecting it to /dev/null. Error
output however should still be forwarded.

### DIFF
--- a/ansible/roles/roundcube/tasks/deploy_roundcube.yml
+++ b/ansible/roles/roundcube/tasks/deploy_roundcube.yml
@@ -120,7 +120,7 @@
   cron:
     name: Roundcube daily database housekeeping
     user: '{{ roundcube__user }}'
-    job: '{{ roundcube__git_dest }}/bin/cleandb.sh'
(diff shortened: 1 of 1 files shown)


imrejonk:
Alternatives to discarding stdout would be saving it to a log file (and setting up logrotate), or making the cronjob user-configurable.

SuperSandro2000:
Are errors logged to stderr? If so this is totally fine.

imrejonk:
@SuperSandro2000 I believe so, yes. The `>` operator only redirects stdout, so any stderr output will still end up in an email message.

drybjed (approved):
Good idea, thanks.

PR #1437, opened 2020-07-24
Various improvments

ypid:
I made an extra effort to keep the commits as clean as possible :)

@drybjed: Please wait for my approval when you release a new version. I added myself as a codeowner of the respective file. The background is that I normally buffer up minor fixes until I have a bunch before I open a PR. When you wait for my approval, I could flush my buffers. Alternatively ping me shortly before you make an release (mailing list for example).


tasosalvas (approved):

imrejonk:
Wait... Since Ansible 2.8, we don't have to type statements like these anymore?
```
nodejs__node_upstream: '{{ ansible_local.nodejs.node_upstream|bool
                           if (ansible_local|d() and ansible_local.nodejs|d() and
                               ansible_local.nodejs.node_upstream is defined)
                           else nodejs__distribution_release in [ "jessie", "xenial" ] }}'
```

drybjed:
@imrejonk Yes. :-) There are cases when you need to use `{{ x if condition else y }}` statement, but for simple check if a parameter is defined, like in the Ansible local facts, you can go with `{{ x | d(y) }}` format.

imrejonk:
@drybjed Alrighty, that will save some time as well as keyboard wear. It's good to do reviews like these together, it can be educational :)

ansible/roles/resources/tasks/main.yml
```diff
@@ -297,6 +297,16 @@
          ((item.path|d() or item.dest|d() or item.name|d()) or item))
   tags: [ 'role::resources:paths' ]
 
+- name: Set custom file capabilities
+  capabilities:
+    path:       '{{ item.path  | d(item.name) }}'
+    capability: '{{ item.capability }}'
+    state:      '{{ item.state | d("present") }}'
+  with_flattened:
```
imrejonk, line 305: While `with_flattened` is not (yet) deprecated, Ansible recommends using `loop` with the `flatten` filter: https://docs.ansible.com/ansible/latest/user_guide/playbooks_loops.html#with-flattened

imrejonk (approved):

drybjed (approved):
Thanks for the PR. I tested the `nodejs` condition changes and they seem to work as expected. I'll backport relevant commits to the stable releases.

@ypid: I will send an e-mail about the next release on the `debops-devel` list a day or two before the release will happen. The next one will be on August 2nd, after the Debian point release.

ansible/roles/resources/tasks/main.yml
ypid: Thank you for pointing this out! I will prefer the `loop` `flatten` construct in the future.

ypid:
> Wait... Since Ansible 2.8, we don't have to type statements like these anymore?

For the protocol, this has been discussed further in #1371.

PR #1439, opened 2020-07-27
Update nginx security directives

apapsch:
This adapts the nginx configuration according to [Mozilla recommendations](https://ssl-config.mozilla.org/#server=nginx&version=1.17.7&config=intermediate&openssl=1.1.1d&guideline=5.6):

* updated default cipher suite
* TLSv1.0 and 1.1 removed from default protocols
* default TLS curves expanded

The http block includes the ssl.conf snippet, which avoids having nginx defaults take precedence over desired config in server blocks preceding the one server block containing explicit config directives.

Clients are allowed to choose ciphers, because clients know best if they have support for hardware-accelerated AES.

No dhparam is set if the only protocol is TLSv1.3, because that protocol uses Ephemeral Diffie-Hellman key exchange protocol, which employs one-time keys for the current network session. At the end of the session, the key is discarded. The hard coded protocol identifier is a bit of concern; one may argue that there's no bad coming from specifying dhparam if TLSv1.3 does not use them (removing the need for TLSv1.3 special casing).

Last, HSTS age is raised to 2 years.

### DIFF
--- a/ansible/roles/nginx/defaults/main.yml
+++ b/ansible/roles/nginx/defaults/main.yml
@@ -476,10 +476,14 @@ nginx_pki_hook_action: 'reload'
 # .. envvar:: nginx_ssl_dhparam [[[
 #
 # Path to the file with Diffie-Hellman parameters to use by the webserver.
-nginx_ssl_dhparam: '{{ (ansible_local.dhparam[nginx_ssl_dhparam_set]
-                        if (ansible_local|d() and ansible_local.dhparam|d() and
-                            ansible_local.dhparam[nginx_ssl_dhparam_set]|d())
-                        else "") }}'
+nginx_ssl_dhparam: '{{ (""
+                        if nginx_default_tls_protocols|length == 1 and
+                           nginx_default_tls_protocols[0] == "TLSv1.3"
+                        else
+                           (ansible_local.dhparam[nginx_ssl_dhparam_set]
+                            if (ansible_local|d() and ansible_local.dhparam|d() and
+                                ansible_local.dhparam[nginx_ssl_dhparam_set]|d())
+                            else "")) }}'
 
                                                                    # ]]]
 # .. envvar:: nginx_ssl_dhparam_set [[[
--- a/ansible/roles/nginx/defaults/main.yml
+++ b/ansible/roles/nginx/defaults/main.yml
@@ -523,7 +523,7 @@ nginx_default_tls_protocols: '{{ [ "TLSv1.2", "TLSv1.3" ]
 #
 # See also: https://security.stackexchange.com/questions/31772/
 # Set to ``False`` to disable ECC.
-nginx_default_ssl_curve: 'secp384r1:secp256r1:X448:X25519'
+nginx_default_ssl_curve: 'secp384r1:secp256r1'
 
                                                                    # ]]]
 # .. envvar:: nginx_default_ssl_verify_client [[[
--- a/ansible/roles/nginx/defaults/main.yml
+++ b/ansible/roles/nginx/defaults/main.yml
@@ -492,7 +492,7 @@ nginx_ssl_dhparam_set: 'default'
 #
 # Default set of cipher suites to use.
 # Refer to ``nginx_ssl_ciphers`` for details.
-nginx_default_ssl_ciphers: 'bettercrypto_org__set_b_pfs'
+nginx_default_ssl_ciphers: 'mozilla_intermediate'
 
                                                                    # ]]]
 # .. envvar:: nginx_default_tls_protocols [[[
@@ -501,9 +501,9 @@ nginx_default_ssl_ciphers: 'bettercrypto_org__set_b_pfs'
 # version 1.13.0 and up.
 #
 # See also: https://nginx.org/en/docs/http/ngx_http_ssl_module.html#ssl_protocols
-nginx_default_tls_protocols: '{{ [ "TLSv1", "TLSv1.1", "TLSv1.2", "TLSv1.3" ]
+nginx_default_tls_protocols: '{{ [ "TLSv1.2", "TLSv1.3" ]
                                  if ansible_local.nginx.version|d("0.0.0") is version("1.13.0", ">=")
-                                 else [ "TLSv1", "TLSv1.1", "TLSv1.2" ] }}'
+                                 else [ "TLSv1.2" ] }}'
 
                                                                    # ]]]
 # .. envvar:: nginx_default_ssl_curve [[[
@@ -516,7 +516,7 @@ nginx_default_tls_protocols: '{{ [ "TLSv1", "TLSv1.1", "TLSv1.2", "TLSv1.3" ]
 #
 # See also: https://security.stackexchange.com/questions/31772/
 # Set to ``False`` to disable ECC.
-nginx_default_ssl_curve: 'secp384r1'
+nginx_default_ssl_curve: 'secp521r1:secp384r1:prime256v1'
 
                                                                    # ]]]
 # .. envvar:: nginx_default_ssl_verify_client [[[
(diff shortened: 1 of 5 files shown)


apapsch:
With the default settings I now get A+ rating on SSLLabs, with these hosts getting handshake failure:

* IE 11 / Win Phone 8.1 (there is a separate working entry for "Win Phone 8.1 Update")
* Safari 6 / iOS 6.0.1
* Safari 7 / iOS 7.1
* Safari 7 / OS X 10.9
* Safari 8 / iOS 8.4
* Safari 8 / OS X 10.10

The iOS/OS X entries are rather old. iOS >= 9 and OS X >= 10.11 work fine.

I also tested TLSv1.3-only setup, which gives just an A rating, because there is still limited client availability.

ansible/roles/nginx/defaults/main.yml
```diff
@@ -476,10 +476,14 @@ nginx_pki_hook_action: 'reload'
 # .. envvar:: nginx_ssl_dhparam [[[
 #
 # Path to the file with Diffie-Hellman parameters to use by the webserver.
-nginx_ssl_dhparam: '{{ (ansible_local.dhparam[nginx_ssl_dhparam_set]
-                        if (ansible_local|d() and ansible_local.dhparam|d() and
-                            ansible_local.dhparam[nginx_ssl_dhparam_set]|d())
-                        else "") }}'
+nginx_ssl_dhparam: '{{ (""
```
imrejonk, line 479: FWIW, removing the DHE parameter configuration does not have any effect when using TLS 1.3, as custom DHE groups have been dropped in TLS 1.3 (the client chooses from a set of standardized finite-field groups instead). So this change only serves to clean up the generated configuration a bit. However, it might be nice not to have an 'exception' for TLS 1.3 in this regard.

ansible/roles/nginx/defaults/main.yml
```diff
@@ -501,9 +508,9 @@ nginx_default_ssl_ciphers: 'bettercrypto_org__set_b_pfs'
 # version 1.13.0 and up.
 #
 # See also: https://nginx.org/en/docs/http/ngx_http_ssl_module.html#ssl_protocols
-nginx_default_tls_protocols: '{{ [ "TLSv1", "TLSv1.1", "TLSv1.2", "TLSv1.3" ]
+nginx_default_tls_protocols: '{{ [ "TLSv1.2", "TLSv1.3" ]
```
imrejonk, line 511: This change is fine by me, but I think this warrants an entry in the changelog as well as the upgrade notes.

ansible/roles/nginx/defaults/main.yml
```diff
@@ -516,7 +523,7 @@ nginx_default_tls_protocols: '{{ [ "TLSv1", "TLSv1.1", "TLSv1.2", "TLSv1.3" ]
 #
 # See also: https://security.stackexchange.com/questions/31772/
 # Set to ``False`` to disable ECC.
-nginx_default_ssl_curve: 'secp384r1'
+nginx_default_ssl_curve: 'secp521r1:secp384r1:prime256v1'
```
imrejonk: [NCSC-NL](https://english.ncsc.nl/publications/publications/2019/juni/01/it-security-guidelines-for-transport-layer-security-tls) considers `secp521r1` and `prime256v1` insufficient, supposedly because they are not commonly used. Instead they recommend the curves `secp384r1`, `secp256r1`, `x448` and `x25519`. I also don't see `secp512r1` in Mozilla's recommendations. Care to explain why we should default on these curves?

ansible/roles/nginx/templates/etc/nginx/sites-available/default.conf.j2
```diff
@@ -633,13 +633,13 @@ server {
         ssl_certificate           {{ nginx_tpl_ssl_certificate }};
         ssl_certificate_key       {{ nginx_tpl_ssl_certificate_key }};
         ssl_protocols             {{ (item.tls_protocols | d(nginx_default_tls_protocols)) | join(" ") }};
-        ssl_prefer_server_ciphers on;
+        ssl_prefer_server_ciphers off;
```
imrejonk: [NCSC-NL](https://english.ncsc.nl/publications/publications/2019/juni/01/it-security-guidelines-for-transport-layer-security-tls) advises configuring the server to prefer ciphers in a certain order, as it prioritizes the fastest and safest algorithms. Mozilla also recommends this practice when weaker ciphers are selected. I suggest we keep this flag `on` just to be safe.

ansible/roles/nginx/defaults/main.yml
```diff
@@ -492,7 +496,10 @@ nginx_ssl_dhparam_set: 'default'
 #
 # Default set of cipher suites to use.
 # Refer to ``nginx_ssl_ciphers`` for details.
-nginx_default_ssl_ciphers: 'bettercrypto_org__set_b_pfs'
+nginx_default_ssl_ciphers: '{{ "mozilla_modern"
+                               if nginx_default_tls_protocols|length == 1 and
+                                  nginx_default_tls_protocols[0] == "TLSv1.3"
```
imrejonk, line 501: This would cause a fallback to the cipher suites in TLS 1.3, which includes `TLS_AES_128_CCM_8_SHA256`. [NCSC-NL](https://english.ncsc.nl/publications/publications/2019/juni/01/it-security-guidelines-for-transport-layer-security-tls) explicitly considers this cipher to be insufficient because it is "a variant of AES-CCM with a truncated authentication tag, which has a lower security equivalent for integrity protection."

imrejonk (changes requested):

tasosalvas:
Hey there,

I would like to see a paragraph in a  **Default configuration** section in `doc/ansible/roles/nginx/getting-started.rst`, similar to the one [dnsmasq](https://docs.debops.org/en/master/ansible/roles/dnsmasq/getting-started.html) has.

* What the default behavior is, explained in a way that if the end user is happy with it they don't have to learn anything more
* The reasoning the default behavior is based on, with a link to the source it's based on
* A very brief reference to the mechanisms used to configure it further

Your original PR text is an excellent starting point for that purpose, and having just dove into the feature you're just the right person to explain it.

No worries about the million other default behaviors of the role that don't exist there (unless you feel up to it), we're taking it one step at a time :).

ansible/roles/nginx/defaults/main.yml
drybjed: I can handle the Changelog update, so that the patches are easier to  backport to the stable releases.

ansible/roles/nginx/templates/etc/nginx/sites-available/default.conf.j2
drybjed: I agree that we should stick to preferring the server ciphers instead of the client specified ones. This can prevent downgrade attacks and shouldn't be an issue even if that's not relevant anymore with TLSv1.3.

ansible/roles/nginx/defaults/main.yml
drybjed: I think that I'm fine with the empty `nginx_ssl_dhparam`. The condition is explicitly on the Ansible side, and the resulting configuration shows what is really happening (DHE is not used), so there might be less confusion during server debugging, etc.

ansible/roles/nginx/templates/etc/nginx/sites-available/default.conf.j2
drybjed: After reading about changes in TLSv1.3, I guess client-selected ciphers make sens when it is used. But if TLSv1.2 is also available, I would stick to server ciphers. Therefore this needs to be enabled/disabled conditionally, similarly to how DHE params are dealt with. Don't forget about the instance in the `ssl.conf` snippet as well.

drybjed (changes requested):
Nice set of changes. There are a few issues to be dealt with, check the comments.

I think that this really shows that we need a `crypto` role which could consolidate ciphers, protocols, curves in one place and expose these details as local facts, in `openssl` and `gnutls` formats. Then other roles could use these facts to configure themselves accordingly. Putting this in the `pki` role is probably not a good idea.

ypid (changes requested):
Thanks for your work. I agree with the general direction of it. Some review threads need to be addressed.

Some input on the changes. I have been using [some for over a year already](https://github.com/ypid/ypid-ansible-inventory/blob/master/group_vars/public_ypid_paranoid_debops/public_ypid_paranoid_debops_nginx.yml), namely max HSTS of 2 earth years and dropping TLS 1.0 and 1.1. I have stayed away from TLS 1.3 although it has quite a few nice improvements. My only reason for this is that new protocols (and implementations) are not tested as much. But I am fine TLS 1.3 with TLS 1.2 as fallback now as DebOps default. For my self-hosting, I might start deploying TLS 1.3 some time in the future as well.

ansible/roles/nginx/defaults/main.yml
apapsch: Mozillas modern recommendation does not specify an explicit ciphersuite, leaving it to nginx default, which is `HIGH:!aNULL:!MD5`. Running `openssl ciphers 'HIGH:!aNULL:!MD5'` gives me a list of ciphers, which does not include `TLS_AES_128_CCM_8_SHA256`, only those TLSv1.3 ciphers:

* `TLS_AES_256_GCM_SHA384`
* `TLS_CHACHA20_POLY1305_SHA256`
* `TLS_AES_128_GCM_SHA256`

Thus I saw no problem for TLSv1.3, but please correct me if I'm wrong. The story for TLSv1.2 is different: the above mentioned openssl command prints many ciphers, including AES-CCM ones. Mozillas modern recommendation is TLSv1.3-only, therefore `mozilla_modern` shouldn't be used with TLSv1.2 enabled. Though debops would not forbid it.

ansible/roles/nginx/templates/etc/nginx/sites-available/default.conf.j2
apapsch: Per drybjeds suggestions, I'll disable server-preferred ciphers only if TLSv1.3 is the only protocol in use.

ansible/roles/nginx/defaults/main.yml
apapsch: What goes into ssl-config.mozilla.org is first brainstormed on a wiki page. It is linked (somewhat inconspicuously) in the footer of ssl-config. https://wiki.mozilla.org/Security/Server_Side_TLS mentions:

> TLS curves: X25519, prime256v1, secp384r1

Indeed I must have copied the `secp521r1:secp384r1:prime256v1` from the wrong place, will fix it to include secp384r1, secp256r1, x448, x25519 and prime256v1.

apapsch:
> I would like to see a paragraph in a Default configuration section in doc/ansible/roles/nginx/getting-started.rst, similar to the one dnsmasq has.

I will add it.

> I think that this really shows that we need a crypto role which could consolidate ciphers, protocols, curves in one place and expose these details as local facts, in openssl and gnutls formats. Then other roles could use these facts to configure themselves accordingly. Putting this in the pki role is probably not a good idea.

Indeed. What Mozilla describes as their modern recommendation, is more than just a cipher suite. While differences to intermediate/old are small, having a consolidating structure could help keep special cases low on maintenance. For example, the hard coded TLSv1.3 special case for `ssl_dhparam` would need to be expanded for a potential TLSv1.4. If on the other hand the structure would specify `dhe_enabled` (or similar), a TLSv1.4 would not require changes in existing code.

ansible/roles/nginx/defaults/main.yml
imrejonk: `openssl ciphers 'HIGH:!aNULL:!MD5'` gives me a long list of ciphers, including these CCM8 ciphers:
```
ECDHE-ECDSA-AES256-CCM8
DHE-RSA-AES256-CCM8
ECDHE-ECDSA-AES128-CCM8
DHE-RSA-AES128-CCM8
DHE-PSK-AES256-CCM8
AES256-CCM8
PSK-AES256-CCM8
DHE-PSK-AES128-CCM8
AES128-CCM8
PSK-AES128-CCM8
```
This is with OpenSSL version 1.1.1d-0+deb10u3.

apapsch: > openssl ciphers 'HIGH:!aNULL:!MD5' gives me a long list of ciphers, including these CCM8 ciphers:

I see, following Mozilla recommendation to the letter is not suitable for debops, where options are more finegrained. When you enable `mozilla_modern` cipher suite, debops does not forbid to use TLSv1.2, even though Mozilla modern recommendation is TLSv1.3-only (which would defang the CCM8 ciphers).

The security hole needs to be fixed along with the usability problem. I'm currently trying to create a cipher suite name that `openssl ciphers` expands to just TLSv1.3 ciphers, which can then be used for the `mozilla_modern` value in `nginx_ssl_ciphers`. Nginx should then fail if configured with TLSv1.2-only (test pending).

apapsch: I haven't found (quickly) how to specify disabling all TLSv1.2 and below ciphers. The webserver error could also be entirely avoided, if I just copy `nginx_ssl_ciphers.mozilla_intermediate` to `mozilla_modern`. This would not change a thing for TLSv1.3-only, while TLSv1.2-only or dual would not have the usability bug and potential security hole. If you haven't any more concerns, I'll go with copying.

ansible/roles/nginx/defaults/main.yml
apapsch: After a bit more digging I found there's some concern over the rigidity of `prime256v1` because it's a NIST curve. See here for the rabbit hole: https://github.com/nodejs/node/issues/1495

I'm now removing `prime256v1` from the debops list. Just NCSC-NL recommended curves remain.

apapsch: Openssl on Debian Buster only supports one of the four NCSC-NL curves: `secp256r1` (see `openssl ecparam -list_curves`). Configuring just the one curve does not seem to limit availability (checked SSLLabs and two physical devices). One curve it is.

apapsch:
I have named `Default configuration` as `Security defaults`, because a user might expect more docs than just security defaults with the former. Though maybe it should be called `Default configuration`, because renaming it later would break URLs with anchor.

tasosalvas (approved):
Yeah I guess it does look better named _Security Defaults_ for now. It could be a subsection but no need to nest it in an empty section.

The anchor will work fine once we have a second default behavior description and it becomes a subsection.

My reasoning for _Default configuration_ was precisely to make that lack of doc for the rest of the defaults apparent, so future contributors can easily spot what's missing and where it goes. I'm going to first address it in a first-time-contrib handholding doc before we start making things look a bit uglier to promote it, though. :)

Anyway, thank you very much for taking the time! The text itself is great.

drybjed (approved):
I tried the changes out on a development webserver, `nginx` seems to work fine. @apapsch, thanks for your research and updated code!

I'll merge the changes to `master` for now, if anybody uses that branch in their environments let me know if there are any issues with it. If there are none for 24h, I'll backport the changes to the stable DebOps releases.

drybjed:
Changes have been backported to stable DebOps branches. They will be available via the Ansible Galaxy Collections and the Python packages on the next point release, in about two weeks.

PR #1440, opened 2020-07-27
[ldap] Fix issues related to network bonding

imrejonk:
These changes fix the following issues related to creating and updating
host entries in the LDAP database, when the host to be added/updated is
configured with network bonding:

- ldap__device_ip_addresses and ldap__device_mac_addresses were not
  lists, but lists of lists. Flattening happened in another variable.
- ldap__device_ip_addresses did not contain IP addresses of bond*
  interfaces.
- ldap__device_mac_addresses can, when the host is configured for
  network bonding, contain duplicate MAC addresses. This is because
  the Linux Ethernet Bonding Driver changes the MAC address of the
  secondary physical interface so that it uses the same MAC address as
  the primary interface (at least in active-backup setups). This causes
  an ldap.TYPE_OR_VALUE_EXISTS error.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -116,7 +116,7 @@ LDAP
 :ref:`debops.ldap` role
 '''''''''''''''''''''''
 
-- Fixed multiple issues with adding and updating hosts to the LDAP directory,
+- Fixed multiple issues with adding and updating hosts to the LDAP directory
   when these hosts were configured for network bonding.
 
 :ref:`debops.owncloud` role
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -113,6 +113,12 @@ LDAP
 - Fixed an issue where the role would attempt to add APT keys from a PGP
   keyserver without installing the :command:`gnupg` package first.
 
+:ref:`debops.ldap` role
+'''''''''''''''''''''''
+
+- Fixed multiple issues with adding and updating hosts to the LDAP directory,
+  when these hosts were configured for network bonding.
+
 :ref:`debops.owncloud` role
 ''''''''''''''''''''''''''''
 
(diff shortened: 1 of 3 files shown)


tasosalvas (dismissed):
I don't think that comma is needed in the changelog but I don't mean to be a grammar nazi... after all you guys are still waiting for those bicycles :D

Would some of these flattenings and unique-ings benefit from `parse_kv` maybe?

imrejonk:
@tasosalvas Zuerst mein Fahrrad zurück! I didn't know the tale of our bicycles was known outside of the Netherlands and Germany :)

Regarding your `parse_kv` comment, I'll give it a try this afternoon.

drybjed (approved):
Thanks for finding and fixing this. I'll backport the changes to the stable releases as well.

imrejonk:
@tasosalvas Looking at the `parse_kv_config` and `parse_kv_items` [documentation](https://docs.debops.org/en/master/ansible/roles/ansible_plugins/getting-started.html#custom-ansible-filter-plugins), it looks like these plugins are used to parse lists of dictionaries. I don't think that would be useful in this case as the IP and MAC address variables deal with a list of lists instead.

tasosalvas:
@imrejonk oh it goes much deeper than that with `parse_kv`, it weighs, it merges, it knows "present" from "absent", it can totally be useful with just keys and no values. No worries though.

drybjed:
The `parse_kv_config` and `parse_kv_items` filters are useful in more complex situations, where you want to control positioning of elements in a config file, or generate multiple files and be able to modify them conditionally. But for simple lists of items such as in this case they are overkill. And if they were used here, on the other end the role would have to map the resulting dictionaries from the list to get to the actual values again.

PR #1447, opened 2020-07-31
[pki] Ensure Subject CN falls back to default_domain

tumd:
The only time 'CN=...' would be anything other than $(hostname -f) was
if the name of the realm was specifying a domain-name (i.e. including a
'.'), or if the subject-variable was including the CN specifically in
the realm configuration. This behavior is however not according to what
the documentation says.

This is said in the documentation;

    subject, acme_subject

      Optional. The Distinguished Name of the certificate, specified as a
      list of DN elements. If not specified, a CommonName based on the
      default domain of the given PKI realm will be used. Empty string
      elements of the list will be ignored.

This patch should now make the documentation true and set the value of
pki_default_domain as CN as a fallback.

Consider the following PKI Realm;
```
pki_ca_domain: 'fqdn.tld'
pki_group_realms:
  - name: 'tinc'
    acme: False
    default_domain: 'test.tinc.fqdn.tld'
    subject_alt_names:
      - 'DNS:{{ ansible_fqdn }}'
```
Before this PR the above would produce a certificate containing;
Subject: CN = test.fqdn.tld
Subject Alternative Name: 
  DNS:test.tinc.fqdn.tld, DNS:test.fqdn.tld

After this PR;
Subject: CN = test.tinc.fqdn.tld
Subject Alternative Name: 
  DNS:test.tinc.fqdn.tld, DNS:test.fqdn.tld

The only difference is the CN, but since all domains exists in SAN the certificates should be interchangeable.
Ie. depending on the application that's going to use such certificate this may or may not be a problem. If the application don't have support for SAN, this may be a problem.
A such case may be setting up a Galera cluster with SST over socat+TLS. See: https://jira.mariadb.org/browse/MDEV-16915

I haven't tested this patch overwhelmingly well. But it seem to do what the documentation says - which I would argue is the more correct behavior than the current.

### DIFF
--- a/ansible/roles/pki/files/usr/local/lib/pki/pki-realm
+++ b/ansible/roles/pki/files/usr/local/lib/pki/pki-realm
@@ -191,7 +191,6 @@ initialize_environment () {
     config["acme_root_ca_file"]="mozilla/DST_Root_CA_X3.crt"
 
     config["pki_default_ca_bundle"]="/etc/ssl/certs/ca-certificates.crt"
-    config["pki_default_subject"]="CN=$(hostname -f)"
     config["pki_default_fqdn"]="$(hostname -f)"
     config["pki_default_domain"]="$(get_dnsdomainname)"
     config["pki_default_subdomains"]="_wildcard_"
@@ -217,7 +216,6 @@ initialize_environment () {
     config["subdomains"]=""
     config["subject_alt_names"]=""
 
-    config["acme_default_subject"]="CN=$(get_dnsdomainname)"
     config["acme_default_domain"]="$(get_dnsdomainname)"
     config["acme_default_subdomains"]="www/ftp/mail/smtp/imap"
 
@@ -419,7 +417,7 @@ run_pki_hooks () {
 
         export PKI_SCRIPT_REALM="${config['name']}"
         export PKI_SCRIPT_FQDN="${config['pki_default_fqdn']}"
-        export PKI_SCRIPT_SUBJECT="${config['subject']:-${config['pki_default_subject']}}"
+        export PKI_SCRIPT_SUBJECT="${config['subject']:-CN=${config['pki_default_domain']}}"
         export PKI_SCRIPT_DOMAINS="${config['domains']:-${config['pki_default_domain']}}"
         export PKI_SCRIPT_SUBDOMAINS="${config['subdomains']:-${config['pki_default_subdomains']}}"
         export PKI_SCRIPT_PRIVATE_KEY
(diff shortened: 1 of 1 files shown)


drybjed:
I tried out the PR in a test environment and certificates seem to be generated fine.

You mentioned that with the changed CN some services like MariaDB that don't rely on the SubjectAltName fields might have issues with changes. In such case, why not use the FQDN of the host instead for the non-domain PKI realms? It is available as `pki_default_fqdn` shell variable and should work fine with your changes.

As for the documentation suggesting that the CommonName will be based on the domain - I think that the actual code is correct and it's the documentation that is in the wrong, based on the experience with MariaDB and similar services that rely on the CN. In such cases when you configure the database connection over TLS, you will specify the FQDN of the database host; having it be the same in the CN will make such service work fine. I would change the documentation instead to mention the "FQDN" instead of the "domain" here.

tumd:
> I tried out the PR in a test environment and certificates seem to be generated fine.
> 
> You mentioned that with the changed CN some services like MariaDB that don't rely on the SubjectAltName fields might have issues with changes. In such case, why not use the FQDN of the host instead for the non-domain PKI realms? It is available as `pki_default_fqdn` shell variable and should work fine with your changes.
> 

In my example I want a group of hosts connected together with tinc and a common tinc-only subnet to exclusively communicate over this vpn interface with a separate and predefined (/etc/hosts) set of sub-domains under tinc.fqdn.tld-domain. Thus I create a separate and common realm for this group of hosts only used exclusively for the services communicating over the VPN.
I.e this tinc-only realm-group is used for internal service-communication for a group of hosts over vpn using an IP and fqdn that's only relevant on each hosts VPN interface, along side the 'domain' realm still being used for services exposed to the users on the standard eth's.
And yes, I run TLS over an encrypted VPN - because why not :)

> As for the documentation suggesting that the CommonName will be based on the domain - I think that the actual code is correct and it's the documentation that is in the wrong, based on the experience with MariaDB and similar services that rely on the CN. In such cases when you configure the database connection over TLS, you will specify the FQDN of the database host; having it be the same in the CN will make such service work fine. I would change the documentation instead to mention the "FQDN" instead of the "domain" here.

I'm not sure I agree. I think the documentation is more correct. 
If you don't specify a FQDN as the realm-name, the CN= will _always_ become CN=$(hostname -f). The only way to override the CN in such cases is by setting `subject: [ 'cn=host.tinc.fqdn.tld' ]` - which to me doesn't seem right.
Instead setting `default_domain` to override the CN seem more logical and user friendly.

Disclaimer: It was a while since I looked at this, thus writing mostly from memory, and I might as well remember things wrong :)

PR #1449, opened 2020-08-01
[filebeat] How can Filebeat be part of ELK? It is not -> Elastic Stack

ypid:
@drybjed You wrote it yourself in the changelog. ELK means "Elasticsearch-Logstash-Kibana". But how can the Beats agents be part of ELK :) ? They are not (which is not nice). That is why Elastic changed the name for good. ELK is legacy for some time now. We should not use such legacy names. They are not precise. https://www.elastic.co/elastic-stack

Other than that, nice work. I will probably look into the role in more detail at some point in time. For now, I only peeked at the diff.

Fixes: #1445

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -34,7 +34,7 @@ New DebOps roles
   entirely in the playbook.
 
 - The :ref:`debops.filebeat` role can be used to install and configure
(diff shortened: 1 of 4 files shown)


tasosalvas (approved):

drybjed (approved):
Agreed. I guess I did that out of habit, but Elastic stack seems better in this case.

PR #1453, opened 2020-08-04
[roundcube] Don't unnecessarily create LDAP object

imrejonk:
The role used to create a uid=roundcube object for the host DN, even
when roundcube__ldap_enabled was set to False. This change ensures that
the debops.ldap dependent task is ignored in that case.


drybjed:
Perhaps scoping the condition to `roundcube__ldap_password_enabled` variable would be better in this case. The roundcube account in LDAP is only required to lookup user DNs for password changes, and if this is disabled the account is not needed but LDAP still could be enabled for address book access.

imrejonk:
@drybjed That would indeed be a better solution, here you go :)

drybjed (approved):
Looks good, thanks! I'll backport this to stable releases as well.

PR #1455, opened 2020-08-05
New set of DebOps scripts

drybjed:


ypid (dismissed):
I went ahead and pushed small fixes to your branch in order to save time. Let me know if that works for you.

This looks good. Just some input about the interface to Ansible. I would try to keep the `run` and `check` subcommands a small wrapper around the Ansible commands, potentially passing all further arguments for the Ansible based subcommands directly without further parsing. That is just from my usage of DebOps. I just call ansible-playbook directly.

For me, the real benefit we can provide with the DebOps scripts is https://github.com/debops/debops-tools/issues/164.

drybjed:
> I went a head and pushed small fixes to your branch in order to save time. Let me know if that works for you.

Thanks. I'm not sure how you managed to do that, I guess write status on th `debops/debops` repository gives that access.

I actually wanted to siwtch the owner email to `debops-devel`, but decided against it. I guess that I should trust my gut feeling more. :)

> This looks good. Just some input about the interface to Ansible. I would try to keep the `run` and `check` subcommands a small wrapper around the Ansible commands, potentially passing all further arguments for the Ansible based subcommands directly without further parsing. That is just from my usage of DebOps. I just call ansible-playbook directly.

That's what I plan to do. The `debops run` would be an equivalent of the current `debops` command, and `debops check` would be the same but with `--diff --check` added automatically for convenience.

> For me, the real benefit we can provide with the DebOps scripts is [debops/debops-tools#164](https://github.com/debops/debops-tools/issues/164).

We're starting from scratch so we can make things better - for example I'd like to `git init` the project directory automatically, etc. For now I'll try to fix the Travis-CI issues before merging the branch to `master`.

tasosalvas (dismissed):
I'm making a note to update [CLI Doc](https://docs.debops.org/en/master/user-guide/scripts.html) with the new usage. I might be a while, though... still crunching like a madman on my deployment.

PR #1458, opened 2020-08-10
[nslcd] Configure idle_timelimit

imrejonk:
idle_timelimit used to be disabled by default, which meant that nslcd
would never close a connection. The LDAP server however would (after 900
seconds as configured by olcIdleTimeout). This caused nslcd to log lots
of error messages like "ldap_result() failed: Can't contact LDAP server"

This change sets the default idle_timelimit to 600 seconds.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -201,6 +201,14 @@ LDAP
 - Fixed multiple issues with adding and updating hosts to the LDAP directory
   when these hosts were configured for network bonding.
 
+:ref:`debops.nslcd` role
+''''''''''''''''''''''''
+
+- Enabled idle_timelimit to make sure that connections to the LDAP server are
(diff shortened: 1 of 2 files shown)


imrejonk:
Fun fact: setting idle_timelimit dropped the average nslcd error rate in our infrastructure from ~100 to 0 per day :)
Screenshot from 2020-08-10 13-43-07
(screenshot is of a Graylog graph)

drybjed (approved):
Excellent, thanks for fixing this! :+1: I'll backport the fix to the stable releases as well.

I would have thought that `nslcd` would have a sane default instead of no limit for connection timeout... I wonder if some other services managed by DebOps might have the same unnoticed problem.

imrejonk:
@drybjed I've been looking for similar error messages from other services (filtering on syslog severity 3), so far nothing. The nslcd error is an interesting one as the service itself continues to function properly without any other symptoms.

PR #1459, opened 2020-08-11
Nginx and docs fixes and improvements

ypid:
Fixes: #1439
cc: @apapsch


apapsch:
Thanks for catching the corner case with ssl.conf! This and the other changes look all good to me. Re "TLS curve": I picked the term up from Mozilla wiki, but I agree being precise by mentioning ECC is better, lest we contribute to another great confusion as happens with "SSL" mentioned everywhere when it's really TLS.

drybjed (approved):
Good changes, thanks. I'll backport these as well.

ypid:
You are welcome, thanks.

> but I agree being precise by mentioning ECC is better, lest we contribute to another great confusion as happens with "SSL" mentioned everywhere when it's really TLS.

@apapsch Than we are on the same page. Lets push into the right direction :)

PR #1463, opened 2020-08-12
[dhcrelay] Add debops.dhcrelay role

imrejonk:
`debops.dhcrelay` manages the ISC DHCP Relay Agent on Debian Stable for
forwarding of DHCPv4 messages between networks. The init script for
the isc-dhcp-relay package in Buster (and Bullseye, for that matter)
currently does not support DHCPv6, so this role currently only supports
managing dhcrelay for DHCPv4 relaying. A workaround for this can be
implemented in DebOps if there is demand for it.

This role replaces the dhcrelay functionality in `debops.dhcpd`. The
dhcrelay functionality in `debops.dhcpd` will be removed in the next
role update.

### DIFF
--- a/ansible/roles/dhcrelay/tasks/main.yml
+++ b/ansible/roles/dhcrelay/tasks/main.yml
@@ -1,0 +1,40 @@
+---
+# Copyright (C) 2020 CipherMail B.V. <https://www.ciphermail.com/>
+# Copyright (C) 2020 DebOps <https://debops.org/>
+# SPDX-License-Identifier: GPL-3.0-or-later
+
+- import_role:
+    name: 'global_handlers'
+
+- name: Install ISC DHCP relay packages
+  package:
+    name: '{{ (dhcrelay__base_packages + dhcrelay__packages)|flatten }}'
+    state: 'present'
+  register: dhcrelay__register_packages
+  until: dhcrelay__register_packages is succeeded
+
+- name: Divert ISC DHCP relay defaults
+  dpkg_divert:
+    path: '/etc/default/isc-dhcp-relay'
+
+- name: Configure ISC DHCP relay defaults
+  template:
+    src: 'etc/default/isc-dhcp-relay.j2'
+    dest: '/etc/default/isc-dhcp-relay'
+  notify: [ 'Restart isc-dhcp-relay' ]
+
+- name: Make sure that Ansible local facts directory exists
+  file:
+    path: '/etc/ansible/facts.d'
+    state: 'directory'
+
+- name: Save local facts
+  template:
+    src: 'etc/ansible/facts.d/dhcrelay.fact.j2'
+    dest: '/etc/ansible/facts.d/dhcrelay.fact'
+    mode: '0755'
+  register: dhcrelay__register_facts
+
+- name: Update Ansible facts if they were modified
+  action: setup
+  when: dhcrelay__register_facts is changed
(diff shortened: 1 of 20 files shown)


tasosalvas (changes requested):
Missing [upgrade](https://github.com/debops/debops/blob/master/docs/news/upgrades.rst) instructions for people using the feature via `debops.dhcpd`.

imrejonk:
@tasosalvas Good point, I added the upgrade notes.

tasosalvas (dismissed):
Looks great, thanks!

imrejonk:
And I also added the JANE test for `debops.dhcrelay` to .gitlab-ci.yml. Already tested successfully against our GitLab runners.

ansible/roles/dhcrelay/handlers/main.yml
```diff
@@ -0,0 +1,9 @@
+---
+# Copyright (C) 2020 CipherMail B.V. <https://www.ciphermail.com/>
+# Copyright (C) 2020 DebOps <https://debops.org/>
+# SPDX-License-Identifier: GPL-3.0-or-later
+
+- name: Restart isc-dhcp-relay
+  service:
+    name: 'isc-dhcp-relay'
+    state: 'restarted'
```
drybjed: The handlers should now be added in the `global_handlers` role, in a separate file. Don't forget to import that role in the task list.

ansible/roles/dhcrelay/tasks/main.yml
```diff
@@ -0,0 +1,40 @@
...
+    state: 'present'
+  register: dhcrelay__register_packages
+  until: dhcrelay__register_packages is succeeded
+
+- name: Divert ISC DHCP relay defaults
+  command: dpkg-divert --quiet --local
+           --divert /etc/default/isc-dhcp-relay.dpkg-divert
+           --rename /etc/default/isc-dhcp-relay
+  args:
+    creates: '/etc/default/isc-dhcp-relay.dpkg-divert'
```
drybjed: Perhaps you could convert this task to use the `dpkg_divert` Ansible module?

ansible/roles/dhcrelay/templates/etc/ansible/facts.d/dhcrelay.fact.j2
```diff
@@ -0,0 +1,28 @@
...
+        os.access(os.path.join(path, cmd), os.X_OK)
+        for path in os.environ['PATH'].split(os.pathsep)
+    )
+
+
+output = {'installed': cmd_exists('dhcrelay')}
+
+if output['installed']:
+    output['version'] = subprocess.check_output(
+        ['dhcrelay', '--version'], stderr=subprocess.STDOUT
```
drybjed, line 25: Ah, so that's the source of the role name... Sounds good then. I would go with `dhcp_relay`, but if the command line name exists, that's fine with me. :+1:

drybjed (changes requested):
Nice work, especially the documentation. :-) I know that you lifted the tasks and handlers from the `dhcpd` role, but perhaps you could use the occasion to modernize the code a bit. :-) Check the comments for suggestions.

ansible/roles/dhcrelay/templates/etc/ansible/facts.d/dhcrelay.fact.j2
imrejonk: Yeah, naming things is one of the two hard things in IT, next to cache invalidation and off-by-one errors :)
I did indeed name this role dhcrelay after the command line name, just as with dhcpd. It's also a bit shorter than dhcp_relay and maybe less ambiguous (there could be more software out there that does DHCP relaying).

imrejonk:
@drybjed Thanks for the feedback! All your comments should be addressed now. I also added Galaxy meta and Ansible collection information and updated the JANE_LOG_PATTERN.

imrejonk:
Whoops, forgot to add `- import_tasks: 'dhcrelay.yml'` to `global_handlers/handlers/main.yml`...
Fixed now.

drybjed (approved):
Nicely done, thanks!

PR #1467, opened 2020-08-14
[icinga] add variables for defining config files on the master node

lewer:
This is a (simpler) try for achieving this: https://github.com/debops/debops/pull/1418

### DIFF
--- a/ansible/roles/icinga/defaults/main.yml
+++ b/ansible/roles/icinga/defaults/main.yml
@@ -132,6 +132,15 @@ icinga__master_nodes: '{{ lookup("dig", "_icinga-master._tcp." + icinga__domain 
                                  "flat=0", wantlist=True) }}'
 
                                                                    # ]]]
+
+# .. envvar:: icinga__master_node_inventory_hostname [[[
+#
+# The host where tasks that must execute on the master node will be delegated.
+# It's used when ``icinga__master_*_configuration`` is set and configuration
+# files are written on the master node.
+icinga__master_delegate_to: '{{ icinga__master_nodes.0.target }}'
+
+                                                                   # ]]]
 # .. envvar:: icinga__director_nodes [[[
 #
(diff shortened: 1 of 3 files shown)


drybjed (changes requested):
Nice work. Does that solve your use case, @lewer?

You could also add a task to allow removal of remote configuration files. There's probably nothing in the `zones.d/` by default that would warrant support for configuration diverts/reverts, I imagine.

Is the configuration syntax the same as the `icinga__configuration` variables? You can add a note on that in the variable description.

lewer:
Thanks @drybjed for your advices, I've modified the code.

Yes, this solves my use case.

One thing I'm not sure: `icinga__master_nodes.0.target` is the DNS record of the master node, and it might be resolvable from the host but not from the controller, wich would make the delegation fail. To address this, an other variable `icinga__master_node_inventory_hostname` would probably be required... I'm not concerned by this, but do you think someone could?

drybjed:
@lewer Good point with the separate variable for the master node, you should add it and set the first node as the default. Ansible usually handles the delegation via DNS records well enough, but in rare cases that it's incorrect having a variable to define will be useful.

ansible/roles/icinga/defaults/main.yml
```diff
@@ -758,6 +758,49 @@ icinga__combined_configuration: '{{ icinga__default_configuration
...
+# .. envvar:: icinga__master_combined_configuration [[[
+#
+# Variable which combines Icinga master node configuration from previous
+# variables.
+icinga__master_combined_configuration: '{{ icinga__master_configuration
+                                           + icinga__master_group_configuration
+                                           + icinga__master_host_configuration }}'
+                                                                   # ]]]
+                                                                   # ]]]
+
```
drybjed: Remove this empty line, otherwise it will show up in the folded view in `vim`.

drybjed (approved):
Nice work, thanks. :-)

PR #1468, opened 2020-08-15
Switch for omitting installation of nginx packages.

ser:
- As discussed on IRC
- This PR should not affect any current installations

### DIFF
--- a/ansible/roles/nginx/defaults/main.yml
+++ b/ansible/roles/nginx/defaults/main.yml
@@ -54,6 +54,31 @@ nginx_host_allow: []
 nginx__deploy_state: 'present'
 
                                                                    # ]]]
+# .. envvar:: nginx__packages_installation [[[
+#
+# There is a deployment scenario possible where you install nginx packages from
+# sources or manually. This option allows you to omit nginx packages
+# installation by the role. Use with caution, as it can have unexpected
+# consequences. Notice: you must define ``nginx_default_tls_protocols``
+# manually, for example:
+#
+# ``nginx_default_tls_protocols: [ "TLSv1.2", "TLSv1.3" ]``
+#
+# Possible options:
+#
+# ``True``
+#   Default. Required packages will be installed.
+#
+# ``False``
+#   Installation of packages will be omitted.
+#
+#   .. warning:: Designed for expierenced devops only! Example use case includes
+#      a scenario where nginx binaries are prepared by
+#      https://virtubox.github.io/nginx-ee/ automation.
+#
+nginx__install_packages: True
+
+                                                                   # ]]]
 # .. envvar:: nginx_base_packages [[[
 #
 # List of Debian packages installed by this role
--- a/ansible/roles/nginx/tasks/main.yml
+++ b/ansible/roles/nginx/tasks/main.yml
@@ -37,13 +37,14 @@
 - name: Check if nginx is installed
   stat:
     path: '/usr/sbin/nginx'
+  when: nginx__install_packages
   register: nginx_register_installed
 
 - name: Ensure base packages are installed
   package:
     name: '{{ q("flattened", nginx_base_packages) }}'
     state: 'present'
-  when: (nginx__deploy_state in [ 'present' ])
+  when: (nginx__deploy_state in [ 'present' ]) and (nginx__install_packages)
   register: nginx__register_packages_present
   until: nginx__register_packages_present is succeeded
 
@@ -52,6 +53,7 @@
     name: '{{ q("flattened", nginx__flavor_packages) }}'
     state: '{{ "present" if (nginx__deploy_state == "present") else "absent" }}'
   register: nginx__register_packages_flavor
+  when: nginx__install_packages
   until: nginx__register_packages_flavor is succeeded
 
 - name: Make sure that Ansible local facts directory is present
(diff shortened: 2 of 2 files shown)


drybjed:
Sounds good, you just need to fix all the errors reported by Travis-CI first... :-) You don't need to check for `True` in that comparsion, `var|bool` and `not var|bool` work just fine.

drybjed:
@ser Nicely done! Do you want to squish all the fixes into one commit via `git rebase -i`, or should I do it?

ser:
I will do it, thanks for patience :)

ser:
@drybjed do i really have to fix all of debops with my PR? :1234:

ansible/roles/nginx/defaults/main.yml
```diff
@@ -53,6 +53,31 @@ nginx_host_allow: []
...
+# sources or manually. This option allows you to omit nginx packages
+# installation by the role. Use with caution, as it can have unexpected
+# consequences. Notice: you must define ``nginx_default_tls_protocols``
+# manually, for example:
+#
+# ``nginx_default_tls_protocols: [ "TLSv1.2", "TLSv1.3" ]``
+#
+# Possible options:
+#
+# ``True``
```
tasosalvas, line 69: Hi, I'm the doc guy who spoils everyone's fun, nothing personal :)

* "Unexpected consequences" is the opposite of documentation. Please make it clearer until nothing is unexpected.
* You don't need a whole boilerplate to document a boolean. I know `nginx__deploy_state` takes this approach, but check out `nginx_pki` for a shorter one.
* `warning` is used on `__deploy_state` because it **does not do the job it's expected to**. If your option does, explain its usage in regular paragraphs.
* Not a telegram. Full sentences. :P
* Oh is this the option for "expierenced" devops? lol when your doc is ready it will be for everyone.

> Boolean. Defauts to `True`. If set to `False`, the role will omit installation ...
   This is useful in cases where manual installation is desired, such as ...
   Note that `nginx_default_tls_protocols` will not be populated and...

tasosalvas (changes requested):

ansible/roles/nginx/defaults/main.yml
ser: I can't forecast everything what can happen, as options are unlimited, depending on the way nginx binary is installed. I used that phrase to discourage people to use that switch, I really hope I achieved that reading your comments :+1:

tasosalvas: Wait, do you mean to **approve merging** that switch, or to just **use** it?

ser: I wrote ```use that switch``` above, I have not written anything about ```approval```. I have no idea how did you think that out, to be honest :)

drybjed:
@ser No, I'm currently fixing the `ansible-lint` issues. Give me a few hours.

ansible/roles/nginx/defaults/main.yml
```diff
@@ -53,6 +53,31 @@ nginx_host_allow: []
...
+                                                                   # ]]]
+# .. envvar:: nginx__packages_installation [[[
+#
+# There is a deployment scenario possible where you install nginx packages from
+# sources or manually. This option allows you to omit nginx packages
+# installation by the role. Use with caution, as it can have unexpected
+# consequences. Notice: you must define ``nginx_default_tls_protocols``
+# manually, for example:
+#
+# ``nginx_default_tls_protocols: [ "TLSv1.2", "TLSv1.3" ]``
```
drybjed, line 65: Is that step actually needed? The fact script installed by the role should be able to get `nginx` version from `/usr/sbin/nginx`, is the binary from `nginx-ee` installed elsewhere? In that case perhaps we can rely on `$PATH` to find the correct `nginx` binary instead.

ser: The role gave errors when I tested it out when no nginx was installed at all. So maybe instead of **must** I could use wording **should**, indeed and add a phrase ```when the path is different than /usr/sbin/nginx```.

ansible/roles/nginx/defaults/main.yml
tasosalvas: Okay, I'll try to be clear and very verbose. I might have confused you because I thought we were having a snarky, fun back-and-forth.

That red icon in the beginning of my comments is me requesting changes, before your PR can be merged into `debops`.

@drybjed helped you meet the code quality standards. Right now I'm volunteering my time to help you bring it up to the doc quality standards, so we can include your feature.

To address your design goal of "discouraging people to use the switch", no, it's not the way we do things. There are features much more complex than toggling a package installation, and we don't discourage people from using them. We **enable** them to.

Options are not unlimited. When the switch is `False` but the role still runs, it means `nginx` has somehow ended up on the machine, possibly after the `bootstrap` playbook has run but before the `site` playbook runs and errors out. This fact is not even mentioned in the documentation you provided.

As I mentioned in my original request for changes, the stuff for "experienced devops" users happens when developing the feature. When we're done, it will be a toggle like any other one, with clear instructions on how to implement it.
If you can't provide those instructions, then we don't know *anybody* who knows how to use the feature. If you don't even know how to use it, there's no reason to implement it.

If you need help implementing any of my requested changes, I'm here. If you have thoughts on better ways to do stuff than what I'm proposing, we can discuss those too, as we just did the subject of deterring people from using the toggle.

I'm going to bed now. The blockquote in my original message has an example of what the doc should look like to be consistent with the style of the rest of the project. Have a go and if you get stuck somewhere post here and I'll try to address it tomorrow.

drybjed:
I just got an idea that instead of using a separate variable to decide of the role should install `nginx` packages or not, we can use `nginx__deploy_state: "config"` to control this behaviour. I'll look how many changes are required for this to work.

PR #1471, opened 2020-08-18
[dhcpd] Rewrite large parts of the role

### LINKED ISSUE #1018: [debops.dhcpd] Seemingly unrelated AnsibleUndefinedVariable errors
Tested with:
Python 3.7.3 and 2.7.16
Ansible v2.7.13
Debops v1.0.7

I'm getting errors like these when running `debops.dhcpd`:

```
TASK [debops.dhcpd : Configure DHCP server] *************************************************************
ok: [orbit.gwl.amsw.nl] => (item=etc/default/isc-dhcp-server)
failed: [orbit.gwl.amsw.nl] (item=etc/dhcp/dhcpd.conf) => changed=false 
  item: etc/dhcp/dhcpd.conf
  msg: 'AnsibleUndefinedVariable: ''postfix__fqdn'' is undefined'
```

It looks like [the dhcpd.conf template](https://github.com/debops/debops/blob/master/ansible/roles/debops.dhcpd/templates/etc/dhcp/dhcpd.conf.j2) somehow tries to include all host and group variables and then fails to template some variables that are set from other role defaults.

My `debops.dhcpd` configuration is split into two files:

**ansible/inventory/group_vars/gwl/dhcpd.yml**
```
---
dhcpd_authoritative: True
dhcpd_global_default_lease_time: '7200'
dhcpd_global_max_lease_time: '7200'
dhcpd_nameservers: [ '194.109.6.66', '194.109.9.99', '194.109.104.104' ]

dhcpd_keys:

  - key: 'gwl'
    algorithm: 'hmac-sha256'
    secret: '{{ lookup("password", secret + "/dhcpd/tsig-keys/gwl") }}'

dhcpd_zones:

  - zone: 'gwl.amsw.nl'
    primary: '94.142.244.11'
    key: 'gwl'

  - zone: '7.205.10.in-addr.arpa'
    primary: '94.142.244.11'
    key: 'gwl'

dhcpd_subnets:

  - subnet: '10.205.7.0'
    netmask: '255.255.255.128'
    routers: [ '10.205.7.1' ]

    pools:

      - range: '10.205.7.10 10.205.7.62'

dhcpd_options: |
  ddns-update-style standard;
  ddns-domainname "gwl.amsw.nl";
```

**ansible/inventory/host_vars/orbit.gwl.amsw.nl/dhcpd.yml**
```
---
dhcpd_interfaces: [ 'br0' ]
```

imrejonk:
The `debops.dhcpd` role has been largely rewritten in order to modernize
it and to support both DHCPv4 and DHCPv6 servers on the same host. Care
has been taken to improve the documentation and the readability of the
Jinja templates.

Furthermore, the DHCP Relay Agent functionality has been removed. That
functionality can now be found in the `debops.dhcrelay` role. The iPXE
server part is mostly unchanged.

Closes #1018


tasosalvas (dismissed):
<3 You had me at "readability", you kept me with `versions: ['buster']` :D

My approval, as usual, pertains to doc and style.

ansible/roles/dhcpd/defaults/main.yml
```diff
@@ -1,274 +1,314 @@
...
 #
-# List of network interfaces to listen on for DHCP requests
-# If this list is empty, Ansible will try to guess correct interfaces
-# automatically
-dhcpd_interfaces: []
+# The length in seconds that will be assigned to a lease if the client
+# requesting the lease does not ask for a specific expiration time. This is used
+# for both DHCPv4 and DHCPv6 leases (it is also known as the "valid lifetime" in
+# DHCPv6).
+dhcpd__default_lease_time: '43200'
```
drybjed: Is this 12 hours? Perhaps you could write it as:
```yaml
'{{ 60 * 60 * 12 }}'
```
to be easier to understand by a reviewer. The same with max lease time.

ansible/roles/dhcpd/tasks/main.yml
```diff
@@ -1,53 +1,71 @@
...
-- include: dist_vars.yml
-- include: install.yml
-
-- name: Get list of nameservers configured in /etc/resolv.conf
-  shell: set -o nounset -o pipefail -o errexit &&
-         grep -E '^nameserver\s' /etc/resolv.conf | awk '{print $2}' || true
+- name: Divert original configuration
+  command: dpkg-divert --quiet --local
+           --divert {{ item }}.dpkg-divert
+           --rename {{ item }}
```
drybjed: Perhaps you could use the `dpkg_divert` module here instead of the `command` module?

ansible/roles/dhcpd/tasks/main.yml
```diff
@@ -1,53 +1,71 @@
...
+      protocol: 'DHCPv4'
+    - filename: 'dhcpd6.conf'
+      protocol: 'DHCPv6'
+  notify: [ 'Restart isc-dhcp-server' ]
 
-- name: Make sure that included files exist
+- name: Configure ISC DHCP Server defaults
+  template:
+    src: 'etc/default/isc-dhcp-server.j2'
+    dest: '/etc/default/isc-dhcp-server'
```
drybjed, line 62: You need to specify the `mode` parameter (same goes for other `template` tasks).

ansible/roles/dhcpd/tasks/main.yml
```diff
@@ -1,53 +1,71 @@
 ---
-# Copyright (C) 2014-2018 Maciej Delmanowski <drybjed@gmail.com>
-# Copyright (C) 2014-2018 DebOps <https://debops.org/>
-# SPDX-License-Identifier: GPL-3.0-only
+# Copyright (C) 2020 CipherMail B.V. <https://www.ciphermail.com/>
+# Copyright (C) 2020 DebOps <https://debops.org/>
+# SPDX-License-Identifier: GPL-3.0-or-later
 
-- import_role:
-    name: 'secret'
```
drybjed, line 7: It might be useful to keep the `secret` role import so that DHCP keys could be stored there and shared between the `dhcpd` and `bind` roles, when the latter is introduced. Then the users could use lookups to retrieve the keys in both roles with the same contents.

ansible/roles/global_handlers/handlers/main.yml
```diff
@@ -13,6 +13,8 @@
 
 - import_tasks: 'dhcp_probe.yml'
 
+- import_tasks: 'dhcpd.yml'
+
```
drybjed, line 17: Don't forget to import the `global_handlers` role in the `dhcpd` role tasks.

drybjed (changes requested):
Nice work on the role rewrite. The issues on Travis-CI are due to new release of `ansible-lint` which brings additional tests. I'm currently fixing them, should be ready soon.

Interesting that the Jinja macro imports seems to work within a looped `template` task. I remember that at some point this functionality was removed from Jinja due to issues, I wonder if that was reintroduced or the problem was something else.

I'll need to try out the new role before merging, I'll let you know how that goes. :-)

ansible/roles/global_handlers/handlers/main.yml
imrejonk: Crap, our GitLab CI runner actually caught this. I made this PR at the end of the day, didn't want to wait for GitLab CI to report on its status :/

ansible/roles/dhcpd/tasks/main.yml
imrejonk: Agreed. Is it preferred to import the `secret` role in tasks/main.yml or in the playbook?

drybjed: In the `tasks/main.yml` file, please.

imrejonk:
@drybjed About the looped Jinja macros: if you're talking about #1018, I believe that I have solved that during the rewrite. The problem was that in DebOps before version 1.1.0, the macros were imported [`with context`](https://jinja.palletsprojects.com/en/2.11.x/templates/#import-visibility), like so:
`{% import 'macros.j2' as tpl with context %}`

Somehow this resulted in a situation where all inventory variables were passed to the Jinja template. This would break if your inventory held variables that were not resolvable by the `debops.dhcpd` role, like `postfix__mailname: '{{ postfix__fqdn }}'`.
During the rewrite I went with importing macros without context (the default). I'm only passing the required parameters to the print_* functions.

ansible/roles/dhcpd/tasks/main.yml
```diff
@@ -1,53 +1,80 @@
...
-
-- name: Make sure that IPv6 lease file exists
-  command: touch /var/lib/dhcp/dhcpd6.leases creates=/var/lib/dhcp/dhcpd6.leases
-  when: dhcpd_ipversion == '6'
+  notify: [ 'Restart isc-dhcp-server' ]
+
+- name: Make sure that Ansible local facts directory exists
+  file:
+    path: '/etc/ansible/facts.d'
+    state: 'directory'
```
drybjed, line 69: The `mode` parameter needs to be specified as well. `0755` will be fine here.

drybjed (approved):
I tested the role and it seems to work fine. Thanks, @imrejonk! :+1: I'll handle the Changelog conflict on my end.

Personally I would like to take a crack at the `dhcpd` later and convert it to DebOps custom filters, but that's for a later date, I'm currently working on other stuff like LDAP.

PR #1473, opened 2020-08-19
Fixes for tests introduced in 'ansible-lint' 4.3.0

drybjed:


ypid (commented):
Thank you! I should probably clean that up a bit at some point. I am also not sure if a role for Kodi (client application) is really in in scope of DebOps (server deployments). I will clean that up when I eventually start working on Jellyfin.

tasosalvas:
My hero!

Issue #1461, opened 2020-08-11
Specification for parse_kv_items and parse_kv_config


ypid:
`parse_kv_items` and `parse_kv_config` was introduced by @drybjed and is now a core part of DebOps. People working with DebOps need a clear understanding of it. Documentation what the filters do are contained in the actual role documentation but are specific to the role then. This will require people writing new roles to actually look into the code to figure out what it does and how to use it in their own roles. As already proposed by @tasosalvas, specification of the filters is needed defining exactly what their behavior is. I went for a hands-on approach when the filters where introduced and wrote the unit tests to reverse and "document" what they do. I would still be interested in a design document about our direction.

And maybe a "Not invented here" approach, but no better alternative comes to mind for me currently to solve all of the things that the filters do.

https://github.com/debops/debops/blob/master/ansible/roles/ansible_plugins/filter_plugins/debops_filter_plugins.py

I just want to track this here. Again, no pressure. I know maintaining a project is hard and this is a FOSS project. My idea is more that people interested in DebOps maybe can reverse the filters as well and not write unit tests this time but think about the problem and come up with a design document.

drybjed:
Ah, the "magic sauce" of DebOps. :-) I'll try to write a rough description of why the filters were created and their generalized functionality. This will probably end up in the "Developer guide" section of the documentation since the contents of the `ansible_plugins` role will be moved to `ansible/plugins/` directory at some point to be accessible via Ansible Collection mechanisms.

@tasosalvas Did you write anything about the plugins yet? We can then combine the documentation together. If not that's fine as well.

tasosalvas:
I have notes and a plan, I've been putting it off as work stuff had really piled up. I will try to get a first draft together. (just woke up)

drybjed:
Good morning. :) OK then, I'll wait for your draft then. Let's see what you got. And don't worry, you don't need to do it right away if something else needs your attention.

tasosalvas:
I fully agree with @ypid, digging into `parse_kv_config` really untied my hands as a debops user. The unit tests helped immensely.

I have a few paragraphs on:
* the input types the filter accepts
* the resulting mappings reserved attributes
* merging/weighing behavior
* probably some more stuff mixed with the `bind` role's notes

I haven't turned them into `.rst` yet. It's not too much work, I just have some trouble task switching when I get into something.
No need to duplicate this part, though, I got it.

Let's talk design, yeah.


## The goal
> How I learned to stop worrying and love `parse_kv_*`.

[...]

PR #1474, opened 2020-08-20
[doc] User guide: Universal Configuration

tasosalvas:
Main issue thread: #1461

This is the user tutorial, developer style guide coming soon.


tasosalvas:
Just pointing out that the user doc is meant to be a _funnel_ in terms of design: No outgoing links until the end of the document.

docs/user-guide/universal-configuration.rst
```diff
@@ -0,0 +1,410 @@
+.. Copyright (C) 2015-2020 Maciej Delmanowski <drybjed@gmail.com>
```
drybjed: You can remove my copyright from here. :-) This just for this particular file, not the concept of `parse_kv_*` filters themselves.

docs/user-guide/universal-configuration.rst
```diff
@@ -0,0 +1,410 @@
...
+There are two different recursion behaviors for item fields, differentiated by
+whether values of multiple fields will be merged or not.
+
+If an item's ``value`` field contains a list, the configuration filters will
+recurse into it, so any of the documented configuration syntax can be used in
+it as well.
+
+When merging items with the same ``name`` whose ``value`` fields contain lists,
+those lists **will not be merged**.
+Only the last appearing ``value`` will be used.
```
drybjed: That's true for the `parse_kv_items` filter, but not for the `parse_kv_config` filter. In the latter case, values of lists are additive so that multiple elements can be added afterwards. If you want to "reset" a value, you need to use a string instead. This behaviour is used for example in the [`postfix` role](https://docs.debops.org/en/master/ansible/roles/postfix/defaults-maincf.html#lists-are-merged-together) to manage `/etc/postfix/main.cf` lists.

drybjed (changes requested):
Awesome work @tasosalvas, love it! I like the reference to Universal Paperclips. :-) Looking forward for the rest.

docs/user-guide/universal-configuration.rst
tasosalvas: Uh this makes my brain hurt. The end user cannot differentiate between the two filters without peeking at the source, can they? I was hoping to reveal the fact that we're talking about two filters only in the role dev doc. :(

drybjed: I guess you can still do that later. I think that you can describe the behaviour of `parse_kv_config` in the user guide (list values are merged) because it will be commonly encountered in the correct places, for example inside configuration files. And you can leave the specifics of `parse_kv_items` o the dev doc.

tasosalvas: I'll note the difference between `parse_kv_items` and `_config` by mentioning both can happen and to check the role doc, for now, and I'll think about unifying behavior in the `config` rewrite.

Can I consider `_config` to be the default unless otherwise noted?

drybjed: Sure, `parse_kv_config` can be used as the default behaviour.

The `_config` is for single configuration files, `_items` is used outside in the task to handle multiple configuration files in `conf.d/` style setups where the contents are handled by `_config`. That's the common use case. Different behaviour of these filters allows for more complex setups, for example `/etc/postfix/master.cf` where `parse_kv_items` is used in a single configuration file which has muti-level configuration. But that's an exception.

docs/user-guide/universal-configuration.rst
```diff
@@ -0,0 +1,414 @@
...
+to create flowerpots instead.
+
+
+.. _universal_configuration_weight:
+
+Reordering items
+~~~~~~~~~~~~~~~~
+
+Our flowerpot maximizer is almost ready!
+However, due to role constraints I just made up, list item order is important.
```
ypid: Probably better to rephrase that to something like: "However, due to made up role constraints, the list item order is important."

docs/user-guide/universal-configuration.rst
```diff
@@ -0,0 +1,414 @@
...
+       state: 'absent'
+
+     - name: 'flowerpot'
+       material: 'clay'
+       value: 64
+       weight: -50
+
+
+And here we go. Now the role knows to prioritize ``flowerpot`` production,
+without ``production facilities`` hogging all the clay!
```
ypid: "production facilities" is not a keyword here. Quotes could be used but are not needed either for my taste.

docs/user-guide/universal-configuration.rst
```diff
@@ -0,0 +1,414 @@
...
+
+
+.. _universal_configuration_dependent:
+
+The *dependent* pattern
+~~~~~~~~~~~~~~~~~~~~~~~
+
+*Dependent* configuration lists are used when roles are loaded as dependencies
+of other roles. They look like ``role__dependent_list`` and are included
+towards the end of ``role__combined_list`` variables.
```
ypid, line 272: I don’t understand "are included towards the end of ``role__combined_list`` variables.". Do you mean "towards the end of the default variables"? "default variables" is a term that, I just checked, is commonly used in the docs.

docs/user-guide/universal-configuration.rst
```diff
@@ -0,0 +1,414 @@
...
+
+
+.. _universal_configuration_recursion:
+
+Recursion
+~~~~~~~~~
+
+If an item's ``value`` or the special ``options`` field contains a list,
+the configuration filters will recurse into it, so any of the documented
+configuration syntax can be used in it as well.
```
ypid, line 384: To my knowledge, the filters don’t yet recurse into it. Only a second level is currently supported. But I am very in favor of actual recursion (ref: #835). Maybe include this in a note.

ypid (changes requested):
Very well done @tasosalvas! And nice "story line" behind the actual documentation :+1: It resonates with me :)

docs/user-guide/universal-configuration.rst
tasosalvas: It is the value of a `name` field in the initial definition of the example. I give the same treatment to `flowerpot`.

I should unquote neither or both. They do need some kind of wrapping, imo, they are the unique key names.
Maybe just **strong**?

docs/user-guide/universal-configuration.rst
tasosalvas: I mean the entry point of the filter, the list that adds all the other lists together.

I define them in the first example as `paperclip_maximizer__combined_products` and then move to `role__combined_list` without explicitly stating so when things get a bit more advanced.

I would call the `defaults/main.yml` document as a whole "default variables", in this scope I'd call them `role__default_list`.

If you think it needs more clarity, I probably should add a code block showing default, user and dependent get added into combined. I skipped that as I thought it might be too verbose, but it _will_ be clearer visually that way.

docs/user-guide/universal-configuration.rst
tasosalvas: Oops, good catch.

Yes, I will be specific about our current implementation for now, so the doc can be added before the rewrite, and things will get more elegant afterwards.

docs/user-guide/universal-configuration.rst
tasosalvas: Ok, I was breaking up the perspective for comedic effect but I will make it consistent.

tasosalvas:
Monday, though, I'm writing my first CV in years atm :)

docs/user-guide/universal-configuration.rst
ypid: >  is the value of a name field in the initial definition of the example.

Not in this exact form. I only see `production capacity`.

tasosalvas: Ah so. Yes that's what I need to fix :)

tasosalvas:
So, I addressed all requests for changes and added all the other notes I had to the `ansible_plugins` documentation.

My rationale was that my current notes were about __what__ the plugins were doing, and that was the proper place for them to go. The __Writing Roles__ doc I mentioned will be more about __why__ and __how__. 

It felt to me a bit sloppy to just add a whole Writing Roles doc containing doc for only one tool. Let's make it another issue, where I'll brief you gentlemen, then we can brainstorm on what goes in there and do it right without delaying this PR. I think it will be useful as it is.

Suggestions, corrections, opinions are very welcome, of course, but I un-drafted it and when you think it's good enough it can be merged.

I'll start the Writing Roles issue in the next couple of days.

drybjed (dismissed):
Looks good to me. Nice work, @tasosalvas! When @ypid accepts the changes as well, I'll merge it.

tasosalvas:
Fixed some formatting on `ansible_plugins`

docs/ansible/roles/ansible_plugins/getting-started.rst
```diff
@@ -39,21 +39,98 @@ Jinja templates:
...
+    ``options`` or other defined ``merge_keys``.
+
+    Examples:
+
+    .. code-block:: python
+
+      empty={
+        'some_param':  'other_param',
+        'empty_param': ['param2', 'param1']
+      }
```
ypid: I would stick to the YAML syntax here as this is how users will interact with this Python filter.

ypid (changes requested):

docs/ansible/roles/ansible_plugins/getting-started.rst
tasosalvas: Yes, you're absolutely right, I was on autopilot :)

drybjed: The dictionary needs to be specified like this, but it can be shown better. This should probably be something like...
```yaml
{{ variable | parse_kv_items(empty={
                             'some_param': 'other_param',
                             'empty_param': [ 'param2', 'param1']}) }}
```

tasosalvas: I'm expanding the example with results as well, I had trouble understanding it now even though I did when I wrote it!  :D

tasosalvas: Um, actually... the usage I see in the codebase are inline python supplied in `parse_kv_items(args)`... I guess the items could be built as yaml but I think python is the most direct usage. What do you think?

tasosalvas: I think something like:
```python
parse_kv_items(
  empty={
    'some_param':  'other_param',
    'empty_param': ['param2', 'param1']
  }
)
```

```yaml
- name: foo
  other_param: bar

# Will result in

- name: foo
  some_param: bar
  other_param: bar
```

tasosalvas:
@ypid I __reject__ your stale review and double down on the python by renaming "mappings" to "dicts" in those sections :P

tasosalvas:
Oops lol instead of the yaml spec link I had pasted a completely unrelated postgresql link

docs/ansible/roles/ansible_plugins/getting-started.rst
tasosalvas: @drybjed do you want me to wrap the ``parse_kv_items`` example in jinja? I only now saw your comment.

drybjed: @tasosalvas This is how the example will most likely be used within Ansible roles, so I would say wrapping it in Jinja tags should help explain the intended usage.

tasosalvas:
Ok, next time I'll go easy on the amends and learn to squash in the end :)

docs/ansible/roles/ansible_plugins/getting-started.rst
tasosalvas: Colorization sure looks prettier this way

tasosalvas:
@ypid are we good?

docs/ansible/roles/ansible_plugins/getting-started.rst
ypid: I was wrong. I attribute that to being tried when I reviewed this. Thanks for taking the opportunity to improve the docs even further!

ypid (approved):
All good. Thanks. And sorry for the delay.

ypid:
> Oops lol instead of the yaml spec link I had pasted a completely unrelated postgresql link

You are not running Qubes OS, aren’t you? Because this is kinda a Qubes OS joke. Background is that for inter VM* clipboard copies, you have to press 4 shortcuts instead of 2 and people occasionally forget one step and then paste their "old" clipboard :)

*Which you would also do in this use case in a properly isolated Qubes OS setup.

tasosalvas:
> You are not running Qubes OS

A friend of mine is preparing to move to it after I told them about it, having heard it from you :) I'm considering it too but I don't have the metal to try it out right now.

No, I blame `kitty` which also does some clipboard-safety magic

drybjed:
And merged. @tasosalvas, thanks for writing the docs! I wonder if I should backport the PR to the stable releases so that this documentation shows on the stable branches as well... Thoughts?

tasosalvas:
No functionality was changed, I think it can safely be backported. How far back, only you know :)

drybjed:
I don't usually backport the documentation changes to make the stable branches easier to parse, but ok, I can backport this one. At least the documentation will be consistent across the releases that way.

imrejonk:
Holy cow @tasosalvas, what an absolute gem of documentation you wrote here. I had a good laugh at the paperclip and flowerpot example, while learning a lot about Universal Configuration and filter plugins in DebOps. Can't start my day much better than this, thanks :smile:

PR #1475, opened 2020-08-20
playbooks: Add tool for safe upgrade and reboot.

apapsch:
The title says it all :-)


tasosalvas:
Huh, we didn't have this? I like.

You'll need the copyright statement in the SPDX format to pass `make test`, check the other playbooks.

nqb:
I suggest
* Using Ansible `reboot` module to reboot in place of the command used, try something like that:

```yaml
- name: Reboot a machine with shutdown command in unusual place
  reboot:
    search_paths:
     - '/lib/molly-guard'
```
* Rename this playbook with a more explicit name to specify that reboot could occur.
* Adding space between each task for better readability

apapsch:
> Using Ansible reboot module to reboot in place of the command used, try something like that

I currently have no machine which requires reboot and was only able to confirm manually that `/lib/molly-guard/reboot` does its thing without confirmation.

apapsch:
I cheated and touched `/var/run/reboot-required` manually. The playbook worked fine, rebooting the machine.

ansible/playbooks/tools/upgrade-reboot.yml
```diff
@@ -0,0 +1,31 @@
+---
+
+# upgrade - upgrade packages which are safe to be updated.
+# This means no packages will be removed.
+# The machine is rebooted if necessary. Skip the reboot
+# tag to avoid that.
+# Copyright (C) 2020 Ellen Papsch <ellenpapsch@gmail.com>
+# SPDX-License-Identifier: GPL-3.0-only
```
drybjed, line 3: Please move the SPDX identifier and Copyright string to the top of the playbook, under the `---` line.

ansible/playbooks/tools/upgrade-reboot.yml
```diff
@@ -0,0 +1,31 @@
...
+
+- hosts: debops_all_hosts
+  become: true
```
drybjed: You should use `True` / `False` as boolean identifiers for consistency with other roles and playbooks in DebOps.

ansible/playbooks/tools/upgrade-reboot.yml
```diff
@@ -0,0 +1,31 @@
+---
+
+# upgrade - upgrade packages which are safe to be updated.
+# This means no packages will be removed.
+# The machine is rebooted if necessary. Skip the reboot
+# tag to avoid that.
+# Copyright (C) 2020 Ellen Papsch <ellenpapsch@gmail.com>
+# SPDX-License-Identifier: GPL-3.0-only
+
+- hosts: debops_all_hosts
```
nqb: Some other notes:
- There is no name for your play
- Most of other playbooks use:
  * `hosts: [ 'debops_all_hosts' ]`
  * an explicit `gather_facts: False` instruction
  * an environment block like:
```yaml
  environment: '{{ inventory__environment | d({})
                   | combine(inventory__group_environment | d({}))
                   | combine(inventory__host_environment  | d({})) }}'
```

ansible/playbooks/tools/upgrade-reboot.yml
```diff
@@ -0,0 +1,31 @@
...
+  become: true
+  tasks:
+
+    - name: Upgrade safe packages with refreshed cache
+      apt:
+        update_cache: yes
+        upgrade: safe
```
drybjed: You can add a `when: ansible_pkg_mgr == 'apt'` condition here, then we could expand the playbook for more OSes if somebody is interested. However that won't work without fact gathering. Your choice.

The `safe` string should be quoted.

nqb (changes requested):

nqb:
I wonder if we should not have a dedicated playbook to reboot nodes, imported in your play. Because you may want to reboot your nodes in other situations than upgrade. And right now, you have to run a ad-hoc commands.

I used this playbook my side with the `--limit` option of Ansible:

```yaml
---

- name: Reboot DebOps hosts secured by molly-guard
  hosts: [ 'debops_all_hosts' ]
  become: True

  gather_facts: False

  environment: '{{ inventory__environment | d({})
                   | combine(inventory__group_environment | d({}))
                   | combine(inventory__host_environment  | d({})) }}'
  
  vars_prompt:
    
    - name: 'reboot_confirmation'
      prompt: 'Are you sure you want to reboot all these nodes ?'
      default: False
      private: no

  tasks:
    
    - name: Reboot using custom location for shutdown command
      reboot:
        search_paths:
          - '/lib/molly-guard'
      when: reboot_confirmation|bool
```
I didn't add it upstream because I'm not sure if it is safe to add such a playbook in case someone run it on its all infrastructure by error.

ansible/playbooks/tools/upgrade-reboot.yml
```diff
@@ -0,0 +1,31 @@
...
+    - name: Check if reboot is required
+      stat:
+        path: '/var/run/reboot-required'
+        get_checksum: no
+      register: reboot_required
+
+    - name: Reboot a machine with shutdown command in unusual place
+      reboot:
+        search_paths:
+          - '/lib/molly-guard'
```
drybjed, line 34: You should specify the other paths used by default, `['/sbin', '/usr/sbin', '/usr/local/sbin']`. This will alow the playbook to be used on hosts without `molly-guard` installed.

ansible/playbooks/tools/upgrade-reboot.yml
```diff
@@ -0,0 +1,31 @@
...
+        update_cache: yes
+        upgrade: safe
+
+    - name: Check if reboot is required
+      stat:
+        path: '/var/run/reboot-required'
+        get_checksum: no
+      register: reboot_required
+
+    - name: Reboot a machine with shutdown command in unusual place
```
drybjed: I would rename the task to "Reboot a machine when required". In the Ansible output the "unusual place" bit won't be apparent.

drybjed (changes requested):
Nice work. There are a few details I'd like to see fixed before merging this. You can also squash the commits when all issues are resolved, if you like.

drybjed:
@nqb, playbooks in the `tools/` subdirectory are not included in the `site.yml` playbook, so there's no worry that they will be accidentally executed.

I think that your playbook could also be added as `tools/reboot.yml`, the prompt should make it safe enough.

nqb:
> I think that your playbook could also be added as `tools/reboot.yml`, the prompt should make it safe enough.

Ok !

@apapsch, do you want to add my playbook with yours and import it into your `upgrade-reboot` playbook ?

drybjed:
No, no importing. Both playbooks have different use cases, @apapsch playbook runs and upgrade and reboots **if necessary**, and @nqb playbook reboots after user confirmation.

ansible/playbooks/tools/upgrade-reboot.yml
apapsch: I refrained from `when: ansible_pkg_mgr == 'apt'`. I guess if someones wants to use the playbook with Fedora (or whatever) they also want the rest of debops to work with it and that would be quite an undertaking.

drybjed (approved):
Nicely done, thanks.

I tested the playbook on an existing LXC container and it seems to work fine, however in the end the reboot task failed due to `systemd` being sluggish with some jobs. I'm not really sure how the Ansible `reboot` module behaves and if it tries to reconnect after a few failed attemts. This probably can be tweaked later if needed.

PR #1484, opened 2020-08-26
[owncloud] Make Nextcloud 18.0 the default release

lewer:
Nextcloud 17.0 will become end of life in 2020-09, this PR sets 18.0 as the default release. There's no change as 18.0 was already supported.


drybjed:
@lewer You need to update the Nextcloud version [here](https://github.com/debops/debops/blob/master/ansible/roles/owncloud/defaults/main.yml#L426-L428) so that the new installs use 18.0 instead of 17.0.

lewer:
> @lewer You need to update the Nextcloud version [here](https://github.com/debops/debops/blob/master/ansible/roles/owncloud/defaults/main.yml#L426-L428) so that the new installs use 18.0 instead of 17.0.

oops... I forgot to add the file in the commit... :sleeping:

drybjed (approved):
Nicely done, thanks. :-)

PR #1490, opened 2020-08-28
Updated 'bootstrap-ansible' path in CI and switch to newer pip package

drybjed:


imrejonk (approved):

drybjed:
OK, so `python-pip-whl` package does not replace the `python-pip` package, it's just that the latter was removed from newer OS releases. So we need to deal with this differently and just not install `python-pip` directly in roles - let the `python` role handle things by itself.

PR #1493, opened 2020-08-29
Docs fixes and [homeassistant] Serve static files directly by the webserver

ypid:


tasosalvas (dismissed):

ansible/roles/homeassistant/defaults/main.yml
```diff
@@ -162,6 +162,13 @@ homeassistant__basic_auth_name: 'homeassistant'
 #
 # List of user accounts which will be allowed access to Home Assistant.
 homeassistant__basic_auth_users: '{{ ansible_local.core.admin_users|d([]) }}'
+
+                                                                   # ]]]
+# .. envvar:: homeassistant__webserver_user [[[
+#
+# Name of the webserver user account which will be granted read only access to
+# the www directory outside of the Home Assistant configuration directory.
+homeassistant__webserver_user: '{{ ansible_local.nginx.user|d("www-data") }}'
```
SuperSandro2000, line 171:
```suggestion
homeassistant__webserver_user: '{{ ansible_local.nginx.user | d("www-data") }}'
```

ypid: Thanks for reviewing. I would say the omitted spaces are backed by [DEP 2 - DebOps code standards](https://docs.debops.org/en/master/dep/dep-0002.html).

ansible/roles/dropbear_initramfs/templates/etc/initramfs-tools/scripts/local-bottom/debops_dropbear_initramfs.j2
```diff
@@ -1,9 +1,9 @@
+{# Copyright (C) 2015-2017 Robin Schneider <ypid@riseup.net>
+ # Copyright (C) 2017 DebOps <https://debops.org/>
+ # SPDX-License-Identifier: GPL-3.0-only
+ #}
 #!/bin/sh
```
drybjed: I think that in the case of shell scripts, the Copyright and SPDX license strings should be put under the shebang (`#!/bin/sh`). This should allow the `shellcheck` linter to properly detect such scripts. In that case you cannot put this in the Jinja comment tags because shell linting will break... So basically reverse this change. :-)

drybjed (changes requested):
Man, so many misspellings... I wonder if wrong detection of keypresses on my keyboard resulted in a shadow "Shift" pressed with `THe`. I need to switch it asap.

ansible/roles/dropbear_initramfs/templates/etc/initramfs-tools/scripts/local-bottom/debops_dropbear_initramfs.j2
ypid: I see what you did there with the other shell script templates. They are valid shell scripts even before being templated. This works in some cases. I looked at the loop in this file and I don’t see an easy way to make it valid before.

You already disabled `shellcheck` for this file in 8cb601b82c7395237ae00e58f74d5a628d3c2938. I would leave it like this.

ypid: But I agree with you that for consistency, this commit should be undone. I removed it.

drybjed (approved):
Looks good, thanks. The shell script in the `homeassistant` role will have to be rewritten before the role is integrated with the rest of DebOps, so that it can be properly vetted by `shellcheck`.

ypid:
Thanks for merging. For the protocol, the shell script in the `dropbear_initramfs` role would be affected. GitHub does not handle the case well where I removed this commit entirely by rebasing.

PR #1496, opened 2020-08-31
Fix file permissions for nullmailer configs

giannoug:
After running DebOps on a clean installation, nullmailer seems to be broken.

Sending a mail returns an error (for Google's sake):
```
echo "test" | sendmail test@example.com
nullmailer-inject: nullmailer-queue failed: 18
```

Looking at nullmailer's logs, it seems that some configuration files have wrong permissions and it can't start:
```
[...]
nullmailer-send: Error: Could not read config file "/etc/nullmailer/defaulthost": Permission denied
[...]
```

During DebOps execution, there are some warnings regarding those configs:
```
[WARNING]: File '/etc/nullmailer/idhost' created with default permissions '600'. The previous default was '666'.
Specify 'mode' to avoid this warning.
[WARNING]: File '/etc/nullmailer/helohost' created with default permissions '600'. The previous default was '666'.
Specify 'mode' to avoid this warning.
[WARNING]: File '/etc/nullmailer/defaulthost' created with default permissions '600'. The previous default was '666'.
Specify 'mode' to avoid this warning.
[WARNING]: File '/etc/nullmailer/maxpause' created with default permissions '600'. The previous default was '666'.
Specify 'mode' to avoid this warning.
[WARNING]: File '/etc/nullmailer/pausetime' created with default permissions '600'. The previous default was '666'.
Specify 'mode' to avoid this warning.
[WARNING]: File '/etc/nullmailer/sendtimeout' created with default permissions '600'. The previous default was '666'.
Specify 'mode' to avoid this warning.
```

Not sure if mode 666 is the appropriate permission here, but I have tested it and it works. The /etc/nullmailer/remotes file [seems to have mode 600](https://github.com/debops/debops/blob/master/ansible/roles/nullmailer/defaults/main.yml#L388). Any ideas regarding the permissions are welcome.

### DIFF
--- a/ansible/roles/nullmailer/defaults/main.yml
+++ b/ansible/roles/nullmailer/defaults/main.yml
@@ -346,34 +346,42 @@ nullmailer__configuration_files:
 
   - dest: '/etc/nullmailer/idhost'
     content: '{{ nullmailer__idhost }}'
+    mode: '0644'
     state: '{{ "present" if nullmailer__idhost else "absent" }}'
 
   - dest: '/etc/nullmailer/helohost'
     content: '{{ nullmailer__helohost }}'
+    mode: '0644'
     state: '{{ "present" if nullmailer__helohost else "absent" }}'
 
   - dest: '/etc/nullmailer/defaulthost'
     content: '{{ nullmailer__defaulthost }}'
+    mode: '0644'
     state: '{{ "present" if nullmailer__defaulthost else "absent" }}'
 
   - dest: '/etc/nullmailer/defaultdomain'
     content: '{{ nullmailer__defaultdomain }}'
+    mode: '0644'
     state: '{{ "present" if nullmailer__defaultdomain else "absent" }}'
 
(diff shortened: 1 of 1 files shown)


imrejonk (changes requested):
Please don't set permissions to 0666, doing so will allow all users of the system to change the Nullmailer configuration. The Ansible message is misleading as the default is actually 0644.

There is a bug in Ansible 2.9.12 that is causing the 0600 chmod. See [this debops-users post](https://lists.debops.org/hyperkitty/list/debops-users@lists.debops.org/thread/EVH3GOJFDWUZ62AUIK747PEAAF5WOJCL/).

giannoug:
I've updated the PR accordingly

drybjed (approved):
Good idea, thanks. I think that in the future the role tasks could be modified to not `omit` the mode and just use `0644` as the default unless otherwise specified. But let's go with this PR instead for consistency.

PR #1508, opened 2020-09-11
[fail2ban] fix role on ubuntu 20.04

Keridos:
This fixes fail2ban v0.11 on Ubuntu 20.04. I added maps in the defaults/main.yml for the distribution specific stuff and moved the iptables-xt_recents actions to templates, because they need distribution dependant changes, too.

This might need to be extended when Debian 11 releases and if it ships with fail2ban v0.11, too

/edit:
Fixed the travis errors.


drybjed (approved):
Thanks, looks good. I tested the changes on Debian Buster, shouldn't be any adverse effects. I'll backport the PR to the stable releases as well.

I feel that `fail2ban` role needs a general overhaul, but I don't want to get distracted right now. Perhaps when `firewalld` role is introduced in the future, `fail2ban` role will get a proper rewrite as well.

PR #1518, opened 2020-09-23
[lvm] Scan Software RAID devices by default

imrejonk:
This change ensures that LVM scans Linux Software RAID devices
(/dev/md*) for LVM volume groups.


drybjed (approved):
Sounds good, thanks. I'll backport this change to the stable releases as well.

PR #1522, opened 2020-09-27
Minor stuff and setup the NetBox Request Queue Worker

ypid:


ansible/roles/netbox/handlers/main.yml
```diff
@@ -17,3 +17,10 @@
     state: 'restarted'
     enabled: True
   when: netbox__app_internal_appserver|bool
+
+- name: Restart netbox Request Queue Worker
+  service:
+    name: 'netbox'
```
drybjed: It should probably be named `netbox-rq`.

ypid: Yes, that makes more sense :) Fixed.

drybjed (approved):
Nicely done, thanks.

PR #1524, opened 2020-10-03
[owncloud] Nextcloud 18 support

### LINKED ISSUE #1492: [debops.owncloud] error_page support for Nextcloud 18+
According to https://github.com/nextcloud/text/issues/820 the 

```
error_page 403 /core/templates/403.php;
error_page 404 /core/templates/404.php;
```

in nginx config is deprecated.

If the lines are present there is this annoying workspace circle moving forever on the top, as long as there is no README.md inside the directory.

The workspace error appeared with nextcloud >= 18.

ypid:
Closes: #1492

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -74,6 +74,13 @@ LDAP
 
 - Default LVM2 configuration for Debian Stretch and Buster has been added.
 
+:ref:`debops.owncloud` role
+'''''''''''''''''''''''''''
+
+- Drop Nextcloud 16 and 17 support because it is EOL. You need to upgrade Nextcloud
+  manually if you are running version 17 or below. The role now defaults to
+  Nextcloud 18 for new installations.
+
 :ref:`debops.postgresql` role
 '''''''''''''''''''''''''''''
 
--- a/ansible/roles/owncloud/defaults/main.yml
+++ b/ansible/roles/owncloud/defaults/main.yml
@@ -414,9 +414,8 @@ owncloud__variant_name_map:
 #
 # * ownCloud ``10.3``
 #
-# * Nextcloud ``16.0``
-# * Nextcloud ``17.0``
 # * Nextcloud ``18.0``
+# * Nextcloud ``19.0`` (untested)
 #
 # For Nextcloud refer to the `Nextcloud Maintenance and Release Schedule <https://github.com/nextcloud/server/wiki/Maintenance-and-Release-Schedule>`_.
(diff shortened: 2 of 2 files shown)


ypid:
Tested it myself for a few days, updated my other instances yesterday. Users are not complaining so I guess that is good feedback :) Ready for merging.

drybjed:
@ypid Thanks, I'll try and test the update process ASAP.

Since you mention multiple instances - do you have them somehow federated? I was wondering how Nextcloud could be scalled horizontally and federation seems to be a way to do it - set up multiple instances on separate subdomains and join them via federation mechanism. I'm not sure though if various functions like user searches, shared files, Talk, calendars, etc. are propagated automatically across the federation, or they stay in their respective "silos"?

ypid:
Thanks for that. FYI, the update process is not supported/done by the role. You have to do it manually, same as doing point release updates. After that manual step, I run the role to let it update the webserver config and small things like that. Not ideal, I know.

About federation, no, I don’t use it much. Nextcloud can be nicely scaled. I heart that there is an instance with multiple million users. But I don’t have details.

tasosalvas (approved):

imrejonk (approved):
Thanks @ypid! I upgraded my Nextcloud instance from 17.0.10 to 18.0.10 today. Everything looks good so far, haven't tested ownCloud. I'm using MariaDB as a database backend.

I did encounter a few issues during the upgrade that are unrelated to this role. The database upgrade failed because of some faulty migration (see issue https://github.com/nextcloud/server/issues/23174), but I was able to fix this by adding the missing table column. Another thing was that the Nextcloud updater didn't warn me about an incompatibility between Nextcloud 18 and a specific version of the user_saml app I'm using. I had to fix that afterwards by updating the app to another specific version that supports Nextcloud 18, which turned out to be a bit more work than anticipated because apps.nextcloud.com is currently offline (and has been for some time it seems).

ypid:
You are welcome :)

> I did encounter a few issues during the upgrade that are unrelated to this role.

I encountered only one small issue, but that could be easily solved by some `occ` command.

drybjed:
> apps.nextcloud.com is currently offline (and has been for some time it seems)

Ah, so that's why there were no applications listed. Let's hope they bring it back then.

drybjed (approved):

imrejonk:
> > apps.nextcloud.com is currently offline (and has been for some time it seems)
> 
> Ah, so that's why there were no applications listed. Let's hope they bring it back then.

It's back!

PR #1526, opened 2020-10-05
Ensure to add _netdev to choosen options

gardouille:
Otherwise, with a config like this : 
``` yml
nfs__group_shares:
  - path: "/media/nfs/shared"
    src: 'nas.example.org:/shared'
    default_options: False
    options: [ 'defaults', '…', 'noatime', 'noauto' ]
```

I only got this entry in `/etc/fstab` : 
```
nas.example.org:/shared /media/nfs/shared nfs4 _netdev 0 0
```

With the modification, i now have : 
```
nas.example.org:/shared /media/nfs/shared nfs4 defaults,…,noatime,noauto,_netdev 0 0
```


drybjed (approved):
Good idea, thanks. I'll backport the changes as well.

PR #1534, opened 2020-10-16
[gitlab_runner] add new options for docker executor

lewer:
This PR allows to configure more options for gitlab runner's docker executor, as described here: https://docs.gitlab.com/runner/configuration/advanced-configuration.html

It does not try to be smart (no discovery of good values for cpu, memory,...), it only lets the user provide values, if he wants, for these options.

### DIFF
--- a/ansible/roles/gitlab_runner/templates/etc/gitlab-runner/config.toml.j2
+++ b/ansible/roles/gitlab_runner/templates/etc/gitlab-runner/config.toml.j2
@@ -111,6 +111,39 @@ metrics_server = "{{ gitlab_runner__metrics_server }}"
 {%       if (runner.docker_tls_cert_path|d() or gitlab_runner__docker_tls_cert_path) %}
     tls_cert_path = "{{ runner.docker_tls_cert_path | d(gitlab_runner__docker_tls_cert_path) }}"
 {%       endif %}
+{%       if runner.docker_tls_verify is defined %}
+    tls_verify = {{ runner.docker_tls_verify | bool | lower }}
+{%       endif %}
+{%       if runner.docker_memory is defined %}
+    memory = "{{ runner.docker_memory }}"
+{%       endif %}
+{%       if runner.docker_memory_swap is defined %}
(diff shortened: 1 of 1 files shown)


imrejonk:
Thanks for the PR. While it is nice to be able to add these options easily, I don't think that this is the way to do it. It might be more straightforward to let the user define these options themselves in a free-form kind of way (say, a dictionary or a list of option names and values). This would also require less maintenance in the future.

lewer:
What about a **docker_options** key in the runner config, which would be a list like:

```
docker_options:
  - name: 'image'
    value: 'ubuntu:18.04'

  - name: 'shm_size'
    value: 0
```
 
and would produce in _/etc/gitlab-runner/config.toml_:

```
[runners.docker]
  image: "ubuntu:18.04"
  shm_size = 0

```

imrejonk:
I think that would be fine, but even better is using [parse_kv_items](https://docs.debops.org/en/master/ansible/roles/ansible_plugins/getting-started.html#configuration-filters). You can then set default configuration values (if needed) and let people add additional values or override default values themselves. Refer to [Universal Configuration](https://docs.debops.org/en/master/user-guide/universal-configuration.html) in the DebOps user manual to see how this works from a user's perspective.

Would you like to attempt using `parse_kv_items` for this? You can find an example in the `debops.icinga_web` role. I'd like to point you to the relevant [documentation](https://docs.debops.org/en/master/ansible/roles/icinga_web/defaults-detailed.html#icinga-web-ref-ini-configuration), [task](https://github.com/debops/debops/blob/69fbcd12966fc67b6a9372387db3c3d68394d86e/ansible/roles/icinga_web/tasks/main.yml) and [template](https://github.com/debops/debops/blob/69fbcd12966fc67b6a9372387db3c3d68394d86e/ansible/roles/icinga_web/templates/etc/icingaweb2/template.ini.j2).

drybjed (approved):
I'll merge this for now. The `gitlab_runner` role configuration can be redesigned in DebOps 3.0 during the next release cycle.

PR #1536, opened 2020-10-21
[postgresql_server] Update postgresql.conf template to only select wal_keep_segments up to postgres 13

growse:
Postgresql 13 renamed `wal_keep_segments` to `wal_keep_size`. Having `wal_keep_segments` in `postgresql.conf` prevents pg13 from starting.


drybjed (approved):
Thanks for finding this and providing a fix, I'll backport it to the previous releases as well.

PR #1537, opened 2020-10-24
[nginx] Default template preserve support < 1.7.5

asyncopation:
Currently the default template will fail nginx `checkconfig` due to new syntax being used without condition.

This PR is small update to first check the nginx version (as we do elsewhere in the template) before using new syntax.


drybjed (approved):
Thanks! I'll backport it to the previous releases as well.

PR #1538, opened 2020-10-25
[fcgiwrap] Fix AnsibleFilterError in check mode

imrejonk:
This solves an issue where the 'Generate systemd units' task fails with
this error: `msg: 'AnsibleFilterError: Version comparison: LooseVersion
instance has no attribute ''version'''`

The task tries to template a systemd service file, but the template
needs fcgiwrap__register_version to do a version comparison. This
variable is never initialized in check mode. This commit changes that by
forcing Ansible to run the 'Check the fcgiwrap version' in check mode as
well.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -322,6 +322,11 @@ LDAP
 - Fixed the configuration support on Ubuntu Focal due to bantime feature
   changes in the :command:`fail2ban` v0.11.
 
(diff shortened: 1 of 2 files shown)


tasosalvas (approved):

drybjed (approved):
Thanks, I'll backport it to previous releases as well.

PR #1539, opened 2020-10-26
[grub] Also output to VGA when ttyS0 is in use

imrejonk:
The debops.grub role previously disabled the VGA console when activating
the serial console with grub__serial_console. These changes make sure
that the VGA console keeps working, which is nice if you're in the
datacenter and only have a VGA display to work with.


tasosalvas (approved):

drybjed (approved):
That's a handy trick, thanks! It will be backported as well.

PR #1543, opened 2020-10-30
[iscsi] Provide default for discovered_portals

imrejonk:
This sets an empty list as the default value of
``ansible_local.iscsi.discovered_portals``, preventing an error
condition when the debops.iscsi local facts have not been saved (i.e. on
the first run against a host).


tasosalvas (approved):

drybjed (approved):
Looks good, thanks.

PR #1546, opened 2020-11-02
[opendkim] Remove dependency on pyOpenSSL

imrejonk:
This dependency was required in Ansible < 2.8.0 because these versions
were unable to use the `cryptography` module, but DebOps is nowadays
developed against Ansible 2.9. pyOpenSSL was used only to generate
private RSA keys for the `debops.opendkim` role. Switching to
`cryptography` is also a security precaution and the Python
Cryptographic Authority [recommends](https://github.com/pyca/cryptography/blob/master/docs/faq.rst#why-use-cryptography)
doing so.

The openssl_privatekey module in `debops.opendkim` will use
`cryptography` when pyOpenSSL is not installed. The Ansible package
depends on this module, so no further changes are required.

Note that this breaks compatibility with the Ansible 2.7 in Debian
Stable, but I believe we aren't trying to maintain compatibility with
that version anymore anyhow.

### DIFF
--- a/docs/ansible/roles/opendkim/defaults-detailed.rst
+++ b/docs/ansible/roles/opendkim/defaults-detailed.rst
@@ -31,12 +31,11 @@ opendkim__keys
 --------------
 
 The ``opendkim__*_keys`` variables define what DomainKeys are created and used
-by OpenDKIM. The private keys are generated on the Ansible Controller (the
-``python-openssl`` package is required), stored in the
-:file:`secret/opendkim/domainkeys/` directory (see :ref:`debops.secret` role for
-details) and copied to the remote hosts.  The role can install the same private
-key on multiple hosts, which can be useful in environments with multiple SMTP
-servers handling the same domains.
+by OpenDKIM. The private keys are generated on the Ansible Controller, stored in
+the :file:`secret/opendkim/domainkeys/` directory (see :ref:`debops.secret` role
+for details) and copied to the remote hosts. The role can install the same
+private key on multiple hosts, which can be useful in environments with multiple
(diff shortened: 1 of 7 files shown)


docs/ansible/roles/opendkim/defaults-detailed.rst
```diff
@@ -31,12 +31,11 @@ opendkim__keys
...
 The ``opendkim__*_keys`` variables define what DomainKeys are created and used
-by OpenDKIM. The private keys are generated on the Ansible Controller (the
-``python-openssl`` package is required), stored in the
-:file:`secret/opendkim/domainkeys/` directory (see :ref:`debops.secret` role for
-details) and copied to the remote hosts.  The role can install the same private
-key on multiple hosts, which can be useful in environments with multiple SMTP
-servers handling the same domains.
+by OpenDKIM. The private keys are generated on the Ansible Controller, stored in
+the :file:`secret/opendkim/domainkeys/` directory (see :ref:`debops.secret` role
+for details) and copied to the remote hosts.  The role can install the same
```
tasosalvas: I recently read a style guide recommending two spaces after a full stop and was bewildered... I don't think we do this here, though.

tasosalvas (dismissed):

docs/ansible/roles/opendkim/defaults-detailed.rst
imrejonk: Indeed we don't. Fixed, thanks for spotting this.

drybjed (approved):
Looks good to me, thanks. I'll backport the changes as well to previous releases.

PR #1547, opened 2020-11-03
[lvm] Fix check mode bug in fs creation task

imrejonk:
This fixes an issue where the role would fail in check mode. The role
tries to simulate creating a filesystem, but this failed when the
underlying LVM volume did not actually exist (which is to be expected
when running in check mode).

### DIFF
--- a/ansible/roles/lvm/tasks/manage_lvm.yml
+++ b/ansible/roles/lvm/tasks/manage_lvm.yml
@@ -55,6 +55,7 @@
   when: lvm__logical_volumes|d(False) and
         item.vg|d() and item.lv|d() and item.size|d() and
         item.state|d('present') != 'absent'
+  register: lvm__register_logical_volumes
 
 - name: Manage filesystems
   filesystem:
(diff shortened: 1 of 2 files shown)


tasosalvas (dismissed):

ansible/roles/lvm/tasks/manage_lvm.yml
```diff
@@ -67,7 +68,9 @@
   when: lvm__logical_volumes|d(False) and
         item.vg|d() and item.lv|d() and item.size|d() and
         item.state|d('present') != 'absent' and
-        ((item.mount|d() and item.fs|d(True)) or item.fs|d())
+        ((item.mount|d() and item.fs|d(True)) or item.fs|d()) and
+        (not ansible_check_mode or
+         (ansible_check_mode and not lvm__register_logical_volumes.changed))
```
ypid:
```suggestion
         (ansible_check_mode and lvm__register_logical_volumes is not changed))
```

I propose this syntax for two reason:

* https://docs.ansible.com/ansible/latest/user_guide/playbooks_tests.html
* Pythonic/PEP

ypid (dismissed):
LGTM. I have one minor style suggestion.

ansible/roles/lvm/tasks/manage_lvm.yml
imrejonk: Thanks for the suggestion, I like your style :)

ypid (approved):

drybjed (approved):
Thanks for the fix. I'll backport it as well.

PR #1548, opened 2020-11-03
[lvm] Mount LVM volumes by default

imrejonk:
This makes the default behaviour match the documentation: the role now
automatically takes care of mounting a filesystem on an LVM volume if
the mount point is specified with ``item.mount``. This previously
required setting the ``item.fs`` parameter to ``True`` as well.


tasosalvas (approved):

drybjed (approved):
Thanks. I won't backport this since it might cause issues with filesystems.

PR #1550, opened 2020-11-04
[php] Document pm_status and pm_status_path fields

apapsch:
These directives are used in the template `pool.conf.j2`, but are not documented.


tasosalvas (approved):

drybjed (approved):
Thanks for maintaining the documentation. :)

PR #1551, opened 2020-11-04
do not create home dir for snmpd

JohannesKleine:
this change allow deployment on Ubuntu 20.04

### DIFF
--- a/ansible/roles/snmpd/tasks/main.yml
+++ b/ansible/roles/snmpd/tasks/main.yml
@@ -68,6 +68,7 @@
     name: '{{ snmpd_user }}'
     groups: '{{ snmpd_proc_hidepid_group }}'
     append: True
+    create_home: False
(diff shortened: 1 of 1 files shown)


imrejonk:
Hi @JohannesKleine, thanks for the PR. What error message did you get when you ran the role against Ubuntu 20.04? And how does your change resolve the issue?

I think that the issue you are trying to solve has something to do with the values of snmpd_user and snmpd_group ([here](https://github.com/debops/debops/blob/master/ansible/roles/snmpd/defaults/main.yml#L31)). It might be better if you update those.

drybjed (approved):
I think that it's a good solution. The `snmpd` APT package should create the `/var/lib/snmp/` home directory by itself, so Ansible doesn't need to mess with it. The task itself is about adding the `snmpd` service account to the `procadmins` UNIX group, not the user account itself.

I'll backport the fix as well.

PR #1552, opened 2020-11-05
[docker_server] Allow venv to inherit from global pip list

ghost:
When the docker-python virtual environment is created, it has limited pip modules in it.
However, it needs `docker` and `docker-compose` pip modules for
/usr/local/bin/docker-python to function properly.
The solution is to add: `virtualenv_site_packages: True` to the `pip` task in the
`docker_server` role which allows the virtual environment to inherit from the 
global site-packages pip list, which includes `docker` and `docker-compose`.

See:
https://docs.ansible.com/ansible/latest/collections/ansible/builtin/pip_module.html
https://docs.debops.org/en/master/ansible/roles/docker_server/docker-virtualenv.html

### DIFF
--- a/ansible/roles/docker_server/tasks/main.yml
+++ b/ansible/roles/docker_server/tasks/main.yml
@@ -88,6 +88,7 @@
     version: '{{ item.version | d(omit) }}'
     virtualenv: '{{ docker_server__virtualenv }}'
     state: '{{ item.state | d("present") }}'
+    virtualenv_site_packages: True
   with_flattened:
     - '{{ docker_server__default_pip_packages }}'
     - '{{ docker_server__pip_packages }}'


imrejonk:
Thanks for the PR. Please let us know what versions of Ansible, Debian and DebOps you are using.

I tried to reconstruct the sequence of events that causes your problem:

1. I believe that you are running Buster or later, so `docker_server__upstream` is `False`.
2. The task 'Install Python packages in virtualenv' creates a virtualenv in `/usr/local/lib/docker/virtualenv`.
3. The 'docker' and 'docker-compose' packages are not installed in this virtualenv because of step 1. In fact, no packages are installed at all, making the virtualenv quite useless.
4. The `docker` and `docker-compose` commands (and their Python packages) are installed from the Debian repositories by the 'Install required packages' task.

So the root issue seems to be the creation of the virtualenv in the first place. I've opened another pull request (#1563) that addresses this. I'm closing this PR for now, feel free to reopen it if I'm overlooking something.

wordtracker:
Hi, I've run into this issue again for the 3rd time in 2 years. I reckon 'docker' and 'docker-compose' modules should be installed by default in the virtualenv as it is mainly used to manage docker. Though am aware that its stated purpose is to make sure that the Python modules used by upstream Docker don't affect the host Python environment. 

My current workaround is to put the following in ansible/inventory/group_vars/all/docker_server.yml
```
docker_server__default_pip_packages: [ 'docker', 'docker-compose' ]
```

This at least generates the modules inside the virtualenv (if the virtualenv doesn't exist already). I also set:
``` 
vars:
    ansible_python_interpreter: "/usr/bin/env docker-python"
```

However for some reason I'm still getting:
```
An exception occurred during task execution. To see the full traceback, use -vvv. The error was: ModuleNotFoundError: No module named 'docker'
fatal: [es1]: FAILED! => {"changed": false, "msg": "Failed to import the required Python library (Docker SDK for Python: docker above 5.0.0 (Python >= 3.6) or docker before 5.0.0 (Python 2.7) or docker-py (Python 2.6)) on es1's Python /usr/local/bin/docker-python. Please read the module documentation and install it in the appropriate location. If the required library is installed, but Ansible is using the wrong Python interpreter, please consult the documentation on ansible_python_interpreter, for example via `pip install docker` (Python >= 3.6) or `pip install docker==4.4.4` (Python 2.7) or `pip install docker-py` (Python 2.6). The error was: No module named 'docker'"}
```

So it looks like '/usr/bin/env docker-python' which points to '/usr/local/bin/docker-python', which in turn points to '/usr/local/lib/docker/virtualenv/bin/python' is still not picking up the modules.

wordtracker:
Some additional context:
1) Currently running Debian 11 (bullseye) with docker_server__upstream set to True (not False) as I wish to use upstream.
2) The task 'Install Python packages in virtualenv' creates a virtualenv in /usr/local/lib/docker/virtualenv.
3) The 'docker' and 'docker-compose' packages are not installed in this virtualenv so I have to add this requirement to docker_server.yml. *This is the part I'm questioning*. 

So I agree about the root issue being the creation of the virtualenv in the first place and note the fix (thanks). However this is a follow on issue in that if the virtualenv is created, the necessary modules don't seem to be installed as one would expect.

PR #1557, opened 2020-11-15
Fix NetBox upgrade; Improvements in various roles

ypid:
Next round of me submitting my open changes to DebOps. I have tested them for some time and just done another self-review.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -152,7 +152,7 @@ Updates of upstream application versions
   has been updated to ``v1.4.1``.
 
 - In the :ref:`debops.netbox` role, the NetBox version has been updated to
-  ``v2.9.4``.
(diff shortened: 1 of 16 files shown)


imrejonk (approved):
LGTM. I really like the shorter syntax that's now possible with recent Ansible versions, that's gonna save a lot of typing. Also kudos for the docker_registry documentation.

ypid:
@imrejonk Thanks. Don’t give me more credit about the docker_registry docs than I deserve. I implemented this feature and then needed to debug it again because something did not work in my setup and then I had to almost reverse engineer it so I now wrote the docs :)

drybjed (approved):
Nice set of changes, thanks. I'll check what commits can be backported to stable releases later.

PR #1563, opened 2020-11-21
[docker_server] Change when virtualenv is created

imrejonk:
The role always created a virtualenv, by default in
/usr/local/lib/docker/virtualenv. This is however not necessary when
using the docker.io and docker-compose packages from the Debian
repositories.

These changes ensure that the virtualenv is, by default, only created
when Docker is installed from upstream.

Ref. https://github.com/debops/debops/pull/1552

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -194,6 +194,14 @@ LDAP
 
 - The DHCP Relay Agent functionality has been moved to :ref:`debops.dhcrelay`.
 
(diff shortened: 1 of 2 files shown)


ypid (approved):
I checked this on my machines. It has not caused an issue but I can follow your points and this change looks good to me.

drybjed (approved):
Looks good, thanks. I'll backport it to the previous releases too.

PR #1564, opened 2020-11-21
[apt_install] Install CPU microcode by default

imrejonk:
The role now installs CPU microcode packages on physical hosts by
default. These firmware updates correct CPU behaviour and mitigate
vulnerabilities like Spectre and Meltdown.

### DIFF
--- a/ansible/roles/apt_install/defaults/main.yml
+++ b/ansible/roles/apt_install/defaults/main.yml
@@ -56,7 +56,7 @@ apt_install__archive_areas_map:
 # List of package archive areas which are currently available. This list is
 # used to conditionally enable packages for installation, depending on
 # availability of a given archive area.
-apt_install__archive_areas: '{{ ansible_local.apt.archive_areas
+apt_install__archive_areas: '{{ ansible_local.apt.components
                                 |d(apt_install__archive_areas_map[apt_install__distribution] | d([])) }}'
 
                                                                    # ]]]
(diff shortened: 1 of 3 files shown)


ansible/roles/apt_install/defaults/main.yml
```diff
@@ -248,6 +250,24 @@ apt_install__conditional_whitelist_packages:
...
+  - name: 'amd64-microcode'
+    # Non-free microcode firmware for AMD CPUs. These patches mitigate security
+    # issues like Spectre and fix other types of incorrect processor behaviour.
+    # The package only needs to be installed on physical machines. To protect
+    # virtual machines, be sure to set the correct CPU flags in QEMU:
+    # https://www.qemu.org/docs/master/system/target-i386.html#important-cpu-features-for-intel-x86-hosts
+    whitelist: '{{ apt_install__conditional_whitelist_packages }}'
+    state: '{{ "present" if ansible_virtualization_role == "host" and
+                            "AuthenticAMD" in ansible_processor
+                         else "absent" }}'
```
ypid: Those packages are non-free, so installation will fail without those areas enabled.

```suggestion
    state: '{{ "present"
               if (ansible_virtualization_role == "host" and
                   "AuthenticAMD" in ansible_processor and
                   ansible_local.apt.nonfree|d()|bool
               else "absent" }}'
```

Technically, this is firmware so it would better fit in `apt_install__firmware_packages` which also ~~supports~~ has `area`. I quickly checked the code, it seems that `ansible_local.apt.archive_areas` is currently not available. You might play with this and get it to work.

Also note the coding style.

ypid (changes requested):

ansible/roles/apt_install/defaults/main.yml
imrejonk: Thanks for the feedback. I'm trying to get this to work with `apt_install__firmware_packages` and `ansible_local.apt.archive_areas`, but this is quite difficult because of all the distribution/release mappings in the `apt` and `apt_install` roles. I only plan on testing my changes on Debian stable, and I assume you and other DebOps developers do too. Should I just ignore the Devuan, Raspbian and Ubuntu references?

ypid: > Should I just ignore the Devuan, Raspbian and Ubuntu references?

In my eye you can. We provide support for others on best effort basis :) Yes, I also only test against Debian stable.

> but this is quite difficult

I also think so now that I checked this out again today. If you have a better idea, I guess it can be discussed.

imrejonk: I took another look at this. The templating in `debops.apt_install` is complex but very useful, as it allows you to add arbitrary parameters to packages in `apt_install__firmware_packages` (and other lists in `apt_install__all_packages`) that are then compared against the parameters in `apt_install__condition_map`. So I just tweaked those parameters. I also added the `ansible_local.apt.components` fact that lists the repository components (main, contrib, non-free and so on). This is confusingly called 'archive areas' in `apt_install`. Ah well.

These changes should ensure CPU microcode installation on Devuan and Ubuntu as well, but I have only tested against Debian stable.

drybjed: Confusingly, hmm... Well, Debian Wiki [mentions "components"](https://wiki.debian.org/DebianRepository), while Debian Policy [mentions "archive areas"](https://www.debian.org/doc/debian-policy/ch-archive.html). The breaker of the tie is the `sources.list(5)` manual which mentions "components" but not "areas". So I'd say, good choice on the fact name.

drybjed (approved):
Nicely done, thanks. I'll backport this as well.

PR #1566, opened 2020-11-22
[etckeeper] Install etckeeper from backports

### LINKED ISSUE #1560: [etckeeper] Use py3-compatible version from buster-backports
`debops.etckeeper` installs etckeeper from the Debian Stable repository, which is currently at version 1.18.10-1 for Debian 10. This package however depends on Python 2. The role does not install etckeeper by default when a Python 2 environment is not present on the host (see [etckeeper__enabled](https://docs.debops.org/en/master/ansible/roles/etckeeper/defaults/main.html#envvar-etckeeper__enabled)).

I propose that, on Debian 10, we install etckeeper from buster-backports. The package in that repository (version 1.18.14-1~bpo10+1) depends on Python 3 instead. This should allow default installation of etckeeper on all hosts, not just the hosts that have a Python 3 environment.

Let me know what you think of this and I'll put some effort in.

imrejonk:
Closes #1560

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -194,6 +194,13 @@ LDAP
 
 - The DHCP Relay Agent functionality has been moved to :ref:`debops.dhcrelay`.
 
+:ref:`debops.etckeeper` role
+''''''''''''''''''''''''''''
+
+- The role now installs etckeeper on all hosts by default, not just on hosts
+  that have a Python 2 environment. etckeeper is also installed from
+  buster-backports instead of the main Debian 10 repository.
+
 :ref:`debops.fhs` role
 ''''''''''''''''''''''
 
@@ -2083,8 +2090,7 @@ User management
 ''''''''''''''''''''''''''''
 
 - The installation of :command:`etckeeper` will be disabled by default in
-  Python 3.x-only environments. See :ref:`role documentation
-  <etckeeper__ref_python3only>` for more details.
+  Python 3.x-only environments.
 
 :ref:`debops.gitlab` role
 '''''''''''''''''''''''''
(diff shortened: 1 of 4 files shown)


CHANGELOG.rst
```diff
@@ -194,6 +194,13 @@ LDAP
 
 - The DHCP Relay Agent functionality has been moved to :ref:`debops.dhcrelay`.
 
+:ref:`debops.etckeeper` role
+''''''''''''''''''''''''''''
+
+- The role now installs etckeeper on all hosts by default, not just on hosts
+  that have a Python 2 environment. Etckeeper is also installed from
```
ypid:
```suggestion
  that have a Python 2 environment. etckeeper is also installed from
```

ypid (changes requested):
LGTM. I only ask for some style changes:

* Consistent form of "etckeeper". It seems the authors want it to be always written as lowercase. Ref: https://etckeeper.branchable.com/README/
* Consistent commit message. Use `[etckeeper]` instead of `[etc_keeper]`.

ypid:
And not to forget `docs/ansible/roles/etckeeper/getting-started.rst`.

CHANGELOG.rst
imrejonk: Crap, I keep writing etc_keeper... Thanks for spotting this.

imrejonk:
Docs are updated as well now

ypid (approved):
LGTM. Thank you. So the Python 3 only future is near :)

drybjed (approved):
Sounds good. I'll fix the Changelog conflict and add the `etckeeper` `apt_preferences` entry in the common playbook.

PR #1567, opened 2020-11-23
Add apache config to monit defaults

vaisov:


drybjed (approved):
Good idea, thanks. I'll backport it to the stable releases as well.

PR #1568, opened 2020-11-25
[nullmailer] Prevent unnecessary installation

### LINKED ISSUE #652: nullmailer shouldn't install itself only to be removed by postfix in the same run
Test case
========
My remote server is a clean Ubuntu 16.04.
Here's an excerpt from `inventory/hosts`
```
[debops_all_hosts]
EXAMPLE.domain.com

[debops_service_postfix]
EXAMPLE.domain.com
```

What happens
============
When I run `$ debops `initially, the nullmailer role installs nullmailer and sets up its configuration.

But a little while later, the postfix role purges nullmailer.

What I expected to happen
========
The nullmailer role detects that this host will have postfix installed and does not install itself.

Maybe you could check if the current host is part of debops_service_postfix?
But some systems don't directly depend on the postfix service but have it as a dependency.
Could you have postfix run before nullmailer?

imrejonk:
This change prevents `debops.nullmailer` from installing nullmailer when
the host is part of the `debops_service_postfix` group.

Closes #652

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -355,6 +355,14 @@ LDAP
   alternative NTP service implementation and ``systemd-timesyncd`` is not
   available.
 
+:ref:`debops.nullmailer` role
+'''''''''''''''''''''''''''''
+
+- The role will no longer install nullmailer if the host is also configured to
+  install Postfix. This prevents a situation where nullmailer will get installed
+  as part of the common playbook, only to be removed later on in the playbook
+  run.
+
 :ref:`debops.owncloud` role
 ''''''''''''''''''''''''''''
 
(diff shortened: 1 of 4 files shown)


drybjed:
This change moves the condition check from the remote host to the Ansible inventory, additionally depending on another roles' inventory group. This forces users that don't use DebOps playbooks to define the inventory groups in a specific way, which I would want to avoid. I know that there are probably more examples of this behaviour elsewhere, and they probably should be eliminated over time.

imrejonk:
As discussed during the Jitsi meeting, and as pointed out by @drybjed above, this is not the way we want to solve this issue. Users can set `nullmailer__enabled` to `False` for now.

PR #1570, opened 2020-12-02
Add debian bullseye to php list

oz123:
This needed to get the role working on the current debian testing.


drybjed (approved):
Good idea, thanks. I'm not sure why Travis-CI test is stalling, but it should work fine. I'll backport the change as well.

PR #1574, opened 2020-12-04
monit code style update and update the structure of the CODEOWNERS file

ypid:
@drybjed I have no relevant fixes for the next release. This PR can also be considered after the release.

@tasosalvas: I propose to add you as codeowner of lib/images. In case that does not make sense for you just comment.

@vaisov: DebOps has adopted this new style. See #1371
Updates: #1567

This change to the `CODEOWNERS` file needs to be discussed:

* Use `**SOFTWARE**` as a more maintainable way to match files.

  The percentage of false-positives should be pretty small. Arguably, if a file outside of your rule has the name of your role in it, you might be interested in changes to this file as well. This now also includes docs which was previously not covered.

* Drop the use of trailing root slash in CODEOWNERS.

  It is uncommon to have in gitignore files. My main reason for this change is to make Vim file complete (`C-x` `C-f`) and `gf` work which comes in handy to validate paths.

### DIFF
--- a/CODEOWNERS
+++ b/CODEOWNERS
@@ -31,6 +31,7 @@
 /docs/                                              @drybjed @tasosalvas
 *.rst                                               @drybjed @tasosalvas
 *.md                                                @drybjed @tasosalvas
+lib/images/                                         @drybjed @tasosalvas
 docs/news/releases.rst                              @drybjed @tasosalvas @ypid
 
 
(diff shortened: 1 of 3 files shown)


drybjed (approved):
Sounds good, although I think that having a role entry in multiple software categories (in the documentation, not the codeowners file) is fine.

I'll probably not backport this change, since development is done in the `master` branch. If you think that backporting might be useful, let me know.

PR #1580, opened 2020-12-14
Add H5BP settings from https://github.com/h5bp/server-configs-nginx

vaisov:
All settings are optional and are enabled only with nginx__h5bp_enabled
set to true

### DIFF
--- a/ansible/roles/nginx/tasks/nginx_configs.yml
+++ b/ansible/roles/nginx/tasks/nginx_configs.yml
@@ -46,6 +46,31 @@
   when: (item.name|d() and item.state|d('present') != 'absent' and
          (item.delete is undefined or not item.delete|bool))
 
+- name: Remove h5bp nginx maps if requested
+  file:
+    dest: '/etc/nginx/conf.d/map_{{ item.name }}.conf'
+    state: 'absent'
+  with_flattened:
+    - '{{ nginx__h5bp_maps }}'
+  notify: [ 'Test nginx and reload' ]
+  when: (item.name|d() and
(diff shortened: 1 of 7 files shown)


ansible/roles/nginx/tasks/nginx_configs.yml
```diff
@@ -131,3 +156,41 @@
...
+    dest: '/etc/nginx/conf.d/h5bp/{{ item.name }}'
+    owner: 'root'
+    group: 'root'
+    mode: '0644'
+  with_flattened:
+    - '{{ nginx__h5bp_files }}'
+  notify: [ 'Test nginx and reload' ]
+  when: (item.name|d() and item.state|d('present') != 'absent' and
+         (item.delete is undefined or not item.delete|bool) and
+         (nginx__h5bp_enabled|bool))
```
ypid, line 196: I would prefer when the https://github.com/h5bp/server-configs-nginx gets cloned below `/usr/local/src` on the remote system and then the files are copied in place. Search for `/usr/local/src` in the DebOps repo for examples.

See also: https://docs.debops.org/en/master/dep/dep-0003.html

ypid (changes requested):

ypid:
What advantages do the snippets have over the already supported way (https://docs.debops.org/en/master/ansible/roles/nginx/defaults-detailed.html#http-security-headers)? Keep in mind that this role is planned to be reworked to [Universal Configuration](https://docs.debops.org/en/master/user-guide/universal-configuration.html).

PR #1582, opened 2020-12-18
Update to NetBox 2.10.1 and fix issue with "Perform database installation or migration"

ypid:
The clean install is probably still broken. But at least this PR fixes the upgrade to v2.10 :)


drybjed (approved):
Thanks! This change also fixes and issue with NetBox login page infinitely redirecting to a resource that cannot be accessed.

ypid:
You are welcome. Yes, the infinite redirect was also my problem.

PR #1583, opened 2020-12-19
Use HTTPS everywhere (mechanical edit using util from https-everywhere)

ypid:
Mechanical edit which replaces `http://` by `https://` for URLs where it is known that the web server supports HTTPS.

```Shell
node ~/src/EFForg/https-everywhere/utils/rewriter/rewriter.js .
```

A few changes were reset manually before the commit.

Ref: https://github.com/EFForg/https-everywhere/tree/master/utils/rewriter

### DIFF
--- a/.yamllint
+++ b/.yamllint
@@ -10,12 +10,12 @@
 # Project Homepage: https://github.com/adrienverge/yamllint
 #
 # How to override rules in files:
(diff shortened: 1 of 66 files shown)


imrejonk (approved):
LGTM! That's a nice tool to use for these edits as well.

drybjed (approved):
Nice work, thanks!

PR #1586, opened 2020-12-23
[icinga_web] add LDAP support for authentication and groups

lewer:
This PR adds support for LDAP authentication and groups.

It configures options described here: https://icinga.com/docs/icinga-web-2/latest/doc/05-Authentication/#active-directory-or-ldap-authentication

It was successfully tested on my infra (with manual LDAP management; I don't use debops for that. But I was able to create the uid=icingaweb user with **icinga_web__dependent__ldap_tasks**)

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -239,6 +239,11 @@ LDAP
 - The role will now activate both the serial console and the (previously
   disabled) native platform console when ``grub__serial_console`` is ``True``.
 
+:ref:`debops.icinga_web` role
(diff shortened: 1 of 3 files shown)


imrejonk:
This is really nice, thanks for the PR! We've been configuring our LDAP settings through the Icinga Web interface, but this is obviously better.

I am currently testing this against our setup. I have a few suggestions which I'll propose tomorrow.

imrejonk:
@lewer Here are my proposed changes: https://github.com/lewer/debops/pull/1
Let me know what you think!

PR #1587, opened 2020-12-30
[icinga_web] add support for x509 module

lewer:
This PR adds support for Icinga Certificate Monitoring (https://icinga.com/docs/icinga-certificate-monitoring/latest/)

It requires a MySQL/MariaDB database (currently not compatible with PostgreSQL)

It installs the module from the git repo, setup the mysql database and creates the needed icingaweb resources, so that the user is able to monitor his certificates in Icinga Web.

The monitoring configuration (https://icinga.com/docs/icinga-certificate-monitoring/latest/doc/10-Monitoring/) is left to the user with Icinga Director, although another PR could probably automatically create needed objects/services in Icinga Director.

### DIFF
--- a/ansible/playbooks/service/icinga_web.yml
+++ b/ansible/playbooks/service/icinga_web.yml
@@ -22,7 +22,9 @@
           - '{{ php__keyring__dependent_apt_keys }}'
           - '{{ nginx__keyring__dependent_apt_keys }}'
           - '{{ postgresql__keyring__dependent_apt_keys if (icinga_web__database_type == "postgresql") else [] }}'
-          - '{{ mariadb__keyring__dependent_apt_keys if (icinga_web__database_type == "mariadb") else [] }}'
+          - '{{ mariadb__keyring__dependent_apt_keys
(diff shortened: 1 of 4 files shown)


drybjed (commented):
The PR looks good, nice job. I'll backport this as well to the stable releases.

Question: Icinga documentation mentions scheduling the x509 scan via a systemd service (I would assume a service + timer units), or a cron job. Are you doing that in your setup? Or can this be done via Icinga Director internally and separate external exection of scan jobs is not needed?

lewer:
Good point. This scheduling thing isn't crystal clear to me. As far as I understand:
you can either define a schedule together with the job definition in icingaweb UI image
and then when you run `icingacli x509 jobs run`, it scans the jobs which are currently due

or you can define a cron job `icingacli x509 scan --job jobname`
This second option seems more natural to me, but it requires to know _jobname_, and can only be set by the user. 

In my setup this is not a real problem because the infrastructure is stable, I'm OK with a manual scan from time to time when a new host arrives.

drybjed:
Sounds good then. I'll add a note about this in the role documentation. I'll wait for the LDAP PR to be updated to work on this as well.

PR #1591, opened 2021-01-03
[gitlab] Fix gitlab shell compile and sshusers system group

pedroluislopez:
- Use system group prefix when LDAP support is enabled.
- Fix GitLab shell compile for newer versions. Now it uses make.


drybjed (approved):
Nice work, thanks. I'll backport it as well.

ypid:
PS: @pedroluislopez There is a easier way to do it, see #1371 :)

PR #1597, opened 2021-01-16
[gitlab] Some fixes for Gitlab pages and add access control

pedroluislopez:
- Fix gitlab pages configuration. The configuration in gitlab.yml is only used for compose the URL's in the GitLab front and the role installs nginx with https. So, the correct configuration in gitlab.yml file is to use port 443 and https.
- Fix gitlab systemd services deploy. GitLab Pages could be activated after the first installation. So, it's necessary drop "repository checkout changed" condition in systemd services deploy.
- Improvement GitLab Pages configuration. For version greater or equal than 12.8 it use a default file configuration instead of parameters in systemd service.
- Added support for GitLab Pages access control for version greater or equal than 12.10.

### DIFF
--- a/ansible/roles/gitlab/tasks/configure_gitlab-pages.yml
+++ b/ansible/roles/gitlab/tasks/configure_gitlab-pages.yml
@@ -110,3 +110,13 @@
   become: True
   become_user: '{{ gitlab_user }}'
   when: gitlab_register_pages_checkout is changed
+
+- name: Setup GitLab Pages default variables
+  template:
+    src: 'etc/default/gitlab-pages.j2'
+    dest: '/etc/default/gitlab-pages'
+    owner: 'root'
+    group: 'root'
+    mode: '0644'
+  when: gitlab_version is version_compare("12.8", operator="ge", strict=True)
+  notify: [ 'Restart gitlab', 'Restart gitlab.slice' ]
--- a/ansible/roles/gitlab/tasks/configure_gitlab-pages.yml
+++ b/ansible/roles/gitlab/tasks/configure_gitlab-pages.yml
@@ -116,7 +116,7 @@
     src: 'etc/default/gitlab-pages.j2'
     dest: '/etc/default/gitlab-pages'
     owner: 'root'
-    group: 'root'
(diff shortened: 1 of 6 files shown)


ansible/roles/gitlab/tasks/configure_gitlab-pages.yml
```diff
@@ -110,3 +110,13 @@
...
   become_user: '{{ gitlab_user }}'
   when: gitlab_register_pages_checkout is changed
+
+- name: Setup GitLab Pages default variables
+  template:
+    src: 'etc/default/gitlab-pages.j2'
+    dest: '/etc/default/gitlab-pages'
+    owner: 'root'
+    group: 'root'
+    mode: '0644'
```
drybjed: The file contains secrets, so it should have more strict permissions. I would suggest changing the mode to `0640` and the group to `{{ gitlab_group }}` so that GitLab user can read the contents.

ansible/roles/gitlab/templates/etc/default/gitlab-pages.j2
```diff
@@ -0,0 +1,18 @@
+{# Copyright (C) 2011-2015 GitLab B.V. <https://gitlab.com/>
+ # Copyright (C) 2015-2017 Maciej Delmanowski <drybjed@gmail.com>
```
drybjed: You don't need to include my copyright here. :-)

drybjed (changes requested):
Looks good to me, just update the configuration file permissions.

ansible/roles/gitlab/tasks/configure_gitlab-pages.yml
pedroluislopez: Done!

ansible/roles/gitlab/templates/etc/default/gitlab-pages.j2
pedroluislopez: It's a key-value file, he. But, ok, i've set my copyright :D

drybjed (approved):

PR #1598, opened 2021-01-17
Fix netbox-rq service restart and maintenance

ypid:


README.md
```diff
@@ -177,8 +184,8 @@ contribute to DebOps.
 
 ## Licensing
 
-The DebOps project is licensed under the [GNU General Public License 3.0][gpl-3.0].
-You can find full text of the license in the [LICENSE][license] file.
+The DebOps project is licensed under the [GNU General Public License 3.0 or later][GPL-3.0-or-later].
+You can find full text of the license in the [GPL-3.0-or-later.txt] file.
```
drybjed: Could this identifier be called `LICENSES/GPL-3.0-or-later.txt` so that users can find the correct path visually? I'm not sure if `/` can be used in references in Markdown...

README.md
```diff
@@ -1,3 +1,10 @@
+<!--
+Copyright (C) 2018-2020 Maciej Delmanowski <drybjed@gmail.com>
+Copyright (C) 2018-2020 DebOps <https://debops.org/>
+
+SPDX-License-Identifier: GPL-3.0-or-later
+-->
```
drybjed, line 6: This file is used to generate the readme for the Python package via `pandoc`, and the contents will show up on Ansible Galaxy and PyPI. I'm not sure if they will correctly interpret the HTML comment in a Markdown file. I'm fine with keeping this detail in `.reuse/dep5` file.

ansible/roles/global_handlers/handlers/netbox.yml
```diff
@@ -8,7 +8,6 @@
     name: 'gunicorn@netbox'
     state: 'restarted'
   when: (not netbox__app_internal_appserver|bool and
-         ansible_local|d() and ansible_local.gunicorn|d() and
          ansible_local.gunicorn.installed|bool)
```
drybjed: This might break if `gunucorn` fact is not present. You could wrap this in `|d()` to avoid that:
```
(ansible_local.gunicorn.installed|d())|bool
```

ansible/roles/global_handlers/handlers/netbox.yml
```diff
@@ -8,7 +8,6 @@
     name: 'gunicorn@netbox'
     state: 'restarted'
   when: (not netbox__app_internal_appserver|bool and
```
drybjed: Moving NetBox handlers to global handlers might not be a good idea, since they refer to the default variables in the `netbox` role. Did you check that when you run other DebOps playbooks that don't have `netbox` role included, they don't blow up and Ansible doesn't complain about an undefined variables? I specifically did not move these handlers due to them referring to an internal role variables.

.gitignore
```diff
@@ -1,6 +1,6 @@
 # Copyright (C) 2017-2019 Maciej Delmanowski <drybjed@gmail.com>
 # Copyright (C) 2019-2019 DebOps <https://debops.org/>
-# SPDX-License-Identifier: GPL-3.0-or-later
+# SPDX-License-Identifier: CC0-1.0
```
drybjed, line 3: Yeah, I suppose `.gitignore` cannot be copyrighted as such, the same probably goes for `.dockerignore`.

drybjed (changes requested):
Thanks. There are some changes I'm not sure about, see the comments below.

README.md
ypid: As you wish, sounds good to me and it is covered by the spec:

https://spec.commonmark.org/dingus/?text=%5Bfoo%5D%3A%20%2Furl%20%22title%22%0A%0A%5Bfoo%5D%0A

https://spec.commonmark.org/0.29/#link-reference-definitions

README.md
ypid: https://git.fsfe.org/reuse/tool is the role model right? :) https://git.fsfe.org/reuse/tool/raw/branch/master/README.md

I have used HTML comments in Markdown for some time and have not noticed issues.

Also, I agree with REUSE that when possible, the info should be in the file itself.

ansible/roles/global_handlers/handlers/netbox.yml
ypid: Good catch :+1:
But the extra parenthesis are probably not needed.

ansible/roles/global_handlers/handlers/netbox.yml
ypid: I just tried it (after fixing the missing import of the handlers), another role using global handlers did not blow up. But now I know the reason why you left the NetBox role out so I undid this.

ypid:
Good review :+1: I should have addressed all of your feedback.

README.md
drybjed: Alright then, let's see what breaks. :)

drybjed (approved):

PR #1604, opened 2021-01-21
system_users: use correct variable to enable/disable ACLs

Alphix:
The system_users role sets a default value for `system_users__acl_enabled` (documented [here](https://docs.debops.org/en/master/ansible/roles/system_users/defaults/main.html))

https://github.com/debops/debops/blob/cf694ada55def977c0ebac03c5f2a6f9343a2336/ansible/roles/system_users/defaults/main.yml#L31

but then checks for `system_users__home_acl_enabled` instead:

https://github.com/debops/debops/blob/cf694ada55def977c0ebac03c5f2a6f9343a2336/ansible/roles/system_users/tasks/main.yml#L182


drybjed (approved):
Excellent, thankt you! I'll backport this as well.

PR #1605, opened 2021-01-21
etherpad: log to file by default

bleuchtang:
Etherpad role prepare everything to allow etherpad to log into /var/log/etherpad-server/ ([dir creation](https://github.com/debops/debops/blob/master/ansible/roles/etherpad/tasks/main.yml#L129), [logrotate](debops/debops/blob/master/ansible/roles/etherpad/defaults/main.yml#L457) ) but by default the service log into syslog. This PR tel to systemd to log into /var/log/etherpad-server/production.log

This can break things if people parse syslog to extract etherpad log. I can add a variable to make this PR not enabled by default.

### DIFF
--- a/ansible/roles/etherpad/defaults/main.yml
+++ b/ansible/roles/etherpad/defaults/main.yml
@@ -87,6 +87,12 @@ etherpad_src_dir: '{{ (ansible_local.fhs.src | d("/usr/local/src"))
 etherpad_log_dir: '{{ (ansible_local.fhs.log | d("/var/log"))
                       + "/" + etherpad_system_name }}'
                                                                    # ]]]
+# .. envvar:: etherpad_log_to_file [[[
+#
+# By default Etherpad log into syslog, set to true if you whant to log into a
+# file instead of syslog.
+etherpad_log_to_file: False
+                                                                   # ]]]
                                                                    # ]]]
 # Basic configuration [[[
 # -----------------------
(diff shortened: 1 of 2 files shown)


ypid:
Have you seen https://hedgedoc.org/ btw? Seems to me that the etherpad role has not received much updates over the last few years. Maybe it is time for a switch.

drybjed:
@bleuchtang: I think having a variable to control this might be best, please add it if you can.

@ypid: `hedgedoc` is interesting, and it might be good to have two NodeJS-based projects as test cases for DebOps. There are currently [packaging effors in Debian](https://wiki.debian.org/Javascript/Nodejs/Tasks/HedgeDoc) but it seems that they won't be finished for Debian Bullseye. [Manual installation instructions](https://docs.hedgedoc.org/setup/manual-setup/) seem simple enough.

bleuchtang:
> @bleuchtang: I think having a variable to control this might be best, please add it if you can.
> 
I will

> @ypid: `hedgedoc` is interesting, and it might be good to have two NodeJS-based projects as test cases for DebOps. There are currently [packaging effors in Debian](https://wiki.debian.org/Javascript/Nodejs/Tasks/HedgeDoc) but it seems that they won't be finished for Debian Bullseye. [Manual installation instructions](https://docs.hedgedoc.org/setup/manual-setup/) seem simple enough.

I wrote a role debops-compatible for codimd, I can share it if you want.

ypid:
That is always welcome. You can put it into https://github.com/debops/debops/wiki/DebOps-roles-in-development or if it is fully ready open a PR. So you wrote a CodiMD role and at the same time use EtherPad. Already have a favorite/strategy?

drybjed (approved):

PR #1606, opened 2021-01-22
Fix for NetBox supperuser account creation and small LDAP and mail cleanup

drybjed:


ypid (dismissed):
The NetBox fix looks good to me. Thanks!

PR #1609, opened 2021-01-23
[owncloud] add variable for release channel

bleuchtang:
Since Nexcloud 17 we should change the default channel to stable ([default channel for CE](https://nextcloud.com/release-channels/)) . This PR add a variable to set the channel outside the role and define this variable to stable for nexcloud > 17.0.

### DIFF
--- a/ansible/roles/owncloud/defaults/main.yml
+++ b/ansible/roles/owncloud/defaults/main.yml
@@ -819,6 +819,12 @@ owncloud__role_config:
     state: '{{ "present" if (owncloud__temp_path != "") else "absent" }}'
     value: '{{ owncloud__temp_path }}'
 
+# .. envvar:: owncloud__release_channel
+#
+# Release channel
+owncloud__release_channel: '{{ "stable" if (owncloud__variant == "nextcloud" and
+                                owncloud__release is version("17.0", ">="))
+                            else "production" }}'
 
 # .. envvar:: owncloud__role_recommended_config
 #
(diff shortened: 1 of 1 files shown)


ansible/roles/owncloud/defaults/main.yml
```diff
@@ -819,6 +819,12 @@ owncloud__role_config:
     state: '{{ "present" if (owncloud__temp_path != "") else "absent" }}'
     value: '{{ owncloud__temp_path }}'
 
+# .. envvar:: owncloud__release_channel
+#
+# Release channel
+owncloud__release_channel: '{{ "stable" if (owncloud__variant == "nextcloud" and
+                                owncloud__release is version("17.0", ">="))
+                            else "production" }}'
```
ypid: Indention should match. See the other if conditions in that file.

Also "Release channel" as variable description does not add much value to the variable name. Can you add more context? Reference to the URL would be fine. I try to limit the redundancy we have to maintain here.

ypid (changes requested):
Good catch! I confirm that `production` is a legacy setting. I tested it in the GUI and as soon as I select another value, `production` disappears. It still worked so I guess they have `stable` as fallback.

ansible/roles/owncloud/defaults/main.yml
```diff
@@ -414,9 +412,11 @@ owncloud__variant_name_map:
 #
 # Supported releases:
 #
-# * ownCloud ``10.4``
+# * ownCloud ``10.3``
 #
-# * Nextcloud ``19.0``
+# * Nextcloud ``16.0``
+# * Nextcloud ``17.0``
+# * Nextcloud ``18.0``
```
ypid: Something is wrong with this commit.

bleuchtang: Grrr, sorry.

ypid (approved):
LGTM, thanks!

drybjed (approved):
Nicely done, thanks. I'll backport it as well.

PR #1610, opened 2021-01-24
[pki] Fix a warning from ansible

Alphix:
flush_handlers doesn't support "when:" conditionals


drybjed (approved):
Sounds good. I'm not even sure why I added a condition here, but it's probably a no-op at this point. I'll backport this change as well.

PR #1611, opened 2021-01-24
[ldap] Make iface name matching configurable

Alphix:
Right now the interface names are hardcoded in the lookup templates.
I have some nspawn containers with macvlan interfaces (mv-*). They
did not get any IP/MAC addr added to LDAP (which means that no device
entry is created at all since IP/MAC is mandatory in the LDAP schema).

Instead of simply adding "mv-*" interfaces to the lookup scripts, how
about making the lookup configurable, given that there is an everchanging
list of possible interface names.


drybjed (approved):
Sounds good, thanks. I tested this and the change works as expected. I'll backport it as well.

PR #1618, opened 2021-01-26
[python] Install 'python-pip' conditionally

drybjed:


drybjed:
The Travis-CI tests will be removed soon since they take more and more time to complete. The PR should be fine, so I'll merge it.

PR #1621, opened 2021-01-28
[postfix] Disable rarely-used DSN and ETRN

imrejonk:
This disables the rarely-used DSN and ETRN commands, which are used for
Delivery Status Notifications (RFC 3464) and Remote Message Queue
Starting (RFC 1985), respectively. Both commands are optional SMTP
service extensions. Disabling them should improve security and
spam-resistance, and should not have noticeable side effects. GMail has
both commands disabled, for example.

A DebOps user can enable these commands again if needed.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -351,6 +351,22 @@ LDAP
   since Postfix 3.0. This allows specifying parameter values that contain
   whitespace.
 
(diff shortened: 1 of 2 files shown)


drybjed (approved):
Sounds good, thanks. I'll backport it as well.

PR #1624, opened 2021-02-02
[redis_server] Add a way to overwrite save option

scibi:
By default there are 3 values of `save` option:

```yaml
redis_server__default_base_options:
  - name: 'save'
    value: [ '900 1', '300 10', '60 10000' ]
    dynamic: True
```

I need to overwrite it but using something like this doesn't work - new values are appended:

```yaml
redis_server__group_configuration:
  - name: 'main'
    options:
      - name: 'save'
        value:
          - '1200 1'
        dynamic: True
```

This PR makes it possible to disable default values:

```yaml
redis_server__group_configuration:
  - name: 'main'
    options:
      - name: 'save'
        value:
          - '1200 1'
          - name: '900 1'
            state: 'absent'
          - name: '300 10'
            state: 'absent'
          - name: '60 10000'
            state: 'absent'
        dynamic: True
```

Let me know if there is a better way to do it ;)

### DIFF
--- a/ansible/roles/redis_server/templates/etc/redis/instance/ansible-redis-dynamic.conf.j2
+++ b/ansible/roles/redis_server/templates/etc/redis/instance/ansible-redis-dynamic.conf.j2
@@ -37,7 +37,7 @@ readarray -t redis_dynamic_config_map << EOF
 {%           endif %}
 {%         endif %}
 {%       elif element.value is iterable and element.value is not string and element.value is not mapping %}
-{{ '{}={}'.format(element.name, element.value | map(attribute='name') | list | join(' ')) }}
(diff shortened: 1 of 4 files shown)


drybjed:
By default the `parse_kv_config` filter appends new elements to a list, but you can use an empty string to reset a list and create it from scratch:
```yaml
redis_server__group_configuration:
  - name: 'main'
    options:
      - name: 'save'
        value: ''
        dynamic: True

      - name: 'save'
        value:
          - '1200 1'
        dynamic: True
```
But your addition is good, so I'll merge it. You could update the role documentation to indicate this as a possibility, maybe provide an example. If not, I can do it a bit later.

drybjed (approved):

PR #1628, opened 2021-02-07
[system_users] fix python version check

Alphix:
When running the system_users role on a controller with only python3
against a host with python2, the wrong getent_passwd.py script
will be exeucted.

The reason is that the "Gather local Ansible user details" task is
delegated to localhost, but the ansible_python_version variable
comes from the host, not the controller.


drybjed (approved):
That's a neat way to handle this issue, thanks! I'll backport this as well.

PR #1629, opened 2021-02-07
[slapd]: update defaults

Alphix:
One minor change and one comment update in the slapd defaults

### DIFF
--- a/ansible/roles/slapd/defaults/main.yml
+++ b/ansible/roles/slapd/defaults/main.yml
@@ -990,7 +990,7 @@ slapd__structure_tasks:
       pwdInHistory: '5'
       pwdCheckQuality: '1'
       pwdMinLength: '10'
-      pwdExpireWarning: '3600'
+      pwdExpireWarning: '1209600'
       pwdGraceAuthNLimit: '5'
       pwdLockout: 'FALSE'
       pwdLockoutDuration: '300'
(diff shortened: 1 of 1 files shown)


ansible/roles/slapd/defaults/main.yml
```diff
@@ -737,6 +737,7 @@ slapd__default_tasks:
     state: 'exact'
 
     # Warning: changes in this attribute require slapd restart
+    # (fixed for slapd 2.5, see https://bugs.openldap.org/show_bug.cgi?id=6035)
```
drybjed: This comment isn't really useful here - what exactly was fixed? We have to go to the URL to see that the restart for changes in `olcAuthzRegexp` attribute is not needed on `slapd` 2.5. That will be great, when `slapd` 2.5 is finally released. :-)

I suppose that if you want to amend the warning, I would go with something like:
```yaml
# Warning: OpenLDAP v2.4.x requires a restart when this attribute is changed
```

drybjed (changes requested):
The warning time change is a good idea - I think that I based it on some pwdPolicy guide, perhaps Zytrax one. Your change makes more sense. :-) I'll backport it as well.

The comment isn't really needed, I think. Either remove it or change it as suggested.

ansible/roles/slapd/defaults/main.yml
Alphix: Ok, done and updated branch force-pushed

drybjed (approved):

PR #1631, opened 2021-02-07
[netbox] Fix NetBox crash when it tried to send Emails

ypid:


drybjed (approved):
Thanks for fixing the bug! The error with Docker image creation is already fixed in the `master` branch, so I'm merging the PR.

PR #1633, opened 2021-02-08
[sshd]: change defaults so port doesn't have to be specified twice

Alphix:
I wanted to change the default sshd port of a host, so I set sshd_ports
in the ansible inventory and ran debops. Promptly locked myself out
of the host. Turns out the port needs to be specified twice (in
sshd__ports and sshd__ferm_ports.

This changes the default in keeping with DRY and is (IMHO) a lot
more user-friendly. The only loss is that ferm will by default
use a port number rather than a name.


drybjed (approved):
Good idea. I'll backport it to the stable releases as well.

Issue #1617, opened 2021-01-26
Create DebOps role template using Copier


ypid:
I checked out [Copier](https://github.com/copier-org/copier) today which is the thing I was waiting for when it comes to project template management as it cannot only do the initial creation, but also update instantiated projects as the template involves (using git history). I propose we create a template for how a DebOps role should look like and use it to keep our roles up-to-date in the monorepo. Currently, Copier has a bug with applying template diffs when the template is not the repo root (https://github.com/copier-org/copier/issues/315). Keep this in mind. It has cost me a bit of time. So I suggest we create a separate repo for now and later merge it into the mono repo.

drybjed:
I don't really use any kind of role templating, all of my roles are done from scratch, sometimes basing parts of them here and here from existing ones. I'm not sure how these roles could be templatized by Copier, but if you have some ideas then sure - show an example. :)

ypid:
I will do that when I next write a role from scratch or heavily refactor one :) That is the natural time I think to work on this.

ypid:
Hm, while working with Copier, it did not life up to the quality assumptions I have for a software with version `5.x`. Lets hope this improves and this thing stabilizes a bit. Anyway, I have a draft in #1639. It already offloads a bit of manual work for generating new roles:

* Set the copyright year to the current year.
* Set copyright holder.
* Create role with defaults and tasks.
* Create playbook and move it to the correct place.

For updating roles, it can:

* Keep the `COPYRIGHT` file up-to-date.
* Keep the `meta/defaults.yml` file up-to-date.

(Update mode seems to have issues currently with the fact that the role is not the root of the repo it seems.)

I will leave it at that for now due to a few issues with Copier.

PR #1639, opened 2021-02-14
[docs] Add "Creating a new DebOps role" using Copier dev docs

ypid:
See https://github.com/debops/debops/issues/1617

I guess it could be merged when someone tested this. I probably still should add a hint to the docs, that this is currently just an experiment.

### DIFF
--- a/docs/developer-guide/new-role.rst
+++ b/docs/developer-guide/new-role.rst
@@ -1,0 +1,51 @@
+.. Copyright (C) 2021 Robin Schneider <ypid@riseup.net>
+.. Copyright (C) 2021 DebOps <https://debops.org/>
+.. SPDX-License-Identifier: GPL-3.0-or-later
+
+Creating a new DebOps role
+==========================
+
+Historically, a new role was created by copying an existing role and doing an
+initial search and replace. Which has the following problems:
+
+- Which role is up to the latest standards and thus suitable for creating the next role?
+- Requires manual work to setup a new role skeleton before work can be started (search and replace).
+- For new contributers, this is even more difficult.
+- The role standards are going to evolve thus the roles need to be updated.
+
+DebOps uses `Copier`__ to help address those issues.
+
+  .. __: https://copier.readthedocs.io/en/stable/
+
+Setup
+-----
+
+- Install copier.
+- Clone https://github.com/ypid/debops-template.git next to the DebOps monorepo (:file:`~/.local/share/debops/debops-template`).
+
+Create a new rule
+-----------------
+
+To create a role called ``new_role``, in :file:`~/.local/share/debops/debops/ansible/roles` in the DebOps monorepo, run:
+
+.. code-block:: shell
+
+   copier -d "author_fullname=$(git config --get user.name)" -d "author_email=$(git config --get user.email)" copy ../../../debops-template/ new_role
+
+And answer the questions.
+
+Manual post tasks:
+
+- The generated role has quite a bit of meat on the bone. Just remove the parts you don’t need.
+- Search for ``TODO`` in the generated role skeleton and address them.
+- Integrate the playbook so that it gets called. See how other roles are integrated in :file:`ansible/playbooks/` in the monorepo.
+- Write the actual role ;-)
+
+Update a role to the latest role standards
+------------------------------------------
+
+In :file:`ansible/roles/${my_role}`, run:
+
+.. code-block:: shell
+
+   copier --force
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -111,6 +111,7 @@ infrastructure environments.
    developer-guide/contributing
    developer-guide/contribution-workflow
    developer-guide/contributing-docs
+   developer-guide/new-role
    developer-guide/testing
    developer-guide/development-model
    dep/index


boucman:
Nice, I plan to create y first role pretty soon... I'll definitely play the naive tester for you.

Is the plan to integarte the content of debops-template into the monorepo eventually ?
If yes, maybe the templates could be move to the monorepo in this PR instead of keeping a separate repo ?

docs/developer-guide/new-role.rst
```diff
@@ -0,0 +1,50 @@
...
+
+DebOps uses `Copier`__ to help address those issues.
+
+  .. __: https://copier.readthedocs.io/en/stable/
+
+Setup
+-----
+
+- Install copier.
+- Clone https://github.com/ypid/debops-template.git next to the debops monorepo (or symlink it next to it).
```
boucman: Not sure if cloning is the right way to go here...
The role will probably not be kept in sync with the template repo, it's meant to be it's own thing

maybe give a git-archive command, or point to github's generated tarballs...

docs/developer-guide/new-role.rst
```diff
@@ -0,0 +1,50 @@
...
+
+Create a new rule
+-----------------
+
+In :file:`ansible/roles`, run:
+
+.. code-block:: shell
+
+   copier -d "author_fullname=$(git config --get user.name)" -d "author_email=$(git config --get user.email)" copy ../../../debops-template/ new_role
```
boucman, line 33: I assume new_role is the short name of the new role ? this probably needs to be stated explicitely (esp. if there are naming rules...)

docs/developer-guide/new-role.rst
```diff
@@ -0,0 +1,50 @@
...
+
+And answer the questions.
+
+Manual post tasks:
+
+- Search for ``TODO`` in the generated role skeleton and address them.
+- Integrate the playbook so that it gets called.
```
boucman: How do I do that ? Do I need to modify the monorepo for that ?

docs/developer-guide/new-role.rst
ypid: Please read the Copier docs first ;-)

docs/developer-guide/new-role.rst
ypid: I updated the docs to indicate that this will create a role called `new_role` which should obviously be replaced.

docs/developer-guide/new-role.rst
ypid: Added a hint.

ypid:
> Nice, I plan to create y first role pretty soon... I'll definitely play the naive tester for you.

This makes you the ideal tester than :)

> Is the plan to integarte the content of debops-template into the monorepo eventually ?
If yes, maybe the templates could be move to the monorepo in this PR instead of keeping a separate repo ?

See https://github.com/debops/debops/issues/1617#issue-794558295

docs/developer-guide/new-role.rst
```diff
@@ -0,0 +1,50 @@
...
+Setup
+-----
+
+- Install copier.
+- Clone https://github.com/ypid/debops-template.git next to the debops monorepo (or symlink it next to it).
+
+Create a new rule
+-----------------
+
+To create a role called ``new_role``, in :file:`ansible/roles`, run:
```
boucman: Wouldn't it make more sense to have this run from the root of the debops project tree ?

docs/developer-guide/new-role.rst
```diff
@@ -0,0 +1,50 @@
...
+
+.. code-block:: shell
+
+   copier -d "author_fullname=$(git config --get user.name)" -d "author_email=$(git config --get user.email)" copy ../../../debops-template/ new_role
+
```
boucman, line 34: There is an assumption here about layout, but I'm not sure what it is

in other words, the previous line (about cloning next to the monorepo) is not clear enough. my monorepo is in ~/.local/share/debops
so I went to ~/tmp, made a symlink to my monorepo, cloned debops-template next to it and now I don't have the right layout.... (i'll figure it out, but the instructions need to be more specific

boucman (commented):
Generic remarks 
* Q:Name of project the role manages, in the preferred form.
  That was not clear to me... You assume that the role is about packaging an external project into debops, and that was not my case.
  Once I figured that out, I was good, but that deserves a better explanatory text
* Error on first try
  `mv: impossible de déplacer 'playbook/owadmin.yml' vers '/home/jerros/comite_expertise/adminsys-ecs/ansible/playbooks/service/': N'est pas un dossier` 
  Apparently the directory playbooks/services is not created by debops-init. So you can't assume it's here
* The created playbook imports all sorts of role I didn't need. I'm not sure what the right approch is here... should copier ask ? should they be commented as an example? Not sure... but I had to remove them and nothing told me to review them...
* I'm not sure the license text you generate is ok... Debops does not have a copyright on my code and my code is not part of debops (unless I submit it). It kinda make sense to default to GPLv3+ since that what debops uses so it's probably what the user will need, but other licenses are legal too (especially if the code is not distributed) 


# In the task file
* The role creates all sort of model tasks (cloning git, creating a virtualenv) which are interesting and need to be here somehow, but are not interesting in my case. Removing them was a bit tricky since I needed to figure out what variables to suppress in defaults/main.yml.. Maybe ask what features need to be added one by one in copier ?
* The role seems to rely on various template files (I removed most of them in my case). These were not part of the template.
* I didn't see any `TODO` markers... maybe I missed them ?

docs/developer-guide/new-role.rst
ypid: I thought about that:

Pro root:

* Does not relay on the task trick to move the playbook outside of the generated role scope.

Pro inside role:

* The role directory acts as the "boundary" for Copier where the `.copier-answers.yml` file can be kept. This file cannot be in the root because it should be possible to manage multiple independent roles in the same monorepo. `copier update` needs a path where it can be called.

If you can demonstrate a better way to integrate this into the Monorepo, please do.

docs/developer-guide/new-role.rst
ypid: Should be fixed now.

docs/developer-guide/new-role.rst
boucman: yeah... about that.I was kinda tricked by the copier-answers/yml
In our case, copier is more or less a one-shot thing, you change so much stuff when writing a role that you can't really update using copier

Also, if I want to create mutiple roles, the .copier-answer offers me my previous answers, which is usually not what I want

ypid:
Thanks for your feedback. I hope I addressed it. "The role seems to rely on various template files" will be addressed later.

> Error on first try

Good catch. I was not specific enough in the docs. It needs to be run in the monorepo.

> The created playbook imports all sorts of role I didn't need. I'm not sure what the right approch is here... should copier ask ?

> Maybe ask what features need to be added one by one in copier ?

My take was that deleting is easier than copying together common parts from other roles. One reason for this is that if you copy from other roles, you would have to rename the variables and references from the old role name to the new one. One can forget this (has happened, it is in the git history).

So I left quite a bit of "meat on the bone". It could be made configurable but that I am not sure about maintaining that. Also, what are the common cases? Python stack, LAMP Stack, Ruby on Rails Stack?

> I'm not sure the license text you generate is ok..

Tracked in #1643.

Not sure if we need an option to remove the "This Ansible role is part of DebOps."? The template is specifically for roles that are intended for DebOps so even in "the staging area" the roles could already have to make merging easier.

> The role seems to rely on various template files (I removed most of them in my case).

Good catch. That is work in progress I would say. I guess I will at least add an Ansible local facts skeleton.

> I didn't see any TODO markers... maybe I missed them ?

Seems like it. The generated role has at least one:

```Shell
$ rg TODO
meta/main.yml
32:    - TODO
```

boucman:
yes, I see your point about removing is easier than adding, but it's not trivial because you have to figure out what each variable does and which one you can remove.

I'm not sure if there is a way to get the best of both worlds though. Especially since, indeed, adding variables isn't easy either.

docs/developer-guide/new-role.rst
```diff
@@ -0,0 +1,51 @@
...
+And answer the questions.
+
+Manual post tasks:
+
+- The generated role has quite a bit of meat on the bone. Just remove the parts you don’t need.
+- Search for ``TODO`` in the generated role skeleton and address them.
+- Integrate the playbook so that it gets called. See how other roles are integrated in :file:`ansible/playbooks/` in the monorepo.
```
boucman, line 41: That doesn't really help...
AFAICT debops does not have an easy "plug-in" to add external playbooks (I'm not sure how feasible it would be)

I asked on IRC and the answer was to write my own `site.yml` that would be like upstream's one but would import my playbook.

I think it would make more sense if `site.yml` would import an empty `external.yml` playbook that external users could override...

docs/developer-guide/new-role.rst
```diff
@@ -0,0 +1,51 @@
...
+- Write the actual role ;-)
+
+Update a role to the latest role standards
+------------------------------------------
```
boucman, line 45: Roles are so heavily edited, I'm not sure if any form of automatic upgrade makes sense... that would break way too often.

docs/developer-guide/new-role.rst
ypid: I also have a top level `site.yml` that includes the debops `site.yml` as one part of it if that helps. See https://github.com/ypid/ypid-ansible-common/blob/master/playbooks/ypid-playbooks/site.yml

What would you rather like to see in the docs? Should I check in Copier tasks if `ansible/playbooks` is available and if not, leave the generated playbook in the role directory so you can do what you want with it when you don’t intent to have the role merged into DebOps directly?

docs/developer-guide/new-role.rst
ypid: I had something in the back of my mind when I wrote this. I already excluded heavily edited files from having Copier try to update it using its own mechanism. I agree with you, that would be stupid.

```yaml
_skip_if_exists:
  - defaults/main.yml
  - tasks/main.yml
```

My idea is to use [migrations scripts](https://copier.readthedocs.io/en/latest/configuring/#migrations).

PR #1640, opened 2021-02-14
[apt] Update security suite name

imrejonk:
Debian uses the suffix '-security' instead of '/updates' for its
security repository suites since Debian Bullseye.

Ref: https://www.debian.org/releases/bullseye/errata


drybjed (approved):
Nicely done, thanks! This will be backported to the stable releases as well.

PR #1642, opened 2021-02-15
[sudo] Use @includedir with sudo 1.9.1+

imrejonk:
Sudo 1.9.1 changed the includedir syntax from #includedir to
@includedir. This resulted in a duplicate line in /etc/sudoers after
running the debops.sudo role. The fix is to conditionally add the line
with '#' or '@' based on the version of sudo.


drybjed (approved):
Great work as usual, thank you. I'll backport the fix to the stable releases as well.

PR #1644, opened 2021-02-18
[sssd] Add new role

Alphix:
sssd is a spinoff from the FreeIPA project and provides a modern alternative
to the venerable nslcd/nscd combination. sssd provides additional
features such as offline caching of credentials/users/groups and (optional)
integration with Kerberos, IPA, AD, etc.

This role is mostly in preparation for some Kerberos integration that I'm working on.


drybjed:
Thanks for working on this! I need to review the PR, but I wanted to write about the playbook first. I think that you should create the `bootstrap-sss.yml` playbook as a separate from the `bootstrap-ldap.yml` playbook. The users that prefer `sssd` will probably never ever touch the `nscd` and `nslcd` in their infrastructure, so there's no need for additional complexity here. That's why these roles are not present in the `common.yml` playbook as well. This also means that the `nsswitch__daemon` variable can be dropped entirely.

ansible/playbooks/service/sssd.yml
```diff
@@ -0,0 +1,37 @@
+---
+# Copyright (C) 2019 Maciej Delmanowski <drybjed@gmail.com>
+# Copyright (C) 2019 DebOps <https://debops.org/>
+# Copyright (C) 2021 David Härdeman <david@hardeman.nu>
```
drybjed, line 3: Please update the copyright information in the new files, drop my copyright (I haven't even touched these files yet :-) ) and change the year of the DebOps copyright to 2021. You can also move the DebOps copyright under your own, since you are the initial contributor.

drybjed (changes requested):
Awesome work so far, especially the documentation. I added some comments already, one other thing I would like to see is if you could support more YAML values like integers, booleans and lists in the `sssd.conf` configuration file, via properly handling them in the template. It's not crucial, but it would be nice to have.

I'd like to hear also your ideas about introducing Kerberos support, but that probably can be done in a Discussion thread or on the mailing list.

Alphix:
> Thanks for working on this! I need to review the PR, but I wanted to write about the playbook first. I think that you should create the `bootstrap-sss.yml` playbook as a separate from the `bootstrap-ldap.yml` playbook. The users that prefer `sssd` will probably never ever touch the `nscd` and `nslcd` in their infrastructure, so there's no need for additional complexity here. That's why these roles are not present in the `common.yml` playbook as well. This also means that the `nsswitch__daemon` variable can be dropped entirely.

Ok, I've written patches to create a separate `bootstrap-ldap-sssd.yml` which is basically a copy of `bootstrap-ldap.yml` but with `nslcd`/`nscd` replaced by `sssd`. The naming is intentional, since it would make it easier in the future to change the default by moving the current `bootstrap-ldap.yml` to e.g. `bootstrap-ldap-nslcd.yml` and letting `bootstrap-ldap-sssd.yml` take its place (e.g. with a symlink).

Two questions:
1) Should I push the changes as additional commits or do you prefer a rebase?
2) The `sssd` role currenly has Ansible tasks to uninstall `(u)ns(l)cd`. Should I keep it like this? If so, I think `nslcd` should grow the ability to remove `sssd` to make both roles idempotent?

ansible/playbooks/service/sssd.yml
Alphix: I think the copyright notices are largely correct. Almost every single file in the new service is based on copy-pasta from other existing roles (just as an example, `roles/sssd/templates/etc/sssd/sssd.conf.j2` is based on `roles/rsyslog/templates/etc/rsyslog.conf.j2` with a diff that's about four lines, ignoring comments).

In those cases, I've kept the original copyright notices and added my own since it's a derivative work. Also not clear to me that the order of the copyright notices is relevant, but I'd be happy to change them if it's important.

drybjed:
You can do additional commits in this PR, it's fine.

The `ldap` in `bootstrap-ldap.yml` playbook is from the name of the service used in `/etc/nsswitch.conf`. So I would still go with `bootstrap-sss.yml` as the playbook name; this will also help since there's currently no name completion of playbooks in the shell (maybe it will be added in the future).

Adding support for uninstallation to both `sssd` and `nslcd`/`nscd` might be interesting, sure. You should check out the `dpkg_cleanup` role which can help you with this by preparing `dpkg` hooks that will remove files, revert them, etc. on the package removal. You can check the `nullmailer` role for an example usage.

ansible/playbooks/service/sssd.yml
drybjed: In that case move the DebOps copyright under personal copyrights and update the year to 2021 since the code will be included in the project.

Alphix:
> You can do additional commits in this PR, it's fine.

Ok, done.

> The `ldap` in `bootstrap-ldap.yml` playbook is from the name of the service used in `/etc/nsswitch.conf`. So I would still go with `bootstrap-sss.yml` as the playbook name; this will also help since there's currently no name completion of playbooks in the shell (maybe it will be added in the future).

I've changed the naming.

> Adding support for uninstallation to both `sssd` and `nslcd`/`nscd` might be interesting, sure. You should check out the `dpkg_cleanup` role which can help you with this by preparing `dpkg` hooks that will remove files, revert them, etc. on the package removal. You can check the `nullmailer` role for an example usage.

I'll leave this for a separate pull request later.

ansible/playbooks/service/sssd.yml
Alphix: Ok, I think I've updated all the copyright declarations.

Alphix:
> Awesome work so far, especially the documentation. I added some comments already, one other thing I would like to see is if you could support more YAML values like integers, booleans and lists in the `sssd.conf` configuration file, via properly handling them in the template. It's not crucial, but it would be nice to have.

Not 100% sure I follow. Do you want more of the things are are in the `sssd__default_configuration` mega-variable to be separate configuration variables in `roles/sssd/default/main.yml` (which could then be type-cast as part of the inclusion in `sssd__default_configuration`) or do you want the `roles/sssd/templates/etc/sssd/sssd.conf.j2` to be better at supporting non-string option values?

If you mean the latter, do you have a good example of an existing template which does it?

> I'd like to hear also your ideas about introducing Kerberos support, but that probably can be done in a Discussion thread or on the mailing list.

I'll start a discussion thread and post my two WIP Kerberos roles sometime after `sssd` has been merged. Basically, what I have right now is a `kadmind` role which corresponds roughly to the `slapd` role (i.e. it configures the server-side support). The `kadmind` role uses the `krb5-kdc-ldap` Debian package to store Kerberos principals in the LDAP DIT. The `kerberos` LDAP schema provides both `structural` and `auxiliary` `objectClasses` for Kerberos principal storage, so principals will be added to device/user `dn`s (for "default" principals) or as separate subentries (for additional instances).

Then I have a separate `kerberos` role which acts like the `ldap` role in that it can be used by other roles as a dependent play to add principals and keytabs. So far I've tested it with `slapd` (to offer Kerberized LDAP access via `-Y GSSAPI`) and I'm working on expanding it to `sshd`, `nfs` and `sssd`. `nginx` would also be nice, but the `GSSAPI` support in the Debian version of `nginx` seems to be a sad story.

drybjed:
> > Awesome work so far, especially the documentation. I added some comments already, one other thing I would like to see is if you could support more YAML values like integers, booleans and lists in the `sssd.conf` configuration file, via properly handling them in the template. It's not crucial, but it would be nice to have.
> 
> Not 100% sure I follow. Do you want more of the things are are in the `sssd__default_configuration` mega-variable to be separate configuration variables in `roles/sssd/default/main.yml` (which could then be type-cast as part of the inclusion in `sssd__default_configuration`) or do you want the `roles/sssd/templates/etc/sssd/sssd.conf.j2` to be better at supporting non-string option values?

I mean the latter. The former is and old-style way to have override-able options and is not needed with `parse_kv_config` and `parse_kv_items` filters available in DebOps.

> If you mean the latter, do you have a good example of an existing template which does it?

You can see a good example in the [`journald` configuration file template](https://github.com/debops/debops/blob/master/ansible/roles/journald/templates/etc/systemd/journald.conf.d/ansible.conf.j2). Correctly distinguishing between `0`, `1`, `True` and `False` in Jinja is tricky, but doable. The linked template has a good standardized way to do this, and it helps that it generates an INI-style configuration file. For `sssd` with multiple sections, you can put that filtering loop in a macro and use it repeatedly for each value to parse it.

Does `sssd` support lists of options? The `services` configuration option looks suspiciously like a list to me... If it could be parsed as a list, then we could enable/disable various services in a separate variable programatically.

> I'll start a discussion thread and post my two WIP Kerberos roles sometime after `sssd` has been merged. Basically, what I have right now is a `kadmind` role which corresponds roughly to the `slapd` role (i.e. it configures the server-side support). The `kadmind` role uses the `krb5-kdc-ldap` Debian package to store Kerberos principals in the LDAP DIT. The `kerberos` LDAP schema provides both `structural` and `auxiliary` `objectClasses` for Kerberos principal storage, so principals will be added to device/user `dn`s (for "default" principals) or as separate subentries (for additional instances).

Good to hear that you want to integrate it with LDAP. I read about it some time ago but it's hard to implement properly, I was bummed that you cannot really do multiple principals in one LDAP object, I guess subtrees it will have to be. Do you plan to go with MIT or Heimdal? It seems to me that Heimdal has better support in Debian, especially that you can use [`slapo-smbk5pwd` overlay](https://manpages.debian.org/stretch-backports/slapd-smbk5pwd/slapo-smbk5pwd.5.en.html) to sync passwords between LDAP, Samba and Kerberos.

> Then I have a separate `kerberos` role which acts like the `ldap` role in that it can be used by other roles as a dependent play to add principals and keytabs. So far I've tested it with `slapd` (to offer Kerberized LDAP access via `-Y GSSAPI`) and I'm working on expanding it to `sshd`, `nfs` and `sssd`. `nginx` would also be nice, but the `GSSAPI` support in the Debian version of `nginx` seems to be a sad story.

Interesting. I haven't really worked with a kerberos-based environment so I would have to check it out. The `nginx` package in Debian has support for PAM authentication, so I think that we could enable LDAP auth in it this way, maybe even Kerberos, not sure.

drybjed:
@Alphix I'm currently testing this PR in a development environment, and I encountered an issue with group information - either after some time, or after a host reboot, `sssd` stops showing the proper group membership, for example my `drybjed` account stops being reported as a member of `admins` group. This happens with enumeration both enabled or disabled and breaks other things, for example `sudo` stops working because the admin account is not present in the `admins` group.

Did you encounter that issue with `sssd` and did you found any fixes for this? This won't stop the PR merge in DebOps of course, I just add a warning about this issue in the documentation.

Alphix:
> You can see a good example in the [`journald` configuration file template](https://github.com/debops/debops/blob/master/ansible/roles/journald/templates/etc/systemd/journald.conf.d/ansible.conf.j2). Correctly distinguishing between `0`, `1`, `True` and `False` in Jinja is tricky, but doable. The linked template has a good standardized way to do this, and it helps that it generates an INI-style configuration file. For `sssd` with multiple sections, you can put that filtering loop in a macro and use it repeatedly for each value to parse it.

Interesting, I'll have a look later.

> Does `sssd` support lists of options? The `services` configuration option looks suspiciously like a list to me... If it could be parsed as a list, then we could enable/disable various services in a separate variable programatically.

Yeah, at least `services` and `domains` might be useful to turn into separate yaml lists.

> > I'll start a discussion thread and post my two WIP Kerberos roles sometime after `sssd` has been merged. Basically, what I have right now is a `kadmind` role which corresponds roughly to the `slapd` role (i.e. it configures the server-side support). The `kadmind` role uses the `krb5-kdc-ldap` Debian package to store Kerberos principals in the LDAP DIT. The `kerberos` LDAP schema provides both `structural` and `auxiliary` `objectClasses` for Kerberos principal storage, so principals will be added to device/user `dn`s (for "default" principals) or as separate subentries (for additional instances).
> 
> Good to hear that you want to integrate it with LDAP. I read about it some time ago but it's hard to implement properly, I was bummed that you cannot really do multiple principals in one LDAP object, I guess subtrees it will have to be.

It's not as bad as it sounds because users can still have their default principal as aux properties and most services already have subtree entries below the host entry. So it's really just <user>/<instance> type principals that'll end up as separate leaf entries.

> Do you plan to go with MIT or Heimdal? It seems to me that Heimdal has better support in Debian, especially that you can use [`slapo-smbk5pwd` overlay](https://manpages.debian.org/stretch-backports/slapd-smbk5pwd/slapo-smbk5pwd.5.en.html) to sync passwords between LDAP, Samba and Kerberos.

MIT, just because it's what I'm used to. Not really sure either has better support in Debian. I think `slapo-smbk5pwd` is a bit of a red herring ATM. The reason being that a site which uses LDAP and Kerberos and stuff for their Unix environment would probably also want to run Samba as an AD DC. And if you run Samba4 as an AD DC, it'll only use its own internal LDAP/Kerberos stuff.

I think `FreeIPA`s approach is to do cross-realm trust. Haven't really looked into it TBH.

> > Then I have a separate `kerberos` role which acts like the `ldap` role in that it can be used by other roles as a dependent play to add principals and keytabs. So far I've tested it with `slapd` (to offer Kerberized LDAP access via `-Y GSSAPI`) and I'm working on expanding it to `sshd`, `nfs` and `sssd`. `nginx` would also be nice, but the `GSSAPI` support in the Debian version of `nginx` seems to be a sad story.
> 
> Interesting. I haven't really worked with a kerberos-based environment so I would have to check it out. The `nginx` package in Debian has support for PAM authentication, so I think that we could enable LDAP auth in it this way, maybe even Kerberos, not sure.

LDAP auth, yes. Kerberos, not really (unless I'm severely mistaken). But it's a minor issue, I think Apache has support (`libapache2-mod-auth-kerb`).

Alphix:
> @Alphix I'm currently testing this PR in a development environment, and I encountered an issue with group information - either after some time, or after a host reboot, `sssd` stops showing the proper group membership, for example my `drybjed` account stops being reported as a member of `admins` group. This happens with enumeration both enabled or disabled and breaks other things, for example `sudo` stops working because the admin account is not present in the `admins` group.

Hm. How reproducible? And how often can you hit it? Maybe you can crank up the debug level? (needs to be done per-section in sssd.conf, generates quite a lot of log spew)
 
> Did you encounter that issue with `sssd` and did you found any fixes for this?

Nope, haven't encountered anything like it. I've only done testing using a bunch of VMs so far (though some I've kept around for a week or two at least). And my testing has mostly been limited to making sure that UID/GID mapping works as expected.

> This won't stop the PR merge in DebOps of course, I just add a warning about this issue in the documentation.

Yeah, I meant to say (but forgot all about it) that we should probably slap a great big "experimental, here be dragons" banner on top of the `sssd` role for the time being, whether we find a solution for your problem or not.

drybjed:
> > @Alphix I'm currently testing this PR in a development environment, and I encountered an issue with group information - either after some time, or after a host reboot, `sssd` stops showing the proper group membership, for example my `drybjed` account stops being reported as a member of `admins` group. This happens with enumeration both enabled or disabled and breaks other things, for example `sudo` stops working because the admin account is not present in the `admins` group.
> 
> Hm. How reproducible? And how often can you hit it? Maybe you can crank up the debug level? (needs to be done per-section in sssd.conf, generates quite a lot of log spew)

At the moment just on one host, so I'm not sure if it's reproductible. Since it seems that it works fine for you, perhaps it's something witthin my environment. Let's see if somebody else reports that problem.

I updated the configuration format a bit and changed some settings. The new changes are included in https://github.com/debops/debops/pull/1649, what do you think?

drybjed:
The PR has been merged in https://github.com/debops/debops/pull/1649, closing this one.

Alphix:
Cool. Haven't had time to test your changes, but I've read through them and they all look good to me.

PR #1646, opened 2021-02-23
[ci] Install Cryptography from Debian in Vagrant

imrejonk:
Cryptography is an Ansible dependency and was previously installed as a
Python package. I think we all prefer installing it from the Debian
archive. There's another reason for this since the release of
Cryptography 3.4, because it now requires Rust toolchain 1.45.0+ in
order to build. The Rust toolchain version in Debian 10 is too old for
that.


ypid (approved):
Nice, LGTM.

drybjed (approved):
Good idea, thanks. I'll backport it as well.

PR #1648, opened 2021-02-27
Patchset 5: Minor fixes; Now require the use of debops.extrepo in DEP 3 for upstream repos

ypid:


docs/dep/dep-0003.rst
```diff
@@ -119,10 +119,11 @@ Debian Software Repository, upstream repositories MUST require a condition
 that enables them.  This should give the local system administrators an option
 to install either Debian or upstream version of the package as they see fit.
 Configuration required to enable these repositories (installation of an OpenPGP
-key and configuration of APT sources) should be performed using tools available
-in Ansible, and not using external scripts provided by the vendor. The OpenPGP
-key used to sign the packages MUST be imported by its key fingerprint. Those
-measures should ensure idempotency and reliability.
+key and configuration of APT sources) should be performed using the
+``debops.extrepo`` role. If the software vendor repository is missing
+in the `extrepo default list`__, it should be added to it.
```
drybjed, line 124: This change limits the usefulness of DebOps to Debian only, since `extrepo` currently doesn't provide repository details to other distributions like Ubuntu, Raspbian and the like. Also, `extrepo` at the moment doesn't provide a way to select a different repository mirror (for example NeuroDebian mirrors) or different application version from multiple available (for example Elastic). There is a way to achieve this with `apt-cacher-ng`, but that's more involved than necessary for most cases.

I would add a paragraph after this one that use of Ansible-based repository management tasks is still allowed when needed, but `extrepo` approach should be preferred.

drybjed (approved):
Thanks for removing the hanging spaces. :) I'll add the mentioned note to the DEP file and backport as many commits as feasible.

ypid:
Thanks for maintaining, sounds good to me :)

docs/dep/dep-0003.rst
ypid: I agree with you, "SHOULD" is better for the migration phase. There are still rough edges that need work in extrepo.

> currently doesn't provide repository details to other distributions like Ubuntu

At least `extrepo` is packed for Ubuntu 20.04 and up, but I guess nobody actively decided that, it just entered from Debian :) See that https://packages.ubuntu.com/search?keywords=extrepo&searchon=names&suite=all&section=all

> doesn't provide a way to select a different repository mirror

I would say when someone wants to use a different mirror, then they will likely also run their own cache or mirror anyway. Am I missing something?

Tracked in: https://salsa.debian.org/extrepo-team/extrepo-data/-/issues/11

> different application version from multiple available (for example Elastic)

Tracked in: https://salsa.debian.org/extrepo-team/extrepo-data/-/issues/12

drybjed: > I would say when someone wants to use a different mirror, then they will likely also run their own cache or mirror anyway. Am I missing something?

No, you're right - you explained this use case at the last meeting we had, how `apt-cacher-ng` can be used to provide a local mirror. But still, setting that up is significantly more effort than, for example, switching major release via an inventory variable.

Let's see how `extrepo` developers will deal with these issues, since they will definitely be relevant over time.

PR #1650, opened 2021-02-28
[postldap] Use last mailservice schema

tblein:
The current postldap role default setup is using filter that are not in accordance with the current mailservice.schema.
This try to correct it. 

For users:
- main address from `mailAddress` attribute
- alias addresses from `mailAlternateAddress` attribute
- delivering of email location from `mailHomeDirectory` attribute

For aliases (expected in `postldap__ldap_groups_dn`):
- standalone alias with the `mailAlias` structural object
- attach to existing organizational unit with the `mailDistributionList` auxiliary object

Sorry for the long time to provide it. I did not found the time to implement it completely until lately.

### DIFF
--- a/ansible/roles/postldap/defaults/main.yml
+++ b/ansible/roles/postldap/defaults/main.yml
@@ -332,7 +332,7 @@ postldap__postfix__dependent_lookup_tables:
       (and are responsible for them), the virtual_mailbox_maps configuration is used.
     no_log: '{{ postldap__no_log|d(True) }}'
     connection: '{{ postldap__postfix_ldap_connection }}'
-    search_base: '{{ postldap__ldap_people_dn|join(",") }}'
+    search_base: '{{ postldap__ldap_base_dn|join(",") }}'
     query_filter: '(&
                     (objectClass=mailRecipient)
                     (|
--- a/ansible/roles/postldap/defaults/main.yml
+++ b/ansible/roles/postldap/defaults/main.yml
@@ -139,6 +139,7 @@ postldap__mailbox_base: '/var/vmail'
 # Note: these lookups are recursive.
 postldap__virtual_alias_maps:
   - '$alias_maps'
+  - 'ldap:/etc/postfix/ldap_virtual_groups_maps.cf'
   - 'ldap:/etc/postfix/ldap_virtual_alias_maps.cf'
                                                                    # ]]]
                                                                    # ]]]
@@ -308,6 +309,24 @@ postldap__postfix__dependent_lookup_tables:
                     )
                   )'
     result_attribute: 'mailAddress'
+
+  - name: 'ldap_virtual_groups_maps.cf'
+    state: 'present'
+    comment: |
+      The virtual_groups_maps setting is used to find the delivery addresses of
+      the members of a group (mailAlias or mailDistributionList) (configured in
+      the attribute mailForwardTo).
+    no_log: '{{ postldap__no_log|d(True) }}'
+    connection: '{{ postldap__postfix_ldap_connection }}'
+    search_base: '{{ postldap__ldap_groups_dn|join(",") }}'
+    query_filter: '(&
+                     (|
+                       (objectClass=mailAlias)
+                       (objectClass=mailDistributionList)
+                     )
+                     (mailAddress=%s)
+                   )'
+    result_attribute: 'mailForwardTo'
 
   - name: 'ldap_virtual_mailbox_maps.cf'
     state: 'present'
--- a/ansible/roles/postldap/defaults/main.yml
+++ b/ansible/roles/postldap/defaults/main.yml
@@ -326,7 +326,7 @@ postldap__postfix__dependent_lookup_tables:
     connection: '{{ postldap__postfix_ldap_connection }}'
     search_base: '{{ postldap__ldap_people_dn|join(",") }}'
     query_filter: '(&
-                    (objectClass=inetOrgPerson)
+                    (objectClass=mailRecipient)
                     (|
                       (mailAddress=%s)
                       (mailAlternateAddress=%s)
@@ -365,7 +365,7 @@ postldap__postfix__dependent_lookup_tables:
     search_base: '{{ postldap__ldap_base_dn|join(",") }}'
     query_filter: '(|
                     (&
-                      (objectClass=inetOrgPerson)
+                      (objectClass=mailRecipient)
                       (|
                         (mailAddress=%s)
                         (mailAlternateAddress=%s)
@@ -429,7 +429,7 @@ postldap__postfix__dependent_lookup_tables:
     search_base: '{{ postldap__ldap_base_dn|join(",") }}'
     query_filter: '(|
                     (&
-                      (objectClass=inetOrgPerson)
+                      (objectClass=mailRecipient)
                       (|
                         (mailAddress=%s)
                         (mailAlternateAddress=%s)
--- a/ansible/roles/postldap/defaults/main.yml
+++ b/ansible/roles/postldap/defaults/main.yml
@@ -118,13 +118,23 @@ postldap__vmail_posix_gidnumber: '{{ ansible_local.postldap.vmail_posix_gidnumbe
 # All mailboxes are stored directly in the file system of the server.
 #
 # This setting describes the prefix that the virtual delivery agent prepends to all pathname results from
-# `$virtual_mailbox_maps` table lookups.
+# ``ldap_virtual_mailbox_maps`` table lookups.
 # The directories under this base contain the users MailDirs.
 # This directory must be created in advance and must be writeable by all potential mail users.
 # To achieve this, either use the same GID across all mail users and set group ownership to that GID,
 # or make the directory world writable.
 postldap__mailbox_base: '/var/vmail'
 
+                                                                   # ]]]
+# .. envvar:: postldap__virtual_mailbox_maps_attribute [[[
+#
+# The attributes needed to reconstruct the mailbox path use by the ``ldap_virtual_mailbox_maps`` lookup table.
+postldap__virtual_mailbox_maps_attribute: 'uid'
+                                                                   # ]]]
+# .. envvar:: postldap__virtual_mailbox_maps_format [[[
+#
+# The relative path format to the user mailbox use by the ``ldap_virtual_mailbox_maps`` lookup table.
+postldap__virtual_mailbox_maps_format: '/%u/Maildir/'
                                                                    # ]]]
 # .. envvar:: postldap__virtual_alias_maps [[[
 #
@@ -321,7 +331,6 @@ postldap__postfix__dependent_lookup_tables:
     comment: |
       As we only want to accept mail, where we know the recipients
       (and are responsible for them), the virtual_mailbox_maps configuration is used.
-      The path for the mailbox is retrieved from the mailHomeDirectory attribute
     no_log: '{{ postldap__no_log|d(True) }}'
     connection: '{{ postldap__postfix_ldap_connection }}'
     search_base: '{{ postldap__ldap_people_dn|join(",") }}'
@@ -336,7 +345,8 @@ postldap__postfix__dependent_lookup_tables:
                       (authorizedService=mail:receive)
                     )
                   )'
-    result_attribute: 'mailHomeDirectory'
+    result_attribute: '{{ postldap__virtual_mailbox_maps_attribute }}'
+    result_format: '{{ postldap__virtual_mailbox_maps_format }}'
 
   - name: 'ldap_virtual_mailbox_domains.cf'
     state: 'present'
--- a/ansible/roles/postldap/defaults/main.yml
+++ b/ansible/roles/postldap/defaults/main.yml
@@ -312,7 +312,6 @@ postldap__postfix__dependent_lookup_tables:
                        (objectClass=mailDistributionList)
                        (objectClass=mailRecipient)
                      )
-                     (objectClass=mailRecipient)
                      (|
                        (mailAddress=%s)
                        (mailAlternateAddress=%s)
@@ -338,7 +337,6 @@ postldap__postfix__dependent_lookup_tables:
                     (objectClass=mailRecipient)
                     (|
                       (mailAddress=%s)
-                      (mailAlternateAddress=%s)
                     )
                     (|
                       (authorizedService=all)
--- a/ansible/roles/postldap/defaults/main.yml
+++ b/ansible/roles/postldap/defaults/main.yml
@@ -371,28 +371,26 @@ postldap__postfix__dependent_lookup_tables:
     no_log: '{{ postldap__no_log|d(True) }}'
     connection: '{{ postldap__postfix_ldap_connection }}'
     search_base: '{{ postldap__ldap_base_dn|join(",") }}'
-    query_filter: '(|
-                    (&
-                      (objectClass=mailRecipient)
-                      (|
-                        (mailAddress=%s)
-                        (mailAlternateAddress=%s)
-                      )
-                      (|
-                        (authorizedService=all)
-                        (authorizedService=mail:send)
-                      )
-                    )
-                    (&
-                      (objectClass=account)
-                      (uid=%u)
-                      (host=%d)
-                      (|
-                        (authorizedService=all)
-                        (authorizedService=mail:send)
-                      )
-                    )
-                  )'
+    query_filter: '(&
+                     (|
+                       (&
+                         (objectClass=mailRecipient)
+                         (|
+                           (mailAddress=%s)
+                           (mailAlternateAddress=%s)
+                         )
+                       )
+                       (&
+                         (objectClass=account)
+                         (uid=%u)
+                         (host=%d)
+                       )
+                     )
+                     (|
+                       (authorizedService=all)
+                       (authorizedService=mail:send)
+                     )
+                   )'
     result_attribute: ''
     special_result_attribute: 'member'
     leaf_result_attribute: 'uid'
--- a/ansible/roles/postldap/defaults/main.yml
+++ b/ansible/roles/postldap/defaults/main.yml
@@ -344,7 +344,6 @@ postldap__postfix__dependent_lookup_tables:
                        (authorizedService=mail:receive)
                      )
                    )'
-    result_attribute: ''
     special_result_attribute: 'member'
     leaf_result_attribute: 'mailAddress'
 
@@ -438,7 +437,6 @@ postldap__postfix__dependent_lookup_tables:
                        (authorizedService=mail:send)
                      )
                    )'
-    result_attribute: ''
     special_result_attribute: 'member'
     leaf_result_attribute: 'uid, mailAddress, mailAlternateAddress'
     size_limit: 1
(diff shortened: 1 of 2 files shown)


drybjed:
@tblein: You might want to see [this thread on the `debops-devel` mailing list](https://lists.debops.org/hyperkitty/list/debops-devel@lists.debops.org/thread/LX4TCWSKS4HAOLUURRAOSDLJJMDHKUU6/) to see another production setup and compare notes.

drybjed:
I think that the `inetOrgPerson` object class should be changed to `mailRecipient` in various filters. It will enforce the relationship with its attributes even more, and lets users create mail accounts from different object classes as well.

tblein:
> @tblein: You might want to see [this thread on the `debops-devel` mailing list](https://lists.debops.org/hyperkitty/list/debops-devel@lists.debops.org/thread/LX4TCWSKS4HAOLUURRAOSDLJJMDHKUU6/) to see another production setup and compare notes.

Since I wasn't register to the `debops-devel` I didn't see it. I will read the thread (and register to the list).

@imrejonk feel free to comment/correct my draft proposition: I have few users on my system and can adapt to a more general setup.

imrejonk:
@tblein Thanks for the PR! I've been wanting to improve this role as well. I'll review your changes tomorrow.

ansible/roles/postldap/defaults/main.yml
```diff
@@ -291,45 +292,63 @@ postldap__postfix__dependent_lookup_tables:
...
-                      (mail=%s)
-                      (mailAlias=%s)
-                    )
-                    (|
-                      (authorizedService=all)
-                      (authorizedService=mail:receive)
-                    )
-                  )'
-    result_attribute: 'mail'
+                     (objectClass=inetOrgPerson)
```
imrejonk: This should be changed to the `mailRecipient` object class.

ansible/roles/postldap/defaults/main.yml
```diff
@@ -291,45 +292,63 @@ postldap__postfix__dependent_lookup_tables:
   - name: 'ldap_virtual_alias_maps.cf'
     state: 'present'
     comment: |
-      The virtual_alias_maps setting is used to find the target mailbox
-      (configured in the attribute mail, aka the mailbox or maildrop).
+      The virtual_alias_maps setting is used to find the final delivery address
+      (configured in the attribute mailAddress).
     no_log: '{{ postldap__no_log|d(True) }}'
     connection: '{{ postldap__postfix_ldap_connection }}'
     search_base: '{{ postldap__ldap_people_dn|join(",") }}'
```
imrejonk: This should be changed to `postldap__ldap_base_dn`, to allow service accounts (under the ou=Hosts tree) to have aliases as well.

ansible/roles/postldap/defaults/main.yml
```diff
@@ -291,45 +292,63 @@ postldap__postfix__dependent_lookup_tables:
...
+    connection: '{{ postldap__postfix_ldap_connection }}'
+    search_base: '{{ postldap__ldap_groups_dn|join(",") }}'
+    query_filter: '(&
+                     (|
+                       (objectClass=mailAlias)
+                       (objectClass=mailDistributionList)
+                     )
+                     (mailAddress=%s)
+                   )'
+    result_attribute: 'mailForwardTo'
```
imrejonk, line 372: I have experimented with virtual_alias_maps and LDAP groups a lot lately. As I didn't like the idea of having separate LDAP lookups for aliases and groups (which can also be used together, e.g. an alias for a group), I came up with the following lookup table that solves all problems I could come up with:
```
- name: 'ldap_virtual_alias_maps.cf'
  state: 'present'
  comment: | 
    The virtual_alias_maps setting is used to find the final delivery
    addresses, given an alias or distribution list address.
  no_log: '{{ postldap__no_log|d(True) }}'
  connection: '{{ postldap__postfix_ldap_connection }}'
  search_base: '{{ postldap__ldap_base_dn|join(",") }}'
  query_filter: '(&
                   (|
                     (objectClass=mailAlias)
                     (objectClass=mailDistributionList)
                     (objectClass=mailRecipient)
                   )
                   (|
                     (mailAddress=%s)
                     (mailAlternateAddress=%s)
                   ) 
                   (|
                     (authorizedService=all)
                     (authorizedService=mail:receive)
                   ) 
                 )'
  result_attribute: 'mailForwardTo'
  special_result_attribute: 'member'
  leaf_result_attribute: 'mailAddress'
```
What's nice about it:
- The number of LDAP lookups and lookup tables is limited.
- It works with mailAlternateAddress attributes in personal and group objects.
- For group objects, the right-hand side of the table consists of the mailForwardTo attributes, as well as mailAddress attributes in member objects. There is no need to copy email addresses to multiple objects, just add the personal object as a member. If you need to add recipients that don't have an LDAP object, just add them as mailForwardTo attributes.

ansible/roles/postldap/defaults/main.yml
```diff
@@ -291,45 +292,63 @@ postldap__postfix__dependent_lookup_tables:
...
+                      (mailAlternateAddress=%s)
                     )
                     (|
                       (authorizedService=all)
                       (authorizedService=mail:receive)
                     )
                   )'
-    result_attribute: 'mail'
-    result_format: '/%d/%u/Maildir/'
+    result_attribute: 'mailHomeDirectory'
```
imrejonk: I personally prefer this:
```
result_attribute: 'uid'
result_format: '/%u/Maildir/'
```
We should probably make this configurable.

ansible/roles/postldap/defaults/main.yml
```diff
@@ -360,8 +379,8 @@ postldap__postfix__dependent_lookup_tables:
                     (&
                       (objectClass=inetOrgPerson)
```
imrejonk: This should be changed to the `mailRecipient` object class.

ansible/roles/postldap/defaults/main.yml
```diff
@@ -378,7 +397,7 @@ postldap__postfix__dependent_lookup_tables:
                       )
                     )
                   )'
-    result_attribute: 'uid, mail, mailAlias'
+    result_attribute: 'uid, mailAddress, mailAlternateAddress'
```
imrejonk: This lookup should return uid's that are allowed to send mail from a specified email address. It should not return email addresses.

I think it would be desirable if a member of an LDAP group is also allowed to send mail from the group address. Suppose you have a group with mailAddress/mailAlternateAddress 'group@example.com' and a member with uid 'user' and mailAddress/mailAlternateAddress 'user@example.com', then this lookup should return 'user' when querying either 'group@example.com' or 'user@example.com'.

I solved this by using these parameters:
```
result_attribute: ''
special_result_attribute: 'member'
leaf_result_attribute: 'uid'
```

imrejonk (changes requested):

ansible/roles/postldap/defaults/main.yml
tblein: To have it we need to set two variables one for `result_attribute` and one for `result_format`. Am I right ?
In the documentation we could give examples to cover the three discussed possibilities:

- `/var/vmail/<uid>/Maildir `
- `/var/vmail/<domain>/<localpart>/Maildir `
- `<mailHomeDirectory>`

If you have any suggestion of how to do it properly I am open to suggestions.

Is it possible to have a mix ? If mailHomeDirectory is defined for the user, use it, otherwise use a scheme. I do not know enough of LDAP to know whether it is easy to implement.

drybjed: In theory you could have two separate Postfix lookup tables, one with `(mailHomeDirectory=*)`, another with a different filter. But that's probably too complex for our purposes, I would keep just two separate variables to define the attribute and format. As for the variable names, how about `postldap__virtual_mailbox_maps_attribute`, `postldap__virtual_mailbox_maps_format` with mention that these are specifically for the `ldap_virtual_mailbox_maps.cf` lookup table.

tblein: Here it is, both as variables. As default, I kept imrejonk settings that might be more common.

Last question for this point : where do we set the corresponding documentation for the different use case ? I see at least three possibilities:
- in the `getting-started` page in a paragraph under "Overriding lookup table configuration";
- in the `guides` page
- close to the variable definition in a paragraph before/after.

ansible/roles/postldap/defaults/main.yml
```diff
@@ -291,23 +301,30 @@ postldap__postfix__dependent_lookup_tables:
...
-                      (authorizedService=mail:receive)
-                    )
-                  )'
-    result_attribute: 'mail'
+                     (|
+                       (objectClass=mailAlias)
+                       (objectClass=mailDistributionList)
+                       (objectClass=mailRecipient)
+                     )
+                     (objectClass=mailRecipient)
```
imrejonk: That is one `(objectClass=mailRecipient)` too many ;)

ansible/roles/postldap/defaults/main.yml
```diff
@@ -318,18 +335,18 @@ postldap__postfix__dependent_lookup_tables:
     connection: '{{ postldap__postfix_ldap_connection }}'
     search_base: '{{ postldap__ldap_people_dn|join(",") }}'
```
imrejonk: This search base should also be `postldap__ldap_base_dn`, so that service accounts (under the ou=Hosts tree, for example) can have mailboxes as well.

ansible/roles/postldap/defaults/main.yml
```diff
@@ -318,18 +335,18 @@ postldap__postfix__dependent_lookup_tables:
...
     query_filter: '(&
-                    (objectClass=inetOrgPerson)
+                    (objectClass=mailRecipient)
                     (|
-                      (mail=%s)
-                      (mailAlias=%s)
+                      (mailAddress=%s)
+                      (mailAlternateAddress=%s)
```
imrejonk: This part:
```
(|
  (mailAddress=%s)
  (mailAlternateAddress=%s)
)
```
can be replaced with:
```
(mailAddress=%s)
```
because aliases are already handled by the ldap_virtual_alias_maps.cf lookup table.

ansible/roles/postldap/defaults/main.yml
```diff
@@ -378,7 +395,9 @@ postldap__postfix__dependent_lookup_tables:
                       )
                     )
                   )'
```
imrejonk: The query_filter attribute of the ldap_smtpd_sender_login_maps.cf lookup table can be deduplicated a little bit. The authorizedService matches are in there twice right now. I propose this filter:
```
query_filter: '(&
                 (|
                   (&
                     (objectClass=mailRecipient)
                     (|
                       (mailAddress=%s)
                       (mailAlternateAddress=%s)
                     ) 
                   ) 
                   (&
                     (objectClass=account)
                     (uid=%u)
                     (host=%d)
                   ) 
                 )
                 (|
                   (authorizedService=all)
                   (authorizedService=mail:send)
                 )
               )'
```

imrejonk (changes requested):

ansible/roles/postldap/defaults/main.yml
imrejonk: Thanks for updating the role, @tblein. Are the changes I'm proposing still compatible with your setup?

About documenting the mailbox location possibilities: I think these should go in the 'default variables' section. The getting started guide should be updated to reflect the new role defaults.

ansible/roles/postldap/defaults/main.yml
```diff
@@ -118,13 +118,23 @@ postldap__vmail_posix_gidnumber: '{{ ansible_local.postldap.vmail_posix_gidnumbe
 # All mailboxes are stored directly in the file system of the server.
 #
 # This setting describes the prefix that the virtual delivery agent prepends to all pathname results from
-# `$virtual_mailbox_maps` table lookups.
+# ``ldap_virtual_mailbox_maps`` table lookups.
```
drybjed: You should revert this change. The `$virtual_mailbox_maps` variable is a Postfix configuration option, not a specific table. The maibox base specified here will be added to the mailbox paths from any lookup table specified here, not just LDAP (think MySQL, or other databases).

drybjed (commented):
The proposed changes look good to me. There are some cosmetic issues with documentation, but I can fix that before merging the PR. I'll wait for the other reviewers to accept the changes before merging. Since there's a lot of them, these will not be backported to the stable releases.

ansible/roles/postldap/defaults/main.yml
tblein: @imrejonk  your changes are perfectly compatible with my setup. Thanks for your concern.

In the same time the final delivery of the emails on my system are not handled by postfix but by dovecot (through lmtp) to have the sieve filters. Therefore, the mail home is set in dovecot settings.

tblein:
On my side, I am not  sure to be able to work on the documentation before next week. The end of the week is quite busy on my side. @drybjed if you do not have time to do it, I will try to find time to make some proposition next week.

ansible/roles/postldap/defaults/main.yml
tblein: I did not test it enough.

I have an entry like the following one :
```
dn: ou=IT Department,ou=Groups,dc=tblein,dc=eu
objectClass: organizationalUnit
objectClass: mailDistributionList
objectClass: authorizedServiceObject
authorizedService: all
ou: IT Department
mail: itdept@mydomain.net
mailAddress: itdept@mydomain.net
mailForwardTo: admin1@mydomain.net
mailForwardTo: admin2@mydomain.net
```

The lookup table returns also the `mailAddress` of the group:

```
# postmap -q itdept@mydomain.net ldap:/etc/postfix/ldap_virtual_alias_maps.cf

itdept@mydomain.net,admin1@mydomain.net,admin2@mydomain.net
```

Therefore, it ends in a delivery error to the itdept@mydomain.net, while the mail is correctly delivered to the other addresses.

@imrejonk any idea ?

drybjed: Mail aliases forward e-mail messages from an e-mail address (`itdept@mydomain.net`) to other e-mail addresses (`admin1@mydomain.net`, `admin2@mydomain.net`).

Mail accounts accept e-mail messages directed to them (`itdept@mydomain.net`).

You cannot combine the two cases together in one Postfix lookup, otherwise you get a loop where mail for `itdept@mydomain.net` is fowarded to `itdept@mydomain.net`. This is not an issue on the LDAP side, but on the Postfix side, mail aliases need to be looked up separately from normal e-mail addresses.

ansible/roles/postldap/defaults/main.yml
imrejonk: That's good to hear!

Regarding LMTP delivery to Dovecot: we do that as well (for the same reason as you, sieve filtering), so I guess that the virtual_mailbox_maps setting in my main.cf doesn't actually do anything. See [this blog post](https://www.ciphermail.com/blog/eating-our-own-dog-food.html) if you're interested in the mail setup at my place of work.

We set Dovecot's mail_home like this: `dovecot_vmail_home: '/var/vmail/%u'`

imrejonk (dismissed):
These changes look good to me, apart from the documentation. The role is now very similar to the customized version I currently have in use. I will open one or more PR's after this one has been merged to propose some other (minor) changes.

ansible/roles/postldap/defaults/main.yml
tblein: When you use authentication with the email address it is no more working since there is a mismatch between the login used (the email address) and the return `uid` and postfix refuse to send the message.

Therefore, `ldap_smtpd_sender_login_maps` lookup table should return `mailAddress` and `mailAlternateAddress`.

tblein:
I tested more in depth the update in my environment and found some problems:

- `mailForwardTo` problem with the suggestion of @drybjed to use two different lookups as I proposed in the beginning. @imrejonk is not working in my case.
- authentication with email addresses is not working. Difficult to pinpoint, since there is an additional bug introduced by 59a59d3e3730367ec65fc6b83f698fa5876c0df9 (`%u` should be `%U`). 

I will propose updates in the beginning of next week or during the weekend, and a bug report for `saslauthd` if not solved.

drybjed:
@tblein What's the difference between `%u` and `%U` in saslauthd, changed case? I tried to find some details in the documentation, but I'm unsuccessful so far.

tblein:
According to https://blog.sys4.de/cyrus-sasl-saslauthdconf-man-page-en.html
```
%u
    %u is replaced by the complete user string.
%U
    If the string is an address (%u), %U will be replaced by the local part of that address.
```

So in our case the %u will return the full address (user@domain.net) and %U only the first part (user).

ansible/roles/postldap/defaults/main.yml
```diff
@@ -416,7 +416,7 @@ postldap__postfix__dependent_lookup_tables:
                    )'
     result_attribute: ''
     special_result_attribute: 'member'
-    leaf_result_attribute: 'uid'
+    leaf_result_attribute: 'uid, mail'
```
tblein: This to allow the use of the email as login. Therefore, I propose to use the `mail` attribute since it is authentication. However, It might be better to use `mailAddress` and `mailAlternateAddress`.

drybjed: Agreed, I think that `mailAddress` and `mailAlternateAddress` might be good here. I'm thinking about adding a new attribute in the schema, `mailReceiveAddress`, for read-only or historical addresses of a given user. These should not be allowed to be used as logins, so having only the two mentioned attributes would work well with that.

Conversely, `saslauthd` role might need to be updated to allow SMTP AUTH using these attributes instead of `inetOrgPerson`, but that probably be done in a separate PR.

ansible/roles/postldap/defaults/main.yml
```diff
@@ -321,10 +320,34 @@ postldap__postfix__dependent_lookup_tables:
                        (authorizedService=mail:receive)
                      )
                    )'
-    result_attribute: 'mailForwardTo'
+    result_attribute: ''
```
drybjed: I'm not sure if an empty `result_attribute` needs to be set? What happens if it's omitted entirely from the lookup table?

tblein: If I remove the corresponding line in the final lookup file, it is still retrieving working without complaint using `postmap`.
It should be safe to remove all the empty `result_attribute` entries.

tblein:
Hello

Anything to do to improve the proposition ?
Thanks for your feedbacks.

drybjed:
@tblein I think that the PR is fairly complete, I just need to test it before merging. I'll try to do it over the weekend.

drybjed:
@tblein I finally tested your proposed changes in a development environment. Awesome work, hoever I decided to switch the mailbox layout to `/%d/%u/Maildir` to keep it synchronized with the current defaults in the `dovecot` role. I also cleaned up the comments a bit.

I also found an issue with allowing the `mailAddress` to be used for authentication - if an user account has more e-mail addresses, using each one for authentication creates a "separate account" in Roundcube, each one seeing the same mail storage though. I'm not sure how to deal with that yet, I think this is an issue between Dovecot and Roundcube, not related to Postfix. Something to experiment in the future, I suppose.

tblein:
@drybjed Thanks for the inclusion.

Good point for Roundcube. Certainly something to do on Roundcube/Dovecot side.

PR #1652, opened 2021-03-02
Correct section tag for docker.tmpfs

anzil:
- ensure use of correct section keyword (`runners.docker.tmpfs` instead of `runners.docker_tmpfs`), see https://docs.gitlab.com/runner/executors/docker.html#mounting-a-directory-in-ram
- add space before =


drybjed (approved):
Thanks for the fix. I'll backport this to the stable releases as well.

PR #1656, opened 2021-03-08
Saslauthd uid correction

tblein:
A small confusion between the '%u' and '%U' token that was preventing the identification of sender using email addresses.

https://blog.sys4.de/cyrus-sasl-saslauthdconf-man-page-en.html

At the same time I propose to restrict the LDAP lookup for email sender to be closer to the mailservice schema: look only for `mailAddress` and `mailAlternateAddress` of `mailRecipient` object class.


drybjed (approved):
Thanks for finding the error and providing a fix. I'll backport it to the stable releases as well.

PR #1659, opened 2021-03-14
Fix debops-default not finding playbooks in <projectdir>/debops

boucman:
When cloning debops as a subdirectory of a <projectdir> it is expected that all command line tools will search in that
subdirectory.

the find_playbook function did not search for playbooks in that directory, making debops-defaults non functional in that case.


drybjed (approved):
Thanks for the bugfix, I'll backport it to stable releases. The `ansible-lint` issues should be already fixed in the `master` branch.

PR #1660, opened 2021-03-14
Correctly react to various ways python files are reported by "file"

boucman:
Under debian/testing, python files are reported as text/x-script.python instead of test/x-python

this breaks check-pep8. this fixes allows both mime types to work

related debian bug : https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=985209


drybjed (approved):
Thanks for the fix, I'll backport it to the stable releases. The `ansible-lint` errors should be fixed now in the `master` branch so I'll merge this PR.

PR #1661, opened 2021-03-17
add lxc__default_container_backing_store variable

boucman:
The backing store is a parameter that is used only on the lxc-create CLI
As such it is not possible to set a default value via lxc.conf or
any other mechanism on the LXC side

This provides a new role-global variable to set a default for the backing
store


drybjed (approved):
Sounds good, thanks. This will be backported to previous releases.

Issues found by `ansible-lint` should be fixed in the `master` branch, so after checking locally I'll merge the PR.

boucman:
Awesome, thx a lot

PR #1667, opened 2021-03-24
fix to support dependent roles

kosli:
it should be probably be fixed in other roles too (e.g. users)


drybjed (approved):
Looks good, thanks! I'll backport the fix as well.

PR #1670, opened 2021-04-02
Update netboot and firmware version in ipxe role variables

velvetant:


drybjed (approved):
Thanks for the update! I'll backport it to the previous stable releases as well.

PR #1673, opened 2021-04-12
Allow configuration of autopostgresql backup permissions

Cadair:
This seems to work locally for me.


drybjed (approved):
Looks good, thanks. I'll backport the change to the stable releases as well.

Since the backup script is executed as `root`, you could also add a custom group to `chgrp` the finished files into, and do it only if it's different than `root`.

Issue #1637, opened 2021-02-10
[debops.elasticsearch] ElasticSearch 7.x Unrecognized VM option 'UseConcMarkSweepGC'


alinalexandru:
```
Unrecognized VM option 'UseConcMarkSweepGC'
Error: Could not create the Java Virtual Machine.
Error: A fatal exception has occurred. Program will exit.
```
It seems to be related to java 11.
I will submit a PR soon. 

What version is debops supporting?
https://www.elastic.co/support/matrix#matrix_jvm

drybjed:
At the moment the `elasticsearch` role is configured to support v6.x, but I want to switch the repository management to the `extrepo` role, which will result in installation of Elasticsearch 7.x by default (or newer, later on).

The JVM configuration options can now be configured using a `conf.d/` style directory, so I would switch the role to use that mechanism instead of replacing the whole file. Although I'd like to keep support for the v6.x releases for now, and I'm not sure if `conf.d/` can be used there; most likely not. I plan to refresh the Elastic roles in the current release cycle, so I'll take a look at this soon.

drybjed:
The issue should be fixed by https://github.com/debops/debops/pull/1678, closing.

PR #1678, opened 2021-04-18
Refresh Elastic stack (Elasticsearch, Kibana, Filebeat)

drybjed:


alinalexandru:
Probably https://github.com/debops/debops/issues/1637 is also fixed. Right?

drybjed:
Yes, this should be fixed. I'll close the issue, thanks for remembering it. :-)

PR #1683, opened 2021-04-21
better solution: no more warnings here

stefangweichinger:


drybjed (commented):
These changes essentially revert the previous ones. The `locale.setlocale()` function sets the environment variables for the current process (script), but it doesn't do the same for the subprocess started by the script. That's why you had to change the selector to `[1]`, becaue the `sudo --version` command was executed, again, with `de_DE.UTF-8` locale you are probably using.

I've made the https://github.com/debops/debops/pull/1685 fix which should deal with locales in a more Pythonic way and use them correctly with `subprocess` call.

PR #1686, opened 2021-04-24
[netbox] Upgrade to NetBox v2.11.1 

ypid:
Related to: https://github.com/netbox-community/netbox/pull/6272


drybjed (approved):
Thanks for the upkeep! I tested the upgrade from the current version of NetBox in DebOps and it seems to work fine.

I'll bump up the version to `v2.11.2` since it was released a few days ago.

PR #1687, opened 2021-04-24
ci: Switch to Conventional Commits convention

ypid:
See https://www.conventionalcommits.org/en/v1.0.0/

I have been using this for a few weeks now and like it. At least it is standardized.


drybjed:
It is interesting, however I would like to see the linting implemented locally instead of relying on GitHub. That way we could maybe do it in a pre-commit hook at some point. Any interesting tools in PyPI or preferably Debian itself that can lint commits? Looks like `gitlint` package will be available in Debian Bullseye, we could probably add configuration for it in the repository.

ypid:
Seems I did not write much context but I am happy you already like the general idea of it. Maybe lets start with a soft migration phase, to get contributors to adapt from our way of writing commits, later it can be a hard requirement with it being mentioned in the docs.

I also don’t like the way of testing either, but I did not find another tool. I am sure I will come across one. I just wanted to propose it in some machine readable way.

Maybe lets start with  #1691 and leave this PR open until I found a better linting tool and we defined the "scopes" and "types" and so on?

drybjed:
Sure, we can go with a soft approach, not requiring the `gitlint` test to pass for the moment. But it can still be added and when we ask people about the commits they can check if the test passes locally.

ypid:
I had a second look at `gitlint` now, looks like it is flexible enough to use it :) And it is in Debian :)

PR #1699, opened 2021-05-03
add letsencrypt API V2 to pki role, add more information on usage out…

prk0ghy:
In this commit I added the new letsencrypt V2 API. I could not find it in Master nor in stable-2.2


drybjed (approved):
Nice work, thanks. :-)

PR #1705, opened 2021-05-05
LXC-4.x: do not drop sys_admin capability

stefangweichinger:
This is a draft only, to document things.
I still have to do more tests with fresh VMs, both buster and bullseye.

### DIFF
--- a/ansible/roles/lxc/tasks/main.yml
+++ b/ansible/roles/lxc/tasks/main.yml
@@ -340,6 +340,21 @@
   lineinfile:
     path: '/var/lib/lxc/{{ item.name | d(item) }}/config'
     regexp: '^lxc\.mount\s+=\s+'
+    line: 'lxc.mount = /var/lib/lxc/{{ item.name | d(item) }}/fstab'
+    insertafter: '^lxc\.rootfs\.backend\s+=\s+'
+    state: 'present'
+    mode: '0644'
+  loop: '{{ lxc__containers }}'
+  when: ansible_local|d() and ansible_local.lxc|d() and
+        (item.name | d(item)) not in ansible_local.lxc.containers|d() and
+        item.state|d("started") == "started" and item.fstab|d() and
+        lxc__version is version("4.0", operator="lt" strict=True)
+  tags: [ 'role::lxc:containers' ]
+
+- name: Include fstab parameter in LXC container configuration
+  lineinfile:
+    path: '/var/lib/lxc/{{ item.name | d(item) }}/config'
+    regexp: '^lxc\.mount\.fstab\s+=\s+'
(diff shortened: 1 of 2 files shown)


ansible/roles/lxc/tasks/main.yml
```diff
@@ -340,7 +340,7 @@
   lineinfile:
     path: '/var/lib/lxc/{{ item.name | d(item) }}/config'
     regexp: '^lxc\.mount\s+=\s+'
-    line: 'lxc.mount = /var/lib/lxc/{{ item.name | d(item) }}/fstab'
+    line: 'lxc.mount.fstab = /var/lib/lxc/{{ item.name | d(item) }}/fstab'
```
drybjed, line 358: This change should probably be versioned, LXC 3.x supports the `lxc.mount` variant, and 4.x supports the `lxc.mount.fstab` variant. Also, you might need to update the regexp accordingly.

stefangweichinger: Ack. I will try to come up with some useful suggestion.

boucman:
why do we systematically need cap_sys_admin ?

stefangweichinger:
> why do we systematically need cap_sys_admin ?

I can't tell .. I just found that LXC-4.x seems to need that change (with Debian 11.x "Bullseye").

stefangweichinger:
Did a quick test with 2 target hosts "deb10" and "deb11" running "buster" and "bullseye".
Created 2 lxc-containers "deb10_fstabtest" and "deb11_fstabtest" on them, with host_vars:

```
lxc__containers:

  - name: 'deb10_fstabtest'
    distribution: 'debian'
    release: 'buster'
    ssh: False
    fstab: |
      /srv/shared/lxc-opt opt none bind 0 0
```

and

```
lxc__containers:

  - name: 'deb11_fstabtest'
    distribution: 'debian'
    release: 'bullseye'
    ssh: False
    fstab: |
      /srv/shared/lxc-opt opt none bind 0 0
```

# result

```
deb10 # tail -1 /var/lib/lxc/deb10_fstabtest/config
lxc.mount = /var/lib/lxc/deb10_fstabtest/fstab

deb11 # tail -1 /var/lib/lxc/deb11_fstabtest/config 
lxc.mount.fstab = /var/lib/lxc/deb11_fstabtest/fstab
```

stefangweichinger:
https://discuss.linuxcontainers.org/t/lxc-4-0-6-in-debian-11-x-lxc-cap-drop-sys-admin/11100

drybjed (approved):
Well, there's no news after two weeks, so I guess we can go with this change for now since it looks like this will be required for Debian Bullseye. I'll backport this PR to the stable releases since there's no change for older LXC versions.

stefangweichinger:
ack, thx

PR #1706, opened 2021-05-06
[mariadb_server] Allow excluding databases in backups

pedroluislopez:
Allow excluding databases in backups using regular expressions

Signed-off-by: Pedro Luis López Sánchez <plopez@neuromobilemarketing.com>


drybjed (approved):
Nice work. The example regex is incomplete, so I'll fix it. :)

PR #1711, opened 2021-05-15
[apache] Use 'dont-create' to not upload template.

julien-lecomte:
According to documentation of "dont-create", the snippet already exists on host. To quote documentation: "This special type assumes the snippet file is already present and does not try to create it."

Yet, the "Create sites-available configuration" task from the apache role will create a vhost file.


drybjed (approved):
Thanks for catching that up. I'll backport it to older releases as well.

drybjed:
The failed tests seem to be an issue with Microsoft APT repositories. I'll test the change locally and merge it if there are no issues.

PR #1713, opened 2021-05-19
add version detection for ansible 4 (changed format of --version output)

tlotze:


stefangweichinger:
works for me in a first quick test. Thanks.

drybjed (approved):
Thanks for the quick fix! I'll backport it to the stable releases as well.

drybjed:
The Docker test failed because Dockerfile installs `debops` from PyPI, which does not have a fix included yet. This is an issue with the Docker image build pipeline, I'll try to solve it later. Merging this PR since it fixes the issue.

Issue #1712, opened 2021-05-18
[debops.lxc] : port updated connection plugin lxc_ssh.py to debops


stefangweichinger:
The `debops.lxc` role uses the ansible connection plugin `lxc_ssh.py` from [here](https://github.com/andreasscherbaum/ansible-lxc-ssh)

Recently the PR https://github.com/andreasscherbaum/ansible-lxc-ssh/pull/31 was merged, and it would be nice to port the latest version into debops.

@stefangweichinger refactored the script in https://github.com/andreasscherbaum/ansible-lxc-ssh/pull/34

@drybjed : pls advise: you want me to run it through `pycodestyle` ? At a first glance there are many long lines etc, which rules are to apply?

I will then try to create a PR for that.

stefangweichinger:
this is done, right?

PR #1714, opened 2021-05-19
commit 77a4a86 from https://github.com/andreasscherbaum/ansible-lxc-ssh

stefangweichinger:
belongs to #1712

### DIFF
--- a/ansible/plugins/connection_plugins/lxc_ssh.py
+++ b/ansible/plugins/connection_plugins/lxc_ssh.py
@@ -1,4 +1,5 @@
 # Copyright 2016 Pierre Chifflier <pollux@wzdftpd.net>
+# SPDX-License-Identifier: GPL-3.0-or-later
 #
 # SSH + lxc-attach connection module for Ansible 2.0
 #
--- a/ansible/plugins/connection_plugins/lxc_ssh.py
+++ b/ansible/plugins/connection_plugins/lxc_ssh.py
@@ -18,7 +18,6 @@
 #
 # You should have received a copy of the GNU General Public License
 # along with Ansible.  If not, see <http://www.gnu.org/licenses/>.
-#
 
 from __future__ import absolute_import, division, print_function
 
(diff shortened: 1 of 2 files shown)


ansible/plugins/connection_plugins/lxc_ssh.py
```diff
@@ -1,5 +1,4 @@
 # Copyright 2016 Pierre Chifflier <pollux@wzdftpd.net>
-# SPDX-License-Identifier: GPL-3.0-or-later
```
drybjed: SPDX identifier is required by DebOps.

stefangweichinger: re-added in 455757438

drybjed:
@stefangweichinger Looks like you didn't pass the changes through `pycodestyle`, you can do it locally by running `make pep8` command in the DebOps monorepo.

stefangweichinger:
Yeah, I see. You're strict at the line length also for ported scripts?
I wonder if we can "fork" the script and fix that in our copy, I wonder if upstream accepts shortened comments etc

stefangweichinger:
btw: I am trying to use [black](https://black.readthedocs.io/en/stable/integrations/github_actions.html) for reformatting and that tool could also be used within a github action. Maybe interesting to you:

https://black.readthedocs.io/en/stable/integrations/github_actions.html

stefangweichinger:
https://github.com/andreasscherbaum/ansible-lxc-ssh/pull/35 was merged right now.

https://github.com/andreasscherbaum/ansible-lxc-ssh/commit/6ce02212f6d2acdcd9f4eaf5460fa1a6bb873c1a

Committing the latest script in a minute.

drybjed (approved):
Thanks for working with upstream and updating the plugin. I'll backport the changes to the stable releases as well, to keep them synchronized.

stefangweichinger:
> Thanks for working with upstream and updating the plugin. I'll backport the changes to the stable releases as well, to keep them synchronized.

Glad I could help.

Issue #1718, opened 2021-05-23
[tinc] ansible's ipaddr filter does not support "/32"


julien-lecomte:
`tinc/templates/etc/tinc/network/tinc-up.j2` uses ansible `ipaddr` filter in the following form:

```
{% set tinc__tpl_addresses = [] %}
{% if item.value.address|d() %}
{%   set _ = tinc__tpl_addresses.extend(([ item.value.address ]
                                         if item.value.address is string
                                         else item.value.address)
                                        | ipaddr('host/prefix')) %}
{% endif %}
```

After some searching and testing, I've noticed that `ipaddr` the following behavior:

| input           | filter                | output          |
|-----------------|-----------------------|-----------------|
| 192.168.73.1    | ipaddr('host/prefix') | (blank)         |
| 192.168.73.1    | ipaddr('host')        | 192.168.73.1/32 |
| 192.168.73.1    |                       | 192.168.73.1    |
| 192.168.73.1/32 | ipaddr('host/prefix') | (blank)         |
| 192.168.73.1/32 | ipaddr('host')        | 192.168.73.1/32 |
| 192.168.73.1/32 |                       | 192.168.73.1/32 |
| 192.168.73.1/24 | ipaddr('host/prefix') | 192.168.73.1/24 |
| 192.168.73.1/24 | ipaddr('host')        | 192.168.73.1/24 |
| 192.168.73.1/24 |                       | 192.168.73.1/24 |

The first blank seems to be problematic, at least for me.
I suggest changing the filter to ipaddr('host').

PR #1719, opened 2021-05-23
Tinc filter host

julien-lecomte:
cf. https://github.com/debops/debops/issues/1718


drybjed (approved):
Sounds good to me. The `ipaddr('host')` filter is definitely more forgiving for the users, let's make their life easier. :-) I'll backport this to the stable releases as well.

PR #1720, opened 2021-05-25
[opendkim] Restore compatibility with Ansible < 2.10

imrejonk:
This omits the `regenerate` parameter in opendkim__keys when using
Ansible < 2.10, as it is not supported prior to that release.


drybjed (approved):
Thanks! I'll backport these changes to older releases as well.

PR #1722, opened 2021-05-27
[ferm] Disable mgmt of iptables backend on Stretch

imrejonk:
Commit 0909d02f7 introduced management of the iptables backend symlink
using the 'alternatives' system, but this caused an error on Debian 9
systems:

```
TASK [ferm : Manage iptables backend using alternatives]
fatal: [unifi.ciphermail.com]: FAILED! => changed=false
  msg: Specified path /usr/sbin/iptables-nft does not exist
```

This commit restores the old behaviour on Debian 9.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -270,6 +270,12 @@ General
 
 - The EteSync playbook is now included in the default DebOps playbook.
 
+:ref:`debops.ferm` role
+'''''''''''''''''''''''
+
+- The management of the :command:`iptables` backend symlink using the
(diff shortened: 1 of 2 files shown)


drybjed:
Thanks, good catch. However, could you change this to a list? I think `[ "wheezy", "jessie", "stretch", "trusty", "xenial" ]` should cover all cases that might be relevant.

imrejonk:
Here you go! I didn't originally include the old Ubuntu distribution codenames as I haven't tested this change on them, and Wheezy and Jessie have been EOL for some time now.

drybjed (approved):
Thanks for the fix. :-)

PR #1723, opened 2021-05-28
Fix expected token err '','', got ''integer''.

julien-lecomte:
When running `debops service/ifupdown -K -l host.xyz.example.com`, I get the following error:

```
TASK [ifupdown : Generate network interface configuration] 
fatal: [host.xyz.example.com]: FAILED! => 
  msg: 'template error while templating string: expected token '','', got ''integer''. String: {{ item.value.mode | d(0644) }}'
```

The only ifupdown config in my inventory is `ifupdown__external_interface: "eth0"` for said host.

### DIFF
--- a/ansible/roles/ifupdown/tasks/main.yml
+++ b/ansible/roles/ifupdown/tasks/main.yml
@@ -86,7 +86,7 @@
                                                                                                | d(item.key) }}'
     owner: 'root'
     group: 'root'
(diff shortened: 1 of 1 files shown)


drybjed (approved):
Thanks for the fix, I'l backport it to the stable releases as well. The error with Docker-based test is known and shouldn't impact this change, so I'm merging the PR.

PR #1724, opened 2021-05-29
[minidlna] First working version

julien-lecomte:
Allows user to install and setup a working MiniDLNA server. Registers ports in /etc/services, and setups the ferm firewall.

Issues I'd like to iron out:
- [ ] ferm currently creates two rule files (150_rule_minidlna_allow_tcp.conf, 150_rule_minidlna_allow_udp.conf). I don't know if there would be a method to only have one.
- [x] Also, why is it a 150xyz file and not a 100xyz, or even a 50xyz since I specify 50 as weight.
- [ ] No dependencies are in the meta file; same as all other roles. I guess this is normal.

### DIFF
--- a/ansible/roles/sysctl/templates/etc/sysctl.d/parameters.conf.j2
+++ b/ansible/roles/sysctl/templates/etc/sysctl.d/parameters.conf.j2
@@ -20,9 +20,11 @@
 {{ value -}}
 {%   endif %}
 {% endmacro %}
+
 {% if item.comment|d() %}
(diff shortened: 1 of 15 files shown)


ansible/roles/sysctl/templates/etc/sysctl.d/parameters.conf.j2
```diff
@@ -20,6 +20,7 @@
 {{ value -}}
 {%   endif %}
 {% endmacro %}
+
```
julien-lecomte, line 23: to be removed.

PR #1725, opened 2021-06-01
Fix version matchin bug.

prk0ghy:
In python virtualenv installations "ansible-playbook --version"
output in first line is "ansible-playbook [core 2.11.1]"
previous code matched "[core" as version

### DIFF
--- a/bin/debops
+++ b/bin/debops
@@ -33,6 +33,7 @@ from __future__ import print_function
 import sys
 import os
 import subprocess
+import re
 try:
     import configparser
 except ImportError:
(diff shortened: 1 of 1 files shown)


drybjed:
Did you test this on Ansible installations outside of a vrtualenv (for example as `pip3 --user` or installed via a `.deb` package)? How about older Ansible versions that use the legacy format?

prk0ghy:
Can we talk on jitsi?

Am Di., 1. Juni 2021 um 22:30 Uhr schrieb Maciej Delmanowski <
***@***.***>:

drybjed:
Sure, Heading to https://meet.jit.si/debops now.

PR #1727, opened 2021-06-02
Minidlna

julien-lecomte:
Allows user to install and setup a working MiniDLNA server. Registers ports in /etc/services, and setups the ferm firewall.

### DIFF
--- a/ansible/playbooks/srv/minidlna.yml
+++ b/ansible/playbooks/srv/minidlna.yml
@@ -1,0 +1,1 @@
+../service/minidlna.yml
--- a/ansible/roles/minidlna/defaults/main.yml
+++ b/ansible/roles/minidlna/defaults/main.yml
@@ -1,0 +1,147 @@
+---
+# .. vim: foldmarker=[[[,]]]:foldmethod=marker
+
+# .. Copyright (C) 2021 Julien Lecomte <julien@lecomte.at>
+# .. Copyright (C) 2021 DebOps <https://debops.org/>
+# .. SPDX-License-Identifier: GPL-3.0-only
+
+# .. _minidlna__ref_defaults:
+
+# .. contents:: Sections
+#    :local:
+#
+# .. include:: ../../../../includes/global.rst
+
+# Packages and installation [[[
+# -------------------------
+
+# .. envvar:: minidlna__base_packages [[[
+#
+# List of APT packages to install for :command:`minidlna` support.
+minidlna__base_packages: [ 'minidlna' ]
+
+                                                                   # ]]]
+# .. envvar:: minidlna__packages [[[
+#
+# List of additional APT packages to install during :command:`minidlna`
+# configuration.
+minidlna__packages: []
+
+                                                                   # ]]]
+                                                                   # ]]]
+# MiniDLNA configuration [[[
+# ----------------------
+
+# .. envvar:: minidlna__http_port [[[
+#
+# Port for HTTP traffic. Defaults to 8200.
+minidlna__http_port: 8200
+
+                                                                   # ]]]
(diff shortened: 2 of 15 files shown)


ansible/playbooks/srv/minidlna.yml
```diff
@@ -0,0 +1,29 @@
+---
```
drybjed: Looks like this is a real file? It should be a symlink to `../service/minidlna.yml` instead, so that the two playbooks don't get desynchronized in the future.

ansible/roles/sysctl/templates/etc/sysctl.d/parameters.conf.j2
```diff
@@ -20,6 +20,7 @@
 {{ value -}}
 {%   endif %}
 {% endmacro %}
+
```
drybjed: Why this change is here?

ansible/roles/minidlna/defaults/main.yml
```diff
@@ -0,0 +1,147 @@
...
+# List of media directories to scan and present to end user.
+#
+# If you want to restrict a media_dir to a specific content type, you can
+# prepend the directory name with a letter representing the type (A, P or V),
+# followed by a comma, as so:
+#
+# * "A" for audio    (eg. media_dir=A,/var/lib/minidlna/music)
+# * "P" for pictures (eg. media_dir=P,/var/lib/minidlna/pictures)
+# * "V" for video    (eg. media_dir=V,/var/lib/minidlna/videos)
+# * "PV" for pictures and video (eg. media_dir=PV,/var/lib/minidlna/digital_camera)
```
drybjed: The examples here should reflect what needs to be put in the variable, so don't add `media_dir=` since this will be prefixed in the template automatically.

ansible/roles/minidlna/defaults/main.yml
```diff
@@ -0,0 +1,147 @@
...
+
+                                                                   # ]]]
+# .. envvar:: minidlna__allow [[[
+#
+# List of IP addresses or CIDR subnets which are allowed to connect to the
+# MiniDLNA instances over the network, on all hosts in the Ansible
+# inventory. This variable configures the firewall for all instances at the
+# same time, for individual instance configuration you should modify the
+# :envvar:`minidlna__ferm__dependent_rules` variable directly.
+minidlna__allow: {}
```
drybjed: This should be a list (`[]`) and not a YAML dictionary (`[]`), otherwise the `ferm` role will not parse this correctly.

ansible/roles/minidlna/templates/etc/ansible/facts.d/minidlna.fact.j2
```diff
@@ -0,0 +1,9 @@
+{# Copyright (C) 2021 Julien Lecomte <julien@lecomte.at>
```
drybjed: It would be better to make the fact a Python script. That way users can use `run-parts /etc/ansible/facts.d` to get a nice overview, or other services can benefit from this. The `postwhite` role has a nice example of a simple Python fact script.

The fact scripts should if possible gather information from the host itself. You could check the minidlna configuration file and find out the actual http_port in use.

julien-lecomte: I'll look into that.

julien-lecomte:
I'd like to bump this MR so I can move onto the next ones. Thanks

drybjed:
@julien-lecomte I started looking into the PR and I'll probably make a few changes, for example add the whole contents of the configuration file by default. Expect this to be merged in a few days tops.

PR #1734, opened 2021-06-10
[nginx] Always allow access to ACME challenge dir

imrejonk:
This change ensures that NGINX always allows access to ACME challenge
files, even when the server is configured with restrictive access
controls.


drybjed (approved):
Awesome idea, thanks. I'll backport it to the stable releases as well.

PR #1737, opened 2021-06-10
[apt_cacher_ng] No longer create NGINX webroot

imrejonk:
The NGINX server configuration provided by `debops.apt_cacher_ng` only
functions to set up proxies and a redirect, so no webroot is necessary.


drybjed (approved):
Sounds good, thanks! I'll backport this to the stable releases.

PR #1738, opened 2021-06-10
[apt] allow to add foreign architectures

boucman:
This adds a new set of variables apt__*_extra_architectures that will
be added to the machine, allowing the configuration of multi-architecture
machines

Note that this has only been tested on 2.2 

Signed-off-by: Jérémy Rosen <jeremy.rosen@smile.fr>

### DIFF
--- a/ansible/roles/apt/defaults/main.yml
+++ b/ansible/roles/apt/defaults/main.yml
@@ -26,6 +26,31 @@
 apt__enabled: True
 
                                                                    # ]]]
+# Extra architectures [[[
+# -----------------------
+
+# These lists define extra architectures to be enabled on the host.
+# The main architecture does not need to be defined that way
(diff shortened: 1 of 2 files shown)


ansible/roles/apt/defaults/main.yml
```diff
@@ -25,6 +25,31 @@
 # Enable or disable this role.
 apt__enabled: True
 
+                                                                   # ]]]
+# Extra architectures [[[
+# -----------------------
+
+# These lists define extra architectures to be enabled on the host.
+# The main architecture does not need to be devined that wat
```
drybjed: "devined that wat" :-)

drybjed (approved):
Sounds good, thanks. This will be backported to stable releases.

PR #1740, opened 2021-06-15
Grafana Role

julien-lecomte:
Purpose is to get a Grafana role in debops; all the rest (LDAP, nginx proxy) can be added at a later date.


ypid:
What are the differences to #1124? I assume you based on that work but were are the commits? You squashed them?

julien-lecomte:
Yes, I squashed them but kept the original authors.

The two major changes: always installing on the 3000 port, which is the default; and not using the nginx as a proxy.

ypid:
How do you then provide TLS/HTTPS?

julien-lecomte:
If TLS/HTTPS is desired, it's possible to configure the role to set the correct configuration values.

ypid:
I don’t see a case where one would not want to have HTTPS. It should be setup by default. That level of integration is one of the nice things of DebOps.

julien-lecomte:
I understand your point of view, but I will argue that #1124 had HTTPS support (via nginx) and was not merged for reasons unknown. 

I'd rather have a minimal role that does what it promises, and does it well, and is improved upon according to demand.

.gitlab-ci.yml
```diff
@@ -925,6 +938,14 @@ stages:
     JANE_DIFF_PATTERN: '.*/roles/gunicorn/.*'
     JANE_LOG_PATTERN: '\[gunicorn\]'
 
+'grafana role':
```
julien-lecomte, line 941: to be removed.

julien-lecomte:
This versions has
- previous commits are not squashed,
- universal configuration
- ningx (opt-out) http + https support 
- random admin password on install (and not longer "admin")

julien-lecomte:
Ready for full review + acceptance.

PR #1742, opened 2021-06-16
Fix owncloud ldap

### LINKED ISSUE #1741: [owncloud] error on configure ldap (nextcloud)
Setting up nextcloud with ldap integration fails on first run with:

```
TASK [owncloud : Check default ldap config exists] ***************************************************************************************************************************************************************
fatal: [nextcloud.intern.iris-ev.de]: FAILED! => {"changed": false, "cmd": ["php", "--file", "/srv/www/nextcloud//occ", "ldap:show-config", "s01"], "delta": "0:00:00.318695", "end": "2021-06-16 14:47:43.496665", "msg": "non-zero return code", "rc": 1, "start": "2021-06-16 14:47:43.177970", "stderr": "", "stderr_lines": [], "stdout": "Invalid configID", "stdout_lines": ["Invalid configID"]}
```

This is because the return code on an inexistent ldap config with gives an non null exit code:

```root@nextcloud:~# occ ldap:show-config  xx
Invalid configID
root@nextcloud:~# echo $?
1
```

and ansible fails.

No idea if this changed on recent nextcloud versions (I test with 21).

The solution would be e.g. to add `ignore_errors: True` to `tasks/ldap.yml`

```
- name: Check default ldap config exists
  command: php --file "{{ owncloud__app_home }}/occ" ldap:show-config "{{ owncloud__ldap_config_id }}"
  changed_when: False
  when: (not ansible_check_mode)
  register: owncloud__register_ldap_default_config_exits
  become: True
  become_user: '{{ owncloud__app_user }}'
  ignore_errors: True
```

jankowa:
fixes #1741


ypid:
This PR mixes a few changes together and I am not sure that is intended as you did not state it in the PR title or description. Can you check?

Edit: I checked your actual change. Can you explain why `ignore_errors` is needed?

drybjed:
Agreed. @jankowa, you need to rebase the `master` branch on the upstream version, then rebase the `fix_owncloud_ldap` off of the `master` branch. You can update the PR via `--force` afterwards.

jankowa:
oh, sorry. yes, I messed up my branch ;-) and didn't realized that. 
going to make a new pull request.

ypid:
FYI: A new pull request is not required.

Note my review note that I added to my comment above:

> I checked your actual change. Can you explain why ignore_errors is needed?

jankowa:
> FYI: A new pull request is not required.
> 
> Note my review note that I added to my comment above:
> 
> > I checked your actual change. Can you explain why ignore_errors is needed?

I explained it in #1741: the non-zero exit code from `ldap:show-config` with an unknown non existent config id aborts the task.

ypid:
Sorry, looked at the commit message.

https://github.com/debops/debops/blob/master/ansible/roles/owncloud/tasks/run_occ.yml already has a few of those tricks in a central location. Can you rework your fix in that file?

jankowa:
> Sorry, looked at the commit message.
> 
> https://github.com/debops/debops/blob/master/ansible/roles/owncloud/tasks/run_occ.yml already has a few of those tricks in a central location. Can you rework your fix in that file?

Hy ypid, today i came through this again. Actually I don't now how to reworking this using the run_occ.yml. If I understood this correctly the both tasks `ldap.yml` and `run_occ.yml` are executed separately. So as far as I see we have to fix the issue in https://github.com/debops/debops/blob/master/ansible/roles/owncloud/tasks/ldap.yml

ypid:
You are right. It uses the `command` module directly. It could be changed to `run_occ.yml` but that is probably more complex than it has to be. You can resubmit this. But please rework it by removing the `ignore_errors` and use something like `failed_when`. `run_occ.yml` has examples.

PR #1743, opened 2021-06-22
feat(sshd): Add sshd__ferm_interface to limit access to SSH on iface

ypid:


drybjed (approved):
Sounds good, thanks! I'll backport this to the stable releases as well.

PR #1729, opened 2021-06-06
[rsyslog] allow forward in clear tcp anon mode

bleuchtang:
This PR allow to forward logs to rsyslog server in plain tcp and anonymous mode (use in local or trusted network).
Exemple:

    rsyslog__default_forward:
      - myrsyslog.local:
        target: 'myrsyslog.local'
        port: '514'
        protocol: 'tcp'
        driver: 'ptcp'
        drivermode: '0'
        driverauthmode: 'anon'

PR #1749, opened 2021-06-29
[rsyslog] Re-enable streamDriverPermittedPeers

imrejonk:
Commit 2cadbbbd23900fa093fe63c2131ab04c922eecc6 added a conditional that
resulted in streamDriverPermittedPeers being set only when the
(undocumented) `drivermode` parameter was set in the `rsyslog__forward`
rule entry. As streamDriverPermittedPeers specifies the name of the
remote peer as used in TLS certificate validation, this option should
always be set when securing the syslog traffic with TLS, which is what
this commit takes care of.

This commit also documents the new rsyslog__forward parameters, and adds
one new default variable: `rsyslog__default_driver_mode`. I have taken
the liberty to rename element.driver, element.drivermode and
element.driverauthmode to better match the naming in the
defaults/main.yml file.

/cc @bleuchtang

Ref: https://github.com/debops/debops/pull/1729

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -34,6 +34,13 @@ Added
 
   .. __: https://wiki.debian.org/Multiarch/HOWTO
 
+:ref:`debops.rsyslog` role
+''''''''''''''''''''''''''
+
+- It is now possible to override the default ``netstream_driver``,
+  ``driver_mode`` and ``driver_authmode`` parameters in every
+  :ref:`rsyslog__ref_forward` forwarding rule.
+
 :ref:`debops.sshd` role
 '''''''''''''''''''''''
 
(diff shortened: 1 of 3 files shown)


drybjed (approved):
@imrejonk Thanks for improving the modification and adding documentation. This will be backported as well to the stable branches.

I'm wondering how we could avoid fixes like this in the future. On first glance the problem seems to be me not testing the change thoroughly (I don't have good test environment to play with at the moment) and accepting changes without updated documentation. It's probably something to talk about at the next DebOps meeting.

bleuchtang:
Argh.. I just saw this, really sorry for my previous bad PR...

I just came across this pull request because I wonder if, with this changes, we can remove rsyslog__pki variable here:
https://github.com/debops/debops/blob/e7e077837d1c5ae291226ec06d889ad7c9d00370/ansible/roles/rsyslog/defaults/main.yml#L669

It seem useless now as we can forward without certificates in ptcp mode.

imrejonk:
No worries, it can happen.
We could remove the `rsyslog__pki` variable from that conditional, but the result would be that the role configures syslog forwarding over plaintext TCP (aka reliable syslog service, [RFC 3195](https://www.rfc-editor.org/rfc/rfc3195.html)) when the pki configuration is absent. I consider that undesirable from a security perspective.

bleuchtang:
Yes, so maybe we can add another variable like rsyslog__forward_enabled_plain set to false by default.

imrejonk:
Yes or just set rsyslog__default_netstream_driver to gtls always, and let the user manually override it. Go ahead with a PR if you’re interested in this.

PR #1753, opened 2021-07-05
Correct naming of sshd__register_host_certs var

growse:
`sshd_config.j2` referenced a variable `sshd__host_key_cert_files` which didn't exist. Looks like this was meant to be `sshd__register_host_certs`.


drybjed (approved):
Excellent, thanks for the fix! It will be backported to the stable releases as well.

PR #1754, opened 2021-07-05
[redis_server] Facts version and passwordless istances

tumd:
Fixed version not present in the output.

If no password has been specified in `redis_server__auth_password` no
instances would be presented.


drybjed (changes requested):
I'm not about the `None` value for password, did you test this with other applications that use Redis, like Roundcube or NetBox? Usually an empty password is denoted by an empty string, maybe that could be a better value here?

drybjed (approved):
I tested the change with NetBox, using `Null` as a value seems to work OK. Thanks! I'll backport this as well.

PR #1758, opened 2021-07-07
[system_users] Fix create_home parameter

imrejonk:
There were typos in the Ansible task causing a `create_home: False`
setting to create the home directory anyway.


drybjed (approved):
Thanks! It must have been a hold out from the time the `createhome` → `create_home` change in the Ansible `user` module happened. I'll backport this to the stable releases.

PR #1761, opened 2021-07-07
[lvm] Allow custom lvcreate options

oxmie:
The _ansible_ [`lvol` modules](https://docs.ansible.com/ansible/latest/collections/community/general/lvol_module.html) supports custom options ([`opts`](https://docs.ansible.com/ansible/latest/collections/community/general/lvol_module.html#parameter-opts)) for the `lvcreate` command. This is really nice to have in DebOps, because I will gave the _LVM_ much more power.

P.S.: In ansible/roles/lvm/defaults/main.yml, I just fixed a very little typo

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -41,6 +41,9 @@ Added
 
   .. __: https://man7.org/linux/man-pages/man7/lvmthin.7.html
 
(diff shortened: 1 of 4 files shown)


drybjed (approved):
Looks good, thanks. This will be backported to the stable releases.

PR #1762, opened 2021-07-08
[docs] Add note to pki-realms.rst about DH params

imrejonk:
Related to `debops.pki`
Some Java-based applications don't like the DH params at the end of
default.crt. I figured this out when debugging a failed update to
Graylog 4.1: https://github.com/Graylog2/graylog2-server/issues/10997


drybjed (approved):
Sounds good, thanks.

PR #1478, opened 2020-08-20
[resolvconf] Remove variables from fact script

imrejonk:
This commit removes the 'domain', 'nameservers' and 'search' variables
from the resolvconf Ansible local facts script. `ansible_domain`,
`ansible_dns.nameservers` and `ansible_dns.search` can be used instead.
References in `debops.docker_server` and `debops.dnsmasq` have been
updated to reflect this change.

Closes #1469

PR #1763, opened 2021-07-08
[resolvconf] Fix resolving upstream nameservers

yethee:
This patch prevents override variable `nameservers` and lost information from the primary resolver configuration.

Related to #1478

Currently (at least at versions 2.3.0, 2.3.1) upstream nameservers may not be included in the fact, because value of the `nameservers` is not relevant.

```py
if (('127.0.0.1' in nameservers or
        '127.0.0.53' in nameservers) and
        upstream_nameservers):
    output['upstream_nameservers'] = upstream_nameservers
```

### DIFF
--- a/ansible/roles/resolvconf/templates/etc/ansible/facts.d/resolvconf.fact.j2
+++ b/ansible/roles/resolvconf/templates/etc/ansible/facts.d/resolvconf.fact.j2
@@ -60,9 +60,7 @@ if os.path.isfile(systemdresolved_path):
 
 for element in resolvconf_files:
     try:
-        nameservers = read_resolv_file(element)
-
-        for server in nameservers:
+        for server in read_resolv_file(element):
             if not server.startswith('127.'):
                 upstream_nameservers.append(server)
 


drybjed:
I'm not really sure what might be the issue you are trying to fix. Can you explain a bit more about the environment you found this in? Do you use Docker, `systemd-resolved`, `dnsmasq` or some other DNS cache/proxy?

yethee:
I faced with this issue with configuration of docker, with default options of role `docker_server`.

In my case `ansible_local.resolvconf.upstream_nameservers` always missing and `ansible_dns.nameservers` resolves as `['127.0.0.53']` (`systemd-resolved` is used). DNS not working within docker containers when used `127.0.0.53` as DNS server


Host: Ubuntu 20.04

```
$ cat /etc/resolv.conf
# Dynamic resolv.conf(5) file for glibc resolver(3) generated by resolvconf(8)
#     DO NOT EDIT THIS FILE BY HAND -- YOUR CHANGES WILL BE OVERWRITTEN
# 127.0.0.53 is the systemd-resolved stub resolver.
# run "systemd-resolve --status" to see details about the actual nameservers.

nameserver 127.0.0.53
```

drybjed:
Thanks, that explains it. I managed to reproduce the bug, good catch. :-)

PR #1768, opened 2021-07-15
+ custom_log and error_log for fullly customization of CustomLog and ErrorLog Directive

hungpr0:
Hello,

We should have fully customization for `CustomLog` and `ErrorLog` Directive of `apache` role.
This allows us to save log to different path, different filename...


drybjed (approved):
Good idea, thanks. I'll add some documentation bits about the new parameters.

Issue #1769, opened 2021-07-15
[debops.debops.apache] fully customizable Alias Directive


hungpr0:
Hello,

Current, with
```
{% macro get_alias(url_path, fs_directory) %}{# [[[ #}
Alias {{ url_path | quote }} {{ (fs_directory + ("/" if (fs_directory[-1] != "/") else "")) | quote }}
{% endmacro %}

```

Example:
```
apache__vhosts:
  - name: test.local
    document_root: /var/www/html
    alias: /cgi /var/www/cgi
```

expect result:
```
   Alias /cgi /var/www/cgi
```

actual result:
```
  Alias /cgi /var/www/cgi /var/www/html
```

we are unable to have custom file path. It is always `fs_directory`. It is really inconvenient.
This Alias Directive https://httpd.apache.org/docs/2.4/mod/mod_alias.html#alias shows that some examples
```
Alias "/image" "/ftp/pub/image"
```
```
Alias "/icons/" "/usr/local/apache/icons/"
```

It seems it also lacks of support multiple `Alias`
Can we have an improvement for this?

Thanks

PR #1770, opened 2021-07-15
+ [debops.apache] alias_path to customize the alias path of Alias Directive

hungpr0:
For issue https://github.com/debops/debops/issues/1769


drybjed (approved):
Thanks, sounds good. I'll add the documentation bits.

PR #1774, opened 2021-07-21
Enable TCP sysctl performance flags

julien-lecomte:
We already do hardening via sysctl, might as well also propose perfomance improvements.

### DIFF
--- a/ansible/roles/sysctl/defaults/main.yml
+++ b/ansible/roles/sysctl/defaults/main.yml
@@ -105,6 +105,17 @@ sysctl__hardening_ipv6_disabled: False
 sysctl__hardening_experimental_enabled: False
                                                                    # ]]]
                                                                    # ]]]
+
+# Performance parameters [[[
+# --------------------------
+
+# .. envvar:: sysctl__tcp_performance_enabled [[[
+#
+# Should the TCP performance options be applied?
+sysctl__tcp_performance_enabled: False
+
+                                                                   # ]]]
+                                                                   # ]]]
 # Kernel parameters [[[
 # ---------------------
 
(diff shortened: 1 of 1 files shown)


ansible/roles/sysctl/defaults/main.yml
```diff
@@ -235,6 +245,13 @@ sysctl__default_parameters:
                          ansible_virtualization_type == "openvz"))
                   | ternary("present", "absent") }}'
 
+      - name: 'net.ipv4.tcp_slow_start_after_idle'
+        value: '{{ sysctl__tcp_performance_enabled|bool | ternary(0, 1) }}'
```
drybjed, line 250: This `ternary()` function is redundant. If the option is not enabled, it will not be included in the config file. Just change this to `1` so that it's easier to parse by humans. Or should that be `0` to enable this behavior?

ansible/roles/sysctl/defaults/main.yml
```diff
@@ -105,6 +105,16 @@ sysctl__hardening_ipv6_disabled: False
...
+
+# Performance parameters [[[
+# ---------------------
+# .. envvar:: sysctl__hardening_enabled [[[
+#
+# Should the TCP performance options be applied?
+sysctl__tcp_performance_enabled: False
+
+                                                                   # ]]]
+                                                                   # ]]]
```
drybjed, line 118: The additional markers will mess up folding in `vim`. Remove one marker from the previous variable to keep them balanced. Two markers should not have a blank space before them, so that spacing of variables is consistent.

ansible/roles/sysctl/defaults/main.yml
julien-lecomte: The default value is 1. When set to 1, it slow starts. When set to 0 it does not slow start.

julien-lecomte: If we play on the present/absent flag depending on the value of sysctl__tcp_performance_enabled, it means that if we deactivate the enabled flag, then the value is not reset to it's original value. This is also the problem with most other sysctl values.

eg:
sysctl__tcp_performance_enabled: true, => it sets net.ipv4.tcp_slow_start_after_idle to 0.
then
sysctl__tcp_performance_enabled: false, => it sets net.ipv4.tcp_slow_start_after_idle stays at 0.

ansible/roles/sysctl/defaults/main.yml
julien-lecomte: I'm confused. Folding is correct in my vim because of the new section "Performance parameters"

drybjed: Right, I see a new section now. Good point.

drybjed (approved):
Thanks for the PR. I'll make some changes to make this option a bit clearer for future readers.

PR #1776, opened 2021-07-22
[dovecot] Don't hardcode mailbox definitions

Alphix:
Currently the values in /etc/dovecot/conf.d/15-mailboxes.conf (from the Debian
packaging) are hardcoded and cannot be changed via the Dovecot role. Since
the mailbox hierarchy (e.g. legacy setups) and naming (e.g. for localization)
is site dependent, it makes sense to allow these settings to be changed easily.

### DIFF
--- a/ansible/roles/dovecot/defaults/main.yml
+++ b/ansible/roles/dovecot/defaults/main.yml
@@ -585,6 +585,42 @@ dovecot_auth_config_map:
         mode: '0600'
 
                                                                    # ]]]
+# .. envvar:: dovecot_mailbox_definitions [[[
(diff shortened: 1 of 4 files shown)


Alphix:
Hmm, the more time I spend with the Dovecot role, the more convinced I am that it would be better to rewrite the role to use the [universal configuration](https://docs.debops.org/en/master/user-guide/universal-configuration.html) format. There are several things which are hardcoded and/or hard to impossible to achieve with the current design.

Alphix:
See #1780 for a more complete solution.

PR #1777, opened 2021-07-22
Allow user override via envvar

julien-lecomte:
Allows a user to `export SPHINXOPTS=-j4` for example.


drybjed (approved):
Thanks, seems like a good idea.

PR #1780, opened 2021-07-26
[dovecot] Rewrite to use universal configuration

Alphix:
First I wanted to make it possible to change mailbox settings/flags/names by not hardcoding some values (#1776). Then I realised that there were more things I wanted to add which couldn't be added to the configuration file using the current config scheme. For example, a config like this

```
service lmtp {
        unix_listener /var/spool/postfix/private/dovecot-lmtp {
                mode = 0660
                group = postfix
                user = postfix
        }

        user = vmail
}
```

is impossible to express (note the `user = vmail` line at the end).

So in the end I rewrote the dovecot role to use universal config. I must admit that I've only tested the end result and not each intermediate state.


Alphix:
I should add that the documentation for the `dovecot` role will need further work if it is agreed that this is a worthwhile change (I intentionally did not do it yet since I wanted feedback on the general idea before spending more time on updating the documentation to match the new behaviour)

imrejonk:
Hi Alphix,

Thanks for the great work on this PR. I think that `debops.dovecot` can benefit a lot from Universal Configuration. We have quite a complicated Dovecot setup at work, and some parts of that configuration are now done in `dovecot_custom_localconf`, which I consider to be a workaround.

Converting our inventory for use with your adapted role will be lots of work however, and I'm currently in the middle of a big email infrastructure change already. I'm prioritizing our upgrade to Debian 11 (due to be released this Saturday) after that, so the first time I'll be able to really test your changes would be somewhere at the end of September or in October.

If @drybjed decides that this PR can't wait for my full review, then I propose that we take an extra careful look at the extensibility of this rewrite. It should be okay if the principles of Universal Configuration are correctly applied. The reason I'm asking is that I will most likely want to improve the parts of the role that my employer cares about: dsync, sieve and LDAP support, and TLS to encrypt it all. It would be nice if that can relatively easily be done, and without having to change the role defaults again.

Again, great work on this role so far, keep it up :)

drybjed:
I think that we can review this PR later, no need to rush. I'll try to do a review in the coming days; the current weekend will be a busy time due to Debian Bullseye release.

Alphix:
> If @drybjed decides that this PR can't wait for my full review, then I propose that we take an extra careful look at the extensibility of this rewrite. It should be okay if the principles of Universal Configuration are correctly applied. The reason I'm asking is that I will most likely want to improve the parts of the role that my employer cares about: dsync, sieve and LDAP support, and TLS to encrypt it all. It would be nice if that can relatively easily be done, and without having to change the role defaults again.

Hi,

it should be universal enough to cover all of your needs (I also need sieve, LDAP and TLS, and dsync is on my todo list).

One thing I've considered is whether it'd make more sense to create a new role (say, `dovecot_uc`) using the same prefix (`dovecot__`) for the configuration variables. That way the two roles could co-exist in the `debops` repo for an extended period of time, allowing admins to gradually move over rather than force a flag day on everyone.

Personally I also find it easier to review a complete role rather than analyse a large number of piecemel patches moving a role from A to B.

drybjed:
I don't like the idea of adding a rewritten Dovecot role alongside the old one - that way we will end up either supporting both, or the old one will be removed and then what? You will have to rename the new role or be left with `dovecot_uc` role.

We have stable releases, the rewritten role won't be backported to the previous release due to the lots of changes. And it's easy to rewrite the Dovecot role from scratch in a separate fork and then merge the changes to the main project.

Alphix:
I've rebased and pushed two small additional fixes. With those applied, I've successfully setup two test VMs using pretty much only vanilla debops roles using both `buster` and `bullseye`.

`vmtest1`:
```
bootstrap
common
service/slapd
ldap/init-directory
```

`vmtest2`
```
bootstrap
common
bootstrap-sss (not necessary for this test)
service/postfix
service/postconf
service/postldap
service/postscreen
service/dovecot
service/nginx
service/pam
rspamd (local role, in development)
rainloop (dito)
```

Added a LDAP user with the `mailRecipient` `objectClass`, then sent an email to the new user and sucessfully logged in via the `rainloop` web iface and read the email. So LDAP and TLS should be in working order after the `universal config` rewrite.
 
Next, I'll test `sieve` and see what's necessary to support `dsync` (which isn't supported out of the box pre- or post rewrite AFAIK).

And yeah, do feel free to review the PR whenever you get around to it, no rush. I already have the role I need in my local repo, so I'm in no hurry.

Alphix:
Ok, now `dsync` is supported as well (tested by setting up a third vm, `vmtest3`, and making sure that `dovecot__dsync_host` on `vmtest2` points to `vmtest3` and vice versa and that changes are synchronised between the two).

Don't know if the `dsync` functionality is flexible enough for complicated setups, @imrejonk, please remember to verify that (optional features which are not currently supported are e.g. returning `noreplicate` or `mail_replica` fields from the `userdb`). 

Still have to test `sieve` in more detail, but the rewrite should be essentially feature-complete now.

Alphix:
And now I've tested `sieve` as well, and it works, including `dsync`:ing `sieve` filters between servers.

Alphix:
Rebased and re-tested, seems to work.

@imrejonk and @drybjed : any update on when you think you'll have time for a review?

imrejonk:
@Alphix I really want to review your PR and test any changes carefully, but this will have to happen at work and we're currently very busy. Know that this is still on my radar. I appreciate the work you've put into the role thus far and I apologize for keeping you waiting. This PR requires a thorough review.

ansible/roles/dovecot/defaults/main.yml
```diff
@@ -953,8 +1414,18 @@ dovecot__ferm__dependent_rules:
...
-                    if ("pop3" in dovecot_protocols|d([])
-                         and dovecot_pki|d(True))
+                    if ("pop3s" in dovecot__features|d([])
+                         and dovecot__pki|d(True))
+                    else "absent" }}'
+
+  - name: 'dovecot_doveadm'
+    type: 'accept'
+    by_role: 'debops.dovecot'
+    dport: [ '{{ dovecot__dsync_port }}' ]
```
imrejonk: maybe add the doveadm port to /etc/services with (dependent) `debops.etc_services`, and use a human-friendly name here. This could say `dport: [ 'doveadm' ]`, for example, after adding this dependent configuration:
```
# .. envvar:: dovecot__etc_services__dependent_list [[[
#
# Configuration for the :ref:`debops.etc_services` role.
dovecot__etc_services__dependent_list:

  - name: 'doveadm'
    port: '{{ dovecot__dsync_port }}'
    protocols: [ 'tcp' ]
    comment: 'Added by debops.dovecot Ansible role.'

                                                                   # ]]]
```
The playbooks needs to be modified, with this before the `debops.ferm` role:
```
- role: etc_services
  tags: [ 'role::etc_services', 'skip::etc_services' ]
  etc_services__dependent_list:
    - '{{ dovecot__etc_services__dependent_list }}'
```
This will give you nice output in [ss(8)](https://manpages.debian.org/ss.8) as well.

imrejonk:
Hi @Alphix, just wanted to let you know that I have (finally) started reviewing your pull request. I'll propose some changes next week. Sorry for the long wait.

The redesigned role looks really well done on first glance. It's good to see that the dsync and sieve support is now more complete. I'd say we try to include this in the 3.0 release, which is scheduled for next month. One thing that needs more work is of course the documentation. Would you like to put some effort in? Otherwise I can do it as well.

Alphix:
> Hi @Alphix, just wanted to let you know that I have (finally) started reviewing your pull request. I'll propose some changes next week. Sorry for the long wait.

Cool. No problem.
 
> The redesigned role looks really well done on first glance. It's good to see that the dsync and sieve support is now more complete. I'd say we try to include this in the 3.0 release, which is scheduled for next month. One thing that needs more work is of course the documentation. Would you like to put some effort in? Otherwise I can do it as well.

I can update the documentation, no problem. I just wanted to wait until there was some kind of agreement on the actual role before spending time on docs. So I'll wait for your proposed changes and then start working on docs.

ansible/roles/dovecot/defaults/main.yml
```diff
@@ -405,191 +469,593 @@ dovecot__ssl_dh_file: '{{ ansible_local.dhparam[dovecot__dhparam_set]
...
+# .. envvar:: dovecot__mail_plugins [[[
 #
-# List of LMTP ``inet_listeners`` or ``unix_listeners`` which will be enabled.
-# See :ref:`dovecot_lmtp_listeners` for more details.
-dovecot_lmtp_listeners: [ '/var/spool/postfix/private/dovecot-lmtp' ]
+# Default mail plugins for all protocols.
+dovecot__mail_plugins: '{{ [ "$mail_plugins" ]
+                              + [ "notify" if "dsync" in dovecot__features ]
+                              + [ "replication" if "dsync" in dovecot__features ]
+                              + [ "quota" if "quota" in dovecot__features ] }}'
```
imrejonk: This goes wrong if I limit `dovecot__features` to a subset like `[ 'imaps', 'lmtp', 'sieve' ]`:
```
TASK [dovecot : debug] *********************************************************
ok: [mail5.ciphermail.com] => 
  dovecot__mail_plugins: '[''$mail_plugins'', Undefined, Undefined, Undefined]'
```
This works for me:
```
dovecot__mail_plugins: '{{ [ "$mail_plugins" ]
                           + ([ "notify" ]
                              if "dsync" in dovecot__features else [])
                           + ([ "replication" ]
                              if "dsync" in dovecot__features else [])
                           + ([ "quota" ]
                              if "quota" in dovecot__features else []) }}'
```

ansible/roles/dovecot/defaults/main.yml
```diff
@@ -405,191 +469,593 @@ dovecot__ssl_dh_file: '{{ ansible_local.dhparam[dovecot__dhparam_set]
...
-        user: 'postfix'
-        group: 'postfix'
-        mode: '0600'
-
-  protocol:
-    mail_plugins: '{{ dovecot_mail_plugins }}'
+# Enabled mail plugins for the IMAP protocol.
+dovecot__mail_plugins_imap: '{{ dovecot__mail_plugins
+                                 + [ "imap_sieve" if "sieve" in dovecot__features ]
+                                 + [ "imap_quota" if "quota" in dovecot__features ] }}'
```
imrejonk: I'm getting a templating error here when `dovecot__features` is limited to a subset:
```
Unexpected templating type error occurred on ({{ dovecot__mail_plugins + [
"imap_sieve" if "sieve" in dovecot__features ] + [ "imap_quota" if "quota" in
dovecot__features ] }}): can only concatenate str (not "list") to str
```
After fixing `dovecot__mail_plugins` (see above), the template error is gone, but I see undefined list items:
```
TASK [dovecot : debug] *********************************************************
ok: [mail5.ciphermail.com] => 
  dovecot__mail_plugins_imap: '[''$mail_plugins'', ''imap_sieve'', Undefined]'
```
The solution is to apply the same fix as I've proposed for `dovecot__mail_plugins`. Idem for `dovecot__mail_plugins_lda` and `dovecot__mail_plugins_lmtp`.

ansible/roles/dovecot/defaults/main.yml
```diff
@@ -806,6 +1253,13 @@ dovecot__ldap_quota_attribute: 'mailQuota'
 # User default LDAP quota. Use human-readable values, e.g. "2 GB".
 dovecot__ldap_quota_default: '10 GB'
                                                                    # ]]]
+# .. envvar:: dovecot__ldap_trash_field [[[
+#
+# The Dovecot internal field name which corresponds to the trash mailbox, used
+# to control the automatic expunction of mails via the ``mailExpungeTrash``
+# LDAP attribute.
+dovecot__ldap_trash_field: 'namespace/inbox/mailbox/Trash/autoexpunge'
```
imrejonk, line 1244: This used to be `namespace/inbox/Maildir/Trash/autoexpunge`, why the change?

ansible/roles/dovecot/tasks/main.yml
```diff
@@ -61,105 +69,79 @@
...
-    - [ '{{ "20-lmtp.conf" if "lmtp" not in dovecot_protocols else [] }}' ]
-    - [ '{{ [ "20-managesieve.conf", "90-sieve.conf", "90-sieve-extprograms.conf" ]
-            if "managesieve" not in dovecot_protocols else [] }}' ]
+- name: Generate Dovecot main configuration file
+  template:
+    src: '{{ lookup("template_src", "etc/dovecot/dovecot.conf.j2") }}'
+    dest: '/etc/dovecot/dovecot.conf'
+    owner: 'root'
+    group: 'root'
+    mode: '0644'
```
imrejonk: Octal permissions `0644` are not secure anymore now that the dsync password gets stored in dovecot.conf.
I suggest `0640` with `group: 'dovecot'`.

imrejonk:
I'm almost done with the review, I'll submit it later today. This PR is looking really good!
We have a staging mail server setup that I'm using as a playground for testing your redesign. I discovered a few bugs that I'll address in the review, but I got everything working without much hassle. I've tested IMAP access with implicit TLS (TCP port 993), LDAP lookups, dsync replication with TLS and password authentication, LMTP delivery from Postfix, Sieve filtering (using the Roundcube Sieve client) and of course overriding and adding some variables in `dovecot__*_configuration`. At the moment we're not using mail quotas or POP3, but it's nice to know it's in the role. Everything seems to work quite nicely now and I'm planning on running the playbook against our production environment at the beginning of next week. I will of course post an update afterwards to tell you how it went.

Here are the role default overrides that I'm currently using on my `cluster_mail_staging` Ansible inventory group, might be interesting to take a look at:

```
---
# We like tight control over the issued certificates, and want to restrict access to
# realms on a per-service basis as much as possible. Almost every service has
# their own PKI realm configuration.
dovecot__pki_realm: 'mail'

# We're not interested in IMAP+STARTTLS or quota support, but we want dsync.
dovecot__features: [ 'dsync', 'imaps', 'lmtp', 'sieve' ]

# We have a lot of respect for the Dutch National Cyber Security Center.
dovecot__ssl_cipher_list: '{{ dovecot__ssl_cipher_list_ncsc_nl }}'

# Our usernames are unique in our LDAP database and are perfectly fine for
# referencing Maildir locations. I like to keep my paths short.
dovecot__vmail_home: '/var/vmail/%u'

# We use the LDAP username, the mail domain must not be appended or this
# will break mail delivery.
dovecot__auth_default_realm: ''

# This is a workaround for allowing some services in our network to
# programmatically access their own IMAP accounts. These services
# are not "people" in the traditional sense, but as far as this role is
# concerned, they are.
dovecot__ldap_people_dn: '{{ dovecot__ldap_base_dn }}'

# We need a custom user filter here to make LDAP lookups for
# the above-mentioned service accounts work.
dovecot__ldap_user_filter: '(|
                              (&
                                (objectClass=inetOrgPerson)
                                (uid=%u)
                                (|
                                  (authorizedService=all)
                                  (authorizedService=mail:access)
                                )
                              )
                              (&
                                (objectClass=account)
                                (uid=%u)
                                (authorizedService=mail:access)
                              )
                              (&
                                (objectClass=account)
                                (uid=%n)
                                (host=%d)
                                (authorizedService=mail:send)
                              )
                            )'

# Set the "other" dsync host name. These are our actual staging mail servers.
dovecot__dsync_host: '{{ "mail5.ciphermail.com"
                         if ansible_fqdn == "mail6.ciphermail.com"
                         else "mail6.ciphermail.com" }}'

# I prefer shorter filesystem paths every day.
dovecot__dsync_password_path: 'dovecot/doveadm_password'

dovecot__group_configuration:

  - section: 'protocols'
    options:

      # Our usernames don't contain the domain part.
      - name: 'auth_username_format'
        value: '%n'

  - section: 'mailbox_namespaces'
    options:

      - name: 'namespace inbox'
        options:
          - name: 'mailbox Junk'
            options:
              # Our 
              - name: 'auto'
                value: 'subscribe'

  - section: 'plugins'
    options:

      - name: 'plugin'
        options:
          # We have a custom role that creates cluster-wide default Sieve filters
          # in the directories below. I'm thinking about integrating this functionality
          # in debops.dovecot after PR 1780 gets merged.
          # These filters are applied on all incoming mail for all users, and are
          # currently only used by us to move messages marked with
          # 'X-Spam: Yes' to the Junk folder.
          - name: 'sieve_before'
            value: '/etc/dovecot/sieve-before/'
          - name: 'sieve_after'
            value: '/etc/dovecot/sieve-after/'
```

Again, this is looking really good, thanks for all the hard work! :1st_place_medal: :1st_place_medal: :1st_place_medal:

ansible/roles/dovecot/tasks/main.yml
```diff
@@ -61,105 +69,79 @@
...
-    - [ '{{ "auth-ldap.conf.ext" if "ldap" in dovecot_user_accounts else [] }}' ]
-    - [ '{{ "auth-checkpassword.conf.ext" if "checkpassword" in dovecot_user_accounts else [] }}' ]
-    - [ '{{ [ "20-managesieve.conf", "90-sieve.conf", "90-sieve-extprograms.conf" ]
-            if "managesieve" in dovecot_protocols else [] }}' ]
+           --divert /etc/dovecot/dovecot.conf.dpkg-divert
+           --rename /etc/dovecot/dovecot.conf
+  args:
+    creates: /etc/dovecot/dovecot.conf.dpkg-divert
   notify: [ 'Restart dovecot' ]
+  tags: [ "role::dovecot:conf" ]
```
imrejonk: It would be nice if you could use the [`dpkg_divert`](https://github.com/debops/debops/blob/master/ansible/plugins/modules/dpkg_divert.py) module for this.

ansible/roles/dovecot/tasks/main.yml
```diff
@@ -61,105 +69,79 @@
...
-            if "managesieve" not in dovecot_protocols else [] }}' ]
+    - [ '10-auth.conf', '10-master.conf', '10-mail.conf', '10-ssl.conf', '15-lda.conf' ]
+    - [ '20-pop3.conf', '20-imap.conf', '20-lmtp.conf' ]
+    - [ '20-managesieve.conf', '90-sieve.conf', '90-sieve-extprograms.conf' ]
+    - [ 'auth-deny.conf.ext' ]
+    - [ 'auth-system.conf.ext' ]
+    - [ 'auth-sql.conf.ext' ]
+    - [ 'auth-ldap.conf.ext' ]
+    - [ 'auth-passwdfile.conf.ext' ]
+    - [ 'auth-checkpassword.conf.ext' ]
```
imrejonk: You can use the [`dpkg_divert`](https://github.com/debops/debops/blob/master/ansible/plugins/modules/dpkg_divert.py) module to remove diversions as well.

ansible/roles/dovecot/tasks/main.yml
```diff
@@ -61,105 +69,79 @@
...
-            if "managesieve" not in dovecot_protocols else [] }}' ]
+- name: Generate Dovecot main configuration file
+  template:
+    src: '{{ lookup("template_src", "etc/dovecot/dovecot.conf.j2") }}'
+    dest: '/etc/dovecot/dovecot.conf'
+    owner: 'root'
+    group: 'root'
+    mode: '0644'
   notify: [ 'Restart dovecot' ]
+  tags: [ "role::dovecot:conf" ]
```
imrejonk: This dovecot.conf templating task should probably be delayed until after the tasks for creating the 'vmail' user and group and other "dependent" tasks have been executed. It should probably be one of the last tasks. This way, when the playbook is interrupted, there are less errors for the administrator to deal with.

imrejonk (changes requested):
There are a few changes that I'd like you to make. Please let me know if anything about my feedback is not entirely clear.

Overall this redesign looks really good. I like that basically everything is in `dovecot.conf` now instead of in funny-named separate files in `conf.d`. The structure of the `dovecot.conf` file also looks well thought-out to me.

I should note that I tested your changes on the Ansible 2.10 version found in the Debian repositories, and with DebOps stable-2.3. I had to change the DebOps collection names (like `debops.debops.template_src`) back to their previous names to make this work with the older version of DebOps. I recommend against backporting this, but I'm quite sure @drybjed didn't plan to anyway ;)

Alphix:
@imrejonk Cool, thanks for the review. I'm working on updating the documentation right now. I'll go through your suggestions and fix them as well. I hope to finish both tasks before the end of the day.

ansible/roles/dovecot/defaults/main.yml
Alphix: Sure, done

ansible/roles/dovecot/defaults/main.yml
Alphix: Oof....fixed

ansible/roles/dovecot/defaults/main.yml
Alphix: Same as above

ansible/roles/dovecot/defaults/main.yml
Alphix: Unless I'm mistaken....the old value is incorrect? There is no "Maildir" parameter inside the "namespace inbox"?

ansible/roles/dovecot/tasks/main.yml
Alphix: Good idea, fixed

ansible/roles/dovecot/tasks/main.yml
Alphix: Oh, didn't know about that one....fixed

ansible/roles/dovecot/tasks/main.yml
Alphix: Same

ansible/roles/dovecot/tasks/main.yml
Alphix: Sure, I hope the new location is better (didn't know whether the `post_main` hook should go before or after the flushing of the event handlers).

Alphix:
@imrejonk - Rebased and pushed an updated version of the PR, only the last 5 patches are actually new. Still need to update the documentation and agree on whether `dovecot__ldap_trash_field` is incorrect or not :)

Alphix:
So to be clear for anyonw watching, docs are there as well now...just waiting for resolution of one review comment

ansible/roles/dovecot/defaults/main.yml
imrejonk: You're not mistaken, looks like the old value was indeed incorrect. Thanks for pointing it out.

ansible/roles/dovecot/defaults/main.yml
```diff
@@ -56,320 +59,410 @@ dovecot__version: '{{ ansible_local.dovecot.version|d("0.0.0") }}'
...
+# SQL Authentication Settings [[[
+# ---------------------------
 #
-# SQL query string to get the password. This function should return the values
-# ``username``, ``domain`` and ``password``.
-dovecot_sql_password_query: "SELECT userid AS username, domain, password FROM users WHERE userid = '%n' AND domain = '%d'"
 
-                                                                   # ]]]
-# .. envvar:: dovecot_sql_user_query [[[
+# .. envvar:: dovecot__sql__connect [[[
```
imrejonk: This should be `dovecot__sql_connect`. Note the single underscore between `sql` and `connect`. The double underscore is reserved for separating role names from other components of the variable name.
(same goes for the other `dovecot__sql__* variables`)

imrejonk: Pedantic: I noticed that you've put the "defaults-detailed" documentation inside the `defaults/main.yml` file. The convention within DebOps is to separate extended configuration details like variable parameters from this file. Then again I don't see how this could cause trouble for documentation readers, other than that the "defaults" documentation page becomes longer and perhaps more cluttered.

imrejonk (changes requested):

docs/ansible/roles/dovecot/getting-started.rst
```diff
@@ -11,55 +11,83 @@ Getting started
...
+``role::dovecot:conf:ldap``
+  `LDAP` specific configuration subtag.
+``role::dovecot:user``
+  Limited to :command:`dovecot` user configuration tasks.
+``role::dovecot:group``
+  Limited to :command:`dovecot` group configuration tasks.
+
+
+Default setup
+-------------
```
imrejonk: I think it would be better to have the "Default setup" and "Other resources" sections between "Example inventory" and "Example playbook", just to prioritise it more.

imrejonk:
I ran your role against the production mail servers at CipherMail today, it went really well. The cleanup of old configuration files works perfectly.

Documentation looks really good as well! Thanks for putting so much work in. I've started two review conversations that should be easily resolvable for you. After that, this PR gets my stamp of approval :+1:

docs/ansible/roles/dovecot/getting-started.rst
Alphix: Ok done

ansible/roles/dovecot/defaults/main.yml
Alphix: Done

Alphix:
@imrejonk and @drybjed ping? Is there anything left to be done before this PR can be merged?

imrejonk:
It looks really good to me. We’ve been using your revamp to deploy our staging and production environments at work, no problems at all. I was actually looking at your documentation earlier today and that is now high quality as well.

@drybjed merge this before 3.0 pretty please?

imrejonk (approved):

drybjed (approved):
That's really impressive amount of work, thank you! I'll not backport this, due to too many changes, but it will be appropriate to have it in v3.0.0 release. :-)

Alphix:
Yay! Thanks to both of you @drybjed and @imrejonk !

PR #1781, opened 2021-07-26
feat(apt_preferences): Preset: Prefer NeuroDebian but don't upgrade; feat(mariadb_server): Define ports used for Galera in /etc/services

ypid:


drybjed:
@ypid, so what's with that commit message? Are we gonna do JSON now in `git log` and just let people interpret what exactly that commit does?

ypid:
I started using `datalad run` for mechanical edits to record the exact commands I used to make that change. That allows to rerun the command in the future. This in turn allows to review big diffs. Consider I changed thousands of files. Humans are not good reviewing monotone changes as we saw in #1371. Instead, you can review the command and be sure the commit is the result of the command and nothing more. I used `datalad run` also for that commit to signal that I did not only fix the typo in one file, but made sure the same mistake does not exist in any other part of the monorepo.

I know `datalad` because it is in the ecosytem of my beloved `git-annex`.

Refs:

* https://handbook.datalad.org/en/latest/usecases/reproducible_neuroimaging_analysis.html
* https://handbook.datalad.org/en/latest/basics/basics-run.html
* http://docs.datalad.org/en/stable/generated/man/datalad-run.html

drybjed (approved):
Thanks for the updates! I'll backport the eligible ones to the stable releases as well.

PR #1782, opened 2021-07-27
[cryptsetup] Disable checksum computation

anzil:
Using very large files (hundreds of GB) as _ciphertext block device_ in `cryptsetup` leads to prohibitively long checksum computation times in the existence check of task _Check if ciphertext block device exists_ (which uses the [`stat` module](https://docs.ansible.com/ansible/latest/collections/ansible/builtin/stat_module.html#parameter-get_checksum)). 

This PR adds an optional flag to control the computation of the checksum (defaults to `True`).

### DIFF
--- a/docs/ansible/roles/cryptsetup/defaults-detailed.rst
+++ b/docs/ansible/roles/cryptsetup/defaults-detailed.rst
@@ -74,18 +74,6 @@ Each item of those lists is a dictionary with the following documented keys:
 
   Default to :envvar:`cryptsetup__use_uuid`.
 
-.. _cryptsetup__devices_check_with_checksum:
-
-``check_with_checksum``
-  Optional, boolean.
-  Compute checksum of the `ciphertext block device` given by
-  :ref:`item.ciphertext_block_device <cryptsetup__devices_ciphertext_block_device>`.
-
-  Setting this to ``False`` for (very large) regular files used as `ciphertext block device` 
-  avoids long checksum computation times.
-
-  Default to ``True``.
-
 .. _cryptsetup__devices_mode:
 
 ``mode``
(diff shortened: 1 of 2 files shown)


docs/ansible/roles/cryptsetup/defaults-detailed.rst
```diff
@@ -74,6 +74,18 @@ Each item of those lists is a dictionary with the following documented keys:
...
+
+``check_with_checksum``
+  Optional, boolean.
+  Compute checksum of the `ciphertext block device` given by
+  :ref:`item.ciphertext_block_device <cryptsetup__devices_ciphertext_block_device>`.
+
+  Setting this to ``False`` for (very large) regular files used as `ciphertext block device` 
+  avoids long checksum computation times.
+
+  Default to ``True``.
```
ypid: Do we actually need to make this configurable? Is the checksum used anywhere?

anzil: No, the checksum computed by the `stat` module is not used elsewhere in the `cryptsetup` role.

It would surely be possible to disable the checksum by default with `get_checksum: 'False'`. If that is ok I will update the PR accordingly. Please advise.

ypid: Yes, that is what I meant. Checksumming can be disabled. Thank you for bringing that up! I just forgot to set the best options for this module and never noticed …

anzil:
Based on above arguments the PR has been modified. Now, checksum computation is disabled by default.

ypid (approved):
LGTM, thanks!

drybjed (approved):

PR #1785, opened 2021-08-02
WIP [pki] Use dns-ovh certbot

julien-lecomte:
This is a Work-In-Progress / Proof of Concept of a pki certbot with the OVH DNS plugin.

The OVH DNS plugin allows one to generate certificates, and wildcard certificates, for domains via the OVH API by changing the DNS records to match it's challenge.

The current inventory configuration I use is:
```                                 
  pki_enabled: true                                 
  pki_acme_type: 'dns-ovh'                          
  pki_default_domain: 'ivy.example.at'              
  pki_inventory_groups: [ 'network_ivy_example_at' ]
                                                    
  pki_group_realms:                                 
    - name: '{{ pki_default_domain }}' # realm name 
      acme: true                                    
      acme_type: 'dns-ovh'                          
      acme_contacts: [ 'admin@example.at' ]         
      subject:  [ "cn={{ ansible_fqdn }}" ]         
      domains:  [ "{{ ansible_fqdn }} " ]           
      subdomains: [ "_wildcard_" ]                                                             
```

This configuration enables a let's encrypt certificat for "ivy.example.at, *,ivy.example.at"

I haven't yet tested many configurations.

The other option would be a certbot role with a PKI hook.


ansible/roles/pki/defaults/main.yml
```diff
@@ -252,19 +253,22 @@ pki_create_acme_challenge_dir: '{{ True if (ansible_local.nginx.acme|d() and
 # List of APT packages to install by default on the remote host for
 # cryptographic support.
 pki_base_packages: [ 'ssl-cert', 'make', 'ca-certificates',
-                     'gnutls-bin', 'openssl', 'acl' ]
+                     'gnutls-bin', 'openssl', 'acl'
+                   ]
```
drybjed: I'm not sure why indentation here changed, any ideas?

ansible/roles/pki/files/usr/local/lib/pki/pki-realm
```diff
@@ -1187,6 +1188,37 @@ request_acme_tiny_certificate () {
 
 }
 
+request_acme_dns_ovh_certificate() {
+    # We expect the API secrets to be in ./private/dns-ovh-credentials.key
+    local credentials=/etc/pki/realms/${config['name']}/private/dns-ovh-credentials.key
```
drybjed: This is actually a pretty clever way of handling the credentials. I haven't checked all the plugins yet, but I think they expect an auth key and optional `.ini` configuration file with other details, right? Maybe this could be generalized a bit to support other APIs and RFC2136 dynamic updates.

ansible/roles/pki/files/usr/local/lib/pki/pki-realm
```diff
@@ -1187,6 +1188,37 @@ request_acme_tiny_certificate () {
...
+    local letsencrypt=/etc/letsencrypt/live/${config['pki_default_fqdn']}
+    local pkirealm=/etc/pki/realms/${config['pki_default_domain']}
+    local email=""
+
+    if [ -n "${config['acme_contacts']}" ]; then
+        email="-m${config['acme_contacts']}"
+    fi
+ 
+    certbot certonly -n --dns-ovh --agree-tos "${email}" --dns-ovh-credentials ${credentials} -d${config['pki_default_fqdn']}
```
drybjed: The `--dns-ovh` and other such flags could be parametrized to support different API plugins. I think that there's just a handful of different "variants" of options, one being with just `--dns-*-credentials`, other with additional parameters. You can define different sets of parameters and select one command based on the plugin type.

BTW, does the `-d` parameter support multiple domains here? I assume not, is there a way to support that?

drybjed (changes requested):
Very interesting approach. I suppose that if we focus on DNS-based authentication with `certbot`, we don't need to worry about webserver support at all.

ansible/roles/pki/defaults/main.yml
julien-lecomte: Cuz' I screwed up ;-)

ansible/roles/pki/files/usr/local/lib/pki/pki-realm
julien-lecomte: They're all pretty much a flag + ini file except maybe the route53 case. The RFC2136 is also just a ini file and a flag (https://certbot-dns-rfc2136.readthedocs.io/en/stable/)

ansible/roles/pki/files/usr/local/lib/pki/pki-realm
julien-lecomte: Yes, you can specify multiple domains with -d.

In the case of OVH, if you specify foobar.example.com and *.example.com together, it will result in a error stating that foobar is already part of the wildcard

drybjed: No, I mean if it can be done using current `pki` role variables, since the `pki_default_fqdn` is global and based on the host's FQDN. You would probably have to gather domains from the realm configuration and merge them into one string separate by commas, then pass it to the `-d` parameter.

Dealing with wildcards could be done in the script as well, but we can do that later, when basic implementation works. We could filter out any subdomains if a wildcard is requested.

julien-lecomte: Basic implementation works, and does not yet deal with wildcard filter.

ser:
I fully support your efforts, i am using successfully acme.sh with knot-dns server.

ansible/roles/pki/files/usr/local/lib/pki/pki-realm
julien-lecomte: Current implementation now handles gathering the multiple domains then filtering out non-wildcards if needed and passing it along to the certbot.

PR #1792, opened 2021-08-10
[ldap] Update device iface regex to support systemd-nspawn veth

Alphix:
Systemd-nspawn containers which use virtual ethernet (veth) networking
will by default have interfaces named host0 (see
/lib/systemd/network/80-container-host0.network on a normal Debian
installation).


drybjed (approved):
Thanks for the update, I'll backport this to the stable releases as well.

PR #1795, opened 2021-08-11
Add info on used data type

prk0ghy:


imrejonk (changes requested):
I don't think this is correct, `pki_ca_organization` is actually a string. The value we put in at my place of work is 'CipherMail B.V.', for example.

I think the misconception here is that `pki_ca_domain.split(".")` results in a list that consists of the DNS labels in the pki_ca_domain value. The first DNS label is then taken (`| first`) and capitalized (`| capitalize`), as a simple way to extract an organization's name (a string) from their domain name. In my example above, the default `pki_ca_organization`'s value would be 'Ciphermail'.

I suggest that you change your pull request to state that the organization must be specified as a string.

PR #1796, opened 2021-08-12
[dhcpd] Quote host-identifier

imrejonk:
host-identifier parameters that contain periods (e.g. FQDNs) need to be
quoted to prevent this error:

```
/etc/dhcp/dhcpd6.conf line 50: semicolon expected.
    host-identifier option fqdn.fqdn secure7.
```


drybjed (approved):
Thanks for fixing the issue. :-) I'll backport it to the stable releases as well.

PR #1797, opened 2021-08-12
[sssd] sudo support fixes

Alphix:
sssd needs libsss-sudo to be installed and does not require the
separate functionality of sudo-ldap.


drybjed (approved):
Nice change, thanks. I'll backport this to the stable releases as well.

PR #1801, opened 2021-08-15
[debops] support security key SSH keys

Alphix:
OpenSSH supports FIDO U2F security keys via the ed25519-sk and ecdsa-sk
key types. They are denoted as "sk-ssh-*" keys in the output from
"ssh-add".

Example:
sk-ssh-ed25519@openssh.com <gibberish>
instead of:
ssh-ed25519 <gibberish> <comment>


drybjed (approved):
Nice work, thanks! I'll backport this to the stable release too.

PR #1803, opened 2021-08-15
[debops] Fix PKI hook uninstallation

Alphix:
The *_pki_hook_path variables point at the PKI hook directory,
not the actual pki hook files. Unless I'm mistaken, setting
a directory to "absent" would mean that Ansible would recursively
delete the whole directory.

The dovecot role has the same issue, but I didn't touch it for now,
already fixed in my separate dovecot rewrite branch.


drybjed (approved):
Thanks for catching this - indded, the whole PKI hook directory would be deleted. I'll backport this to the stable releases as well.

PR #1804, opened 2021-08-15
[debops] Fix version parsing in several facts.d scripts

Alphix:
I noticed that the facts.d output for redis_server was missing the
version information. The same query seemed to be used in several
places so I took the liberty of changing the other uses as well.

Note: only tested with redis_server.


drybjed (approved):
Thanks for updating the fact scripts. :-) I'll backport the changes to the stable releases.

PR #1805, opened 2021-08-16
automatically upgrade icinga2

lewer:
This commit enables automatic apt upgrade for icinga2 packages.

It would also automatically upgrade icingaweb2 (installed with debops.icingaweb2 role) since it is installed from the same apt repository. Is it a problem?

### DIFF
--- a/ansible/roles/icinga/defaults/main.yml
+++ b/ansible/roles/icinga/defaults/main.yml
@@ -882,4 +882,13 @@ icinga__ferm__dependent_rules:
     by_role: 'icinga'
     name: 'icinga_api'
                                                                    # ]]]
+# .. envvar:: icinga__unattended_upgrades__dependent_origins [[[
+#
+# List of origin patterns managed by the :ref:`debops.unattended_upgrades` role.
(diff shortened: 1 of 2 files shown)


ansible/roles/icinga/defaults/main.yml
```diff
@@ -882,4 +882,13 @@ icinga__ferm__dependent_rules:
...
     name: 'icinga_api'
                                                                    # ]]]
+# .. envvar:: icinga__unattended_upgrades__dependent_origins [[[
+#
+# List of origin patterns managed by the :ref:`debops.unattended_upgrades` role.
+icinga__unattended_upgrades__dependent_origins:
+
+  - origin: 'site=packages.icinga.com'
+    by_role: 'debops.icinga'
+    state: 'present'
```
imrejonk: It would be nice if the state of this origin is `absent` if `icinga__upstream` is `False`.

drybjed (approved):
Nicely done, thanks. I think that upgrades for icigaweb are fine, since users will want to have all the services synchronized. I'll backport this to the stable branches as well.

PR #1810, opened 2021-08-20
[tcpwrappers] Respect tcpwrappers__enabled

imrejonk:
The role will now skip all tasks related to configuring TCP Wrappers
when `tcpwrappers__enabled` is set to `False`.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -172,6 +172,12 @@ debops.boxbackup role
   service without it actually being present on the host. This should now be
   avoided by carefully checking the service status.
 
+:ref:`debops.tcpwrappers` role
+''''''''''''''''''''''''''''''
+
+- The role used to partially configure TCP Wrappers even when
+  ``tcpwrappers__enabled`` was set to ``False``.
+
 Removed
 ~~~~~~~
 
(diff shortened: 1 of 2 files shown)


drybjed:
The `tcpwrappers` role was designed when one of the accepted ways to add configuration to the `/etc/hosts.allow` file was to put a snippet in `/etc/hosts.allow.d/` directory and re-assemble the `/etc/hosts` file. That's why the tasks that create the directory are applied even if the `tcpwrappers` role is disabled, to not break other roles that might create or modify their snippets directly. I think that no roles in DebOps do that currently, so we can drop that requirement.

Having a fact script present in `/etc/ansible/facts.d/` on a given host can tell the administrator that a given role was applied on a host. The role's enabled/disabled status is stored in the fact, so this might be useful to other roles or to the sysadmin to know the state of the host configuration. Even if the `tcpwrappers` role is disabled, I would still like for it to manage its own fact script to indicate its state.

imrejonk:
> Having a fact script present in `/etc/ansible/facts.d/` on a given host can tell the administrator that a given role was applied on a host. The role's enabled/disabled status is stored in the fact, so this might be useful to other roles or to the sysadmin to know the state of the host configuration. Even if the `tcpwrappers` role is disabled, I would still like for it to manage its own fact script to indicate its state.

Good point! Fixed.

imrejonk:
@drybjed would you reconsider this PR for merging?

The role now:
- Still writes local facts when disabled, but does nothing otherwise.
- Works just like before when the role is enabled.
- Removes the `ALL: ALL` line before removing the 'allow' entries when disabling, preventing lockout.
- Doesn't touch /etc/hosts.allow.d when disabling.

PR #1813, opened 2021-08-24
- adding config parameter to deploy_state.

ser:
This additional and optional parameter allows advanced system operators to use untypical sets of nginx packages on their servers. In my case it is nginx-pagespeed which cannot be distributed as a binary package thus is unavailable for automatic installation.


drybjed (approved):
Thanks. Looks like it can be backported to the stable release as well.

Issue #1679, opened 2021-04-19
Log and metrics collection stack


ypid:
Log collection is what I do at work and am passionate about it. This issue can serve as place to discuss and define the architecture of logging in DebOps. I opened this ticket to not spam #1370 anymore.

Scope is server logs (DebOps) but also network devices (syslog).

Logs need to be parsed and sometimes enriched, lets look at actual options in this area:

Log (and metric) parser options:

* [Filebeat module](https://www.elastic.co/guide/en/beats/filebeat/current/filebeat-modules.html) with Ingest pipelines. Should probably be used for log types for which modules exist. This is vendor lookin. At least I see no valid reason why this is (only) implemented directly in Elasticsearch.
* [Logstash](https://www.elastic.co/logstash). A beast and very slow. I don’t see a future for it.
* [vector.dev](https://github.com/timberio/vector): My current favorite. Vector is already very well designed. It already crushes Logstash in terms of performance (startup and throughput), some features are missing but it is already usable and I would say it has a great future ahead. It also supports working with metrics which is good because sometimes, metrics are included in logs that you might handle outside of the log collection.
* [Rsyslog](https://www.rsyslog.com/). Advertisement is true, it is very fast and has a reasonable feature set. But it’s configuration language is difficult to work with, at least for me.

Other options that I only tried and ditched:

* [Logagent](https://sematext.com/logagent/)
* [Apache nifi](https://nifi.apache.org/)

That just as a summary.

PR #1817, opened 2021-08-31
- Telegraf role arrives to DebOps.

ser:
It's a simple role which helps to maintain Telegraf agent/proxy configuration across servers. It uses ```influxdata``` role to get APT packages from, but is independent from existing InfluxDB roles. It's projected to work with future InfluxDB 2.0 role but can be used with existing InfluxDB 1.x installations as well.

### DIFF
--- a/ansible/playbooks/service/telegraf.yml
+++ b/ansible/playbooks/service/telegraf.yml
@@ -1,5 +1,6 @@
 ---
-# Copyright (C) 2020 DebOps <https://debops.org/>
+# Copyright (C) 2021 Dr. Serge Victor <https://dr.sergevictor.eu/>
+# Copyright (C) 2021 DebOps <https://debops.org/>
 # SPDX-License-Identifier: GPL-3.0-or-later
 
 - name: Manage Telegraf instance
--- a/ansible/playbooks/service/telegraf.yml
+++ b/ansible/playbooks/service/telegraf.yml
@@ -1,0 +1,27 @@
+---
+# Copyright (C) 2020 DebOps <https://debops.org/>
+# SPDX-License-Identifier: GPL-3.0-or-later
+
+- name: Manage Telegraf instance
+  collections: [ 'debops.debops' ]
+  hosts: [ 'debops_service_telegraf' ]
+  become: True
+
+  environment: '{{ inventory__environment | d({})
+                   | combine(inventory__group_environment | d({}))
+                   | combine(inventory__host_environment  | d({})) }}'
+
+  roles:
+
+    - role: keyring
+      tags: [ 'role::keyring', 'skip::keyring', 'role::influxdata' ]
+      keyring__dependent_apt_keys:
+        - '{{ influxdata__keyring__dependent_apt_keys }}'
+
+    - role: influxdata
+      tags: [ 'role::influxdata', 'skip::influxdata' ]
+      influxdata__dependent_packages:
+        - '{{ telegraf__influxdata__dependent_packages }}'
+
+    - role: telegraf
+      tags: [ 'role::telegraf', 'skip::telegraf' ]
--- a/ansible/roles/telegraf/defaults/main.yml
+++ b/ansible/roles/telegraf/defaults/main.yml
@@ -2,13 +2,13 @@
 # .. vim: foldmarker=[[[,]]]:foldmethod=marker
 
 # .. Copyright (C) 2021 Dr. Serge Victor <https://dr.sergevictor.eu/>
-# .. Copyright (C) 2021 DebOps <https://debops.org/>
+# .. Copyright (C) 2020 DebOps <https://debops.org/>
 # .. SPDX-License-Identifier: GPL-3.0-or-later
 
 # .. _telegraf__ref_defaults:
 
 # debops.telegraf default variables
-# ========================================
+# =================================
 
 # .. contents:: Sections
 #    :local:
@@ -17,7 +17,7 @@
 
 
 # APT packages [[[
-# -------------------------------
+# ------------
 
 # .. envvar:: telegraf__base_packages [[[
 #
@@ -40,70 +40,50 @@ telegraf__packages: []
 # :file:`/etc/telegraf/telegraf.d` configuration directory.
 # Definition of minimum one input and one output is mandatory.
 
-# .. envvar:: telegraf__plugins_input [[[
+# .. envvar:: telegraf__configuration [[[
 #
-# Input plugins list
-telegraf__plugins_input:
-  - '{{ telegraf__input_proxy_default }}'
+# All plugins you want to run on all hosts
+telegraf__configuration:
+
+  - name: 'input_influxdb_local'
+    config: |
+      [[inputs.influxdb_v2_listener]]
+        service_address = "127.0.0.1:38086"
+    state: 'present'
+
+  - name: 'output_discard'
+    config: |
+      [[outputs.discard]]
+    state: 'present'
                                                                    # ]]]
 
-# .. envvar:: telegraf__plugins_output [[[
+# .. envvar:: telegraf__group_configuration [[[
 #
-# Output plugins list
-telegraf__plugins_output:
-  - '{{ telegraf__output_proxy_default }}'
-
-                                                                   # ]]]
-
-# .. envvar:: telegraf__plugins_aggregator [[[
-#
-# Aggregator plugins list
-telegraf__plugins_aggregator: []
-
-                                                                   # ]]]
-
-# .. envvar:: telegraf__plugins_processor [[[
-#
-# Processor plugins list
-telegraf__plugins_processor: []
+# All plugins you want to run in group.
+telegraf__group_configuration: []
 
                                                                    # ]]]
 
-# .. envvar:: telegraf__input_proxy_default [[[
+# .. envvar:: telegraf_host_configuration [[[
 #
-# Default dummy proxy input
-telegraf__input_proxy_default:
-  name: 'input_proxy_default'
-  content: |
-    [[inputs.influxdb_v2_listener]]
-      service_address = ":38086"
-  state: 'present'
+# All plugins you want to run in a particulr host.
+telegraf__host_configuration: []
 
                                                                    # ]]]
 
-# .. envvar:: telegraf__output_proxy_default [[[
 #
-# Default dummy proxy output
-telegraf__output_proxy_default:
-  name: 'output_proxy_default'
-  content: |
-    [[outputs.influxdb_v2]]
-  state: 'present'
-                                                                   # ]]]
-
 # .. envvar:: telegraf__combined_configuration [[[
 #
 # Full list of Telegraf configuration options passed to the
 # configuration template dir.
-telegraf__plugins_combined: '{{ telegraf__plugins_input +
-                                  telegraf__plugins_output +
-                                  telegraf__plugins_aggregator +
-                                  telegraf__plugins_processor }}'
+telegraf__configuration_combined: '{{ telegraf__configuration
+                                      + telegraf__group_configuration
+                                      + telegraf__host_configuration }}'
 
                                                                    # ]]]
                                                                    # ]]]
 # Configuration for other Ansible roles [[[
-# -----------------------------------------
+# -------------------------------------
 
 # .. envvar:: telegraf__influxdata__dependent_packages [[[
 #
--- a/ansible/roles/telegraf/defaults/main.yml
+++ b/ansible/roles/telegraf/defaults/main.yml
@@ -1,7 +1,8 @@
 ---
 # .. vim: foldmarker=[[[,]]]:foldmethod=marker
 
-# .. Copyright (C) 2020 DebOps <https://debops.org/>
+# .. Copyright (C) 2021 Dr. Serge Victor <https://dr.sergevictor.eu/>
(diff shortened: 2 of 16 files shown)


ansible/roles/telegraf/meta/main.yml
```diff
@@ -0,0 +1,25 @@
+---
+# Copyright (C) 2021 Dr. Serge Victor <https://dr.sergevictor.eu/>
```
drybjed, line 2: You should add this line to all files created in this PR.

ser: Done, thanks for pointing it out.

ansible/playbooks/service/telegraf.yml
```diff
@@ -0,0 +1,28 @@
+---
+# Copyright (C) 2021 Dr. Serge Victor <https://dr.sergevictor.eu/>
+# Copyright (C) 2021 DebOps <https://debops.org/>
+# SPDX-License-Identifier: GPL-3.0-or-later
+
+- name: Manage Telegraf instance
+  collections: [ 'debops.debops' ]
```
drybjed, line 7: The `influxdata` role is not included in the `debops.debops` collection, and can move between various `debops.rolesXY` collections, so you should include all of them here.

ansible/roles/telegraf/defaults/main.yml
```diff
@@ -0,0 +1,96 @@
...
+
+                                                                   # ]]]
+# .. envvar:: telegraf__packages [[[
+#
+# List of additional APT packages ta install with Telegraf.
+telegraf__packages: []
+
+                                                                   # ]]]
+                                                                   # ]]]
+
```
drybjed, line 35: The default is to include two "empty lines" between variables, which might or might not include `vim` fold markers. This is the end of section, so to close it properly two fold marksers are neede here, but the empty lines will show up when the entire file is folded. Remove the empty lines and leave just the ones with fold markers, then it will be just right.

ansible/roles/telegraf/defaults/main.yml
```diff
@@ -0,0 +1,96 @@
...
+  - name: 'input_influxdb_local'
+    config: |
+      [[inputs.influxdb_v2_listener]]
+        service_address = "127.0.0.1:38086"
+    state: 'present'
+
+  - name: 'output_discard'
+    config: |
+      [[outputs.discard]]
+    state: 'present'
```
drybjed, line 57: Are the above plugins useful in normal operations? Because with this setup, when users try to define their own configuration with `telegraf__configuration` variable in the inventory this configuration will be "masked" by Ansible merging the variables in the inventory. It might be better to move this configuration to `telegraf__default_configuration` variable and let users decide to explicitly remove these plugins via `telegraf__configuration` later on.

ansible/roles/telegraf/defaults/main.yml
```diff
@@ -0,0 +1,96 @@
...
+                                                                   # ]]]
+
+# .. envvar:: telegraf__group_configuration [[[
+#
+# All plugins you want to run in group.
+telegraf__group_configuration: []
+
+                                                                   # ]]]
+
```
drybjed, line 66: This line is not needed and will show up in the folded view.

ansible/roles/telegraf/defaults/main.yml
```diff
@@ -0,0 +1,96 @@
...
+      [[inputs.influxdb_v2_listener]]
+        service_address = "127.0.0.1:38086"
+    state: 'present'
+
+  - name: 'output_discard'
+    config: |
+      [[outputs.discard]]
+    state: 'present'
+                                                                   # ]]]
+
```
drybjed, line 59: You should move the fold marker one line down to not create empty lines in the folded view.

ansible/roles/telegraf/defaults/main.yml
```diff
@@ -0,0 +1,96 @@
...
+# .. envvar:: telegraf_host_configuration [[[
+#
+# All plugins you want to run in a particulr host.
+telegraf__host_configuration: []
+
+                                                                   # ]]]
+
+#
```
drybjed, line 74: These two lines are unnecessary.

ansible/roles/telegraf/defaults/main.yml
```diff
@@ -0,0 +1,96 @@
...
+# .. envvar:: telegraf__combined_configuration [[[
+#
+# Full list of Telegraf configuration options passed to the
+# configuration template dir.
+telegraf__configuration_combined: '{{ telegraf__configuration
+                                      + telegraf__group_configuration
+                                      + telegraf__host_configuration }}'
+
```
drybjed, line 82: This empty line should be removed.

ansible/roles/telegraf/defaults/main.yml
```diff
@@ -0,0 +1,96 @@
...
+# Configuration for other Ansible roles [[[
+# -------------------------------------
+
+# .. envvar:: telegraf__influxdata__dependent_packages [[[
+#
+# Configuration for the :ref:`debops.influxdata` Ansible role.
+telegraf__influxdata__dependent_packages:
+  - '{{ telegraf__base_packages }}'
+  - '{{ telegraf__packages }}'
+
```
drybjed, line 94: This empty line should be removed.

ansible/roles/telegraf/meta/main.yml
```diff
@@ -0,0 +1,26 @@
...
+# Copyright (C) 2021 DebOps <https://debops.org/>
+# SPDX-License-Identifier: GPL-3.0-or-later
+
```
drybjed, line 5: You should add collection information so that Ansible can find plugins and modules properly, usually it's formatted as:
```yaml
# Ensure that custom Ansible plugins and modules included in the main DebOps
# collection are available to roles in other collections.
collections: [ 'debops.debops' ]
```

ansible/roles/telegraf/handlers/main.yml
```diff
@@ -0,0 +1,9 @@
+---
+# Copyright (C) 2021 Dr. Serge Victor <https://dr.sergevictor.eu/>
+# Copyright (C) 2021 DebOps <https://debops.org/>
+# SPDX-License-Identifier: GPL-3.0-or-later
+
+- name: Restart Telegraf
+  service:
+    name: "telegraf"
+    state: "restarted"
```
drybjed, line 9: You should put the handler in the `global_handlers` role, that way other roles can integrate with the Telegraf service and reload/restart it if necessary. Also, usually handlers are called with lowercase service name, so "Restart telegraf". Remember to update the `notify` keywords as well.

Does Telegraf provide a way to check configuration for errors? It might be a good idea to check it before restarting to avoid disrupting the service in case of wrong configuration provided by the user.

ansible/roles/telegraf/tasks/main.yml
```diff
@@ -0,0 +1,80 @@
+---
+# Copyright (C) 2021 Dr. Serge Victor <https://dr.sergevictor.eu/>
+# Copyright (C) 2020 DebOps <https://debops.org/>
```
drybjed, line 3: Wrong year, we have 2021 now. :-) Check in other files as well.

ansible/roles/telegraf/tasks/main.yml
```diff
@@ -0,0 +1,80 @@
...
+
+- name: Import DebOps global handlers
+  import_role:
+    name: 'global_handlers'
+
+- name: Divert the original telegraf main configuration file
+  dpkg_divert:
+    path: '/etc/telegraf/telegraf.conf'
+    state: 'present'
+    delete: true
```
drybjed, line 18: In DebOps we use capitalized `True`/`False` for booleans, just like it's done in Python.

Since there is no mechanism to uninstall Telegraf implemented, the role will never remove the diversion. The `delete` parameter is not needed here.

ansible/roles/telegraf/tasks/main.yml
```diff
@@ -0,0 +1,80 @@
...
+    state: 'absent'
+  loop: '{{ telegraf__configuration_combined | parse_kv_items }}'
+  when: item.state|d('present') == 'absent' and item.config|d()
+  notify: [ 'Restart Telegraf' ]
+
+- name: Find plugins which are in filesystem
+  find:
+    paths: /etc/telegraf/telegraf.d
+    patterns: '*.conf'
+  register: telegraf_existing_plugins_filenames
```
drybjed, line 42: You should use namespacing and mark the variables as registers properly, so that they are eaiser to locate in the code. Rename this variable to `telegraf__register_existing_plugins_filenames`.

ansible/roles/telegraf/tasks/main.yml
```diff
@@ -0,0 +1,80 @@
...
+
+- name: Backup and disactivate plugins unexisting in desired configuration
+  command:
+    cmd: mv -f '{{ item.path | quote }}' '{{ item.path | quote }}.inactive'
+    removes: item.path
+  loop: '{{ telegraf_existing_plugins_filenames.files }}'
+  when: item.path | basename | splitext | first not in telegraf__configuration_combined | parse_kv_items | map(attribute="name")
+  notify: [ 'Restart Telegraf' ]
```
drybjed, line 50: Are you moving around files packaged in a `.deb` package? This will break things when the `telegraf` package is upgraded. Is that really needed? Wait, no, you are moving around configuration files... That's even worse, are these configuration files included with the package? Run `dpkg-query -s telegraf` and check if these are marked as Conffiles. If yes, then you should divert them instead so that `dpkg` can properly track them.

ansible/roles/telegraf/templates/etc/ansible/facts.d/telegraf.fact.j2
```diff
@@ -0,0 +1,25 @@
...
+
+def cmd_exists(cmd):
+    return any(
+        os.access(os.path.join(path, cmd), os.X_OK)
+        for path in os.environ["PATH"].split(os.pathsep)
+    )
+
+
+output = {}
+output['installed'] = cmd_exists('telegraf')
```
drybjed, line 23: You could add a version check here, either via `telegraf --version` if that's supported or via `dpkg-query`. With this, users can add configuration depending on the installed version, however to do that properly, the fact script needs to be set up before configuration files are generated.

docs/ansible/roles/telegraf/man_index.rst
```diff
@@ -0,0 +1,21 @@
+:orphan:
+
+.. Copyright (C) 2020 DebOps <https://debops.org/>
```
drybjed, line 3: Looks like not all files contain your copyright and updated year...

docs/ansible/roles/telegraf/real-world-example.rst
```diff
@@ -0,0 +1,124 @@
...
+          interfaces = ["eth0", "bridge0"]
+      state: 'present'
+
+    - name: 'input_zfs'
+      config: |
+        [[inputs.zfs]]
+          poolMetrics = true
+          datasetMetrics = true
+      state: 'present'
+
```
drybjed, line 124: Remove empty line at the end of the file.

docs/ansible/roles/telegraf/real-world-example.rst
```diff
@@ -0,0 +1,124 @@
...
+``state``
+    Optional. By default ``present``. If you declare it as ``absent``, this
+    plugin will be removed from your Telegraf instance configuration and
+    disactivated.
+
+If you configure your instance and later remove one or more of plugins from your
+ansible configuration, it will be automatically disactivated in the host
+and backuped to a filename ending ``.inactive``, for your future reference.  
+
+Examples
```
drybjed, line 57: Moving the "Examples" section before the "Syntax" section can quickly show users that already know Telegraf how to define configuration using the DebOps role, and to easily refresh their memory in the future.

docs/ansible/roles/telegraf/real-world-example.rst
```diff
@@ -0,0 +1,124 @@
...
+    https://docs.influxdata.com/telegraf/
+
+``state``
+    Optional. By default ``present``. If you declare it as ``absent``, this
+    plugin will be removed from your Telegraf instance configuration and
+    disactivated.
+
+If you configure your instance and later remove one or more of plugins from your
+ansible configuration, it will be automatically disactivated in the host
+and backuped to a filename ending ``.inactive``, for your future reference.
```
drybjed, line 55: Do you expect any other sources of Telegraf configuration apart from this role? If no, then deactivating will never happen because configuration files set to `absent` state will be removed by Ansible. The "disactivation" mechanism seems to not be needed in such case.

drybjed (changes requested):
I did an in-depth review of your new role, but I haven't tried it out yet. There are a few things to adress, some cosmetic and some important during execution.

Overall, looks like you did an excellent work for your first DebOps role. :-) BTW, you might need to "attach" the Telegraf playbook to the `site.yml` playbook somewhere, I suggest `ansible/playbooks/agent/` which will be executed at the end of the site playbook.

ser:
Thanks @drybjed !

ypid-geberit:
Maybe worth dropping the hint to you @ser as metrics and logs are getting more tied together (observability). I see people using https://vector.dev/ as a replacement for Telegraf aside from Vector being used as replacement for Logstash, Filebeat, Journalbeat and possibly also Rsyslog.

Ping @ypid: Update #1679 (ping to myself ;-)

PR #1818, opened 2021-09-04
Zabbix agent

julien-lecomte:


ansible/playbooks/srv/all.yml
```diff
@@ -106,3 +106,5 @@
 - import_playbook: libuser.yml
 
 - import_playbook: minidlna.yml
+
+- import_playbook: zabbix_agent.yml
```
drybjed: Perhaps a better fit would be the `agent.yml` playbook, which is executed late in the `site.yml`. It's meant for services like this which might want to contact the central server during deployment.

I will change this before merging the role.

ansible/roles/zabbix_agent/defaults/main.yml
```diff
@@ -0,0 +1,501 @@
...
+# User to use to run Zabbix agent
+zabbix_agent__user: 'zabbix'
+
+                                                                   # ]]]
+# .. envvar:: zabbix_agent__tls_psk_identity [[[
+#
+# Unique, case sensitive string used to identify the pre-shared key.
+zabbix_agent__tls_psk_identity: '{{ lookup("password", secret
+                       + "/zabbix/" + inventory_hostname
+                       + "/identity.key length=32 chars=ascii_letters,digits") }}'
```
drybjed: Perhaps a better path could be:
```
secret + "/zabbix/identities/" + inventory_hostname + "/identity.key"
```
This gives some wiggle room if we want to add more secret stuff in this role later.

I can change this before merging the PR.

ansible/roles/zabbix_agent/tasks/main.yml
```diff
@@ -0,0 +1,100 @@
...
+    unsafe_writes: '{{ True if (core__unsafe_writes|d(ansible_local.core.unsafe_writes|d()) | bool) else omit }}'
+  register: zabbix_agent__register_configuration
+
+- name: Create secret key
+  shell: 'test -f {{ item }} || openssl rand -hex 16 > {{ item }}'
+  args:
+    creates: '{{ item }}'
+  with_items: '{{ secret + "/zabbix/" + inventory_hostname + "/secret.key" }}'
+  delegate_to: localhost
+  become: false
```
drybjed: I don't think that this task is needed. If you use the `zabbix_agent__tls_psk_identity` variable, the lookup inside will be executed and the file will be created automatically. For installation on the remote host, you can use the `copy` task below with the `content:` parameter instead of `src:`, which should do the trick.

ansible/roles/zabbix_agent/templates/etc/zabbix/zabbix_agentd.conf.j2
```diff
@@ -0,0 +1,20 @@
...
+{%   if element.name|d() and element.state|d('present') not in [ 'absent', 'ignore' ] %}
+
+{%     for line in element.comment.split('\n') %}
+{{       '# {}'.format(line) }}
+{%     endfor %}
+{%     set element_comment = ('#' if element.state|d('present') == 'comment' else '') %}
+{{     '{}{}={}'.format(element_comment, element.name, element.value) }}
+{%   endif %}
+{% endfor %}
+
```
drybjed: Empty line can be removed here.

ansible/roles/zabbix_agent/templates/etc/zabbix/zabbix_agentd.conf.j2
```diff
@@ -0,0 +1,20 @@
...
+# {{ ansible_managed }}
+
+# This is a configuration file for Zabbix agent daemon (Unix)
+# To get more information about Zabbix, visit http://www.zabbix.com
+{% for element in zabbix_agent__combined_configuration | parse_kv_config %}
+{%   if element.name|d() and element.state|d('present') not in [ 'absent', 'ignore' ] %}
+
+{%     for line in element.comment.split('\n') %}
+{{       '# {}'.format(line) }}
+{%     endfor %}
```
drybjed: This can be made a bit simpler using the `comment()` filter. Also, if there are no comments, the template generation will break, so you should check if `element.comment` exists.

drybjed (commented):
Nice work with the role. :-) There are a few things I want to check with you before changing them, let me know what you think. I will also try and improve the configuration template a bit, to add support for booleans, etc. But very good job with Universal Configuration!

ansible/roles/zabbix_agent/tasks/main.yml
julien-lecomte: To be sure we're on the same page: the secret key is different than the identity key. The secret key is a file saved in /etc/zabbix_agent and the identity is the value of a configuration key in the zabbix_agent configuration file.

I'll change this to a copy task.

PR #1821, opened 2021-09-05
Change 'alpha' for 'ascii_letters' for password generation filter

julien-lecomte:


drybjed (approved):
Thanks! I'll add a note about security consideration of this change since it touches passwords.

PR #1822, opened 2021-09-06
Set correct value for mariadb__package_map

pbousek:
mariadb__package_map should contain exact values as mariadb__flavor


drybjed (approved):
Thanks, good catch. I'll backport this change to stable releases as well.

PR #1823, opened 2021-09-07
Apt uninstall role

julien-lecomte:


drybjed:
Nice work with the role, but I wonder if it's needed. Do you have a specific use case for it? Since you included it at the end of the `common.yml` playbook it's probably meant to remove some packages installed by default on each host.

I think that in the short term we could just add this functionality to the `apt_install` role - a few lists like `apt_install__remove` and a corresponding `apt` task which explicitly removes packages should be enough. In the long term I want to convert the package lists defined there to use Universal Configuration so that deciding which ones should be installed by default will be much easier.

If you want to remove something specific installed at provisioning time via the `bootstrap.yml` playbook, I think that we could add a similar simple list in the `apt` role, meant just for removing the initially provisioned packages. That way you can define them in the inventory as needed depending on your provisioning setup and they will be removed early on. Any thoughts?

julien-lecomte:
I initially thought that apt_install would suffice for my needs by using `state: absent`, but that didn't work. So, I thought of adding the removal functionality to apt_install, but I would have needed to respect the way lists are built in that role, with the specific lookup rules.

So I thought the easiest and simplest way to get this functionality was a new role.

drybjed:
I see. In that case I can add a simple list of packages to remove in the `apt` role useful during bootstrapping right away, which will close this PR. After that I'll refresh the `apt_install` role to use the new configuration in a day or two, after dealing with other PRs. Sounds good?

julien-lecomte:
Sounds great!

PR #1834, opened 2021-09-10
apt: add keys tag

JohannesKleine:
Sometimes the apt role stops at the first step and the only thing i would like: add some keys.


drybjed (approved):
Sounds good, thanks. By stopping at first step, do you mean the first `apt` task in the role? That might mean issues with the package manager. If you encounter it again, check the APT state on that host, for example via `apt-cache update`.

PR #1840, opened 2021-09-14
[rabbitmq_server] Fix per account vhost creation

gaudenz:
The per account vhost lookup template added a vhost for each vhost name,
node name and tracing setting defined on the account. This is fixed to
create single vhost with the correct settings.


href (approved):

drybjed (commented):
The proposed solution leaked the `omit` value used by Ansible, which shouldn't happen. I designed a different fix added via a separate PR which should solve the issue.

PR #1843, opened 2021-09-16
[reprepro] Add missing architectures

imrejonk:
This adds the expected architectures for Bookworm, and adds the missing
architectures for older releases. The lists of official architectures
were obtained from the release notes. The list of architectures for
Bookworm was obtained from the current Bookworm Release file (which is
subject to change).

Adding the Bookworm architectures is necessary in order to make reprepro
work on Debian 11. This is because Bookworm is configured as the 'next'
release, and if that release doesn't have architecture information the
reprepro commands will fail.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -362,6 +362,12 @@ Fixed
 - Fixed an issue with facts not showing Redis instances correctly when password
   is empty.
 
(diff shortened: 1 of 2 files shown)


drybjed (approved):

Issue #1544, opened 2020-10-30
[debops.postfix] maincf configuration partially appied


francois-claude:
When using this configuration, options 'smtp_sasl_password_maps' and 'smtp_sasl_tls_security_options' are not applied to main.cf on the host. 

`postfix__maincf:

  - name: 'smtp_sasl_auth_enable'
    value: True

  - name: 'smtp_tls_security_level'
    value: 'encrypt'

  - name: 'smtp_sasl_tls_security_options'
    value: 'noanonymous'

  - name: 'relayhost'
    value: '[smtp.gmail.com]:587'

  - name: 'smtp_sasl_password_maps'
    value: [ 'hash:${config_directory}/smtp_sasl_password_maps' ]`

please advise

francois-claude:
the missing values are added if state: 'present' is added to each variable

drybjed:
Documentation has been amended, closing.

PR #1847, opened 2021-09-18
Add missing state parameter

rgevaert:
Without the settings are not applied.

See also #1544


drybjed (approved):
Thanks for updating the documentation. :-)

PR #1851, opened 2021-09-20
[elastic_co] Set default version to 7.x

imrejonk:
I have tested this with a single-node Elasticsearch installation that we
use for a Graylog system. Upgrading from Elasticsearch 6.8 to 7.10 went
smoothly. No configuration changes were necessary in this case.


ypid (approved):

drybjed (approved):
Looks good, thanks. :-)

PR #1858, opened 2021-09-23
Improve handling of openssl backup keys in autopostgresql backup

Cadair:
I am trying to use the encryption support in autopostgreql backup, but I couldn't figure out how I was supposed to pass they key around, or why it wasn't working. I feel like I am maybe missing a trick here, was the intent that I copy the cert over to the host explicitly? At least like this debops does it all for you.

This also tweaks the backup script so that it only removes the unencrypted file if the encrypted file is present in the expected location rather than just relying on the exit status of openssl.

### DIFF
--- a/ansible/roles/postgresql_server/templates/usr/sbin/autopostgresqlbackup.j2
+++ b/ansible/roles/postgresql_server/templates/usr/sbin/autopostgresqlbackup.j2
@@ -291,11 +291,11 @@ encryption() {
     echo "      to $ENCRYPTED_FILE"
     echo "      using cypher $ENCRYPTION_CIPHER and public key $ENCRYPTION_PUBLIC_KEY"
 
-    $tempcrt=/tmp/autopostgresql.crt
-    echo "$ENCRYPTION_PUBLIC_KEY" > $tmpcrt
+    tempcrt=/tmp/autopostgresql.crt
+    echo "$ENCRYPTION_PUBLIC_KEY" > "$tempcrt"
     openssl smime -encrypt -"$ENCRYPTION_CIPHER" -binary -outform DEM \
-      -out "$ENCRYPTED_FILE" -in "$1" "$tmpcrt"
-    rm $tmpcrt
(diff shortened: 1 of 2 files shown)


ansible/roles/postgresql_server/templates/usr/sbin/autopostgresqlbackup.j2
```diff
@@ -290,11 +290,19 @@ encryption() {
     echo "Encrypting $1"
     echo "      to $ENCRYPTED_FILE"
     echo "      using cypher $ENCRYPTION_CIPHER and public key $ENCRYPTION_PUBLIC_KEY"
-    if openssl smime -encrypt -"$ENCRYPTION_CIPHER" -binary -outform DEM \
-      -out "$ENCRYPTED_FILE" \
-      -in "$1" "$ENCRYPTION_PUBLIC_KEY" ; then
+
+    tempcrt=/tmp/autopostgresql.crt
```
drybjed: We can have more than one `autopostgresqlbackup` scripts running. Perhaps it would be better to go with
```bash
tempcrt="$(tempfile)"
```
here so that we know that the filename is unique and not taken by anything else?

drybjed (dismissed):
That's a nice solution, thanks! I'll change the static filename to `tempfile` output and it should work fine.

ansible/roles/postgresql_server/templates/usr/sbin/autopostgresqlbackup.j2
Cadair:
```
In ./ansible/roles/postgresql_server/templates/usr/sbin/autopostgresqlbackup.j2 line 294:
    tempcrt=$(tempfile)
              ^------^ SC2186: tempfile is deprecated. Use mktemp instead.
```

me and shellcheck have not been getting on :rofl:

drybjed:
During testing it was discovered that `openssl` has issues with encrypting large files, a better method needs to be implemented.

PR #1859, opened 2021-09-24
[icinga_web] upgrade businessprocess module

lewer:
Business process module has buggy CSS in its 2.2.0 version (see https://github.com/Icinga/icingaweb2-module-businessprocess/blob/73ad4a25ddfb6e0e0914d65109f3a83a6f876d34/public/css/module.less#L386)

It makes the whole UI crash.

The typo was fixed in version 2.3.1.


drybjed (approved):
Thanks! I'll backport this to the stable releases since it's a bugfix.

PR #1866, opened 2021-10-07
add certbot manual support to pki role

pbousek:


drybjed (approved):
Looks good, thanks. I'll update the certbot documentation in the role a bit to point users to the `pki_acme_type` variable.

PR #1871, opened 2021-10-11
fix: python3-toml is now a DebOps dependency

ypid:


drybjed (approved):
Nice work, thanks! I'll backport this to the stable releases as well.

PR #1872, opened 2021-10-11
feat(netbox): Upgrade to NetBox v3.0.7

ypid:
I have been testing it for a few days. No issues so far.


drybjed (approved):
Great, thanks for checking if the upgrade works! :-)

PR #1874, opened 2021-10-12
[icinga_web] Update modules and watch files

imrejonk:
- Update all modules to the latest versions that are supported with
  Icinga Web 2.8 (the version in Debian 11).

- Fix the version number (and regex) specified in the watch-incubator
  file; the version number was incorrectly set to 0.7.0, a version that
  currently does not exist.

- Create a watch file for the x509 module.


drybjed (approved):
Awesome, thank you! :-) I'll fix the merge conflict manually.

PR #1877, opened 2021-10-15
[apt] Do not disable old backports repositories

imrejonk:
Debian LTS and older releases indeed don't need the -updates
repositories anymore, but the -backports repositories can still be
useful. The golang role for example installs Go from backports on
Stretch.


drybjed (approved):
Sounds good, thanks. I'll make a backport to the stable releases as well.

PR #1878, opened 2021-10-15
[sysctl] Enable forwarding on Docker Server hosts

imrejonk:
Docker Server needs packet forwarding for container networking to work. If this
role runs after the Docker daemon has started, it will reset the packet
forwarding parameters, breaking container networking.

This commit fixes that by enabling packet forwarding on Docker Server hosts.


drybjed (approved):
Sounds like a good solution, thanks. I'll backport this PR to the stable releases.

PR #1889, opened 2021-10-23
[wpcli] Dump versions from 2.3.0 to 2.5.0

### LINKED ISSUE #1820: wpcli role - why does it install version which is two years old?
wpcli role has hardcoded version into defaults - why? 

The docs on main site read how to get current version:

```
curl -O https://raw.githubusercontent.com/wp-cli/builds/gh-pages/phar/wp-cli.phar
```

Why the author decided to hardcode the version instead of getting a current one?

alinalexandru:
Will close #1820


drybjed (approved):
Nicely done, thanks! :-)

PR #1890, opened 2021-10-23
[dnmasq] Remove dvcpv6 tcp protocol from ferm

### LINKED ISSUE #1760: dnsmasq: default ferm firewall rules for dhcpv6 break on debian (at least bullseye)
Hi,

i tried rolling out dnsmasq on an Debian/bullseye testing host and could not get ferm to work correctly.
To reproduce I can simply run:
`debobs -l TESTHOST`
with TESTHOST in groups "debops_service_dnsmasq" and "debops_service_ifupdown" (amongst others).
The command results in an error for task `[ferm : Restart ferm]`:

>     Unable to start service ferm: Job for ferm.service failed because the control process exited with error code.
>    See "systemctl status ferm.service" and "journalctl -xe" for details.

According to the log on TESTHOST this is because:
>  Jul 07 11:10:11 TESTHOST ferm[70577]: Starting Firewall: ferm
>  Jul 07 11:10:11 TESTHOST ferm[70710]: ip6tables v1.8.7 (legacy): invalid port/service `dhcpv6-client' specified
>  Jul 07 11:10:11 TESTHOST ferm[70710]: Try `ip6tables -h' or 'ip6tables --help' for more information.
>  Jul 07 11:10:12 TESTHOST ferm[70580]: Firewall rules rolled back.
>  Jul 07 11:10:12 TESTHOST ferm[70722]:  failed!

Running ferm manually for /etc/ferm/rules.d/140_rule_dhcpv6.conf results in the same error.

The problem seems to be located in 
https://github.com/debops/debops/blob/379eaeca0dab49577b6776428db757262fb86ec6/ansible/roles/dnsmasq/defaults/main.yml#L608

... which specifies TCP and UDP for the default rule. According to `/etc/services`, the Port/Protocol uses UDP only, hence iptables complains.

If I override the default dnsmasq role with an local copy and edit this line to 
`       protocol: [ 'udp' ]`
everything works as expected.

Thanks for your response/efforts.

alinalexandru:
Will close #1760

- In `dhcpd` only udp post is set
- In `ferm` only udp post is set
- It was removed from /etc/services starting Debian 11 and Ubuntu 20

### DIFF
--- a/ansible/roles/dnsmasq/defaults/main.yml
+++ b/ansible/roles/dnsmasq/defaults/main.yml
@@ -605,7 +605,7 @@ dnsmasq__ferm__dependent_rules:
     saddr: [ 'fe80::/10' ]
     daddr: [ 'ff02::1:2' ]
     # https://tools.ietf.org/html/rfc3315#section-13
-    protocol: [ 'udp', 'tcp' ]
+    protocol: [ 'udp' ]
     sport: [ 'dhcpv6-client' ]
     dport: [ 'dhcpv6-server' ]
     interface: '{{ dnsmasq__combined_interfaces | flatten | parse_kv_items


drybjed (approved):
Thanks for the update, I'll backport it to the stable releases as well.

PR #1892, opened 2021-10-24
Fix name of the scrape_yahoo script when called from cron

djs52:
See https://github.com/stevejenkins/postwhite -- this logs an error every week on my system


drybjed (approved):
Oops, sorry about that! I hope that this works better now. I'll make a backport to the stable releases as well.

PR #1893, opened 2021-10-25
[ferm] disable iptables backend symlink on Ubuntu Bionic 18.04

lewer:
Ubuntu Bionic 18.04 is compatible with nftables but **iptables** package does not include `/usr/sbin/iptables-nft` nor `/usr/sbin/iptables-legacy` so it makes sense to just disable it.


alinalexandru:
probably focal needs to be added. 
I have some strange behaviour on the VM running gitlab runner

PR #1894, opened 2021-10-26
Java policy available starting  ES 6

alinalexandru:
That file doesn't exists for versions below 6.


imrejonk (changes requested):
`version_compare` was [renamed](https://docs.ansible.com/ansible/latest/user_guide/playbooks_tests.html#comparing-versions) to `version` in Ansible 2.5.
Otherwise: great PR, thanks for the fix!

drybjed (approved):
Sounds good, thanks. I'll update the `version_compare` filter and make a backport to older releases.

PR #1895, opened 2021-10-26
fix #1888, use iptables-legacy by default

### LINKED ISSUE #1888: use iptables legacy be default
ferm doesn't work well with iptables/ebtables-nft and starting with bullseye it uses legacy tools directly, regardless of the alternatives setting: https://github.com/MaxKellermann/ferm/issues/47

nft tools are used by default which is confusing as they show "empty" rules

I believe ferm role should change alternatives for iptables and ebtables to legacy versions.

532910:
Ferm 2.5.1-1 (bullseye) uses legacy tools
regardless of the alternatives setting:
https://github.com/MaxKellermann/ferm/issues/47
Thus, if ferm is used, iptables should point to iptables-legacy
to avoid getting confused by displaying "empty" rules.

Closes: #1888

Signed-off-by: Sergio E. Nemirowski <sergio@outerface.net>

### DIFF
--- a/ansible/roles/ferm/defaults/main.yml
+++ b/ansible/roles/ferm/defaults/main.yml
@@ -60,7 +60,11 @@ ferm__iptables_backend_enabled: '{{ False
 #
 # Newer OS releases might need to use the legacy variant to be compatible with
 # :command:`ferm` manager.
-ferm__iptables_backend_path: '/usr/sbin/iptables-nft'
+# :command:`ferm` doesn't work well with nft
+# and starting with bullseye, it uses legacy tools directly,
+# regardless of the alternatives setting:
+# https://github.com/MaxKellermann/ferm/issues/47
+ferm__iptables_backend_path: '/usr/sbin/iptables-legacy'
 
                                                                    # ]]]
 # .. envvar:: ferm__base_packages [[[


imrejonk:
Closes: #1888

Edit: ah that doesn't work with GitHub comments, I think you have to put it in the pull request description or the commit message.

imrejonk (dismissed):

532910:
> I think you have to put it in the pull request description or the commit message.

But this was done initially:

```
% git log --pretty=oneline --abbrev-commit -1
1bf66b874 (HEAD, origin/iptables-legacy) fix #1888, use iptables-legacy by default
```

And now it shows up as linked. Perhaps it was some kind of github issue?

ansible/roles/ferm/defaults/main.yml
```diff
@@ -60,7 +60,7 @@ ferm__iptables_backend_enabled: '{{ False
 #
 # Newer OS releases might need to use the legacy variant to be compatible with
 # :command:`ferm` manager.
```
532910, line 62: > Select which iptables backend should be used on the host.
> Newer OS releases might need to use the legacy variant to be compatible with

I tried to figure out how this comment should be updated, but came to the conclusion that this option should be renamed, as it has nothing common with ferm. Yes, on buster ferm follows alternative settings, but this is an error and fixed in bullseye. This way there should be no choice: if ferm is used legacy should be used. [Ferm will never support nftables.](https://github.com/MaxKellermann/ferm/issues/35)

PR #1905, opened 2021-11-02
[sshd] Increase LoginGraceTime to 60s

### LINKED ISSUE #1904: [sshd] Default LoginGraceTime conflicts with UseDNS setting
The default LoginGraceTime, the time after which the OpenSSH server disconnects the client if it fails to present a valid key or password, is 30 seconds in DebOps. The upstream default is 120 seconds. I think the rationale here was that a shorter LoginGraceTime better prevents against a type of DoS attack where the attacker opens up many SSH connections without actually authentication, thereby starving the server.

The `debops.sshd` role by default also enables UseDNS, which tells the server to attempt verifying the remote host name through DNS. This is disabled by upstream by default. I am not sure why we enabled this feature, it could have something to do with the pam_access configuration (although I am not sure if the use of UseDNS actually has any effect there given that PAM is not OpenSSH).

A problem occurs when the DNS resolvers don't respond or are slow to respond. The default configuration in `<resolv.h>` in glibc is to wait five seconds before a different resolver is attempted, and retry each resolver once more before an error is returned. This is configurable through [resolv.conf(5)](https://manpages.debian.org/resolv.conf.5). If you have three resolvers configured (the glibc maximum), the worst-case scenario is that the glibc resolver returns an error after a full 30 seconds. The LoginGraceTime has now expired and the server disconnects you. If the server you are trying to get into provides its own DNS resolver, you now have a nice catch-22 to deal with. Ask me how I know ;)

So this is a kind of RFC, I propose that we set the LoginGraceTime to 60 seconds (30 seconds feels really short to me). This makes the LoginGraceTime the same as the default console LOGIN_TIMEOUT which is configured in /etc/login.defs. Administrators who are targeted by the kind of DoS attacks described above can always shorten it themselves.
I also propose that we disable the UseDNS setting, unless there is a good reason to keep it enabled.

imrejonk:
Mitigates a lock-out issue when DNS resolvers are unreachable.
Closes: #1904

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -474,6 +474,13 @@ debops.reprepro role
 - The rsyslog role always configured the streamDriverPermittedPeers option,
   even when the ``anon`` network driver authentication mode was selected.
 
+:ref:`debops.sshd` role
+'''''''''''''''''''''''
+
+- The default :envvar:`sshd__login_grace_time` has been increased from 30 to 60
+  seconds. This mitigates a lock-out issue when :envvar:`sshd__use_dns` is
+  enabled (the default) and your DNS resolvers are unreachable.
+
 :ref:`debops.sudo` role
 '''''''''''''''''''''''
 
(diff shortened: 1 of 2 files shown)


imrejonk:
Looks like some CI dependency had a breaking update recently...

imrejonk:
Manually running `make test` revealed an error in my changelog entry. This has been fixed.

drybjed (approved):
Good idea, thanks. I'll fix the conflict in the Changelog and backport this to the stable releases.

PR #1910, opened 2021-11-04
[icinga] no longer generate icinga_web password

### LINKED ISSUE #1909: [icinga] Role creates password for icinga_web
The `debops.icinga` role can currently create the password of the Icinga Director API user, using the 'password' lookup. See [`icinga__director_register_api_password`](https://docs.debops.org/en/master/ansible/roles/icinga/defaults/main.html#envvar-icinga__director_register_api_password). The password must however be managed by the `debops.icinga_web` role because it needs to configure the API credentials. Both roles having "write access" to the Icinga Director password has proven confusing (ask @stefangweichinger).

I propose that we only let the `debops.icinga_web` role create the password, so that `debops.icinga` can show a more helpful error if it does not exist.

imrejonk:
The 'file' lookup won't generate a password, but simply fail when it
can't find the Director API password.

Closes #1909

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -398,6 +398,16 @@ Fixed
   needed when the host-identifier contains periods (e.g. fully qualified
   domain names).
 
+:ref:`debops.icinga` role
+'''''''''''''''''''''''''
+
+- The role no longer generates the Icinga Director API password right before
+  trying to log in with it. That is now left to the :ref:`debops.icinga_web`
+  role, which is actually in charge of the API authentication setup. If the
+  password does not yet exist (presumably because the Icinga Director has not
+  been installed yet), the task will report a sensible error message and
+  playbook execution will continue.
+
 :ref:`debops.ipxe` role
 '''''''''''''''''''''''
 
--- a/ansible/roles/icinga/defaults/main.yml
+++ b/ansible/roles/icinga/defaults/main.yml
@@ -289,14 +289,22 @@ icinga__director_register_api_url: 'https://{{ icinga__director_register_api_fqd
 icinga__director_register_api_user: 'director-api'
 
                                                                    # ]]]
+# .. envvar:: icinga__director_register_api_password_file [[[
+#
+# Path to the file on the Ansible Controller that contains the Icinga Director
+# API password.
+icinga__director_register_api_password_file: '{{
+  secret + "/icinga_web/api/" + icinga__director_register_api_fqdn
+  + "/credentials/" + icinga__director_register_api_user + "/password" }}'
+
+                                                                   # ]]]
 # .. envvar:: icinga__director_register_api_password [[[
 #
 # The password of the Icinga 2 Director REST API user used to register the host
 # in Icinga. This variable corresponds to the
 # :envvar:`icinga_web__director_api_password` variable.
-icinga__director_register_api_password: '{{ lookup("password", secret + "/icinga_web/api/"
-                                            + icinga__director_register_api_fqdn + "/credentials/"
-                                            + icinga__director_register_api_user + "/password") }}'
+icinga__director_register_api_password: '{{
+  lookup("file", icinga__director_register_api_password_file) }}'
 
                                                                    # ]]]
 # .. envvar:: icinga__director_register_default_templates [[[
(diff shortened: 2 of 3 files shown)


imrejonk:
Marked as draft: I thought this change was trivial, but there's more to it than meets the eye. I'll have to do some more testing before I'm comfortable with a merge.

drybjed:
This is a chicken and egg issue. We need to install `icinga` (I think...) on a host before we install `icinga_web` with the director. If using the `file` lookup here results in a failed execution of the `icinga` role, the user needs to re-run all the roles separately. Use of `password` lookup in both roles is equivalent - no matter which role creates the password, it should be the same one.

From my point of view this seems more to be a failure of properly documenting something.

imrejonk:
After testing some more, I'm now quite convinced that we should make this change.

There is indeed a chicken-and-egg issue here. The Icinga Director installation needs to be configured before any host that is *not* a master can be registered, because the Icinga Director API credentials are needed for this operation.This is true regardless of which role creates the password in the secrets directory, i.e., `debops.icinga` will fail at the 'Register Icinga node in Icinga Director' task anyway. The problem is in the error output that is presented to the user:

```
TASK [debops.debops.icinga : Register Icinga node in Icinga Director] **********
fatal: [icingaclient.imrejonk.nl]: FAILED! => changed=false 
  censored: 'the output has been hidden due to the fact that ''no_log: true'' was specified for this result'
```

The Icinga Director API password is created by `debops.icinga`, but the register task will fail when the Icinga Director service is not set up yet. The error message is not helpful. When disabling `no_log`, the problem becomes a bit more apparent:

```
TASK [debops.debops.icinga : Register Icinga node in Icinga Director] **********
fatal: [icingaclient.imrejonk.nl]: FAILED! => changed=false 
  elapsed: 0
  msg: 'Status code was -1 and not [201, 422, 500]: Request failed: <urlopen error [Errno 111] Connection refused>'
  redirected: false
  status: -1
  url: https://icinga.imrejonk.nl/director/host
```

Now for the 'file' lookup:

```
[WARNING]: Unable to find '/home/imre/src/imrejonk/debops-
imre/ansible/secret/icinga_web/api/icinga.imrejonk.nl/credentials/director-
api/password' in expected paths (use -vvvvv to see paths)

TASK [debops.debops.icinga : Register Icinga node in Icinga Director] **********
fatal: [icingaclient.imrejonk.nl]: FAILED! => 
  censored: 'the output has been hidden due to the fact that ''no_log: true'' was specified for this result'
```

The warning about the missing password file is a bit more helpful I think. Ideally we would perform some check as part of the register task to see whether the password file exists (much like how `debops.ldap` checks for the presence of the LDAP admin password). We could then either:

1. Fail hard like we do now, but with a helpful error message.
2. Skip the registration task and hope the user runs the `icinga` playbook again after the Director server has been set up, like we do with `debops.ldap`.

Which has your preference?

drybjed:
Making a check for the password file and skipping director registration seems like a better idea. That way the host with the director can be configured with `icinga`, `icinga_db` and `icinga_web` roles in sequence, then `icinga` role can be re-run again. We can mention this in the role documentation as well.

stefangweichinger:
Great to see progress here, I had issues with that register-process multiple times.

imrejonk:
Okay, so I've added a password file presence check. If the Icinga Director API password file does not exist, the register task is allowed to fail (i.e. throw error but continue play). This is helpful for finding out why an Icinga node is not registering.

The register task will still run if the password file does not exist. The idea is that the password can also be supplied in some other way than using a file on the Ansible Controller (it is just a variable, after all). In that case, you want the register task to run as well.

Regarding documentation: that should be covered, the deployment guide already tells you to deploy the Icinga master node first.

drybjed:
I'm not sure if `ignore_errors` is a good way to deal with this. The issue with that keyword is that Ansible will ignore any and all errors encountered in the task, not just "file with password is not found". Since we already check if the variable with the password is empty or not, why not move that check into the `when` keyword? That way the task will be skipped and not emit a warning and notice that it was ignored.

drybjed:
I think that #1997 should improve Icinga deployment enough that using the password check to decide how the `icinga` role behaves shouldn't be needed anymore. I apologize that resolving this took so long, entirely my fault. I would love to talk with @stefangweichinger at some point about the current state of Icinga deployment, let's catch up on IRC at some point. :-)

stefangweichinger:
@drybjed : Sure, why not .. although I don't use it actively right now, that installation idles along without attention.
Didn't rollout anything for months there. And I am prepping for a vacation starting at aug,11th

PR #1913, opened 2021-11-05
[backup2l] Add option to disable mail sending from cron (Fix #1806)

### LINKED ISSUE #1806: [backup2l] Add possibility to disable mail sending from crontab
Packages for backup2l [installs](https://packages.ubuntu.com/focal/all/backup2l/filelist) a [`/etc/cron.daily/zz-backup2l`](https://github.com/gkiefer/backup2l/blob/master/zz-backup2l) which says *Redirect its output if you do not want automatic e-mails after each backup.*

This could be an option offered by Debops role.

mathieumd:
As suggested by @imrejonk.

### DIFF
--- a/ansible/roles/backup2l/templates/etc/cron.daily/zz-backup2l.j2
+++ b/ansible/roles/backup2l/templates/etc/cron.daily/zz-backup2l.j2
@@ -1,0 +1,10 @@
+#!/bin/bash
+
+# The following command invokes 'backup2l' with the default configuration
+# file (/etc/backup2l.conf).
+#
+# (Re)move it or this entire script if you do not want automatic backups.
+#
+# Redirect its output if you do not want automatic e-mails after each backup.
+
+! which backup2l > /dev/null || nice -n 19 backup2l -b{% if backup2l__no_cron_emails %} >/dev/null{% endif %}
--- a/ansible/roles/backup2l/templates/etc/cron.daily/zz-backup2l.j2
+++ b/ansible/roles/backup2l/templates/etc/cron.daily/zz-backup2l.j2
@@ -1,4 +1,9 @@
+{# Copyright (C) 2021 DebOps <https://debops.org/>
+ # SPDX-License-Identifier: GPL-3.0-only
+ #}
 #!/bin/bash
+
+# {{ ansible_managed }}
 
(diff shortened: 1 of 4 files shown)


mathieumd:
For ShellCheck "*This { is literal. Check expression (missing ;/\n?) or quote it.*" ([here](https://github.com/debops/debops/runs/4117633115?check_suite_focus=true)), I'm not sure how to fix them.

ansible/roles/backup2l/templates/etc/cron.daily/zz-backup2l.j2
```diff
@@ -0,0 +1,15 @@
+{# Copyright (C) 2021 DebOps <https://debops.org/>
+ # SPDX-License-Identifier: GPL-3.0-only
+ #}
+#!/bin/bash
```
drybjed: The shell scripts need to be parsable by `shellcheck` so that it can detect issues in the code. You should move the copyright block into the shell script itself (comment with normal `#` characters), and remove the Jinja comment section. Obviously the shebang needs to be first.

ansible/roles/backup2l/templates/etc/cron.daily/zz-backup2l.j2
```diff
@@ -0,0 +1,15 @@
...
+# {{ ansible_managed }}
+
+# The following command invokes 'backup2l' with the default configuration
+# file (/etc/backup2l.conf).
+#
+# (Re)move it or this entire script if you do not want automatic backups.
+#
+# Redirect its output if you do not want automatic e-mails after each backup.
+
+! command -v backup2l > /dev/null || nice -n 19 backup2l -b{% if backup2l__no_cron_emails %} >/dev/null{% endif %}
```
drybjed: This command will not pass through `shellcheck`. You should define a variable with `backup2l__no_cron_emails` as a value, and then on the shell level interpret that value.

I think that compressing logic like this in the shell script is counter-intuitive. The scripts are meant to be understandable by humans that audit the code. I would go with a traditional `if` / `else` block here.

drybjed (changes requested):
The idea for discarding e-mails sounds good but the solution still needs some improvement. See the comments about the shell script.

PR #1917, opened 2021-11-07
[imapproxy] Add new role

Alphix:
imapproxy is useful together with, in particular, webmail clients since
it caches open connections to the IMAP server which improves performance
and cuts down on log spam generated by repeated logins from the webmail
client.

Some stats (15 reloads using snappymail web mail on a local VM setup
using DebOps roles, time measured with Google Chrome developer tools
network finish time on reload):

```
        No proxy        Proxy
        ========        =====
Mean:   529 ms		211 ms
Min:    492 ms          194 ms
Max:    566 ms          261 ms
Stdev:	21 ms           17 ms
```

If this role is merged, I can prepare a separate PR to make the
roundcube role use imapproxy automatically when available.

### DIFF
--- a/ansible/playbooks/app/all.yml
+++ b/ansible/playbooks/app/all.yml
@@ -42,3 +42,5 @@
 - import_playbook: debops_api.yml
 
 - import_playbook: roundcube.yml
+
+- import_playbook: imapproxy.yml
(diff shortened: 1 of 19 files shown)


drybjed (approved):
Awesome job with the role! I reviewed the code and everything seems in order. :) I'll add a GitLab CI test and a Changelog entry.

PR #1919, opened 2021-11-07
[postfix] Put all milter_* parameters in the right section

Alphix:
Postfix supports a large number of milter_* parameters, this makes
sure that they all end up in the "filter" section.


drybjed (approved):
Nice work, thanks! :-) I'll backport this to the stable releases.

PR #1920, opened 2021-11-10
debops-init instead

JOduMonT:
on ubuntu 20.04 I had to execute: debops-init instead


drybjed:
Thanks, but the documentation you used is for the `master` branch in which the scripts were modified, but not released yet. You should check the `stable-2.3` branch of the documentation to see the working examples:

https://docs.debops.org/en/stable-2.3/introduction/getting-started.html

You can switch branches at the bottom left of the documentation page.

PR #1923, opened 2021-11-13
system_groups: add throttle variable

JohannesKleine:


drybjed (approved):
Good to know that setting this via a variable works. Thanks for checking. :) I'll backport this to the stable releases.

PR #1930, opened 2021-11-19
[docker_server] Enable live restore by default

imrejonk:
Live restore reduces downtime for containers during a Docker daemon restart.

https://docs.docker.com/config/containers/live-restore/


drybjed (approved):
Sounds good. I'll backport this to the stable releases as well.

PR #1931, opened 2021-11-19
[grub] Enable serial console by default

imrejonk:
The serial console is incredibly useful when working in the data center: instead of fiddling with the monitor trolley and having to wire up power, display and keyboard, you can just take out your laptop and plug in a USB to serial adapter.

Virtualization tools like Ganeti also depend on the (virtual) serial console of the guest, and cloud platforms like Azure require it for console access and dmesg logging.

I don't see any downsides to enabling it by default.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -257,6 +257,11 @@ Continuous Integration
   and the options changed by the role are set in the
   :envvar:`elasticsearch__default_configuration` variable.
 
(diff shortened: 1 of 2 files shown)


drybjed (approved):
Agreed, serial console will be useful in many environments, especially if networking configuration is broken. I'll backport the change to the stable releases as well.

PR #1933, opened 2021-11-19
Use admin email address facts from debops.core in more roles

imrejonk:


drybjed (approved):
Sounds good to me. I don't think that the note in the `upgrades.rst` file is needed here, the final value will most likely not change at all. I'll remove it, so that the changes can be backported to stable releases.

PR #1935, opened 2021-11-24
[sudo] Make use of sudo__ldap_enabled

imrejonk:
- Let sudo__ldap_enabled decide if package `sudo` or `sudo-ldap` gets installed.
- Let sudo__ldap_enabled decide if the LDAP account object gets created.


drybjed (approved):
Good idea, thanks. These checks are also present in the `sudo__ldap_enabled` variable which makes them redundant here. I'll backport this to the stable releases.

PR #1936, opened 2021-11-24
[sshd] Only create LDAP account when needed

imrejonk:
An LDAP account for sshd is only needed when authorized key lookups are
to be performed over LDAP.


drybjed (approved):
Looks good after taking a closer look. Thanks. :) I'll backport this PR to the stable releases.

PR #1938, opened 2021-11-25
[libvirtd] No longer add root to 'libvirt' group

imrejonk:
The role used to add the user executing the Ansible tasks to the
'libvirt' group, even if that user was root. The root account does not
need 'libvirt' membership in order to access libvirtd. Also, adding the
root account to any group causes an idempotency issue with
`debops.root_account`, which enforces the group membership of the root
account.


drybjed (approved):
Nicely done, thanks. I'll backport this to the stable releases.

PR #1939, opened 2021-11-25
feat(apt_install): Install `intel-microcode` from buster-backports

ypid:
I would be interested if you get the same "downgrading" warning or if that is something wrong with my machines. I did get it on multiple machines. Also of course if we should install `intel-microcode` from buster-backports by default.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -249,6 +249,11 @@ Continuous Integration
 
 - The role defaults have been updated, Bullseye is the new Stable.
 
+:ref:`debops.apt_install` role
+''''''''''''''''''''''''''''''
+
+- Install ``intel-microcode`` from buster-backports for improved security.
+
 :ref:`debops.elasticsearch` role
 ''''''''''''''''''''''''''''''''
 
--- a/ansible/roles/apt_install/defaults/main.yml
+++ b/ansible/roles/apt_install/defaults/main.yml
@@ -459,6 +459,12 @@ apt_install__apt_preferences__dependent_list:
     backports: [ 'wheezy', 'jessie', 'trusty' ]
     reason: 'Better support for container technologies'
     by_role: 'debops.apt_install'
+
+  - package: 'intel-microcode'
+    backports: [ 'buster' ]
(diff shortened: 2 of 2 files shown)


imrejonk:
I get the same downgrade warning when trying to install `intel-microcode` from buster-backports when the version from buster was installed previously. The reason for this is simple: the version of `intel-microcode` in buster-backports is older than the version in buster and buster/updates. It is also older than the version found in bullseye. In other words: the backport is outdated and should not be used because it does not contain [the latest security updates](https://www.debian.org/security/2021/dsa-4934).

```
root@buster:/# apt policy intel-microcode 
intel-microcode:
  Installed: 3.20210608.2~deb10u1
  Candidate: 3.20210608.2~deb10u1
  Version table:
 *** 3.20210608.2~deb10u1 500
        500 http://deb.debian.org/debian buster/non-free amd64 Packages
        500 http://deb.debian.org/debian-security buster/updates/non-free amd64 Packages
        100 /var/lib/dpkg/status
     3.20201118.1~bpo10+1 100
        100 http://deb.debian.org/debian buster-backports/non-free amd64 Packages
```

You mention in your commit that installing `intel-microcode` from backports is needed to get security updates. This is a misconception. The Debian Security Team does publish advisories and updated packages for `intel-microcode`, even though packages in contrib and non-free are not part of the Debian distribution and therefore [not officially covered](https://www.debian.org/security/faq#contrib) by them. See for example the [latest DSA](https://www.debian.org/security/2021/dsa-4934) regarding `intel-microcode`.

Note that the Debian Security Team only covers the packages in `stable/main`, and packages in `oldstable/main` for up to one year after the release of stable. After that, the [Debian LTS Team](https://wiki.debian.org/LTS) takes over. Some packages are excluded from security support, details of which you can find in the [debian-security-support](https://packages.debian.org/search?keywords=debian-security-support) package. The -backports repositories are not covered. Any security-related updates to a backported package are provided on a best-effort basis by the maintainer of that backport. There is no backup if a maintainer fails to update their package.

ypid:
Thanks for your very detailed answer. It was a simple fail on my side. I confused `3.20210608.2~deb10u1` and `3.20201118.1~bpo10+1`. It probably has to do with my preference to 2021-06-08 with hyphens for ISO 8601 dates. I have always considered 20210608 to be less readable than 2021-06-08.

PR #1943, opened 2021-11-29
Fix ansible-lint reported errors

ssbarnea:
This change future-proofs the code by including changes required by
newer versions of Ansible.

Note: these errors were found due to eco pipeline that ansible-lint has.
See https://github.com/ansible-community/ansible-lint/blob/main/playbooks/eco.yml#L35-L37

### DIFF
--- a/ansible/roles/etc_services/tasks/main.yml
+++ b/ansible/roles/etc_services/tasks/main.yml
@@ -40,7 +40,7 @@
     owner: 'root'
     group: 'root'
     mode:  '0644'
-  with_flattened: '{{ etc_services__combined_list }}'
+  loop: '{{ etc_services__combined_list | flatten }}'
   when: (etc_services__enabled|bool and
          ((item.name|d() and item.port|d()) or item.custom|d()) and
          (item.state is undefined or item.state != 'absent'))
--- a/ansible/roles/foodsoft/docs/playbooks/foodsoft.yml
+++ b/ansible/roles/foodsoft/docs/playbooks/foodsoft.yml
@@ -3,4 +3,4 @@
 # Copyright (C) 2016-2017 DebOps <https://debops.org/>
 # SPDX-License-Identifier: GPL-3.0-only
 
-- include: foodsoft-nginx.yml
+- import_playbook: foodsoft-nginx.yml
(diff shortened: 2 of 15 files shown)


drybjed:
Interesting, thanks for the fix.

Can you explain how the `requirements.yml` file is supposed to be used? Can the list of collections be added in the `galaxy.yml` file instead, so that these collections are dependencies and will be installed automatically by `ansible-galaxy`?

ssbarnea:
`ansible-lint` is able to detect dependencies from both`galaxy.yml` and `requirements.yml` if those are present at **root** of your project. Basically you can add runtime deps to both, but more important to `galaxy.yml` one. The `requirements.yml` file can also include testing specific requirements in addition to runtime ones.

Sadly `debops` project does not have these files are root of the repo, so the linter would not be able to find them. Using a symlink is also an option.

Keep in mind that passing custom paths to the linter is not advices as this will not work with other tools that may need a linter config file, or look-up for ansible config files. One notable example is https://marketplace.visualstudio.com/items?itemName=redhat.ansible 

That is why I recommend using a layout that does not need custom locations, so all/most tools will just work without extra headaches.

I think that it may be possible to trick ansible-lint to work only with a root folder symlink `galaxy.yml -> ansible/collections/ansible_collections/debops/debops/galaxy.yml` -- Keep in mind that you cannot chain two symlinks, ansible-galaxy will perform an "install" but not the stuff you want. I tried that myself and discovered entire repo being copied.

drybjed:
As I thought, `requirements.yml` file location at the root of the repo is significant... Sigh, when collections were thought up I raised a concern that using the root of the git repository for collections can create issues when Ansible code was merged with some other project (yes, like DebOps itself, for example) and I wished for Ansible to support its code being namespaced to `ansible/` natively... But oh well, it doesn't.

I found a similar issue with the `galaxy.yml` wile while trying to make installation of collection directly from the git repository work. The `ansible-galaxy` tool supports subdirectories in the URL format, that's why `ansible/galaxy.yml` symlink now exists there.

Will just `requirements.yml` in the root of the repo suffice? I don't want to symlink `galaxy.yml` all over the place, this might cause issues as you mentioned. I'm currently fixing the SPDX test issue, and I will add the required collections to `galaxy.yml`, both the static one and generated one. I'll make a separate PR with fixes.

BTW, did you found out the list of required collections using only `ansible-core`, or is that just a ballpark and more collections might be needed for DebOps to work? I guess that I should try to use it just with `ansible-core` to find that out properly...

PR #1945, opened 2021-11-29
feat(owncloud): Upgrade to Nextcloud 21

ypid:
Sorry for the delay. I am late this time (https://github.com/nextcloud/server/wiki/Maintenance-and-Release-Schedule). You can merge this after a day I would say.


ypid:
I found no issues after upgrading to NC 21 with this.

drybjed (approved):
Nice work, thanks. The `python*` packages in `homeassistant__base_packages` should technically be installed via the `python` role (this would allow for Python 2.7 environments to work as well), but since the `homeassistant` role is not yet in the main DebOps playbook, I'll allow it. :-)

Issue #1525, opened 2020-10-04
mariadb_server: Delete anonymous database user fails on Ubuntu 20.04


vaisov:
When I run the playbook to install MariaDB server it fails on Ubuntu 20.04 with following error:
```TASK [mariadb_server : Delete anonymous database user] ***************************************************************************************************************************************
[WARNING]: Module did not set no_log for update_password
failed: [db.domain.com] (item=mariadb-primary) => {"ansible_loop_var": "item", "changed": false, "item": "mariadb-primary", "msg": "unable to connect to database, check login_user and login_password are correct or /root/.my.cnf has the credentials. Exception message: (1698, \"Access denied for user 'root'@'localhost'\")"}
failed: [db.domain.com] (item=localhost) => {"ansible_loop_var": "item", "changed": false, "item": "localhost", "msg": "unable to connect to database, check login_user and login_password are correct or /root/.my.cnf has the credentials. Exception message: (1698, \"Access denied for user 'root'@'localhost'\")"}
```

Vars are set following:
```mariadb_mysql__flavor: mariadb
mariadb_server__upstream_version: 10.3
```

I guess this could be related to this change: https://code.jpod.cc/debops/debops-upstream/-/commit/e076346b53c123d0184432406b893cea537656bd
So after this change it tries to connect using hostname instead of socket. 

I did work this around by adding `login_unix_socket: /run/mysqld/mysqld.sock` to `debops/roles01/roles/mariadb_server/tasks/secure_installation.yml`.

alemonmk:
Not only `mariadb_server`, `tasks/manage_contents.yml` from `mariadb` is also affected.

Gomez:
Got this problem, too. Is login_unix_socket just a work around or should we add it?

jptreen:
Side point for someone else coming to this fresh who just wants a working installation: In Ubuntu 20.04, 10.3 is the default MariaDB, and I've had no issue switching from
`mariadb_server__flavor: mariadb_upstream` to `mariadb_server__flavor: mariadb`, which installs 10.3 painlessly.

Obviously no help to someone who needs 10.4 or 10.5, but it made for a slightly cleaner config in my case and sidestepped this bug.

tblein:
Same problem on fresh, a Debian Buster install, with MariaDB 10.5 from the official Debian repository.

To solve it, I applied the suggestion of @vaisov for `mariadb_server` role to remove the anonymous database user but also the test database.

```diff
[...]

PR #1950, opened 2021-12-08
[mariadb] add login_unix_socket parameter for ansible mysql connector

bleuchtang:
In mariadb >=10.4 default local connection for root is only available through file socket. Since Debian Bullseye come with mariadb 10.5 we should fix every local connection to pass through this socket.

Documentation of authentification mecanism for mariadb >= 10.4 can be found here:
https://mariadb.com/kb/en/authentication-from-mariadb-104/

I didn't fix every debops connection. For instance I didn't fix librenms connection:
https://github.com/debops/debops/blob/c1db9901ac2dd18cf24d2472fc2ff8c2ef7238ad/ansible/roles/librenms/tasks/main.yml#L110
I don't know if I should fix every connection in debops repository or should we do one PR by specific connection ?!

This PR has been tested on Debian Buster and Bullseye.

Related issue: https://github.com/debops/debops/issues/1525

### DIFF
--- a/ansible/roles/icinga_web/tasks/main.yml
+++ b/ansible/roles/icinga_web/tasks/main.yml
@@ -166,6 +166,7 @@
     name: '{{ icinga_web__database_name }}'
     state: 'import'
     target: '{{ item }}'
+    login_unix_socket: '/run/mysqld/mysqld.sock'
   with_items:
     - '{{ icinga_web__database_schema }}'
     - '/tmp/icingaweb-initial-data.sql'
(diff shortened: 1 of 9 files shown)


bleuchtang:
bump :)

drybjed (approved):
Thanks for fixing the connections, and yes - doing this in a single PR is fine. The socket path should probably be configuration via the `mariadb` role, but let's go with the default for now. It can be updated later, when the MariaDB roles are refreshed.

I'll fix the conflicts and backport the changes to stable releases.

PR #1953, opened 2021-12-16
Accept raw Component parameters to Prosodoy role

BerkhanBerkdemir:
Prosody configuration accepts some parameters under the Component description. With this change, this will be added like what happens in the VirtualHost configuration.

Please check prosody [documentation] for more information about the configuration.

[documentation]: https://prosody.im/doc/configure


BerkhanBerkdemir:
Hi @drybjed! Just wanted to let you know that is there anything I can do for this PR to be merged?

BerkhanBerkdemir:
Any update on this PR from @drybjed?

Alphix:
@BerkhanBerkdemir : without any documentation, how would a user even know there is a raw parameter? (in general it seems like the whole role is in need of a documentation overhaul...)

BerkhanBerkdemir:
> @BerkhanBerkdemir : without any documentation, how would a user even know there is a raw parameter? (in general it seems like the whole role is in need of a documentation overhaul...)

The point you made is right. I will work in my own fork until I write documentation. If anyone is interested in this, then they can have the patch from the fork.

BerkhanBerkdemir:
For future reference, there was also no Prosody component documentation so far as it should be on [`prosody/defaults/main.yml`](https://github.com/debops/debops/blob/master/ansible/roles/prosody/defaults/main.yml#L292-L309).

PR #1956, opened 2021-12-18
[pdns] Add new role debops.pdns

imrejonk:
This adds debops.pdns, a role for managing the PowerDNS Authoritative
Server. Configuration is based on the principles of Universal
Configuration and has been made explicitly forward-compatible up to
PowerDNS 4.5. Currently the role installs PowerDNS from the Debian
repositories, configures the gpgsql (PostgreSQL) backend by default and
performs a few overrides that make pdns more friendly to the DebOps
userbase:

- DNSSEC is automatically enabled for the gpgsql backend, but needs
  additional per-domain metadata to function. Hopefully this makes
  deploying DNSSEC a little more attractive.
- The IP address of the internal stub resolver is set automatically.
- The API key is created automatically when the API is enabled.
- The HTTP port is set to 16836 instead of 8081, which is reserved in
  Debian's /etc/services file.
- If the embedded webserver (for metrics and API usage) is enabled, a
  local NGINX server will be configured as a reverse proxy so that the
  webserver can be accessed securely over TLS.
- Dynamic DNS Update (RFC 2136) is automatically enabled, but needs
  additional per-domain configuration in the form of domain metadata to
  function. This makes pdns a little easier to use together with dhcpd.

### DIFF
--- a/ansible/playbooks/service/pdns.yml
+++ b/ansible/playbooks/service/pdns.yml
@@ -1,40 +1,8 @@
 ---
-# Copyright (C) 2020 CipherMail B.V. <https://www.ciphermail.com/>
-# Copyright (C) 2020 DebOps <https://debops.org/>
+# Copyright (C) 2022 CipherMail B.V. <https://www.ciphermail.com/>
+# Copyright (C) 2022 DebOps <https://debops.org/>
 # SPDX-License-Identifier: GPL-3.0-or-later
 
-- name: Manage PowerDNS authoritative server
-  hosts: [ 'debops_service_pdns' ]
-  become: True
+- import_playbook: pdns-plain.yml
 
-  environment: '{{ inventory__environment | d({})
-                   | combine(inventory__group_environment | d({}))
-                   | combine(inventory__host_environment  | d({})) }}'
-
-  roles:
-
-    - role: etc_services
-      tags: [ 'role::etc_services', 'skip::etc_services' ]
-      etc_services__dependent_list:
-        - '{{ pdns__etc_services__dependent_list }}'
-      when: pdns__api or pdns__metrics
-
-    - role: ferm
-      tags: [ 'role::ferm', 'skip::ferm' ]
-      ferm__dependent_rules:
-        - '{{ pdns__ferm__dependent_rules }}'
-
-    - role: postgresql
-      tags: [ 'role::postgresql', 'skip::postgresql' ]
-      postgresql__dependent_roles:
-        - '{{ pdns__postgresql__dependent_roles }}'
-      when: ('gpgsql' in pdns__backends)
-
-    - role: nginx
-      tags: [ 'role::nginx', 'skip::nginx' ]
-      nginx__dependent_servers:
-        - '{{ pdns__nginx__dependent_servers }}'
-      when: pdns__nginx_allow and (pdns__api or pdns__metrics)
-
-    - role: pdns
-      tags: [ 'role::pdns', 'skip::pdns' ]
+- import_playbook: pdns-nginx.yml
--- a/ansible/playbooks/service/pdns.yml
+++ b/ansible/playbooks/service/pdns.yml
@@ -1,0 +1,40 @@
+---
+# Copyright (C) 2020 CipherMail B.V. <https://www.ciphermail.com/>
+# Copyright (C) 2020 DebOps <https://debops.org/>
+# SPDX-License-Identifier: GPL-3.0-or-later
+
+- name: Manage PowerDNS authoritative server
+  hosts: [ 'debops_service_pdns' ]
+  become: True
+
+  environment: '{{ inventory__environment | d({})
+                   | combine(inventory__group_environment | d({}))
+                   | combine(inventory__host_environment  | d({})) }}'
+
+  roles:
+
+    - role: etc_services
+      tags: [ 'role::etc_services', 'skip::etc_services' ]
+      etc_services__dependent_list:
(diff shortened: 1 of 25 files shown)


imrejonk:
This is currently a draft pull request; I'm comfortable with a merge after I've taken this on the production ride, which will probably be next week. Any preliminary feedback is of course greatly appreciated.

imrejonk:
I should note that I have only tested this role on Bullseye LXC containers so far. Our production authoritative nameservers run on Buster, so I might have to make some adjustments for that. We'll see.

imrejonk:
Another thing to note: I'm importing the playbook from `srv/all.yml` instead of `net/all.yml`. The reason for this is that pdns relies on a backend to function, and the default backend I'm configuring with this role is PostgreSQL Server, whose playbook is imported earlier in `srv/all.yml`. The same goes for the NGINX playbook (needed for the HTTP reverse proxy).

imrejonk:
The production run went fine :rocket: 

@drybjed is it okay if I set the `min_ansible_version` tag in `meta/main.yml` to `0.0.0`? I don't actually know what the minimum Ansible version for this role would be, I've only tested it with 2.10 (and don't plan to test it with anything older).

ansible/roles/pdns/tasks/init_postgresql.yml
```diff
@@ -0,0 +1,21 @@
...
+  register: pdns__register_postgresql_status
+
+- name: Import initial database schema
+  postgresql_db:
+    login_user: '{{ pdns__postgresql_role }}'
+    login_password: '{{ pdns__postgresql_password }}'
+    name: '{{ pdns__postgresql_database }}'
+    target: '{{ pdns__postgresql_schema }}'
+    state: 'restore'
+  when: pdns__register_postgresql_status is changed  # noqa no-handler
```
ypid, line 21: This file looks like a job for `debops.postgresql`. I am not sure if schema import is supported, but the role could be extended.

ansible/playbooks/service/pdns.yml
```diff
@@ -0,0 +1,40 @@
...
+    - role: etc_services
+      tags: [ 'role::etc_services', 'skip::etc_services' ]
+      etc_services__dependent_list:
+        - '{{ pdns__etc_services__dependent_list }}'
+      when: pdns__api or pdns__metrics
+
+    - role: ferm
+      tags: [ 'role::ferm', 'skip::ferm' ]
+      ferm__dependent_rules:
+        - '{{ pdns__ferm__dependent_rules }}'
```
ypid: For example nginx dependent_rules are missing here. I guess with this design of role dependencies on the playbook level, you would have to check all roles what dependencies they have.

ypid (changes requested):
I had a quick look out of interest. Overall it looks very good.

> I've only tested it with 2.10 (and don't plan to test it with anything older).

Debian 11 ships with Ansible 2.10 so that it not an unreasonable version to ask for. I would be fine when you set `min_ansible_version` to 2.10 to document that the role is fully supported starting with 2.10.

ansible/roles/pdns/tasks/init_postgresql.yml
imrejonk: `debops.postgresql` doesn't support schema imports, sadly. I'll take a look at extending it, but it's a bit outside the scope of this PR so I'll open a new one. @drybjed can decide whether the `debops.pdns` role can be merged as-is or if we should do the `debops.postgresql` fix first.

ansible/playbooks/service/pdns.yml
imrejonk: Right, I never thought about this. I see that it's done like this with `debops.owncloud` as well. The downside of this approach is that it basically makes copies of the `service/nginx` playbook. Wouldn't it be cleaner to just instruct the user to add the host to the `[debops_service_nginx]` inventory group as well? Just to prevent code duplication and to keep some separation between the roles?

ypid: This is a tradeoff between maintenance overhead and runtime. I guess we favor runtime but you are right. I am not fully happy with the duplication either.

docs/ansible/roles/pdns/getting-started.rst
```diff
@@ -0,0 +1,125 @@
...
+``True``. Doing so will instruct pdns to send (optionally TSIG-signed)
+notifications of changes to secondaries, which can then initiate zone
+transfers. Notifications are only sent for domains with type MASTER in your
+backend.
+
+Secondary operation
+~~~~~~~~~~~~~~~~~~~
+
+Secondary operation can be enabled by setting :envvar:`pdns__secondary` to
+``True``. Doing so will instruct pdns to periodically check for zone changes at the primary nameservers, and update the local zones accordingly. These checks happen every 'refresh' seconds (as specified by the SOA record) and are only performed for domains with type SLAVE in your backend. Additionally, if the primary
```
Alphix: Weird formatting?

imrejonk: Thanks for pointing this out, fixed.

ansible/playbooks/service/pdns.yml
imrejonk: Addressed in e938bf0f5c36d5c4fa8f21f41f7beec0bfeb9a6f

drybjed (approved):
Excellent work, thanks! I'll update the configuration template a bit to allow comments in YAML text blocks, but otherwise, looks good. :-)

Issue #1946, opened 2021-12-02
PKI regression


bleuchtang:
Commit https://github.com/debops/debops/commit/078ae6295b43c540f7206d9e5dd9518a89311d1e introduce a new variable `acme_type` in config/realm.conf file config. This variable is only setup at the creation of the LE certificate, so LE certificates generated before are not working anymore (because of  https://github.com/debops/debops/blob/master/ansible/roles/pki/files/usr/local/lib/pki/pki-realm#L1352)

I think a workaround is to set `acme_type` variable as `acme-tiny` by default when we whant LE certificats.

cc @julien-lecomte

julien-lecomte:
What's the value of `pki_acme_type` in your inventory? and of config['pki_acme'] from the config file?

bleuchtang:
Default value for `pki_acme_type` (acme-tiny)

and

```
/etc/pki# rgrep pki_acme  
realms/domain/config/realm.conf:config['pki_acme']='true'
realms/domain/config/realm.conf:config['pki_acme_library']='openssl'
```

thoemy:
I can confirm this. I made the following change to `pki-realm`

```patch
--- pki-realm	2021-12-21 09:33:07.505691980 +0000
+++ pki-realm.fixed	2021-12-21 09:32:37.873761393 +0000
@@ -222,7 +222,7 @@
     config["acme_default_domain"]="$(get_dnsdomainname)"
     config["acme_default_subdomains"]="www/ftp/mail/smtp/imap"
 
-    config["acme_type"]=""
+    config["acme_type"]="acme-tiny"
     config["acme_subject"]=""
     config["acme_domains"]=""
     config["acme_subdomains"]=""
```

julien-lecomte:
I haven't had much time lately, but I guess a possible fix would be changing in https://github.com/debops/debops/blob/master/ansible/roles/pki/files/usr/local/lib/pki/pki-realm#L1352 :

~~~
if [ "${config['acme_type']}" == "acme-tiny" ]; then
~~~
to
~~~
if [ "${config['acme_type']}" == "acme-tiny" ] || [ "${config['acme_type']}" == "" ] ; then
~~~

I'd really like to test before doing a PR, but I haven't got a setup that can be configured with a default acme setup
(edited)

thoemy:
The `test` command doesn't seem to like `||`.
```
$ /usr/local/lib/pki/pki-realm run -n example.com
[...]

PR #1960, opened 2021-12-21
Fix PKI regression #1946

julien-lecomte:
Probable solution. Not yet tested.


drybjed (approved):
Looks good, thanks! :-)

PR #1961, opened 2021-12-21
feat(sshd): Keep SSH_CONNECTION env var when running commands with sudo

ypid:


drybjed (dismissed):
Sounds good, although this change needs to be propagated to more playbooks, like `bootstrap-ldap`, `bootstrap-sss` and `common`, where the `sshd` role is also used. I'll do that during merging.

ypid:
Ah, good catch. Fixed.

drybjed (approved):
Works as expected, thanks. I'll fix the merge conflict and make a backport.

ypid:
You are welcome, thanks for merging.

PR #1965, opened 2022-01-02
feat(netbox): Upgrade to 3.1.3 and align with upstream defaults

ypid:


drybjed (approved):
Sounds good, thanks. We could convert the NetBox configuration file to the Universal Configuration format (good example for Python-based configuration is in the `mailman` role). That way the configuration options don't have to be static and can be added as needed by the user.

ypid:
You are welcome. Sounds good to me. Not sure when I will come around to do it yet.

PR #1968, opened 2022-01-04
[ipxe] Update release map for Bullseye 11.2

Alphix:
The 11u1 installer media is already gone from the mirrors.


drybjed (approved):
Thanks! I'll update the watch files as well.

PR #1972, opened 2022-01-05
[debops.nginx] Allow logging to socket

ctx:
This changes allow sending logs to a local or remote socket, eg.
/dev/log or a remote syslog server.

### DIFF
--- a/ansible/roles/nginx/defaults/main.yml
+++ b/ansible/roles/nginx/defaults/main.yml
@@ -169,7 +169,24 @@ nginx_run_path: '/run'
 # .. envvar:: nginx_log_path [[[
 #
 # Directory where :program:`nginx` log files are stored.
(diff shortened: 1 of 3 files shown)


gaudenz:
@drybjed Can you please approve this workflow. The doc typo present in the first version seems to be fixed and the CI should now pass.

ypid:
I like the overall direction of this. I guess your background is to have journald listen on that socket? I only peeked at the changes as I have not done that myself with Nginx, but when I get around to it, I will test it.

gaudenz:
Yes the main motivation for this is to get the logs into the systemd journal. But it should be generic enough to also allow logging to a remote syslog port.

drybjed (approved):
I haven't tested them, but the changes seem correct. Thanks! I'll backport these to the stable releases.

Issue #1911, opened 2021-11-04
debops script breaks if project path contains (but doesn't start with) the user's $HOME path


Alphix:
There's plenty of places in `src/debops/*.py` which attempt to shorten `/home/user/foo` to `~/foo`...for example:

https://github.com/debops/debops/blob/deab69f3a5a12453c59aba703c1d988c3a972351/src/debops/ansibleconfig.py#L26-L34

The problem is that `replace` will replace anywhere in the PATH, and not just if the PATH starts with the user's `$HOME`.

I have a symlink in my home dir linking to my debops devel directory (i.e. `/home/user/debops` is a symlink to e.g. `/mnt/someserver/home/user/debops`).

This means that `debops project refresh` creates an `ansible.cfg` with contents like:

```
...
collections_paths = ansible/collections:~/.ansible/collections:/mnt/someserver~/debops/debops/src/debops/_data/ansible/collections:/usr/share/ansible/collections
...
```

Which breaks running commands like `debops run` (note the erroneous third path in `collections_paths`).

Issue #1915, opened 2021-11-05
debops script doesn't seem to use proper exit statuses


Alphix:
```
$ debops run bootstrap -l vmtest1 -e ansible_user=root
Executing Ansible playbooks:
bootstrap
ERROR! the playbook: bootstrap could not be found
$ echo $?
0
```

I'd have expected any exit value except 0 here...

drybjed:
The error message comes from `ansible-playbook` command. But that's good point, we should catch the exit code and pass it along.

PR #1974, opened 2022-01-08
Debops script fixes

Alphix:
Two commits which fix issue #1911 and issue #1915


drybjed (approved):
Nice work, thanks! I'll add a comment explaining why `unexpanduser` is needed.

PR #1975, opened 2022-01-11
Loop log spam fixes

Alphix:
These patches removes the wall of text that some roles like to generate by changing them over to use ``loop`` instead of ``with_*``


drybjed (approved):
Nice work, thanks. I would go with at least a `name` and `state` keys, but since the configuration is changed in separate tasks that should be obvious already. I'll backport this to the stable releases.

PR #1976, opened 2022-01-11
[sshd] Improve LDAP bind password file generation

Alphix:
First, to echo a password means it'll show up temporarily in the process list.

Second, to first create the file and then set the right permissions is also not
a good idea.

This patch also removes a variable which seems kind of pointless. Roles usually
don't make file paths configurable, and I can't see why anyone would like to
change it. If it's better to keep the variable around, I can rework the patch.

### DIFF
--- a/ansible/roles/sshd/COPYRIGHT
+++ b/ansible/roles/sshd/COPYRIGHT
@@ -2,7 +2,8 @@ debops.sshd - Manage OpenSSH server using Ansible
 
 Copyright (C) 2013-2016 Maciej Delmanowski <drybjed@gmail.com>
 Copyright (C) 2015-2017 Robin Schneider <ypid@riseup.net>
-Copyright (C) 2014-2017 DebOps <https://debops.org/>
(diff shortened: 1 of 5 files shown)


drybjed (changes requested):
Oh man, I don't know what caused me from 7 years ago to do it like this, I guess total lack of experience... Yeah, it's much better to avoid shelling out in this case. However, I think that we can even skip templating and just use the `copy` module with `content` parameter. Do you want to see if that works and update the PR?

Dropping the variable with the path is also fine with me. I guess if somebody requests due to their use case, we can bring it back.

Alphix:
I didn't use ``copy`` + ``content`` cause the [Ansible docs](https://docs.ansible.com/ansible/latest/collections/ansible/builtin/copy_module.html) discourage it (quote: ``If you need variable interpolation in copied files, use the ansible.builtin.template module. Using a variable in the content field will result in unpredictable output.``)

drybjed (approved):
Very well then, let's do this with a template. I'll backport the changes to the stable releases as well.

PR #1977, opened 2022-01-11
[ldap] Some fixes to init-directory.yml for non-local admin user

Alphix:
The init-directory.yml already has some support for specifying a non-local
user as the admin user. Unfortunately, it doesn't work since the "getent"
task will fail.

Also, some cosmetic changes (make sure the task titles are correct even
when using a non-local user and that a user like "ansible" doesn't end
up being named "Ansible Ansible" in the LDAP DIT).


drybjed (approved):
Looks good, thanks. I'll backport the changes to stable releases.

PR #1978, opened 2022-01-11
[system_groups|system_users] Also detect if a host will be LDAP-enabled

Alphix:
This solves a bit of a chicken-or-the-egg problem when bootstrapping the
host where the slapd service will be installed. At the time the bootstrap
script is executed, the host cannot possibly be LDAP enabled, but it is
highly likely that it will be in the future.

I know that using inventory groups wasn't considered kosher in PR #1568,
but I'd argue that:

 a) The Ansible inventory is supposed to be *the* source of truth.

 b) This merely sets a user-friendlier default which doesn't negatively
    affect users who are using stand-alone roles and/or not using the
    DebOps playbooks.

 c) Setting a default is a different scenario than preventing a role
    from doing what it's supposed to do (as in PR #1568).


drybjed (approved):
Looks good to me. Since DebOps now is in a monorepo, any changes in the Ansible inventory group names can be easly synchronized with roles. I'll backport this change to the stable releases.

PR #1955, opened 2021-12-16
Debian Bullseye: dist-upgrade.yml, apt_cacher_ng

ypid:
Closes: #1982

PR #1979, opened 2022-01-14
Fixes before the next DebOps release: Fixes, NetBox netbox__config_custom option

ypid:
Note that #1955 is also only fixes and added support for Debian 11.


drybjed (approved):
Sounds good, thanks. I'll try to backport as much as possible.

PR #1980, opened 2022-01-15
[users]: Fix always creating a home directory

dejarikra:
When "managing additional UNIX groups for UNIX accounts", the `ansible.builtin.users` module is being used. In this usage, the `create_home` argument is not specified, and so it defaults to `True` according to [the documentation](https://docs.ansible.com/ansible/latest/collections/ansible/builtin/user_module.html#parameter-create_home). This will always create a home directory, even if the corresponding `users__*_account` element in the `users__combined_accounts` list specified `create_home: False`.

This PR fixes this bug by adding the `create_home` argument as specified by the corresponding `users__*_account` element in the `users__combined_accounts` list, omitting it if it doesn't exist. (Some "tabbing" is also added to make the spacing consistent after the addition of this new argument)

### DIFF
--- a/ansible/roles/users/tasks/main.yml
+++ b/ansible/roles/users/tasks/main.yml
@@ -96,20 +96,21 @@
   # Ref: https://github.com/ansible/ansible/issues/48722
 - name: Manage additional UNIX groups for UNIX accounts
   user:
-    name:   '{{ item.name }}'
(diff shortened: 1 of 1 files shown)


drybjed (approved):
Thanks for the bugfix, nice work. :+1: I'll backport this to the stable releases as well.

PR #1986, opened 2022-02-01
[netbox] Added ssh key generation for napalm

scibi:


ansible/roles/netbox/defaults/main.yml
```diff
@@ -445,6 +445,18 @@ netbox__config_napalm_username: ''
 # This overwrites potentially existing `dynamic configuration settings <https://netbox.readthedocs.io/en/stable/configuration/dynamic-settings/>`__.
 netbox__config_napalm_password: ''
 
+                                                                   # ]]]
+# .. envvar:: netbox__napalm_ssh_generate [[[
+#
+# Generate ``~/.ssh/id_rsa`` SSH private and public key if it doesn't exist.
```
Alphix, line 451: That should be `` :file:`~/.ssh/id_rsa` ``? Also, is there any particular reason to not use `ed25519` rather than `RSA` as the key type?

drybjed: I think that `ed25519` might not be supported everywhere - this SSH key will be used on network devices, remember.

drybjed (approved):
Looks good, thanks. I'll make a backport as well.

PR #1991, opened 2022-02-05
[dnsmasq] Fix DNS A/AAAA record generation

Alphix:
The dnsmasq detailed documentation says that "host", "a" and "aaaa" are
synonyms which can be used interchangeably (further down, not in the
section touched by this patch).

However, a configuration like this:

```
dnsmasq__dns_records:

  - host: 'a.example.com'
    ip: '192.168.0.1'

  - a: 'b.example.com'
    ip: '192.168.0.2'
```

Currently results in this /etc/dnsmasq.d/host-resource-records.conf:

```
  host-record = a.example.com,192.168.0.1

  dhcp-host = 192.168.0.2
```

The reason is that the template only checks whether entry.host is
defined.


drybjed (approved):

PR #1992, opened 2022-02-05
[apt_install] Autodetect presence of virtual RNG

Alphix:
apt_install currently installs "haveged" unconditionally on e.g.
KVM guests. There is a comment in defaults.yml to the effect that
while the host can provide entropy to the guest, this cannot be
assumed.

This patch enables checking for if a virtual RNG is present in
the guest, and skips the installation of "haveged" in that case.

### DIFF
--- a/ansible/roles/apt_install/defaults/main.yml
+++ b/ansible/roles/apt_install/defaults/main.yml
@@ -274,7 +274,9 @@ apt_install__conditional_packages:
     whitelist: '{{ apt_install__conditional_whitelist_packages }}'
     state: '{{ "present"
                if (ansible_virtualization_role|d("guest") in [ "guest" ] and
-                   ansible_virtualization_type|d("unknown") not in ["lxc", "openvz"]
+                   ansible_virtualization_type|d("unknown") not in ["lxc", "openvz"] and
+                   not (ansible_local|d() and ansible_local.apt_install|d() and
+                        ansible_local.apt_install.have_virtual_rng|d(False)|bool)
                )
                else "absent" }}'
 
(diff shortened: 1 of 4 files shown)


drybjed (changes requested):
Hmm... I don't really like to use tasks to check very specific things on the remote host to not make the playbooks too long (round-trip for each task makes them longer, remember). How about adding this check in the `apt_install.fact`? You need to convert it to a script and install it on the host before the main role beings (just like the current tasks you're added), but otherwise it should be a reliable way to detect presence of `virtio_rng` module. And if you you cannot find `lsmod` itself, you should be confident that this is a container.

Alphix:
Sure, I can do that. An alternative to running `lsmod` would also be to check for the presence of `/sys/module/virtio_rng`. That should work even if `lsmod` isn't available. There might be containers/virtualization environment that don't have `/sys` (just guessing, some containers do some funky stuff), but that shouldn't matter for the purposes of this functionality...

Alphix:
@drybjed : Ok, I've pushed a v2 which I think is more in line with what you'd prefer...please have a look

Also: is there a minimum Ansible version that's supported by DebOps? The reason that I'm asking is that if the minimum is >= 2.8, then chunks like this:

```
ansible_local|d() and ansible_local.apt_install|d() and ansible_local.apt_install.have_virtual_rng|d(False)
```

Could be simplified to:

```
ansible_local.apt_install.have_virtual_rng|d(False)
```

[Reference](https://docs.ansible.com/ansible/latest/user_guide/playbooks_filters.html#providing-default-values)

drybjed (approved):
Nice work, thanks. I'll backport this to the stable releases.

PR #1993, opened 2022-02-05
[ldap-init] Allow plaintext password storage

Alphix:
This introduces two changes to how ldap/init-directory.yml works:

First, it now supports optionally generating a random password.

Second, optional support is added for storing the random password unencrypted
in the /secret directory.

This makes it easier to e.g. run the Ansible Controller in an automated
fashion.

See the docs updates for more details.

### DIFF
--- a/ansible/playbooks/ldap/init-directory.yml
+++ b/ansible/playbooks/ldap/init-directory.yml
@@ -15,9 +15,14 @@
 
   vars_prompt:
 
-    - name: 'admin_plaintext_password'
-      prompt: 'New password for your LDAP user account'
+    - name: 'admin_input_plaintext_password'
+      prompt: 'New password for your LDAP user account (enter=random)'
+      default: ''
       private: True
+
+    - name: 'admin_use_password_store'
+      default: True
+      prompt: 'Use Password Store? (default=yes)'
 
   vars:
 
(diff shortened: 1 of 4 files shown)


Alphix:
Another option would also be to keep the ``vars_prompt`` and generate a random password if no input is empty (which would also work for the non-interactive scenario). That'd reduce the risk of users echoing passwords at the cmd prompt. But first I'd like to know if you agree with the general idea...

drybjed:
Sure, enabling non-interactive mode sounds good. If the admin afterwards changes his/her password, it shouldn't impact anything. Before I merge this though, if you want to still keep `vars_prompt` support, how would you solve non-interactive operation, via `expect`? Unless Ansible has some mechanism for that as well...

Alphix:
Ansible already does the right thing for `vars_prompt` when it is set via the command line (when passed as ``--extra-vars``, the prompt is not presented), or when running from a non-interactive session (it uses the defaults or makes the variable undefined, IIRC).

I'll start working on the documentation changes.

drybjed:
Excellent, looking forward to a final version. :)

Alphix:
> Excellent, looking forward to a final version. :)

Ok, this should be the final version, now with docs as well (and I've tested the three versions....env variable, pass, and plaintext file)

drybjed (approved):
Nicely done, thanks. Since it's a significant change in functionality, I will not backport this to the stable releases.

PR #1997, opened 2022-02-07
[icinga] Improve initial host registration

Alphix:
This patch is an attempt to improve the icinga initial setup by automatically creating the default groups which are explained in [the documentation](https://docs.debops.org/en/master/ansible/roles/icinga/deployment.html#initial-deployment). With this patch applied, it's possible to setup icinga and register all hosts without having to touch the web interface at all.

Also, some more information is created for the host objects when they are created (a short display_name, and IPv4/IPv6 addresses).

This is just a draft because:
a) I want to see if you agree with the change before spending the time to update the documentation
b) I'm not sure if ``icinga__director_combined_host_templates`` should be changed to use universal configuration style

### DIFF
--- a/ansible/roles/icinga/defaults/main.yml
+++ b/ansible/roles/icinga/defaults/main.yml
@@ -626,7 +626,7 @@ icinga__default_configuration:
         value: |
           include_recursive "repository.d"
         state: '{{ "absent"
-                   if (icinga__version is version("2.8.0", ">=")
+                   if (icinga__version is version("2.8.0", ">="))
                    else "present" }}'
 
       - name: 'conf.d'
--- a/ansible/roles/icinga/defaults/main.yml
+++ b/ansible/roles/icinga/defaults/main.yml
@@ -124,6 +124,26 @@ icinga__display_name: '{{ (inventory_hostname_short | d(inventory_hostname.split
                           else ansible_hostname }}'
 
                                                                    # ]]]
+# .. envvar:: icinga__ipv4_address [[[
+#
+# The IPv4 address (or FQDN used to lookup the IPv4 address) of this Icinga 2
+# node. This variable is used during the node registration in the Icinga 2
+# Director (and will be used to contact the node for monitoring purposes).
+icinga__ipv4_address: '{{ ansible_default_ipv4.address
+                          | d(ansible_all_ipv4_addresses | d([]) | first)
+                          | d(icinga_fqdn) }}'
+
+                                                                   # ]]]
+# .. envvar:: icinga__ipv6_address [[[
+#
+# The IPv6 address (or FQDN used to lookup the IPv4 address) of this Icinga 2
+# node. This variable is used during the node registration in the Icinga 2
+# Director (and will be used to contact the node for monitoring purposes).
+icinga__ipv6_address: '{{ ansible_default_ipv6.address
+                          | d(ansible_all_ipv6_addresses | d([]) | first)
+                          | d(omit, true) }}'
+
+                                                                   # ]]]
 # .. envvar:: icinga__domain [[[
 #
 # The main DNS domain used by the role to configure Icinga 2.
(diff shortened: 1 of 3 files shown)


Alphix:
Should probably drop the `ipaddr("public")` filters.... The intention was rather "not loopback/useless IPs", but local/private IPs are fine...

Alphix:
Hmm...right, tested this on a bunch of VMs that already had icinga installed (by deleting everything in the web interface first).

Missed that the additional task will fail when icinga is first configured on the master/web host (since the web interface will be unreachable). Could change it to ignore the error....or move the functionality to `icinga_web`....I'll have to research this a bit further...

Alphix:
See also PR #1910 which discusses similar issues...

Alphix:
> Hmm...right, tested this on a bunch of VMs that already had icinga installed (by deleting everything in the web interface first).
> 
> Missed that the additional task will fail when icinga is first configured on the master/web host (since the web interface will be unreachable). Could change it to ignore the error....or move the functionality to `icinga_web`....I'll have to research this a bit further...

Ok, changed the approach a bit. Now the host templates are created as part of the host registration, which will autocreate the templates at the cost of two extra REST API calls per host registration...

(The last two tasks in `icinga/tasks/main.yml` could be merged)

Alphix:
@imrejonk : could you take a look at this PR?

imrejonk:
Hey, sure thing. I'll give it a try this weekend.

ansible/roles/icinga/defaults/main.yml
```diff
@@ -371,13 +403,14 @@ icinga__director_register_host_vars: {}
...
-  address: '{{ icinga__fqdn }}'
-  imports: '{{ lookup("flattened",
-               (icinga__director_register_default_templates
-                + icinga__director_register_templates
-                + icinga__director_register_group_templates
-                + icinga__director_register_host_templates),
-               wantlist=True) }}'
+  display_name: '{{ icinga__display_name }}'
+  address: '{{ ansible_default_ipv4.address | d(ansible_all_ipv4_addresses | d([]) | first) | d(icinga_fqdn) }}'
+  address6: '{{ ansible_default_ipv6.address | d(ansible_all_ipv6_addresses | d([]) | first) | d(omit, true) }}'
```
imrejonk: It would be great if you could make the `address` and `address6` parameters configurable as well. My production servers have their DNS names configured here, and I would like to keep it that way because it would be one less thing to worry about during a network renumbering.

ansible/roles/icinga/defaults/main.yml
```diff
@@ -111,9 +111,19 @@ icinga__additional_groups:
...
 icinga__fqdn: '{{ ansible_fqdn }}'
 
+                                                                   # ]]]
+# .. envvar:: icinga__display_name [[[
+#
+# The short display name of this Icinga 2 node. This variable is used during
+# the node registration in the Icinga 2 Director and in the zone configuration.
+icinga__display_name: '{{ (inventory_hostname_short | d(inventory_hostname.split(".")[0]))
+                          if (inventory_hostname_short | d(inventory_hostname.split(".")[0]) != "localhost")
+                          else ansible_hostname }}'
```
imrejonk, line 124: I prefer the FQDN for the display_name, but this is still fine with me as you made it configurable.

imrejonk (changes requested):
Nice improvements, thanks! I only tested this with two LXC containers (one Icinga master/director, one Icinga client). It was great to see the host registration work on the first try.

ansible/roles/icinga/defaults/main.yml
Alphix: Yeah, I did it like this because the FQDN is already in the `hostname` variable (JSON parameter `object_name`)...and because it seems to match the most common approach in the Icinga docs....but as you say, it's configurable, so I think it can be left as-is

ansible/roles/icinga/defaults/main.yml
Alphix: Sure, will do

Alphix:
Ok, I implemented the changes suggested by @imrejonk, updated the documentation to reflect the new automagic registration, did a test install and verified that everything works as expected.

imrejonk:
Good work @Alphix! I'll give it a try this weekend.

imrejonk (approved):
Nicely done! Docs also look okay to me.

drybjed (commented):
The changes look good, but I think that the initial template creation could be moved from the `icinga` role to the `icinga_web` role. That way we avoid two additional `uri` round trip requests on each Icinga agent registration and the host (or other) templates are configurable in a central place.

I'll check if this is possible in a test environment and make the necessary modifications as soon as possible.

Alphix:
> The changes look good, but I think that the initial template creation could be moved from the `icinga` role to the `icinga_web` role. That way we avoid two additional `uri` round trip requests on each Icinga agent registration and the host (or other) templates are configurable in a central place.
> 
> I'll check if this is possible in a test environment and make the necessary modifications as soon as possible.

That sounds optimal (if it's doable, I must admit the ``icinga-*`` roles are still a bit confusing to me)

PR #1999, opened 2022-02-07
[dig_srv] Convert users of "dig" -> "dig_srv"

### LINKED ISSUE #1957: q("dig", "_someservice._someprotocol.<somedomain>./SRV") isn't idempotent
Right now there's a number of roles that lookup ``SRV`` DNS RRs....at least:

```
ansible/roles/nullmailer/defaults/main.yml
ansible/roles/rsyslog/defaults/main.yml
ansible/roles/roundcube/defaults/main.yml
ansible/roles/icinga/defaults/main.yml
ansible/roles/gitlab_runner/defaults/main.yml
ansible/roles/ldap/defaults/main.yml
```

Typically they do something like this (example taken from ``debops.nullmailer``:

```
# .. envvar:: nullmailer__smtp_srv_rr [[[
#
# List which contains the result of the :command:`dig` query for SMTP
# server ``SRV`` resource records in the host's domain. See
# :rfc:`6186` for details.
nullmailer__smtp_srv_rr: '{{ q("dig", "_smtp._tcp." + nullmailer__domain + "./SRV", "flat=0")
                             if nullmailer__domain|d() else [] }}'

                                                                   # ]]]
# .. envvar:: nullmailer__smtp_port [[[
#
# SMTP port (default is 25; use 587 for STARTTLS or 465 for Implicit TLS.
nullmailer__smtp_port: '{{ (nullmailer__smtp_srv_rr | selectattr("target", "defined")
                            | map(attribute="port") | list | sort | first)
                           if ("NXDOMAIN" not in nullmailer__smtp_srv_rr and
                               nullmailer__smtp_srv_rr[0])
                           else "25" }}'

                                                                   # ]]]
# .. envvar:: nullmailer__relayhost [[[
#
# The FQDN address of the mail server to which all mail messages will be forwarded
# to by the ``nullmailer`` service.
nullmailer__relayhost: '{{ (nullmailer__smtp_srv_rr | selectattr("target", "defined")
                           | map(attribute="target")
                           | map("regex_replace", "\.$","")
                           | list | sort | first)
                          if ("NXDOMAIN" not in nullmailer__smtp_srv_rr and
                              nullmailer__smtp_srv_rr[0])
                          else ("smtp." + nullmailer__domain) }}'

                                                                   # ]]]
```

The problem is:
- They are not idempotent (since there's no guarantee that multiple ``SRV`` records are always returned in order)
- They're quite a mouthful...the same chunk of template is repeated over and over

The first problem could be solved by always sorting by the ``priority``, ``weight`` and ``target`` fields (in that order from the ``SRV`` RR), i.e. by inserting a bunch of:

```
| sort(attribute="priority")
| sort(attribute="weight")
| sort(attribute="target")
```

But that would have to be duplicated over and over.

Perhaps it would be a good idea to introduce a new `filter_plugin`.....say `dig_srv` that would have a syntax like:
``debops.debops.dig_srv("_foobar._tcp.example.com.", "fallback.example.com", 1234)`` which would return a list of dicts of the results or the fallback fake result.

Unless I'm mistaken, that would allow several roles to be simplified a lot....?

Alphix:
As noted in issue #1957, `q("dig"` has several issues. It ignores sorting order and also doesn't differentiate a DNS server timeout from a NXDOMAIN (and therefore isn't idempotent).

This patch series converts all users of the "dig" lookup to use the new "dig_srv" lookup.

I've tested the following roles:

- icinga
- roundcube
- rsyslog
- nullmailer
- ldap

I have not tested the `gitlab` role because it seems broken ATM (complains about version incompatibilities when installed in a clean Bullseye VM)

Closes: #1957

### DIFF
--- a/ansible/roles/gitlab_runner/defaults/main.yml
+++ b/ansible/roles/gitlab_runner/defaults/main.yml
@@ -139,11 +139,17 @@ gitlab_runner__fqdn: '{{ ansible_fqdn }}'
                                                                    # ]]]
 # .. envvar:: gitlab_runner__gitlab_srv_rr [[[
 #
-# List which contains the result of the :command:`dig` query for Gitlab server
-# ``SRV`` resource records in the host's domain. See :rfc:`6186` for details.
-gitlab_runner__gitlab_srv_rr: '{{ q("dig", "_gitlab._tcp."
-                                    + gitlab_runner__domain
-                                    + "./SRV", "flat=0") }}'
+# List which contains the result of the DNS query for Gitlab server ``SRV``
+# resource records in the host's domain. See :rfc:`6186` for details.
+#
+# If there are no resource records, the role checks if a local Gitlab
+# installation is present and uses the host FQDN as the Gitlab API server
+# address. Finally, ``code.<domain>`` is used as a fallback.
+gitlab_runner__gitlab_srv_rr: '{{ q("dig_srv", "_gitlab._tcp." + gitlab_runner__domain,
+                                    ansible_fqdn
+                                    if (ansible_local|d() and ansible_local.gitlab|d() and
+                                       (ansible_local.gitlab.installed|d())|bool)
+                                    else ("code." + gitlab_runner__domain), 443) }}'
 
                                                                    # ]]]
 # .. envvar:: gitlab_runner__api_fqdn [[[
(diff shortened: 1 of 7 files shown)


drybjed (approved):
Nice work, thanks. I'll backport these to the stable releases as well.

PR #2006, opened 2022-02-09
Fix running debops on Python 3.10

Cadair:


drybjed (approved):
Thanks for the bugfixes! :-)

PR #2009, opened 2022-02-11
[nginx] Remove SPDY option

Alphix:
To quote Wikipedia: "Since 2021, no modern browser supports SPDY.".

So...remove this option, HTTP2 is already supported.

Also, since I had difficulties even parsing that 500 characters wide
line in the template, I took the liberty of rewriting it over a
couple of lines...

### DIFF
--- a/ansible/roles/nginx/defaults/main.yml
+++ b/ansible/roles/nginx/defaults/main.yml
@@ -625,13 +625,6 @@ nginx_hsts_preload: False
 nginx_enable_http2: True
 
                                                                    # ]]]
-# .. envvar:: nginx_enable_spdy [[[
-#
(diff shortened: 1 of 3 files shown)


drybjed (approved):
Nicely done, thanks. And I know that the default nginx template is a mess - this template was a famous example on reddit back in the day showing why DebOps is baaaaaad... :) Thankfully, I learned a thing or two after all this time; it will be rewritten, someday.

Alphix:
> Nicely done, thanks. And I know that the default nginx template is a mess - this template was a famous example on reddit back in the day showing why DebOps is baaaaaad... :) Thankfully, I learned a thing or two after all this time; it will be rewritten, someday.

Cool, I even found the old Reddit threads. I might have a look at rewriting the template some day if I feel like taking on a challenge :D

PR #2010, opened 2022-02-12
[bootstrap-*] Improve mitogen compatibility

Alphix:
Mitogen[1][2] is a great way to speed up Ansible runs, but it doesn't
support running raw tasks. This makes sure that the normal Ansible
strategy is used for raw tasks (in other words, it should have no
effect on non-Mitogen users).

[1] https://mitogen.networkgenomics.com/ansible_detailed.html
[2] https://github.com/mitogen-hq/mitogen


drybjed (approved):
Sounds good to me. I'll backport this to the stable releases, thanks.

PR #2011, opened 2022-02-13
Update dist-upgrade.yml

tumd:
Don't fail the task if grep returns no files.


drybjed (approved):
Sounds good, thanks. I'll backport this to stable releases.

PR #2015, opened 2022-02-14
[grub] Fix password generation

Alphix:
Not sure how this could have worked before, without the "-e" option,
echo won't parse the newlines. With this patch applied I can create
grub users with passwords.

Note that the whole shell thing is still unsafe since it'll leak
the grub password to the process list, a more thorough fix would
be to template the password to a tmp file and then cat that to
grub-mkpasswd-pbkdf2 (I've tested it manually and it works).

That'll have to wait for another PR though.


drybjed (approved):
Thanks. Personally I would go with a `printf` command instead (https://unix.stackexchange.com/a/65819), but if you plan to change the whole mechanism at some point, let's go with this quickfix for now. I'll backport it to the stable releases as well.

PR #2016, opened 2022-02-14
[dpkg_divert] Convert all roles to use the module

Alphix:
I've tested all roles that are affected by this PR (great fun) except for `postgresql_server` (the autobackup tool doesn't exist in Bullseye)

### DIFF
--- a/ansible/roles/apt/tasks/main.yml
+++ b/ansible/roles/apt/tasks/main.yml
@@ -6,6 +6,10 @@
 # Copyright (C) 2014-2018 DebOps <https://debops.org/>
 # SPDX-License-Identifier: GPL-3.0-only
 
+- name: Import custom Ansible plugins
+  import_role:
+    name: 'ansible_plugins'
+
 - name: Import DebOps global handlers
   import_role:
     name: 'global_handlers'
--- a/ansible/roles/apt/tasks/main.yml
+++ b/ansible/roles/apt/tasks/main.yml
@@ -80,22 +80,10 @@
   when: item not in ansible_facts.ansible_local.apt.foreign_architectures|d()
   notify: [ 'Refresh host facts' ]
 
-- name: Check current APT diversions
-  environment:
-    LC_ALL: 'C'
-  shell: set -o nounset -o pipefail -o errexit &&
-         dpkg-divert --list '/etc/apt/*.dpkg-divert' | awk '{print $NF}' || true
-  args:
-    executable: 'bash'
-  register: apt__register_diversions
-  when: apt__enabled|bool
-  check_mode: False
-  changed_when: False
-
 - name: Divert original /etc/apt/sources.list
-  command: dpkg-divert --quiet --local --divert /etc/apt/sources.list.dpkg-divert --rename /etc/apt/sources.list
-  when: (apt__enabled|bool and apt__sources_deploy_state == 'present' and
-         '/etc/apt/sources.list.dpkg-divert' not in apt__register_diversions.stdout_lines)
+  dpkg_divert:
+    path: '/etc/apt/sources.list'
+  when: (apt__enabled|bool and apt__sources_deploy_state == 'present')
 
 - name: Configure APT distribution sources
   template:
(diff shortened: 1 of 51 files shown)


Alphix:
Not 100% sure what ansible-lint wants from me :)

ansible/roles/grub/tasks/main.yml
```diff
@@ -119,16 +119,17 @@
...
 
 - name: Copy /etc/grub.d/10_linux.dpkg-divert to its original location
-  command: cp /etc/grub.d/10_linux.dpkg-divert /etc/grub.d/10_linux  # noqa no-handler
-  when: grub__register_linux_divert is changed
+  copy:
+    src: '/etc/grub.d/10_linux.dpkg-divert'
+    dest: '/etc/grub.d/10_linux'
+    remote_src: True
+    force: False
+  when: (grub__combined_users|length > 0)
```
drybjed, line 137: As an example here, `ansible-lint` wants all `copy` and `template` uses to include the `mode` parameter to explicitly define the file permissions.

ansible/roles/snapshot_snapper/tasks/main.yml
```diff
@@ -36,17 +40,18 @@
...
-    creates: '{{ item }}.dpkg-divert'
+  dpkg_divert:
+    path: '{{ item }}'
   register: snapshot_snapper__updatedb_diverted_register
   when: ansible_os_family in [ "Debian" ]
   loop: '{{ snapshot_snapper__divert_files | flatten }}'
 
 - name: Copy diverted configuration file to original location
-  command: cp {{ item }}.dpkg-divert {{ item }}
-  args:  # noqa no-handler
```
drybjed, line 48: Since this task shouldn't be done as a handler (well, we could add it as a handler and do meta: flush_handlers but that's not really needed here...), make sure to copy the `# noqa no-handler` comment to tell `ansible-lint` to ignore this issue.

Alphix:
Ah, I never got that far because `ansible-lint` errored out, complaining that `dpkg_divert` was an unknown module.

I've [figured out](https://github.com/willthames/ansible-review/issues/16) that setting the `ANSIBLE_LIBRARY` environment variable works around it, so I have the real ansible-lint errors now. I'll go through them and update the PR.

ansible/roles/apt/tasks/main.yml
```diff
@@ -108,12 +100,12 @@
   when: (apt__enabled|bool and apt__sources_deploy_state == 'present')
 
 - name: Remove diversion of original /etc/apt/sources.list
-  shell: rm -f /etc/apt/sources.list ; dpkg-divert --quiet --local --rename --remove /etc/apt/sources.list
-  args:
-    warn: False
+  dpkg_divert:
+    path: '/etc/apt/sources.list'
+    state: 'absent'
+    delete: True
```
drybjed: The combination of creation and removal of diversion can be done in a single `dpkg_divert` task - just use the `state` parameter conditionally (the `delete` parameter does nothing with `state: present`). Do you think that you could merge this and other similar tasks into a single task each time this happens? We can leave it for later if it becomes too complex.

Alphix: I'll have a look

Alphix: Ok, done

Alphix:
@drybjed : I've merged the divert/revert tasks for all roles except for `apt_cacher_ng` (which has the divert and revert in two different task files....I have a separate PR for that role which I'll submit later, once I've tested it).

I've tested all the roles affected by the merger of divert/revert (rather light testing on `icinga` and `freeradius` given that they don't actually do much diversion/reversion out of the box) and it all seems to work.

Wouldn't suggest backporting this PR given how many roles it touches....

drybjed (approved):
Excellent work, thanks! This won't be backported due to large number of changes.

ypid:
Awesome work! I will review it now. I consider this an internal refactoring. I am not sure if it needs to be in the changelog? The changelog is long enough already ;-)

PR #2020, opened 2022-02-17
[apt_cacher_ng] Merge and cleanup tasks

Alphix:
This is a followup to PR #2016 which simplifies `apt_cacher_ng` by merging the task files and the cleaning up the resulting single task file.

Most of the commits shouldn't be controversial, save perhaps for the last three commits, which I can drop if they're not deemed suitable.


drybjed (approved):
The changes look sound, thanks. I'll fix the merge conflicts.

PR #2021, opened 2022-02-17
[grub] Use a template instead of echo for passwords

Alphix:
This is a followup to PR #2015 with a more complete fix.


drybjed (commented):
Passwords can be passed using environment variables. I created a separate pull request to fix the issue.

PR #2023, opened 2022-02-17
[apt_install] Remove ranger from package list

Alphix:
Installing ranger by default seems somewhat niche, more like personal
preference than something that'd be generally or universally useful.

In addition, apt_install__shell_packages already include the installation of
"mc", which (according to Debian popcon) is 25x as popular as "ranger", and
which is also a console file manager.

### DIFF
--- a/ansible/roles/apt_install/defaults/main.yml
+++ b/ansible/roles/apt_install/defaults/main.yml
@@ -239,7 +239,6 @@ apt_install__conditional_whitelist_packages:
   - 'haveged'
   - 'gnupg-curl'
   - 'needrestart'
-  - 'ranger'
   - 'open-vm-tools'
 
                                                                    # ]]]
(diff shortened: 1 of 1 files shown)


drybjed:
Well, I'm using `ranger` 95% of the time I work with a file manager. `mc` is used only for opening tarballs and access to remote filesystems via SSH or FTP... I suppose that if we remove `ranger` from defaults, we could go with `mc` as well. In most environments, a file manager is probably used on just a few machines, the vast majority of VMs/containers will be set up and then left running without touching them. And I can live with `ranger` and `mc` being in the Ansible inventory.

Alphix:
> Well, I'm using `ranger` 95% of the time I work with a file manager. `mc` is used only for opening tarballs and access to remote filesystems via SSH or FTP... I suppose that if we remove `ranger` from defaults, we could go with `mc` as well. In most environments, a file manager is probably used on just a few machines, the vast majority of VMs/containers will be set up and then left running without touching them. And I can live with `ranger` and `mc` being in the Ansible inventory.

`ranger` seems to have a python3 version now, so either the `ranger` cfg should be updated to reflect that or it should be removed....but yeah, personally I think `ranger` and `mc` could both go from the defaults....up to you of course 😄

drybjed (approved):
This role needs to be refreshed since we have a better way to deal with conditional installation than a convoluted "whitelist" setup. For now, lets remove `ranger` as suggested, package overhaul can be done at a later date, but before DebOps v3.1 is released.

PR #2024, opened 2022-02-17
[root_account] Remove fix for Debian bug #794727

Alphix:
The bug was fixed in base-files 9.3 (Aug 2015). Which means it only
affects Jessie and earlier, see:
https://sources.debian.org/patches/base-files/

Jessie = oldoldoldstable at this point, and even the ELTS support
will expire soon, see:
https://wiki.debian.org/LTS/Extended

So this should be safe to remove...DebOps 3.1 will be released
about the time when there's no Jessie ELTS support anymore...

### DIFF
--- a/ansible/roles/root_account/defaults/main.yml
+++ b/ansible/roles/root_account/defaults/main.yml
@@ -133,14 +133,6 @@ root_account__group: '{{ "wheel"
 root_account__shell: ''
 
                                                                    # ]]]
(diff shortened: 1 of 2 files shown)


imrejonk (approved):
Looks good to me! Thanks for taking the time to clean these things up.

drybjed (approved):
Sounds good, thanks. I'll backport this change to the stable releases.

PR #2025, opened 2022-02-17
[misc] Update ssh key lookup

Alphix:
This allows the preseed role to also consider non-RSA keys and
adds ~/.ssh/authorized_keys as a last fallback position in case
no other keys are found.


Alphix:
Hmmm...given that this variable is copied in several places...I wonder if it could be set once in some central place...like in the `core` role...

drybjed (approved):
I was apprehensive about this change a bit, since access to Ansible Controller via SSH might not equate to the same access to the hosts beind it, but... I suppose that if somebody wants to restrict access, they can change the variables. So sure, lets go with this setup. I'll backport it as well.

I don't think there's a good way to centralize this under a single variable. Going with something like a global variable won't work since the roles expect a value. Syncing the changes across multiple roles shouldn't be a problem though, since it's a monorepo.

PR #2026, opened 2022-02-17
[apache] Remove ansible version check

Alphix:
The same check is already in ansible/playbooks/common.yml and
it makes little sense to also check it in one single role.


drybjed (approved):
Good idea, thanks. I'll backport this to the stable releases as well.

PR #2030, opened 2022-02-22
[misc] Obsolete releases cleanup

Alphix:
Many of the ``ansible/roles/*/meta/main.yml`` files contain outdated/misleading information.

I assumed it's better to just say that ``all`` are supported for both Debian/Ubuntu (as that
is the goal of each role, as far as I can tell).
 
Also, for the following releases, ELTS ended/ends at:
- Wheezy: 30 June 2020
- Jessie: 30 June 2022
- Precise Pangolin: 26 April 2019

So, in preparation for the 3.1.0 release (which I assume will be around the time Jessie
ELTS support ends), this PR removes support for these releases.


Alphix:
Not sure what the `ansible-lint` failure is about, but it seems unrelated to this PR?

drybjed:
I looked at the PR a bit closer... From the initial comment I assumed that you would be updating only the `meta/main.yml` files in various roles to change the OS releases to `all` so that this information is reflected on Ansible Galaxy. But now I see that you also decided to remove various options that made the roles work on older releases, like `apt_preferences` configuration, package list changes, and so on.

I'm OK with changing the Ansible Galaxy information to say that all releases are supported, but dropping support for older releases in roles themselves seems to be unnecessary at this point. Do you have any arguments for this? Perhaps we could split this PR to handle the Galaxy details separately from modifying the rest of the role code?

imrejonk:
I'm all for cleaning up code that was used to support distribution releases that are now end-of-life. You can't keep it all around forever, and the EOL date of the release is a nice point in time where we can say that we're also dropping support in DebOps. This helps to keep our code clean and tidy. Weird workarounds can be removed, improving maintainability.

If we wouldn't remove support for obsolete releases after their EOL date, then when would we?

drybjed:
I usually do it when a role gets rewritten or refreshed, but sure, a release EOL is also a good moment to do it. Very well, let me review this PR to see if there are any problems and I'll merge it if everything checks out, but this won't be backported due to size. This means that `v3.0.x` branch is still viable for older OS releases and it will be supported for long enough time.

drybjed:
A quick suggestion for the future: with so much changes it might be good to group logical changes across roles together; in this case changing `ansible/roles/*/meta/main.yml` everywhere is one logical change that can be done in a single commit which then is easy to review. With the current changes split per-role the GitHub web interface became inadequate since it's hard to check if removal of a playbook variable and its counterpart in the role happens in the same commit. And reviewing through the webgui per-commit doesn't let you mark files as reviewed... So I resorted to using `tig` TUI for the review which is pretty handy so far.

One problematic change found already: @Alphix, you removed the entire `etc_services` role from the `common.yml` playbook. This role should still be in common playbook since multiple other roles can utilize it. I'll add it back, don't worry.

drybjed (approved):
Review done and accepted. @Alphix, thank you for going through all the roles and cleaning them up, it's really appreciated. :-)

I started this project on Debian Wheezy, it will be sad to see it go, but it's time to move on to the future.

Alphix:
> I usually do it when a role gets rewritten or refreshed, but sure, a release EOL is also a good moment to do it. Very well, let me review this PR to see if there are any problems and I'll merge it if everything checks out, but this won't be backported due to size. This means that `v3.0.x` branch is still viable for older OS releases and it will be supported for long enough time.

Yeah, I agree, the PR was not meant for backporting (since it's based on the assumption that DebOps 3.1.0 will be released around the time Jessie ELTS support ends, which hasn't happened yet).

Alphix:
> A quick suggestion for the future: with so much changes it might be good to group logical changes across roles together; in this case changing `ansible/roles/*/meta/main.yml` everywhere is one logical change that can be done in a single commit which then is easy to review. With the current changes split per-role the GitHub web interface became inadequate since it's hard to check if removal of a playbook variable and its counterpart in the role happens in the same commit. And reviewing through the webgui per-commit doesn't let you mark files as reviewed... So I resorted to using `tig` TUI for the review which is pretty handy so far.

Yeah, I usually err on the side of too many commits since it's easier to rebase and squash commits together than to do the reverse (at least I find it easier). If you'd asked me, I could've squashed them for you before you started the review :)

drybjed:
No no, separate commits like this are fine. I meant that you could put all `meta/main.yml` changes in one commit and then do a commit per role for each role changes. But it's fine either way, I managed to find a way to make it workable. :)

PR #2031, opened 2022-02-24
[pki] Explicitly use certbot authenticator

anzil:
Certain third-party plugins for `certbot` (e.g. `dns-gandi`) do not support command line arguments like `--dns-gandi` but require `--authenticator dns-gandi` to work properly.

This PR suggests to use the `--authenticator` flag explicitly in the `pki-realm` script, which provides compatibility for built-in *and* third-party Certbot plugins.


drybjed (approved):
Sounds like a good idea, thanks. I'll backport this to the stable releases.

PR #2033, opened 2022-02-26
[docs] Introduce top-level DNS documentation

Alphix:
Since there are several roles which rely on DNS SRV resource records for
autoconfiguration, and the same/similar instructions for how to use such
records are currently spread across the roles, this is an attempt to create a
central doc entry providing an in-depth explanation of how DNS SRV entries can
be used/configured.

### DIFF
--- a/docs/ansible/roles/avahi/defaults-detailed.rst
+++ b/docs/ansible/roles/avahi/defaults-detailed.rst
@@ -88,15 +88,14 @@ specific parameters:
   simpler syntax.
 
 ``type``
-  A string similar to a `SRV record <https://en.wikipedia.org/wiki/SRV_record>`_
-  that defines the service type and protocol (it's similar in a way that only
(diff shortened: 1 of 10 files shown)


Alphix:
I've posted this as a draft PR for now since I want some feedback on whether this is a good idea before spending any more time on it....

drybjed (commented):
Awesome work, @Alphix! Since this is still a draft, I'll wait for you to finish it but so far looks ready to be merged... Let me know when you think it's finished. :-)

Alphix:
@drybjed : great, will do some more spell checking and also verify that the output looks good, etc, then it should be ready to go

drybjed (approved):
I reviewed the PR and made some cosmetic fixes. Nicely done. :) I'll backport this to the stable releases as well, for documentation consistency.

PR #2035, opened 2022-02-28
[ntp] fix disable systemd-timesyncd for bullseye

bleuchtang:
Since Bullseye, systemd-timesyncd is present in systemd services but
with not-found as status.


drybjed (approved):
This was tested already and seems to work, thanks. I'll backport this fix to the stable releases as well.

PR #2038, opened 2022-03-04
replace all -> debops_all_hosts in docs

532910:
'debops_all_hosts' folder should be used for all debops host yamls
instead of 'all'

Signed-off-by: Sergio E. Nemirowski <sergio@outerface.net>

### DIFF
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -643,15 +643,15 @@ if ! [ -d src/controller ] ; then
     sed -i '/ansible_connection=local$/ s/^#//' src/controller/ansible/inventory/hosts
 
     vagrant_controller="$(printf "${SSH_CLIENT}\\n" | awk '{print $1}')"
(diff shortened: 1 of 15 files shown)


drybjed (approved):
Sounds good, thanks. I'll backport this to the stable releases as well.

drybjed:
I think that we need to reconsider this PR and update the entries in at least some places. The `[all]` group is not equivalent to `[debops_all_hosts]`, because Ansible treats the `[all]` group as special - it takes precedence over all other groups automatically. To do the same, the `[debops_all_hosts]` group would have to be marked with a higher priority, which is not default. I'll check the PR and see where we can use the `[all]` group reasonably.

More details about variable merging: https://docs.ansible.com/ansible/latest/user_guide/intro_inventory.html#how-variables-are-merged

PR #2039, opened 2022-03-05
[dist-upgrade] do not depend on SUDO_USER

boucman:
dist-upgrade.yml assumes that the variable ansible_env.SUDO_USER is set
This is not the case when sudo is not used to become root (lxc-ssh in my case)

Signed-off-by: Jérémy Rosen <jeremy.rosen@smile.fr>


drybjed (approved):
Good idea. In the future, the playbook could re reworked into a role which will provide greater flexibility.

PR #2043, opened 2022-03-08
add example for docker_server__custom_daemon_options

stefangweichinger:
add example with a multi-line option


drybjed (approved):
Sounds good, thanks. :-)

PR #2054, opened 2022-03-23
[grub] Use env variables to store user passwords

drybjed:
Closes #2021

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -129,6 +129,13 @@ debops.boxbackup role
   task on systems other than Debian 9 or 10. The patch is not required since
   the ``vagrant-libvirt`` v0.1.0 package.
 
+:ref:`debops.grub` role
+'''''''''''''''''''''''
+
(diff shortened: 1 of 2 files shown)


Alphix:
@drybjed : ehrm....correct me if I'm wrong....but this also puts the password in the process list....?

drybjed:
Ah, you're right - this trick works when the target process knows that passwords can be stored in env variables and can access them directly. I was thinking along those lines when writing this PR, but it seems that the shell replaces the env variables on the command line with their values before executing them... Which makes sense, I guess.

I'm reopening your PR and evaluating again. Ah, it's deleted already. :-)

Alphix:
> Ah, you're right - this trick works when the target process knows that passwords can be stored in env variables and can access them directly. I was thinking along those lines when writing this PR, but it seems that the shell replaces the env variables on the command line with their values before executing them... Which makes sense, I guess.
> 
> I'm reopening your PR and evaluating again. Ah, it's deleted already. :-)

I'll revive the PR....still have it available locally...

PR #2058, opened 2022-03-28
[postconf] Disable EHLO IP address check

imrejonk:
Rejecting messages based on the resolved EHLO hostname is prohibited by
https://datatracker.ietf.org/doc/html/rfc5321#section-4.1.4:

An SMTP server MAY verify that the domain name argument in the EHLO
command actually corresponds to the IP address of the client. However,
if the verification fails, the server MUST NOT refuse to accept a
message on that basis.

Some Office 365 (outlook.com) servers have EHLO hostnames that resolve
to 127.0.0.1, which caused our mail servers to reject a few (legitimate)
messages.

This change disables the EHLO IP address check on the next playbook run.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -176,6 +176,15 @@ debops.boxbackup role
   service without it actually being present on the host. This should now be
   avoided by carefully checking the service status.
 
+:ref:`debops.postconf` role
+'''''''''''''''''''''''''''
+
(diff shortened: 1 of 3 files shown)


drybjed (approved):
Agreed, let's conform to the official RFC, at least by default. Spam detection can be performed at a later stage, for example with `rspamd` which already is in DebOps.

I'll backport this to the stable releases.

PR #2061, opened 2022-03-30
[owncloud] - nextcloud 23 compatibility

jankowa:
owncloud role doesn't work with nextcloud 23 (since newest minor release) anymore because

  `owncloud__do_autosetup`

will be false since 

  `"is not installed" in owncloud__register_occ_check.stdout`

will never be True. The `occ check` command recently writes to stderr instead of stdout.

### DIFF
--- a/ansible/roles/owncloud/tasks/setup_owncloud.yml
+++ b/ansible/roles/owncloud/tasks/setup_owncloud.yml
@@ -89,6 +89,7 @@
   ansible.builtin.set_fact:
     owncloud__do_autosetup: '{{ (owncloud__autosetup and
                                  owncloud__admin_username|d() and
+                                 (owncloud__register_occ_check is not skipped) and
                                  ("is not installed" in owncloud__register_occ_check.stdout) or
                                  ("is not installed" in owncloud__register_occ_check.stderr)) }}'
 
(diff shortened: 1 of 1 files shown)


ansible/roles/owncloud/tasks/setup_owncloud.yml
```diff
@@ -89,8 +89,8 @@
   ansible.builtin.set_fact:
     owncloud__do_autosetup: '{{ (owncloud__autosetup and
                                  owncloud__admin_username|d() and
-                                 (owncloud__register_occ_check is not skipped) and
```
ypid, line 92: Why remove `owncloud__register_occ_check`? I don‘t know exactly why (presumably) I added that condition but there was a reason, potentially check mode.

jankowa: you're totally right. This wasn't intended.

ypid (approved):
LGTM now. But you could squash the two commits into one really ;-) Just make it a habit if you don’t already.

bleuchtang:
Any chance to see that PR merge soon ? Nextcloud 22 EOL is in less than 2 month.

drybjed (approved):
Thanks, and sorry for the late merge. I'll backport this to the stable releases as well.

PR #2062, opened 2022-03-30
[nginx] improve documentation nginx_ocsp_resolvers

jankowa:
clarify that it's not only for ocsp resolving but all
  dns resolving.

### DIFF
--- a/ansible/roles/nginx/defaults/main.yml
+++ b/ansible/roles/nginx/defaults/main.yml
@@ -580,7 +580,8 @@ nginx_ocsp_verify: '{{ nginx_ocsp | bool }}'
                                                                    # ]]]
 # .. envvar:: nginx_ocsp_resolvers [[[
 #
-# List of DNS servers used to resolve OCSP stapling. If it's empty, nginx role
+# List of DNS servers used to resolve OCSP stapling and other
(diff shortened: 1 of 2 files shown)


imrejonk:
Sounds good. I think it’s a good idea to rename the variable as well (and document the rename in the upgrade notes).

jankowa:
> Sounds good. I think it’s a good idea to rename the variable as well (and document the rename in the upgrade notes).

I thought about this possibility too. We could use:

`nginx_resolvers`

and make a default like

`nginx_resolvers: "{{ nginx_oscp_resolvers }}"

for backword compatibility?

We find the variables only in 

`ansible/roles/nginx/defaults/main.yml`
and
`ansible/roles/nginx/tasks/main.yml`

imrejonk:
I don't think we should go down the rabbit hole of trying to keep variable names backward-compatible. We already document all variable name changes in one place (the upgrade notes), so it is trivial for a DebOps user to just modify their inventory whenever they upgrade to the next minor release (which will be 3.1).

drybjed (approved):
Documentation looks good, thanks. As for the variable renaming... I guess we can wait for the `nginx` role redesign to make that move, since they will be at least namespaced with two underscores at that point.

PR #2064, opened 2022-04-03
Issue 2046 fixes

### LINKED ISSUE #2046: Do we still need core__distribution and core__distribution_release?
According to the [documentation](https://docs.debops.org/en/master/ansible/roles/core/defaults/main.html#envvar-core__distribution), the variables `core__distribution` and `core__distribution_release` exist because Ansible fails to autodetect this information on some distributions. The values are placed in local facts and are used by other roles.

I see three problems with this:

1. The values in /etc/ansible/facts.d/core.fact need manual updating after a distribution upgrade. The administrator must remember to do this before running any role which depends on these facts.
2. The documentation does not say *how* or *why* Ansible fails to autodetect this information, nor is there a link to an Ansible bug report.
3. This Ansible bug could have been solved already, which would make this workaround obsolete.

I've never experienced Ansible wrongly detecting my OS distribution, but maybe this is still an issue for some users out there (in which case we should update the documentation). If not, we should consider removing these variables.

imrejonk:
Closes: #2046

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -111,6 +111,16 @@ General
 - Fixed an issue with custom Ansible plugins not working in "standalone" mode
   without the DebOps scripts installed on Ansible Controller.
 
+:ref:`debops.apt` role
+''''''''''''''''''''''
+
+- The role no longer defaults to the ``ansible_local.core.distribution`` and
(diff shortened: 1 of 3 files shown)


drybjed (approved):
Good idea, thanks. I'll backport this to the stable release.

PR #2068, opened 2022-04-12
fix #2067 by dropping the superfluous ipaddr call

### LINKED ISSUE #2067: debops.ldap fails when using first IPv6 address in network
# Background

Thanks to [CIDR](https://en.wikipedia.org/wiki/Classless_Inter-Domain_Routing) there is no real technical reason why the first address in a layer 3 network shouldn't be a valid address, as can be seen when comparing `192.0.2.128/25` and `192.0.2.128/24`; both are versions of *192.0.2.128*, and for any device outside this particular network it will effectively be the same address.
This change has made it into IPv6 actually, addresses like `2001:db8:0123:cdef::/64` are perfectly valid addresses now.

# Issue Description

Jumping to the code, these are the relevant lines of code: https://github.com/debops/debops/blob/526110b3073d4855395cfa5c7879674db6a700f6/ansible/roles/ldap/templates/lookup/ldap__device_ip_addresses.j2#L19-L22

While this may look innocent enough, the issue lies with this behaviour of the *ipaddr* filter:

```bash
ansible -m debug -a 'msg={{ "2001:db8::1/64" | ansible.utils.ipaddr("address/prefix") }}' localhost
```

```text
localhost | SUCCESS => {
    "msg": "2001:db8::1/64"
}
```

```bash
ansible -m debug -a 'msg={{ "2001:db8::/64" | ansible.utils.ipaddr("address/prefix") }}' localhost
```

```text
localhost | SUCCESS => {
    "msg": false
}
```

This behaviour breaks in the following line when *false* is being passed to the *ipv6* filter:

```text
TASK [debops.debops.ldap : Perform LDAP tasks] ***********************************************************************************************
fatal: [example.com]: FAILED! =>
  msg: 'An unhandled exception occurred while templating ''{{ ldap__dependent_tasks if (ldap__fact_configured|bool and ldap__dependent_tasks|d()) else (ldap__default_tasks + ldap__tasks + ldap__group_tasks + ldap__host_tasks + ldap__dependent_tasks) }}''. Error was a <class ''ansible.errors.AnsibleError''>, original message: An unhandled exception occurred while templating ''[{''name'': ''Ensure that {{ ldap__hosts_rdn }} object exists in LDAP directory'', ''dn'': ''{{ [ ldap__hosts_rdn ] + ldap__base_dn }}'', ''objectClass'': [''organizationalStructure''], ''attributes'': {''ou'': ''{{ ldap__hosts_rdn.split("=")[1] }}'', ''description'': ''Servers and other data center equipment''}}, {''name'': ''Create domain object for {{ ldap__device_dn | join(",") }}'', ''dn'': ''{{ [ ldap__device_domain_rdn, ldap__device_branch_rdn ] + ldap__base_dn }}'', ''objectClass'': ''{{ ldap__device_domain_object_classes }}'', ''attributes'': ''{{ ldap__device_domain_attributes }}'', ''state'': ''{{ "present"
    if (ldap__device_enabled|bool and ldap__device_separate_domains|bool) else "ignore" }}''}, {''name'': ''Create device object for {{ ldap__device_dn | join(",") }}'', ''dn'': ''{{ ldap__device_dn }}'', ''objectClass'': ''{{ ldap__device_object_classes }}'', ''attributes'': ''{{ ldap__device_attributes }}'', ''state'': ''{{ "present" if (ldap__device_enabled|bool) else "ignore" }}''}, {''name'': ''Update device object for {{ ldap__device_dn | join(",") }}'', ''dn'': ''{{ ldap__device_dn }}'', ''attributes'': ''{{ ldap__device_attributes }}'', ''state'': ''{{ "present" if (ldap__fact_configured|bool and ldap__device_enabled|bool) else ("exact" if (ldap__device_enabled|bool) else "ignore") }}''}]''. Error was a <class ''ansible.errors.AnsibleError''>, original message: An unhandled exception occurred while templating ''{''cn'': ''{{ ([ ldap__device_self_rdn.split("=")[1] ] + ldap__device_aliases) | unique }}'', ''ipHostNumber'': ''{{ ldap__device_ip_addresses }}'', ''macAddress'': ''{{ ldap__device_mac_addresses
    }}'', ''manager'': ''{{ ldap__device_managers }}''}''. Error was a <class ''ansible.errors.AnsibleError''>, original message: An unhandled exception occurred while templating ''{{ lookup("template", "lookup/ldap__device_ip_addresses.j2", convert_data=False) | from_yaml }}''. Error was a <class ''ansible.errors.AnsibleError''>, original message: An unhandled exception occurred while running the lookup plugin ''template''. Error was a <class ''ansible.errors.AnsibleError''>, original message: Unrecognized type <<class ''bool''>> for ipv6 filter <value>. Unrecognized type <<class ''bool''>> for ipv6 filter <value>'

```

# Upstream

Finding the upstream code is currently a bit of an obstacle since the code moved from *ansible* to *ansible.netcommon* and then onwards to [*ansible.utils*](https://github.com/ansible-collections/ansible.utils/blob/c5d2df6256dc92d4f985ec0b55b31b2ff8d954b8/plugins/plugin_utils/base/ipaddr_utils.py#L98) (see the [release announcement](https://mailchi.mp/redhat/the-bullhorn-50)).
Effectively the code is as follows:

```python
def _address_prefix_query(v):
    if v.size > 2 and v.ip in (v.network, v.broadcast):
        return False
    return str(v.ip) + "/" + str(v.prefixlen)
```

## Deprecation

Something that may be relevant here, but that I was unable to trace back to anywhere are the [*deprecation comments*](https://github.com/ansible/ansible/blame/2cbfd1e350cbe1ca195d33306b5a9628667ddda8/test/support/network-integration/collections/ansible_collections/ansible/netcommon/plugins/filter/ipaddr.py#L514) on the prefix filters.

Since I'm running into problems with those filters specifically it kind of feels like that may be the reason why those comments exist.

# Possible Solution

I cannot discern the purpose of the *ipaddr* filter in the debops codebase, since both address and prefix are known.
My *very* limited testing yields that this should be fine though:

```jinja
{%     for element in hostvars[inventory_hostname][iface].ipv6 %}
{%       set address = element.address + '/' + element.prefix %}
{%       if not address | ipv6('link-local') %}
{%         set _ = ldap__tpl_device_ip_addresses.append(element.address) %}
{%       endif %}
{%     endfor %}
```

benaryorg:
A thorough description of the problem can be found in #2067.
The text below reflects the commit message.

---

This allows usage of the first IPv6 address in a network (e.g. 2001:db8::/64).
Upstream seems to have issues with that usage pattern:

```bash
ansible -m debug -a 'msg={{ "2001:db8::1/64" | ansible.utils.ipaddr("address/prefix") }}' localhost
```

```text
localhost | SUCCESS => {
    "msg": "2001:db8::1/64"
}
```

```bash
ansible -m debug -a 'msg={{ "2001:db8::/64" | ansible.utils.ipaddr("address/prefix") }}' localhost
```

```text
localhost | SUCCESS => {
    "msg": false
}
```

### DIFF
--- a/ansible/roles/ldap/templates/lookup/ldap__device_ip_addresses.j2
+++ b/ansible/roles/ldap/templates/lookup/ldap__device_ip_addresses.j2
@@ -16,7 +16,7 @@
 {% for iface in ldap__tpl_device_interfaces %}
 {%   if hostvars[inventory_hostname][iface].ipv6|d() %}
 {%     for element in hostvars[inventory_hostname][iface].ipv6 %}
-{%       set address = ((element.address + '/' + element.prefix) | ipaddr('host/prefix')) %}
+{%       set address = element.address + '/' + element.prefix %}
 {%       if not address | ipv6('link-local') %}
 {%         set _ = ldap__tpl_device_ip_addresses.append(element.address) %}
 {%       endif %}
@@ -38,7 +38,7 @@
 {%     if hostvars[inventory_hostname][iface].interfaces | intersect(ldap__tpl_device_interfaces) %}
 {%       if hostvars[inventory_hostname][iface].ipv6|d() %}
 {%         for element in hostvars[inventory_hostname][iface].ipv6 %}
-{%           set address = ((element.address + '/' + element.prefix) | ipaddr('host/prefix')) %}
+{%           set address = element.address + '/' + element.prefix %}
 {%           if not address | ipv6('link-local') %}
 {%             set _ = ldap__tpl_device_ip_addresses.append(element.address) %}
 {%           endif %}


drybjed (approved):
It looks like removing the `ipaddr` check here is a good solution, I suppose that the role can trust Ansible facts to be correct IP/netmask string. I'll backport this to the stable releases.

And I feel a bit guilty as the original author of the `ipaddr` filter. :-) The bug is in the interpretation of the information parsed by `netaddr` library, it seems that at the time lack of experience with IPv6 and focus on IPv4 resulted in wrong assumptions on my part.

PR #2070, opened 2022-04-14
[ferm] Expose all options from /etc/default/ferm

monsterry:
Hi debops-Team,

I have an use case where I need to set the fast mode of ferm. At the moment this option is not exposed as ansible variables. This PR exposes all options from `/etc/default/ferm` as variables.


drybjed (approved):

PR #2071, opened 2022-04-17
Remove unsafe compression Prosody module

magicbrothers:
This module got removed from Prosody in 0.10 because of questionably security properties and resource usage.

See https://prosody.im/doc/modules/mod_compression


BerkhanBerkdemir (approved):

drybjed (approved):

PR #2075, opened 2022-04-24
Introducing Miniflux service to DebOps

BerkhanBerkdemir:
Miniflux is a minimalist and opinionated feed reader. It is written in
Golang which is distributed as a single binary. This commit provides a
playbook and role to run Miniflux service on Debian and Ubuntu
distributions.

Even though this is the first commit of the playbook/role, it is not
ready for production use.

For more information about Miniflux, please check http://miniflux.app/
website.


BerkhanBerkdemir:
Previously, I sketched some things to-do before I accept as this work as
done.

- [x] Allow user to install from Git
    @ `miniflux/defaults/main.yml#230`
- [ ] Add service restart handler
    @ `miniflux/defaults/main.yml#237` (?)
- [x] Use miniflux__gecos instead acording to `ansible.builtin.user`
    @ `miniflux/defaults/main.yml#279`
- [ ] Use community.general.ini_file to generate for conf. file
    @ `miniflux/tasks/main.yml#43`
- [x] Daemon should see environment variables
    @ `miniflux/tasks/main.yml#47`
- [ ] Create miniflux__database_url to manage connection string
    @ `miniflux/tasks/main.yml#47`
- [x] Manage Miniflux binary, so that can be used in systemd unit file
- [ ] Can I make PostgreSQL as Miniflux dependency? What happens if
    database is seperate?
- [x] Variables are not in use
    - `miniflux__upstream_gpg_key`
    - `miniflux__upstream_upgrade`

ansible/roles/miniflux/defaults/main.yml
```diff
@@ -0,0 +1,307 @@
...
+# Configuration for the :ref:`debops.golang` Ansible role.
+miniflux__golang__dependent_packages:
+
+  - name: 'miniflux'
+    upstream_type: '{{ miniflux__upstream_type }}'
+    url:
+      - src: '{{ miniflux__upstream_url_mirror + "download/2.0.35/miniflux-" + miniflux__upstream_platform }}'
+        dest: 'releases/{{ miniflux__upstream_platform }}/miniflux/miniflux/2.0.35/miniflux_2.0.35_linux_amd64'
+    url_binaries:
+      - src: 'releases/{{ miniflux__upstream_platform }}/miniflux/miniflux/2.0.35/miniflux_2.0.35_linux_amd64'
```
drybjed: The version could be specified in a separate variable, `miniflux__upstream_version`, this could simplify upgrades. You can also add a watch file (you can find and example in the `docker_registry` role) and when you run the `make versions` command, the script will find out if upstream released a new version.

ansible/roles/miniflux/defaults/main.yml
```diff
@@ -0,0 +1,307 @@
...
+
+                                                                   # ]]]
+                                                                   # ]]]
+                                                                   # ]]]
+
+# TODO: Create documentation for these used variables
+#
+miniflux__comment: 'Miniflux'
+miniflux__bind: '127.0.0.1'
+miniflux__port: '3366'
```
drybjed: Looking forward to it. :-)

ansible/roles/miniflux/tasks/main.yml
```diff
@@ -0,0 +1,55 @@
...
+    mode: '0644'
+  loop:
+    - 'etc/systemd/system/miniflux.service'
+    - 'etc/miniflux.conf'
+  notify: [ 'Reload service manager' ]
+
+- name: Run migrations
+  # environment:
+  #   DATABASE_URL: 'postgres://{{ miniflux__database_user }}:{{ miniflux__database_password }}@{{ miniflux__database_host }}:{{ miniflux__database_port }}/{{ miniflux__database_name }}'
+  command: '/usr/local/bin/miniflux -migrate'
```
drybjed: Is this command idempotent? Expect that the role can be executed multiple times against the host.

drybjed (changes requested):
Nice work so far. :+1: You should add role documentation before it will be merged into DebOps though. Keep it up! :-)

ansible/roles/miniflux/defaults/main.yml
BerkhanBerkdemir: Default Miniflux configuration uses 127.0.0.1:8080. Do I really need to set some variables for `__bind` and `__port`? I was thinking maybe I can use UNIX socket for nginx and Apache?

BerkhanBerkdemir:
I would like to point out what I did so far. As @drybjed pointed out, I created a `watch-miniflux` and moved migration commend to a configuration file in 9ccbe2f8a1e9b3081c166ecc9fc208178f9930a6. I used one unused variable and removed the other. I also wanted to add initial work of CI testing in GitLab.

ansible/roles/miniflux/defaults/main.yml
drybjed: You can use a socket, but it would be nice to have it configurable so that if somebody want to do a distributed setup, they still can. If one variable is enough to define that socket/port connection then you can go with that; otherwise if Miniflux uses bind address and port separately, then can be separate.

BerkhanBerkdemir:
It's been long time no commit I made to this particular feature. The only commits or changes I have so far related to documentation. I can't promise if I can push those changes in the near future. If anyone willing to write documentation and/or add variables, please let me know so that I can share raw documentation I have.

drybjed:
@BerkhanBerkdemir I plan to finish your PR and write documentation in the near future, so sure - show me what you've got. :-)

PR #2076, opened 2022-04-24
rollback changes for docker unit file

prk0ghy:
to 674412fad89a44145d66e394ad70ae4baa7944d9 since this had fixed an empty unit file, remove duplcate line


drybjed:
@prk0ghy This seems to only be a cosmetic change, and you switch from the new syntax that handles the nested YAML dictionaries with defaults to the old one which required separate checks for each level of the YAML dictionary. I don't think that this change is needed. Can you expand on what this is supposed to fix?

drybjed:
I'm closing this PR since the proposed changes don't seem to be necessary anymore.

PR #2083, opened 2022-05-03
Fix Prosody PKI variable name typo

BerkhanBerkdemir:
As also happened in debops#2072, typo in variable name which is fixed.
There is no usage of the variable name that is changed in DebOps.


BerkhanBerkdemir:
It has been in the repository for [4 years](https://github.com/debops/debops/blame/caca01a0ee52e0a44fad9461da39fe0698a6df40/ansible/roles/prosody/defaults/main.yml#L53) :smile:

drybjed (approved):
Thanks for the spelling fix. I don't normally backport variable renames to stable releases to avoid changes required in the repository, but I think that in this case, with variable that's probably not changed in a lot of environments, we can do the backports.

PR #2087, opened 2022-05-06
[apache2_module] Make a Debian only version that does not fatal error

julien-lecomte:
When a Apache2 configuration is broken, current community.general.apache2_module
will fail with a fatal error on command `apache2ctl -M`, even if
"ignore_configcheck" is set to true.

This version removes the apache2ctl dependency and only uses a2query
for check mode.

### DIFF
--- a/ansible/plugins/modules/apache2_module.py
+++ b/ansible/plugins/modules/apache2_module.py
@@ -1,0 +1,208 @@
+#!/usr/bin/python
+# -*- coding: utf-8 -*-
+
+# Copyright (C) 2013-2014, Christian Berendt <berendt@b1-systems.de>
+# Copyright (C) 2022 DebOps <https://debops.org/>
+# SPDX-License-Identifier:  GPL-3.0-or-later
+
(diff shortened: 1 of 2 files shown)


ansible/plugins/modules/apache2_module.py
```diff
@@ -0,0 +1,198 @@
+#!/usr/bin/python
+# -*- coding: utf-8 -*-
+
+# (c) 2013-2014, Christian Berendt <berendt@b1-systems.de>
+# GNU General Public License v3.0+ (see COPYING or https://www.gnu.org/licenses/gpl-3.0.txt)
```
drybjed: You should use the syntax supported by the `reuse` tool, so `SPDX-License-Identifier:` and `Copyright:` headers. Also add your own copyright entry as well.

drybjed (approved):
Nicely done, thanks. I'll backport the module to the stable-3.0 release so that it's available in the released DebOps collection in the future.

PR #2088, opened 2022-05-10
[libvirt] No longer install virtual qemu-kvm pkg

imrejonk:
The /usr/bin/kvm binary is now provided by the qemu-system-x86 package.


drybjed (approved):
Thanks for the update. I'll backport this to the stable releases.

PR #2089, opened 2022-05-11
Adopt fqcn module call for dpkg_divert

ssbarnea:


drybjed (approved):
Looks good. I'll fix the `ansible-lint` errors in a separate PR, merging this so that we can implement the new `ansible-lint@main` pipeline.

PR #2093, opened 2022-05-12
[pdns] No longer set UID/GID on pdns >= 4.3.0

imrejonk:
This is no longer necessary now that the PowerDNS systemd service
performs the setuid/setgid operations:
https://doc.powerdns.com/authoritative/upgrading.html#systemd-service-and-permissions


drybjed (approved):
Sounds good, thanks. :-)

PR #2094, opened 2022-05-12
[grub] Use a file to pass the PW to grub-mkpassword

Alphix:
This is a followup to PR #2054

### DIFF
--- a/ansible/roles/grub/tasks/main.yml
+++ b/ansible/roles/grub/tasks/main.yml
@@ -62,7 +62,7 @@
   no_log: '{{ debops__no_log | d(True) }}'
 
 - name: Save temporary grub-mkpasswd formatted password in secrets
-  template:
+  ansible.builtin.template:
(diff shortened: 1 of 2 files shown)


drybjed (approved):
It's a better solution than the current way to pass the passphrase via a pipe, but still isn't ideal. With this, the `grub` role expects that the Ansible Controller has `grub` software installed which might not be the case in some environments like LXC containers. Perhaps we could find a better way of hashing the `grub` password, for example with
```
openssl enc -pbkdf2 -passin env:variable
```
and some clever templating trick. The current solution dependent on `grub-mkpasswd-pbkdf2` exists because that command doesn't support passing the passphrase securely.

Alphix:
> It's a better solution than the current way to pass the passphrase via a pipe, but still isn't ideal. With this, the `grub` role expects that the Ansible Controller has `grub` software installed which might not be the case in some environments like LXC containers.

Yeah, I agree that it isn't ideal...but I should note that the use of `grub-mkpasswd` on the controller was already there before this patch...so this is still an improvement...

PR #2095, opened 2022-05-12
[pdns] Fix syntax error on local-address option

imrejonk:
This actually fixes two issues:

- Syntax error on the local-address option with pdns >= 4.5.0.
- The local-ipv6 option was actually deprecated in pdns 4.3.0 instead of in
  4.5.0. It was removed in 4.5.0.


drybjed (approved):
Looks good, thanks. I'll backport this to the stable releases.

PR #2098, opened 2022-05-13
[postldap] LDAP fixes

Alphix:
One minor fix (for setups where the LDAP server is only reachable via implicit TLS, no STARTTLS) and one cleanup for the ``postldap`` role.


imrejonk (approved):

drybjed (approved):
Thanks! I'll backport this to the stable releases.

PR #2099, opened 2022-05-13
[dovecot] LDAP fixes

Alphix:
Again, two small LDAP-related fixes to the ``dovecot`` role (fix the no-STARTTLS default and remove an unused variable)


imrejonk (approved):

drybjed (approved):
Looks good, thanks. It will be backported to the stable releases as well.

PR #2101, opened 2022-05-17
[slapd] Add DB index for "dc"

Alphix:
I'm seeing a lot of log spam like this:

	slapd[24293]: <= mdb_equality_candidates: (dc) not indexed

The reason is that postfix (see `/etc/postfix/ldap_virtual_mailbox_domains.cf`)
does a LDAP query with a "dc" comparison for each incoming email
(``(& (objectClass=dNSDomain) (dc=%s) )``).

Adding an index on dc makes the log spam go away (i.e. this is mostly cosmetic).


drybjed (approved):

PR #2107, opened 2022-05-25
Don't try to install firmware-linux-*free when on Proxmox

julien-lecomte:
I didn't find a more elegant solution than this.

It avoids needing to add the following to the inventory (for proxmox hosts only)

      apt_install__firmware_packages:
        - name: 'firmware-linux-free'
          state: "absent"
        - name: 'firmware-linux-nonfree'
          state: "absent"


drybjed (approved):
I was wondering why it was needed, but it's due to the custom kernel used by Proxmox, not its virtualization stack. I think hat besides Proxmox and old OpenVZ there are no cases of conflicts such as this, so the solution is fine for mow.

PR #2109, opened 2022-05-26
Add a reboot role

julien-lecomte:


ansible/playbooks/tools/upgrade-reboot.yml
```diff
@@ -15,26 +17,14 @@
...
-      reboot:
-        search_paths:
-          - '/lib/molly-guard'
-          - '/sbin'
-          - '/usr/sbin'
-          - '/usr/local/sbin'
-      when: reboot_required.stat.exists
-      tags:
-        - reboot
+    - role: reboot
```
drybjed: You could use the `include_role` module to avoid using `pre_tasks` and `roles` separately.

ansible/roles/reboot/tasks/main.yml
```diff
@@ -0,0 +1,27 @@
...
+
+- name: Check if reboot is required
+  ansible.builtin.stat:
+    path: '/var/run/reboot-required'
+    get_checksum: no
+  register: reboot_required
+
+- name: Is reboot required?
+  ansible.builtin.debug:
+    var: reboot_required.stat.exists
```
drybjed: Don't use the `debug` Ansible module in production code.

ansible/roles/reboot/tasks/main.yml
```diff
@@ -0,0 +1,27 @@
...
+
+- name: Reboot DebOps hosts
+  ansible.builtin.reboot:
+    search_paths:
+      - '/lib/molly-guard'
+      - '/sbin'
+      - '/usr/sbin'
+      - '/usr/local/sbin'
```
drybjed: The `search_paths` could be exposed via default variables to allow users to modify them if necessary.

ansible/roles/reboot/tasks/main.yml
```diff
@@ -0,0 +1,27 @@
...
+# .. vim: foldmarker=[[[,]]]:foldmethod=marker
+
+# Copyright (C) 2022 Julien Lecomte <julien@lecomte.at>
+# Copyright (C) 2022 DebOps <https://debops.org/>
+# .. SPDX-License-Identifier: GPL-3.0-only
+
+- name: Check if reboot is required
+  ansible.builtin.stat:
+    path: '/var/run/reboot-required'
+    get_checksum: no
```
drybjed: You can use `False` value to keep the booleans consistent with the rest of the project.

ansible/roles/reboot/tasks/main.yml
```diff
@@ -0,0 +1,27 @@
...
+  register: reboot_required
```
drybjed: Use namespacing in role variables. This variable should be named `reboot__register_required`.

ansible/playbooks/tools/reboot.yml
```diff
@@ -1,9 +1,11 @@
 ---
```
drybjed, line 1: Perhaps this playbook could be put in the `ansible/playbooks/` directory directly. Then we can use it nicely via the command line:
```
debops run site reboot -l host`
debops run common service/nginx reboot -l host
```

drybjed (changes requested):
Hmm, actually with a proper playbook it can be a very handy role for initial deployment. Good idea with adding this. :-) There are some suggested changes in the comments below.

PR #2111, opened 2022-06-02
[dovecot] Fix broken PKI hook script

imrejonk:
Noticed this when Icinga alerted me of an expiring IMAP certificate yesterday: the Dovecot PKI hook script is broken since the role redesign. It's important that we fix this, as this will cause TLS certificates to become untrusted unless the administrator reloads/restarts the Dovecot service or reboots their server on a very regular basis.

I've tested this as follows:

1. Log in to a server that has the broken PKI hook script installed.
2. `# PKI_SCRIPT_DEFAULT_CRT="/etc/pki/realms/SOME_PKI_REALM/default.crt" PKI_SCRIPT_STATE="changed-certificate" /etc/pki/hooks/dovecot`
3. Notice that the Dovecot service did not reload.
4. Fix the PKI hook script.
5. Retry step 2
6. Notice that the Dovecot service did reload this time.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -167,6 +167,14 @@ debops.boxbackup role
   playbook. This fixes an issue with Ansible stopping the site playbook
   execution when it cannot find the ``boxbackup`` role in the Collection.
 
+:ref:`debops.dovecot` role
+''''''''''''''''''''''''''
+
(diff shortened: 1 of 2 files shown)


drybjed (approved):
Good catch, thanks. I'll backport this to the stable release.

PR #2112, opened 2022-06-05
Correct default adddress to current apt packages

ser:
Stable repo has changed according to discussions on https://community.influxdata.com/t/repo-for-ubuntu-jammy/24657

One address covers all distros.


drybjed (changes requested):
@ser, the conversation you linked to points to the fact that the Ubuntu repositories also exist. I think that you should leave the `ansible_distribution` part as-is, so that Ubuntu installations work as expected. So the value should be:

```
deb https://repos.influxdata.com/{{ ansible_distribution|lower }} stable main
```

ser:
bump - just do it, @drybjed

drybjed (approved):
Looks good, thanks. Sorry that it took so long to merge. I'll try to backport this and make a new release soon.

PR #2113, opened 2022-06-05
fix(postgresql): Use FQCN to fix issue with quoting in SQL command; AND minor fixes

ypid:
Ref: ansible-collections/community.postgresql#49


drybjed (approved):
Sounds good. The `community.postgresql` collection is already in DebOps requirements, so that's covered. I'll backport the fixes to the stable release.

PR #2115, opened 2022-06-07
Loop appends to a wrong variable

kravietz:
Apparently a copy-paste error, appends to `rdnss` where it should append to `dnssl`


drybjed (approved):
You're right, thanks for the bugfix! :-) I'll backport this to the stable releases.

PR #2116, opened 2022-06-08
[dokuwiki] User shell as a variable

bleuchtang:
Sometimes we want a real shell for the dokuwiki user. This PR set the dokuwiki user shell as a variable.


drybjed (approved):
Sounds good to me. I'll backport this to the stable releases.

PR #2123, opened 2022-06-13
[python] Decode subprocess output in fact script

snipfoo:
Hello,

When run with Python 3, the `/etc/ansible/facts.d/python.fact` script silently fails on the `.split(' ')` calls (lines 45 and 56) and therefore cannot properly set the `version2` and `version3` facts:
https://github.com/debops/debops/blob/a1a88ab9732fd13d1bed5ce09f10c5e0b2e6f531/ansible/roles/python/templates/etc/ansible/facts.d/python.fact.j2#L53-L56

This is because the subprocess output is a byte string, and it needs to be decoded into a regular string for further manipulation and JSON output.

The fix proposed in this PR simply adds a call to `.decode('utf-8')`, just as other DebOps fact scripts do. E.g., for the `apt` role:
https://github.com/debops/debops/blob/a1a88ab9732fd13d1bed5ce09f10c5e0b2e6f531/ansible/roles/apt/templates/etc/ansible/facts.d/apt.fact.j2#L135-L137

Thank you!

snipfoo

### DIFF
--- a/ansible/roles/python/templates/etc/ansible/facts.d/python.fact.j2
+++ b/ansible/roles/python/templates/etc/ansible/facts.d/python.fact.j2
@@ -42,7 +42,7 @@ if output['installed2']:
         output['version2'] = subprocess.check_output(
             ['python2', '--version'],
             stderr=subprocess.STDOUT
-        ).strip().split(' ')[-1]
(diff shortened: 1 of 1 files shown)


drybjed (approved):
Good catch, thanks! I'll backport this to the stable releases.

PR #2125, opened 2022-06-14
[fhs] Add support for the *all/group/host* pattern to `fhs__*_directories`

snipfoo:
Hello,

This PR is a simple proposal to add support for the [*all/group/host* configuration pattern](https://docs.debops.org/en/stable-3.0/user-guide/universal-configuration.html#the-all-group-host-pattern) to the `fhs__*_directories` variables in the `fhs` DebOps role.

It just add the `fhs__group_directories` and `fhs__host_directories` variables (both empty by default) and combines them into `fhs__combined_directories`, which already combined `fhs__default_directories` and `fhs__directories`.

Thank you!

snipfoo

### DIFF
--- a/ansible/roles/fhs/defaults/main.yml
+++ b/ansible/roles/fhs/defaults/main.yml
@@ -131,8 +131,23 @@ fhs__default_directories:
                                                                    # ]]]
 # .. envvar:: fhs__directories [[[
 #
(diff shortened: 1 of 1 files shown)


drybjed (approved):
Sounds good to me. Thanks! I'll backport this to the stable release.

PR #2126, opened 2022-06-14
Move debops-contrib dropbear to debops

julien-lecomte:
Requires a last proofread, and a last test.


drybjed (approved):
Good work on the integration of the role with the rest of DebOps. Thanks! :-)

PR #2127, opened 2022-06-14
[cron] Fix some issues with cron job parameters

snipfoo:
Hello,

This PR proposes to fix a couple of small issues in the way the DebOps `cron` role handles the parameters of the cron jobs given in the `cron__*_jobs` variables.

* In theory, the `user` and the `backup` parameters only apply to cron files and not to individual cron jobs. This is confirmed by the `cron__combined_jobs.j2` lookup template, which specifically ignores these parameters when building the list of individual jobs for a given cron file:
  https://github.com/debops/debops/blob/a1a88ab9732fd13d1bed5ce09f10c5e0b2e6f531/ansible/roles/cron/templates/lookup/cron__combined_jobs.j2#L57-L61
  However, the [documentation](https://docs.debops.org/en/master/ansible/roles/cron/defaults-detailed.html#cron-jobs) for the `cron__*_jobs` variables incorrectly lists `user` among the job parameters which *“can be specified both in main YAML dictionary, as well as in a dictionary entry on the jobs list.”*
  Also, the `Manage cron jobs` task of the `cron` role called the `ansible.builtin.cron` module with the `backup` parameter of individual jobs (`item.1.backup`, which is always undefined) instead of that of the parent cron file (`item.0.backup`).

* Another issue was in the way individual cron jobs inherited parameters from their parent cron file: the `cron__combined_jobs.j2` lookup template first initialized the job parameters using the YAML definition of the individual job (contained in the `job` dictionary), then overrode them using the parameters from the cron file (contained in the `params` dictionary):
  https://github.com/debops/debops/blob/a1a88ab9732fd13d1bed5ce09f10c5e0b2e6f531/ansible/roles/cron/templates/lookup/cron__combined_jobs.j2#L53-L70
  This prevents one from declaring global parameters for a cron file, with overrides for specific jobs. For instance, the following configuration would not work as expected:
  ```yaml
  cron__jobs:
    'test_cron':
      special_time: 'daily'
      jobs:
        - job: 'touch /tmp/foo'
        - job: 'touch /tmp/bar'
        - job: 'touch /tmp/baz'
          special_time: 'weekly'
  ```
  The three jobs would indeed be scheduled to run daily, as the third job's `special_time: 'weekly'` would be overridden by the file-wide parameter `special_time: 'daily'`.
  As proposed in the third commit of this PR, swapping the order in which the two parameter dictionaries are merged by `cron__combined_jobs.j2` should solve this issue.

Thank you!

snipfoo


drybjed (approved):
I verified the changes, they work as expected. Thanks! I'll backport these fixes to the stable releases.

PR #2128, opened 2022-06-15
Reduce ansible-lint 6.3.0 errors to 35 (and 126 warning(s) on 4521 files.)

julien-lecomte:
Ran multiple selective and global sed for the filters with missing spaces.

The remaining 35 errors seem to have to be done manually. Some errors also seem to be false-positives, such as

    map('regex_replace', '^(imap|pop3)s$', '\\1')

which ansible-lint requires to re-write as:

    map('regex_replace', '^(imap | pop3)s$', '\\1')


drybjed (approved):
Thanks for doing most of the work needed to handle new `ansible-lint` rules. I'll try to fix the rest as soon as possible, so that CI is useful again. :-)

julien-lecomte:
You're welcome! I'll also look into helping out for the remaining cases.

PR #2132, opened 2022-06-15
[pki] add acme_domains on initialization of realm

### LINKED ISSUE #2131: [pki] realm is not initialized with acme_domains
Hello,
I tried to create a ACME/Letsencrypt certificate with multiple 2nd level domains and could not get it to work.
I think the cause is because the realm config is initialized without the --acme-domains parameter in tasks/main.yml:

https://github.com/debops/debops/blob/3edefdba9b382d879a75d7599a3acf5d49a32b9f/ansible/roles/pki/tasks/main.yml#L166-L186

If an suitable parameter is added, i.e.

`--acme-domains "{{ item.acme_domains | d([]) | join('/') }}"`

 the realm conf is updated and it seems to work OK (if one forces a new request creation)

Will try to create a pull request if i get github to cooperate.

SunStrom:
fixes #2131


drybjed (approved):
Thanks for the fix. I'll backport it to the stable releases.

PR #2134, opened 2022-06-18
Add replace capabilities to debops.resources

julien-lecomte:
Context: After installing and running the playbook against my Debian 11 hosts, I always end up with `use_authtok` in one of the lines of `/etc/pam.d/common-password`.

The presence `use_authtok` in said file prevents people, including root, from changing their password. The error message always being `passwd: Authentication token manipulation error`.

By adding `ansible.builtin.replace` capacities to service/resources, this enables the end user to remove erroneous `use_authtok` without having to either create a specific task, or a any other role.

Of course, this is just one use case, many more are possible since this solution is generic.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -38,6 +38,9 @@ New DebOps roles
 - The :ref:`debops.reboot` role can be used to reboot, forcefully or only if
   required, any DebOps host.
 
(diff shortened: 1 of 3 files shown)


drybjed (approved):
Looks like a good idea, thanks. I'll include some documentation for completeness.

PR #2135, opened 2022-06-20
[opensearch] Add new role debops.opensearch

imrejonk:
The new role can only manage local, unsecured installations, which is what my employer needed for our Graylog installation. I decided not to copy the existing `debops.elasticsearch` role with all its features, because OpenSearch has some subtle differences and I have yet to set up a full Elasticsearch/OpenSearch cluster before I can properly implements such things in a role like this. This is a completely separate role with just the things necessary to do what it promises.

Tested with OpenSearch 1.3.3 and 2.0.1.

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -1761,6 +1761,19 @@ stages:
     JANE_DIFF_PATTERN: '.*/roles/opendkim/.*'
     JANE_LOG_PATTERN: '\[opendkim\]'
 
+'opensearch role':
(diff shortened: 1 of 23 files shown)


imrejonk:
The CI failure seems unrelated. The check succeeds when testing locally with ansible-lint 5.4.0.

drybjed (approved):
Nicely done. :-) I'm sure that the role will be improved and expanded in the future, and it will be easier when it's merged with the main project. Good work, keep it up. :)

PR #2136, opened 2022-06-20
[slapd] Make saslauthd role execution configurable

imrejonk:
This adds a new default variable `slapd__saslauthd__enabled` (which defaults to `True` for backwards compatibility) that allows users to skip the saslauthd service installation when running the slapd playbook.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -71,6 +71,11 @@ General
   the :ref:`debops.ldap` role is detected on the host. Currently only user
   authentication and Django ACL system is supported via LDAP groups.
 
+:ref:`debops.slapd` role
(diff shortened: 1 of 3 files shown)


drybjed (approved):
Ansible has some issues with `when:` statements on the playbook level so I try to avoid them, but OK, let's add this one. I wonder why you would want to skip `saslauthd` though, without this using something like
```
ldapsearch -Z '(uid=*)'
```
on the command line fails, and you need to specify the full DN of the account you want to use.

However, I'll change the variable name to `slapd__saslauthd_enabled`, since the variable is not passed to the `saslauthd` role. The current name with two namespace separators is incorrect when applied to the scheme used in DebOps.

PR #2138, opened 2022-06-22
[pki] Change symlink to public cert if needed on renewal

julien-lecomte:
Made the pki renewal script aware of certificats private and public part to self-correct itself.

### DIFF
--- a/ansible/roles/pki/files/etc/letsencrypt/renewal-hooks/deploy/pki-realm-refresh-keys
+++ b/ansible/roles/pki/files/etc/letsencrypt/renewal-hooks/deploy/pki-realm-refresh-keys
@@ -32,6 +32,16 @@ if [ -d "/etc/letsencrypt/live" ] ; then
         chmod 0640 private/key_chain_dhparam.pem
         chown root:ssl-cert private/key_chain_dhparam.pem
 
+        # Verify that public and private parts match
+        pub=$(openssl x509 -noout -modulus -in default.crt | openssl md5)
+        key=$(openssl rsa  -noout -modulus -in default.key | openssl md5)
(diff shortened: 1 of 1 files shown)


drybjed:
This PR solves an issue where a PKI realm is configured without internal CA enabled, which results in a `selfsigned/` set of certificates and the `default.crt` symlink pointing to the `public/cert_dhparam.pem` file. However, the solution here is IMO put in a wrong place - instead of putting this in a `certbot` deploy-hook script, the `pki-realm` script that manages the PKI realms need to check if the CA has changed from `selfsigned/` to `acme/` or some other, and correct the symlink. This way, both `certbot` and `acme-tiny` scripts, as well as any externally provided certificates can function properly.

I'll try to fix this issue in the correct place at some point in the future, as time permits.

julien-lecomte:
I believe that some functionality has to be in the `certbot` deploy-hook script since testing has shown that apache2 server can't restart because of wrong SSL cert pair after a certbot reissue of certs.

drybjed:
@julien-lecomte When the Apache 2 issue happens, is the `default.crt` symlink pointing to the `public/cert_dhparam.pem` file, or `public/cert_intermediate_dhparam.pem` file? The `pki-realm` script uses `certbot` to initialize the `acme/` CA, at that point it should also update the `default.crt` symlink to point it to `public/cert_intermediate_dhparam.pem` file. Afterwards, on a re-issue of the certificate, the symlink will be correct and Apache 2 should correctly restart.

drybjed (approved):
The solution seems reasonable. It will probably be integrated with the `pki-realm` script in a future rewrite. Thanks. :-)

PR #2140, opened 2022-06-22
[resolvconf] Enable full static resolvconf without package

julien-lecomte:
If the idea is OK, then I'll write up documentation and changelog.

Works with the following example:

```
---
# debops/roles/resolvconf/defaults/main.yml
groups:
  debops_service_resolvconf:

vars:
  resolvconf__enabled: true

  resolvconf__deploy_state: 'absent'
  resolvconf__static_enabled: True

  resolvconf__static_content: |
    domain {{ netbase__domain }}
    search {{ netbase__domain }}.
    nameserver 1.1.1.1
```

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -87,6 +87,9 @@ Updates of upstream application versions
 - In the :ref:`debops.netbox` role, the NetBox version has been updated to
   ``v3.2.4``.
 
+- In the :ref:`debops.resolvconf` role, you can now write a fully static 
+  :file:`/etc/resolv.conf` file without the ``resolvconf`` package.
+
 General
 ~~~~~~~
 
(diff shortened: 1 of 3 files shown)


julien-lecomte:
There's currently one issue with the task:

```
- name: Add/remove diversion of /etc/resolvconf/interface-order
  dpkg_divert:
    path: '/etc/resolvconf/interface-order'
    state: '{{ resolvconf__deploy_state }}'
    delete: True
  notify: [ 'Refresh /etc/resolv.conf' ]
  when: (resolvconf__enabled | bool and
         resolvconf__deploy_state in [ 'present', 'absent' ])
```

It should remove diversion if state is absent, but the notify will fail.

julien-lecomte:
Notify on absent issue solved by duplicating task, one for present with handler, one for absent without handler.

drybjed (approved):
Sounds good, thanks. I'll backport this to the stable release.

PR #2143, opened 2022-06-27
Switch 'from/to_json' to 'from/to_yaml' filters

drybjed:
The new Jinja2 3.x release changed the behaviour of 'to_json' filter
- it doesn't accept YAML lists as input anymore, only YAML dictionaries.
To fix this issue, we need to switch the filters to use YAML as
a transport between various places in the code, which permits lists to
be passed between variables.

There will be more places in the code that this might need to happen,
they will be dealt with on later commits.

### DIFF
--- a/ansible/playbooks/templates/debops__tpl_macros.j2
+++ b/ansible/playbooks/templates/debops__tpl_macros.j2
@@ -159,7 +159,7 @@
 {%       endfor %}
 {%     endif %}
 {%   endif %}
-{{ merged_dict | to_json }}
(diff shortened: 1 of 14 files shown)


morremeyer:
Could a new release be tagged that contains this PR? 

Currently, this blocks the upgrade to ansible 6 for everyone affected by #2148.

drybjed:
@morremeyer Sorry for the long delay. You're right, this is a critical fix, I'll try to make a new release tomorrow, with backported changes.

PR #2150, opened 2022-07-08
[resources] Add a way to replace strings in files

drybjed:


drybjed:
The `ansible-lint` errors are confined to the `opensearch` Ansible role, so I'll ignore them in this case.

PR #2159, opened 2022-07-24
[nginx] Document further advantages with waiting for network

Alphix:
The fix which was introduced with PR #2144 also addresses a separate
Debian bug. Document this fact so that the workaround isn't removed
unless and until both issues have been fixed.


drybjed (approved):
Good idea, thanks.

PR #2160, opened 2022-07-24
[rspamd] Fix PEP632 deprecation of distutils

### LINKED ISSUE #2147: rspamd-dkim-keygen script: Deprecation Warning
The script shipped by `rspamd` role to renew DKIM keys, which is launched by Cron `/etc/cron.d/rspamd-dkim-keygen`, has an stderr output:
```
/usr/local/sbin/rspamd-dkim-keygen:46: DeprecationWarning: The distutils package is deprecated and slated for removal in Python 3.12. Use setuptools or check PEP 632 for potential alternatives
  import distutils.spawn
```

So every first day of each months, the user `_rspamd` is receiving the cron error output by mail.

(Which we do not actually receive: it's delivered locally. The role should set an alias `_rspamd` → `root`, in `/etc/aliases`, no?)

Alphix:
The alternative, ``shutil.which``, was introduced with Python 3.3, which
predates Debian stretch, so it should be safe to use without any
fallback to ``distutils``.

Closes: #2147

### DIFF
--- a/ansible/roles/rspamd/files/usr/local/sbin/rspamd-dkim-keygen
+++ b/ansible/roles/rspamd/files/usr/local/sbin/rspamd-dkim-keygen
@@ -42,11 +42,11 @@ import logging
 import tempfile
 import argparse
 import subprocess
-import distutils.spawn
+import shutil
 
 __license__ = 'GPL-2.0 OR Apache-2.0'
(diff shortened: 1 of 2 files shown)


Alphix:
No idea what the failing ansible-lint CI check is about....

drybjed (approved):
Thanks for the update. I'll backport this to the stable releases as well.

PR #2163, opened 2022-07-27
[dovecot] Fix backslashes in regex_replace

Alphix:
Commit ca2803e32de85b09ec1ebc87e16d5151031f708a removed the quoting around
the protocols parameter. The problem is that regex_replace is kind of wonky
and will behave differently depending on whether the var is quoted or not, see:
https://github.com/ansible/ansible/issues/33202

The result is that if there's a match (for e.g. "imaps"), the protocols
will contain a literal "\1" rather than "imap".


drybjed (approved):
Thanks for the heads up! I need to remember that such regexes need to have double quotes for this reason.

PR #2166, opened 2022-08-01
[roundcube] Update to Roundcube 1.6.0

imrejonk:
This also updates the role defaults to accommodate some breaking changes
[1]. One interesting change can be found in the LDAP address book
configuration: Roundcube no longer recognizes the `port` and `use_tls`
options, they should be specified in the `host` option instead. I used
the product filter [2] here to construct the new `host` list from the
existing role variables.

[1] https://github.com/roundcube/roundcubemail/releases/tag/1.6.0
[2] https://docs.ansible.com/ansible/latest/user_guide/playbooks_filters.html#products


drybjed (approved):
Nicely done, thanks. :-)

Issue #2171, opened 2022-08-04
[nginx] 'enabled' servers var is not functional, has no effect


marcidy:
Trying to disable an nginx server by using the `enabled:`  var doens't remove the symlink from `/etc/nginx/sites-enabled`.

To reproduce with a standard server as an example, declare the default welcome server in `host_vars/<host>.yml and run the `nginx` role.

```yaml
nginx_server_welcome:
  enabled: True
  name: [ 'welcome' ]
  welcome: True
  welcome_domain: '{{ nginx_welcome_domain }}'
  csp: "default-src 'none'; style-src 'self' 'unsafe-inline';img-src 'self';"
  csp_enabled: True
```

The server is created as expected.

Change `enabled` to `False`, and run again (can limit tag to `role::nginx:servers`).  

Ansible reports the variable change but takes no action:

```bash
TASK [debops.debops.nginx : Generate nginx server configuration] **********************************************************************************************
ok: [mattarcidy] => (item={'name': ['welcome'], 'enabled': False, 
[...]
```
The site remains enabled (symlink exists in `/etc/nginx/sites-enabled`).

The responsible task is in `nginx/tasks/nginx_servers`.
```yaml
- name: Disable nginx server configuration
  file:
    path: '/etc/nginx/sites-enabled/{{ item.filename
                                       | d(item.name if item.name is string else item.name[0] | d("default")) }}.conf'
    state: 'absent'
  with_flattened:
    - '{{ nginx__servers }}'
    - '{{ nginx__default_servers }}'
    - '{{ nginx__internal_servers }}'
    - '{{ nginx__dependent_servers }}'
    - '{{ nginx_servers | d([]) }}'
    - '{{ nginx_default_servers | d([]) }}'
    - '{{ nginx_internal_servers | d([]) }}'
    - '{{ nginx_dependent_servers | d([]) }}'
  notify: [ 'Test nginx and restart' ]
  when: (item.name is defined and
         ((item.state|d("present") == 'absent') or
          (item.enabled|d() and not item.enabled|bool) or
          (item.delete|d()|bool)))
```

This line in the `when` condition is not effective when enabled is defined AND false:
```yaml
          (item.enabled|d() and not item.enabled|bool) or
```
It evaluates to either:
  (true) and not (true)  == false
or: 
   (false) and not (false) == false

I believe it ought to be:
```yaml
          (item.enabled is defined|d() and not item.enabled|bool) or
```

[...]

PR #2172, opened 2022-08-04
[nginx] Disable servers using `enabled` var

marcidy:
Disabling nginx servers by using `enabled: False` is fixed
in the task 'when' condition.  The removal of symlinks
in `/etc/nginx/sites-enabled` is controlled by this variable, but
the check did not work when the variable is defined and is false.


marcidy:
for #2171

goetzk:
I have some code in may playbooks to work around this - hadn't occurred to me to look for a PR before now.

@drybjed any particular testing you're hoping for before reviewing this? i could try and test in the next few weeks if thats what its waiting on.

PR #2173, opened 2022-08-05
[cgit] Add new role

Alphix:
This adds support for setting up the cgit web frontend for git repos.

See: https://git.zx2c4.com/cgit/about/


Alphix:
Rebased on current master

PR #2175, opened 2022-08-07
[sshd] Fix molly-guard when using dracut

Alphix:
molly-guard is currently installed by default by the sshd role.

On machines using dracut (not default), molly-guard will unfortunately mean
that a reboot is no longer possible, see:
https://bugs.debian.org/992351

This patch ensures that a workaround is in place (noop on systems not
using dracut). Also, I've defined the new variable sshd__combined_packages
to make the resulting tasks/main.yml look a bit nicer.

Not 100% happy about adding three more task steps for this corner-case, but
a non-working reboot is on the other hand a pretty nasty scenario.

### DIFF
--- a/ansible/roles/sshd/tasks/main.yml
+++ b/ansible/roles/sshd/tasks/main.yml
@@ -90,14 +90,31 @@
     state: 'directory'
     mode: '0755'
 
+  # molly-guard + dracut means that the system can't reboot anymore, see:
+  # https://bugs.debian.org/992351
+- name: Ensure that dracut directory exists before installing molly-guard fix
+  ansible.builtin.file:
+    path: '/etc/dracut.conf.d'
+    state: 'directory'
+    mode: '0755'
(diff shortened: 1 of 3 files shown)


ansible/roles/sshd/tasks/main.yml
```diff
@@ -91,14 +91,31 @@
     mode: '0755'
   when: ansible_service_mgr != 'systemd'
 
+  # molly-guard + dracut means that the system can't reboot anymore, see:
+  # https://bugs.debian.org/992351
+- name: Ensure that dracut directory exists before installing molly-guard fix
```
julien-lecomte, line 95: Doesn't this and the following create the directories even for users that don't use dracut?

Alphix: Yep, the problem is that if `dracut` is installed later, a reboot will also blow up...unless the workaround is already in place...

Issue #2177, opened 2022-08-11
Debops Roles02 Minio - Skeleton parameter issue on user role


Izual750:
HI,

**Environment**: 
Here is my Ansible version:
```
ansible [core 2.12.7]
  config file = /home/izual750/Documents/Projects/BSO/cn-cloud/ansible/ansible.cfg
  configured module search path = ['/home/ozual750/.ansible/plugins/modules', '/usr/share/ansible/plugins/modules']
  ansible python module location = /usr/lib/python3/dist-packages/ansible
  ansible collection location = /home/izual750/.ansible/collections:/usr/share/ansible/collections
  executable location = /usr/bin/ansible
  python version = 3.10.4 (main, Jun 29 2022, 12:14:53) [GCC 11.2.0]
  jinja version = 2.11.3
  libyaml = True
```

**Issue**: 
And when I'm running the DebOps Role for Minio, as soon as Ansible reach the step:
`TASK [debops.roles02.minio : Create required UNIX system account]` it faces an issue with skeleton parameter of the user role.

```
<172.16.0.163> (0, b'', b'')
fatal: [prod-storage-03]: FAILED! => changed=false 
  invocation:
    module_args:
      append: true
      authorization: null
      comment: MinIO
      create_home: true
      expires: null
      force: false
      generate_ssh_key: null
      group: minio
      groups: []
      hidden: null
      home: /var/local/minio
      local: null
      login_class: null
      move_home: false
      name: minio
      non_unique: false
      password: null
      password_expire_max: null
      password_expire_min: null
      password_lock: null
      profile: null
      remove: false
      role: null
      seuser: null
      shell: /usr/sbin/nologin
      skeleton: /dev/null
      ssh_key_bits: 0
      ssh_key_comment: ansible-generated on prod-storage-03
      ssh_key_file: null
      ssh_key_passphrase: null
      ssh_key_type: rsa
      state: present
      system: true
      uid: null
      umask: null
      update_password: always
  msg: '[Errno 20] Not a directory: ''/dev/null'''
```

I can confirm /dev/null exists on the remote hosts, and is accessible for the ansible user.
```
bso@prod-storage-03:~$ stat /dev/null
  File: /dev/null
  Size: 0         	Blocks: 0          IO Block: 4096   character special file
Device: 5h/5d	Inode: 4           Links: 1     Device type: 1,3
Access: (0666/crw-rw-rw-)  Uid: (    0/    root)   Gid: (    0/    root)
[...]

PR #2178, opened 2022-08-11
Issue2177: Bug fix with skeleton parameter set as /dev/null for user role

Izual750:
Linked issue: https://github.com/debops/debops/issues/2177


drybjed (approved):
Nothing we can do about it since it's a bug in Ansible `user` module, so let's remove the `skeleton` parameter for now. I'll backport this to the stable releases as well.

Izual750:
Thanks :)

PR #2180, opened 2022-08-17
[owncloud] Update ownCloud to 10.10

Gomez:
Updated repo url, packages name and version.


drybjed (approved):
Changes look fine to me apart from change of the protocol to `http://`. Since the repository is accessible over HTTPS as well, I'll stick to it so that we don't lower the security of the installation. I'll fix any conflicts and backport this change to the `stable-3.0` branch since it will stay supported for some time.

PR #2181, opened 2022-08-20
Add pipeline to run ansible-lint from master

ssbarnea:
See: https://github.com/ansible/ansible-lint/discussions/1403#discussioncomment-2654415

### DIFF
--- a/.ansible-lint
+++ b/.ansible-lint
@@ -12,7 +12,11 @@ exclude_paths:
   - 'lib/debops-api/tests/'
   - '.gitlab-ci.yml'
 
-# Add vitual roles for playbook examples
+# Temporary added until we switch to fully qualified modules
+mock_modules:
+  - cran
+
+# Add virtual roles for playbook examples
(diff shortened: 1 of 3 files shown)


drybjed:
@ssbarnea Since you tried this a second time, I've added the separate workflow in the `debops/debops` repository, via https://github.com/debops/debops/pull/2186 to avoid issues with `reuse`. It seems to be passing, which probably shouldn't happen at the moment due to all the warnings and false-positives. I'm not even sure if the `ansible-lint` from the `main` branch is used, there's no information about the version or even commit id that I can see, any way to correlate that in the workflow run?

ssbarnea:
@drybjed I do have a change on https://github.com/ansible/ansible-lint/pull/2375 that is fixing most of the problems. We still have few type of strings that we cannot format but we only display a runtime warning for these and do nothing. Still, I see that the newsion is able to correctly deal with unaries and with arrays and that there are few minor updates to be made to debops.

PR #2182, opened 2022-08-28
Allow ARM-based Macs to run docker_server role

ghost:
The latest Mac's use 'arm64' arch and this is not present inside the
'docker_server__upstream_arch_map' which means that the correct
'docker_server__upstream_repository' url cannot be created.

The solution is to simply add an option for 'aarch64' which maps to
'arm64' so the correct url can be generated.


drybjed (approved):
Thanks, I'll backport this to the stable releases as well.

PR #2184, opened 2022-08-30
[keyring] Make keyring path configurable

Alphix:
This is useful e.g. when configuring third-party repos, since the GPG
key can be stored in a separate file and referenced in the
/etc/apt/sources.list.d/* entry like this:

deb [signed-by=/usr/share/keyrings/example.gpg] https://example.com/debian/ stable main

Which means that example.gpg won't be trusted to sign *any* package.


drybjed (approved):
Excellent idea, thanks! I'll try to find a good example and add some more documentation. Perhaps the current DebOps roles that utilize `keyring` could be switched to handling keys with this as well. There's just no good way to specify the keyring yet.

PR #2191, opened 2022-09-02
[docs] Add faq about DebOps being slow

aptiko:


drybjed (approved):
Nice work. :-) I'll make some adjustments and expand a bit on round-trips between the Ansible Controller and remote hosts.

PR #2196, opened 2022-09-05
Update OpenSearch to 2.2.1, fix permission issue

imrejonk:
Update OpenSearch to 2.2.1

The /etc/opensearch directory permissions would be set to 0640 on every
update, which caused OpenSearch startup failures. This commit fixes that
issue and makes the applied ownership and permissions more visible.


ypid (approved):
Looks good to me.

PR #2198, opened 2022-09-14
pki: convert intermediate to PEM

muelli:
The pki-realm script tried to convert the intermediate certificate by running openssl to see whether it could read the file. Ubuntu's OpenSSL 3.0.2 15 Mar 2022 (Library: OpenSSL 3.0.2 15 Mar 2022) will happily read a DER file, though, and thus, the test passes which causes the conversion to be skipped.

This change explicitly requests openssl to read a PEM format. If that fails, the conversion will be pursued.

### DIFF
--- a/ansible/roles/pki/files/usr/local/lib/pki/pki-realm
+++ b/ansible/roles/pki/files/usr/local/lib/pki/pki-realm
@@ -1071,8 +1071,8 @@ convert_der_to_pem () {
     local input_file="${1}"
 
     if [ -n "${input_file}" ] && [ -r "${input_file}" ] ; then
-        if ! openssl x509 -in "${input_file}" -noout 2>/dev/null ; then
-            openssl x509 -inform DER -in "${input_file}" -out "${input_file}.tmp"
+        if ! openssl x509 -inform PEM -in "${input_file}" -noout 2>/dev/null ; then
+            openssl x509 -in "${input_file}" -outform PEM -out "${input_file}.tmp"
             mv "${input_file}.tmp" "${input_file}"
         fi
     fi


muelli:
what I realise now: `inform` has been provided before. I don't know whether this was a necessity. So other versions of openssl need to be tested for accepting the certificate even without `inform`.

ypid:
@muelli Does this fix #2250?

muelli:
looks like it has the potential to fix it!

goetzk:
Might also resolve #2232

goetzk:
I've just done the following test:

* Replaced `/usr/local/lib/pki/pki-realm` on my target server with the file in this PR
* Replaced `~/.ansible/collections/ansible_collections/debops/debops/roles/pki/files/usr/local/lib/pki/pki-realm` with the file in this PR
* Moved aside `/etc/pki/realms/example.com` on my target server
* Moved aside `secret/pki/realms/by-host/example.com/example.com` on my ansible controller

Running a playbook which involved debops.pki generated full.pem and intermediate_root.pem without any binary components.
```
# grep 'BEGIN CERT' public/full.pem 
-----BEGIN CERTIFICATE-----
-----BEGIN CERTIFICATE-----
-----BEGIN CERTIFICATE-----
# grep 'BEGIN CERT' public/intermediate_root.pem
-----BEGIN CERTIFICATE-----
-----BEGIN CERTIFICATE-----
```

Before that change I was experiencing the problem as described in the two related bug reports so I'm fairly confident in this solution.

alinalexandru:
I am also affected by this. What else needs to be done for this to be approved ?

ypid (approved):

drybjed (approved):
Thanks for the fix! I'll backport this to stable releases as well.

PR #2200, opened 2022-09-20
Fixed broken jinja syntax

ssbarnea:
Found this while working on improving the linter. It causes ansible
to produce a runtime error.


Alphix:
@ssbarnea : The syntax was fixed as part of the ansible-lint fixes in PR #2214, so I think this can be closed...?

PR #2202, opened 2022-09-25
[pki] Fix missing quotes causing invalid directory creation

MichaelHarder:


drybjed (approved):
I can't the issue or an explanation of it, but the change seems sound. I'll backport it to the stable releases as well.

PR #2208, opened 2022-10-09
DebOps dev patchset 13: Fix debops-contrib dropbear playbook, fix(btrfs): btrfs-tools is now provided by btrfs-progs package

ypid:
Fixes: #2126


drybjed (approved):
Thanks for the fixes. Some of these commits will be useful in stable branches, for example the `nginx` fixes. I'll cherry pick them individually.

Issue #1171, opened 2019-12-04
[debops.slapd] Configure TLS certificates failed


erikschwalbe:
Ubuntu 18.04.3
ansible (2.9.1)
debops (1.2.0)

added new host "ldap" to debops_service_slapd and run debops with default settings

...
TASK [debops.slapd : Configure TLS certificates] ********************************************************************************************************************************************************************************************************************
task path: /home/ansible/.local/share/debops/debops/ansible/roles/debops.slapd/tasks/slapd_tasks.yml:17
The full traceback is:
Traceback (most recent call last):
  File "/tmp/ansible_ldap_attrs_payload_3WWTuV/ansible_ldap_attrs_payload.zip/ansible/modules/ldap_attrs.py", line 399, in main
  File "/usr/lib/python2.7/dist-packages/ldap/ldapobject.py", line 629, in modify_s
    return self.modify_ext_s(dn,modlist,None,None)
  File "/usr/lib/python2.7/dist-packages/ldap/ldapobject.py", line 602, in modify_ext_s
    resp_type, resp_data, resp_msgid, resp_ctrls = self.result3(msgid,all=1,timeout=self.timeout)
  File "/usr/lib/python2.7/dist-packages/ldap/ldapobject.py", line 749, in result3
    resp_ctrl_classes=resp_ctrl_classes
  File "/usr/lib/python2.7/dist-packages/ldap/ldapobject.py", line 756, in result4
    ldap_result = self._ldap_call(self._l.result4,msgid,all,timeout,add_ctrls,add_intermediates,add_extop)
  File "/usr/lib/python2.7/dist-packages/ldap/ldapobject.py", line 329, in _ldap_call
    reraise(exc_type, exc_value, exc_traceback)
  File "/usr/lib/python2.7/dist-packages/ldap/ldapobject.py", line 313, in _ldap_call
    result = func(*args,**kwargs)
OTHER: {'desc': u'Other (e.g., implementation specific) error'}

fatal: [ldap]: FAILED! => {
    "changed": false,
    "details": "{'desc': u'Other (e.g., implementation specific) error'}",
    "invocation": {
        "module_args": {
            "attributes": {
                "olcTLSCACertificateFile": "/etc/pki/realms/domain/CA.crt",
                "olcTLSCertificateFile": "/etc/pki/realms/domain/default.crt",
                "olcTLSCertificateKeyFile": "/etc/pki/realms/domain/default.key"
            },
            "bind_dn": null,
            "bind_pw": "",
[...]

Issue #690, opened 2019-02-15
[debops.apparmor] Migrate debops-contrib.apparmor role into the main playbook


alinalexandru:
Migrate debops-contrib.apparmor role into the main playbook, converting the configuration to use parse_kv_items
Remove the custom grub support in the apparmor role and use debops.grub to enable/disable kernel arguments for apparmor

drybjed:
The AppArmor role should be a part of the `common` playbook, because AppArmor will be enabled by default. That means that it should be activated via `apparmor__enabled` boolean, and disabled by default on older OS releases.

Also, probably using `debops.grub` directly is a bad idea (it was mine originally, don't fret), because that would require adding `debops.grub` to the common playbook, but it could be handled differently - the Debian `grub` packages support configuration in `/etc/default/grub.d/` directory. The `debops.apparmor` role could add its configuration snippet there and include the `update-grub` handler to reconfigure the bootloader on changes.

Alphix:
Note that this is largely done now, the only thing missing is to add the `apparmor` role to the `common` playbook...

ekin-oener-provsn:
With the latest release (3.2.4) `debops.apparmore` role is still not added to the `common` playbook.
I could prepare a PR for this issue, but I am not certain if there is an order to be followed in layer/common.yml.

Since apparmor is enabled by default and no profiles, locales or tunables are configured by default in the role, only `apparmor__packages` will be installed. At least in newer debian releases (>11).

My simple suggestion would be adding it at the end of the common role.

EDIT:
The documentation clearly states, that it is already in common.yml btw.
https://docs.debops.org/en/stable-3.2/ansible/roles/apparmor/getting-started.html#example-inventory

Issue #456, opened 2018-09-27
[debops.slapd] Configure TLS certificates - apparmor getting in the way


mikenowak:
While running this role with pretty much default settings I came across this error

```
TASK [debops.slapd : Configure TLS certificates (first time only)] 
*************************************************************************
fatal: [ldap1.internal.example.jp]: FAILED! => {"changed": true, "cmd": "ldapmodify -Y EXTERNAL -H
 ldapi:/// -f /tmp/tmp.mocM3TIqD0/configure_tlscerts.ldif", "delta": "0:00:00.012040", 
"end": "2018-09-27 18:13:30.458942", "msg": "non-zero return code", "rc": 80, 
"start": "2018-09-27 18:13:30.446902", "stderr": "SASL/EXTERNAL authentication started\nSASL
 username: gidNumber=0+uidNumber=0,cn=peercred,cn=external,cn=auth\nSASL SSF: 
0\nldap_modify: Other (e.g., implementation specific) error (80)", 
"stderr_lines": ["SASL/EXTERNAL authentication started", "SASL username: 
gidNumber=0+uidNumber=0,cn=peercred,cn=external,cn=auth", "SASL SSF: 0", "ldap_modify: Other (e.g., implementation specific) error (80)"],
 "stdout": "modifying entry \"cn=config\"", "stdout_lines": ["modifying entry \"cn=config\""]}
```

It appears that apparmor is getting in the way - perhaps an area for improvement?

```
Sep 27 18:13:30 ldap1 kernel: audit: type=1400 audit(1538061210.453:22): apparmor="DENIED"
 operation="open" profile="/usr/sbin/slapd" name="/etc/pki/realms/domain/private/key.pem" 
pid=17221 comm="slapd" requested_mask="r" denied_mask="r" fsuid=111 ouid=0
```

For the record this is on Ubuntu 18.04 and I don't believe this is officinally supported yet.

drybjed:
I'd like to add AppArmor support in DebOps soon. Ubuntu already has it, and Debian Buster will most likely have it enabled by default. One thing that I have to do first is to fix [this `debops.grub` issue](https://github.com/debops/ansible-grub/issues/8#issuecomment-312424389), then the AppArmor role will be able to use `debops.grub` to enable AppArmor support.

[...]

PR #2209, opened 2022-10-11
[apparmor] Convert to universal configuration and move from contrib

Alphix:
This (quite lightly tested) pull request modernises the apparmor role and moves it from contrib to the official list of roles.

This, in turn, allows apparmor to be used as a dependency for other roles, which allows us to fix various issues caused by the combination of apparmor and customizations made by DebOps roles (e.g. `/etc/pki` for SSL/TLS certs).

This should address at least:
#1171
#690
#456

And is also a prerequisite for another role I'm working on (`bind`).


Alphix:
Note that this PR doesn't add `apparmor` to the `common` playbook yet...

Also: once the PR has been approved, we need to:
* add integration to `slapd`
* add integration to `apt_cacher_ng` and remove `debops-contrib-playbooks/service/apt_cacher_ng.yml`
* add integration to `dnsmasq` and remove `debops-contrib-playbooks/service/dnsmasq.yml`

Alphix:
Merged some of the fixups with their parent commits and also rebased this PR on top of #2214 so that the CI checks can be performed properly.

Alphix:
Rebased, fixing conflicts and also squashed the fixup patches with the parent patches

drybjed (approved):
Great job with the migration, thank you! I agree that this role needs to be in the common playbook, but let's merge it first and test in the wild to make sure it doesn't break different environments.

ypid:
#2325 includes a small fix. Otherwise awesome work @Alphix! I only did a review and light testing for now.

Edit: Ah, #2319 already fixes the bug I found :)

PR #2210, opened 2022-10-11
[ntp] Misc systemd-timesyncd / virtualization fixes

Alphix:


ansible/roles/ntp/defaults/main.yml
```diff
@@ -131,6 +133,7 @@ ntp__base_packages:
   - '{{ "chrony" if (ntp__daemon == "chrony") else [] }}'
   - '{{ "ntp" if (ntp__daemon == "ntpd") else [] }}'
   - '{{ "openntpd" if (ntp__daemon == "openntpd") else [] }}'
+  - '{{ "systemd-timesyncd" if (ntp__daemon == "systemd-timesyncd") else [] }}'
```
gardouille: Installing **systemd-timesyncd** might cause some issue on Debian Buster because it comes with systemd 241 by default.

systemd-timesyncd is only available in buster-backports and require to upgrade systemd to 245.4 (buster-backports).

Alphix: Buster and Stretch both have `systemd-timesyncd`...it's just that it's not in its own package, it's included in the [systemd package](https://packages.debian.org/buster/amd64/systemd/filelist), but yeah, good point, I can make the line above conditional on distro != buster/stretch.

Alphix: Done in a separate commit....

julien-lecomte:
Interesting, I'll have to test this patch out. If I understand correctly, it make `systemd-timesyncd` the default on bullseye, but will it also uninstall openntpd?

Alphix:
> Interesting, I'll have to test this patch out. If I understand correctly, it make `systemd-timesyncd` the default on bullseye, but will it also uninstall openntpd?

Unless I've made a mistake, it won't *change* the NTP daemon on existing installations (see the ``ansible_local.ntp.daemon`` check in the ``ntp__daemon`` variable). It'll change the default for *new* installations (so it won't uninstall openntpd, but rather not install it in the first place)

drybjed (approved):
Sounds good to me, let's go with `systemd-timesyncd` for new installations. But I think this is a big enough change that I don't plan to backport this to the stable releases.

In the future I want to split the `ntp` role into different separate NTP server roles, with `timesyncd` being in the `common` playbook and backing off when a different NTP server is present, similar to how the `nullmailer` role works. But that's for another time.

PR #2213, opened 2022-10-12
Add log-driver to docker_server

julien-lecomte:
Quick and dirty addition so that "none" can be specified as the log-driver in an easy way, like for the storage option.


drybjed (approved):
Looks good, thanks. The Docker configuration needs to be converted to Universal Configuration at some point, then we won't have to add separate options like this anymore. I'll backport it to the stable releases as well.

PR #2214, opened 2022-10-13
[misc] Fixup latest ansible-lint gripes

Alphix:


Alphix:
If I understand the output from the failing check correctly, the `Continuous Integration / ansible-lint` check needs to install the `python3-netaddr` package in the test environment....so this isn't actually an error in the PR...?

drybjed (approved):
Excellent work as always. Thanks for the fixes. :-)

PR #2217, opened 2022-10-20
Add ansible.netcommon to requirements

goetzk:
I've just had an incident where hosts file templating failed because ansible.netcommon wasn't available in my environment. This dependency appears to be missing from requirements.yml so those installing direct from git will be left without.
As in #2067, 'ansible.utils.ipaddr' appears to be recommended now but I'm unsure if there are compatibility reasons for remaining with netcommon which would preclude changing dependency.

### DIFF
--- a/requirements.yml
+++ b/requirements.yml
@@ -8,6 +8,7 @@
 # detect it correctly.
 
 collections:
+  - 'ansible.netcommon'
   - 'ansible.posix'
   - 'community.crypto'
   - 'community.general'


drybjed (commented):
Looking at the [list of `ansible.netcommon` plugins](https://docs.ansible.com/ansible/latest/collections/ansible/netcommon/index.html) included in the collection I don't see anything that would show up in DebOps - these are focused on network devices (switches, routers) rather than servers. Because of that I'm not sure if this collection is really a DebOps dependency.

Can you be more specific about the issue you had in the inventory? Perhaps you mixed together servers and network devices in one Ansible inventory and the issue was outside of DebOps roles and playbooks? This could help make the issue clearer.

goetzk:
I think this was just superceded by https://github.com/debops/debops/issues/2271

PR #2219, opened 2022-10-20
[ifupdown] Fix mismerge from cherry-pick

gaudenz:
When cherry-picking 8d23f1e7143732ae8d3ae873113feb22e47a8ec7 with 437cc2ee76a5bce192ecbb53515bbbc15097dde3 one line in iface.j2 was removed which should not have been removed.


drybjed (approved):
Thanks, and sorry for my mistake. I'll try to be careful while backporting next time.

PR #2220, opened 2022-10-25
[ipxe] Fix yaml syntax.

htgoebel:
This was introduced in 5c355450ff0fdc455b55b8167eced59622c42776.


drybjed (approved):
Ooops! My bad, thanks for noticing. I'll make sure that this is fixed in older backports as well.

PR #2222, opened 2022-10-26
role docker_server: fix typo in task name saving local facts

pescobar:


drybjed (approved):
Thanks for the name fix. :-)

PR #2225, opened 2022-11-02
[Netplan] [WIP] New role to configure network using netplan

pescobar:
Still a PoC but it works to deploy a netplan configuration to my ubuntu22.04 test environment.

Posible improvements:

* Not sure if the handlers should be moved to role `global_handlers`
* Doesn't support `netplan__configuration`, `netplan__group_configuration` and `netplan__host_configuration` (not sure if needed though)
* Doesn't provide local facts
* Doesn't integrate with other roles like ferm/sysctl/kmod


ansible/roles/netplan/COPYRIGHT
```diff
@@ -0,0 +1,20 @@
+debops.ifupdown - Manage network interfaces using Ansible
+
+Copyright (C) 2015-2017 Maciej Delmanowski <drybjed@gmail.com>
+Copyright (C) 2015-2017 Robin Schneider <ypid@riseup.net>
+Copyright (C) 2015-2017 DebOps <https://debops.org/>
+SPDX-License-Identifier: GPL-3.0-only
```
drybjed, line 5: Don't forget to update the Copyright headers in all files. :-)

ansible/roles/netplan/defaults/main.yml
```diff
@@ -0,0 +1,115 @@
...
+
+
+# APT packages [[[
+# ----------------
+
+# .. envvar:: netplan__base_packages [[[
+#
+# List of base APT packages to install.
+netplan__base_packages:
+  - [ 'nplan', 'netplan.io' ]
```
drybjed: It seems that the `nplan` package is virtual and only exists on Ubuntu. I'm not sure if it's needed. If you make sure that the role works on Debian, you get extra points. :-)

ansible/roles/netplan/defaults/main.yml
```diff
@@ -0,0 +1,115 @@
...
+
+                                                                   # ]]]
+# .. envvar:: netplan__renderer [[[
+#
+# Choices are:
+#
+# * ``networkd``
+# * ``NetworkManager``
+#
+netplan__renderer: networkd
```
drybjed: The value here should be quoted since it's a string - just a convention in DebOps.

ansible/roles/netplan/defaults/main.yml
```diff
@@ -0,0 +1,115 @@
...
+#         nameservers:
+#           addresses: [192.168.2.1]
+#         routes:
+#           - to: default
+#             via: 192.168.2.1
+#       eth1:
+#         addresses:
+#           - 192.168.34.3/24
+
+netplan__configuration: {}
```
drybjed, line 102: I would urge you to get familiar with DebOps [Universal Configuration](https://docs.debops.org/en/stable-3.0/user-guide/universal-configuration.html) and use it in this role. Since `netplan` uses YAML for its configuration, a good starting point would be to check out other DebOps roles that deploy YAML configuration. For example the `lxd` role has "preseed" configuration done in YAML using Universal Configuration.

A default configuration defined in the role can be useful - it's usually the default Ethernet interface with DHCP enabled.

ansible/roles/netplan/handlers/main.yml
```diff
@@ -0,0 +1,12 @@
...
+# Copyright (C) 2015-2017 Robin Schneider <ypid@riseup.net>
+# Copyright (C) 2015-2017 DebOps <https://debops.org/>
+# SPDX-License-Identifier: GPL-3.0-only
+
+- name: Check netplan config and apply
+  ansible.builtin.command: /usr/sbin/netplan generate
+  notify: [ 'Apply netplan config' ]
+
+- name: Apply netplan config
+  ansible.builtin.command: /usr/sbin/netplan apply
```
drybjed: The handlers could be moved to the `global_handlers` role, that way other roles could use them in the future without requiring the `netplan` role as a dependency. Make sure that your role imports the `global_handlers` role for this to work.

ansible/roles/netplan/tasks/main.yml
```diff
@@ -0,0 +1,83 @@
...
+    purge: True
+
+- name: Purge existing netplan configs not deployed by this role
+  block:
+
+    - name: Locate existing netplan configurations
+      ansible.builtin.find:
+        paths: /etc/netplan
+        patterns: "*.yml,*.yaml"
+      register: _netplan_existing_configs
```
drybjed: In DebOps, the convention for runtime variables is `<role_name>__register_<variable>` so in this case it would be:
```yaml
register: netplan__register_existing_configs
```

drybjed (changes requested):
Interesting first attempt, I'm looking forward to the improvements. :-)

ansible/roles/netplan/defaults/main.yml
pescobar: I have updated the list of packages to only include `netplan.io`

ansible/roles/netplan/handlers/main.yml
pescobar: done

ansible/roles/netplan/tasks/main.yml
pescobar: done

ansible/roles/netplan/COPYRIGHT
pescobar: Not sure what would be the right copyright header to use. Should I only add my name to the existing list?

ansible/roles/netplan/defaults/main.yml
pescobar: Personally, in sensitive roles like this one which reconfigure the network and can leave the machine inaccesible if a wrong config is applied I prefer to avoid providing a default configuration to make sure that the applied configuration has been defined by the user.

That's also why I apply this conditional before applying the netplan network config

```
  when:
    - netplan__configuration.keys() | length > 0
```

With the current implementation any netplan configuration can be applied using the inventory without modifying the role itself. 

If using universal configuration is mandatory I will look at it

ansible/roles/netplan/COPYRIGHT
drybjed: Add your name, drop me and Robin, leave DebOps. Make sure that you're fine with the license as well.

ansible/roles/netplan/tasks/main.yml
drybjed: Remove the `_` prefix too. I know that it's a convention in Python to use them for private variables, but it's not needed in Ansible.

ansible/roles/netplan/defaults/main.yml
drybjed: I wouldn't call it "mandatory" but encouraged. :-) The problem it can solve here is that the configuration template is strict with bridge, bond, vlan entries. If `netplan` extends this configuration, any new features will have to be explicitly added to the template. With Universal Configuration implementation, you can add any new features in the inventory and the role doesn't need to be updated.

For a first iteration of the role, going with a YAML dictionary is fine though, no complaints. If `netplan` can be installed and doesn't complain about default configuration being present, going with no default seems OK with me as well. But in this case, I expect extensive documentation with multiple examples. :-)

PR #2226, opened 2022-11-03
[python] Allow overriding /etc/pip.conf template

pescobar:
In our environment we need a custom /etc/pip.conf 

This change would allow us to use a custom template without modifying the default DebOps roles


drybjed (approved):
Sounds good. I'm not fond of the lookup myself and perhaps in the future it can be replaced with a proper configuration using Universal Configuration, but for now it should suffice.

PR #2231, opened 2022-11-09
Support PostgreSQL versions 14 and 15

mbehrle:


drybjed (approved):
Nice work, thanks! I'll backport this to the stable releases as well.

PR #2233, opened 2022-11-19
[proc_hidepid] Optimise /etc/ansible/facts.d/proc_hidepid.fact for hosts joined to active-directory domain

gajdusek:
Avoid retrieving all system groups which is very slow if the host is joined to a large AD domain, has winbindd running, and uses winbindd in /etc/nsswitch.conf. It can take several minutes depending on number of domain users. Like 4 minutes for 6k groups in AD.


drybjed (approved):
Makes sense and is much more Pythonic. Thanks! I'll backport this to the stable releases as well.

PR #2235, opened 2022-11-26
Typo fixing

julien-lecomte:
Just some partial typo fixing with controlled codespell commands.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -83,7 +83,7 @@ General
 :ref:`debops.kibana` role
 '''''''''''''''''''''''''
 
-- The role can now manage passwords and other confidental data stored in the
+- The role can now manage passwords and other confidential data stored in the
   Kibana keystore.
 
 :ref:`debops.netbox` role
@@ -96,7 +96,7 @@ General
 :ref:`debops.resources` role
 ''''''''''''''''''''''''''''
 
-- The :ref:`debops.resources` role can now be used to repace a line via the
+- The :ref:`debops.resources` role can now be used to replace a line via the
   ``ansible.builtin.replace`` module.
 
 :ref:`debops.slapd` role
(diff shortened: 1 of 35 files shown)


LICENSES/CC-BY-3.0.txt
```diff
@@ -10,7 +10,7 @@ BY EXERCISING ANY RIGHTS TO THE WORK PROVIDED HERE, YOU ACCEPT AND AGREE TO BE B
 
 1. Definitions
 
-     a. "Adaptation" means a work based upon the Work, or upon the Work and other pre-existing works, such as a translation, adaptation, derivative work, arrangement of music or other alterations of a literary or artistic work, or phonogram or performance and includes cinematographic adaptations or any other form in which the Work may be recast, transformed, or adapted including in any form recognizably derived from the original, except that a work that constitutes a Collection will not be considered an Adaptation for the purpose of this License. For the avoidance of doubt, where the Work is a musical work, performance or phonogram, the synchronization of the Work in timed-relation with a moving image ("synching") will be considered an Adaptation for the purpose of this License.
```
ypid: `LICENSES/*` should probably be excluded ;-) This is managed by the `reuse` command.

ypid (changes requested):

LICENSES/CC-BY-3.0.txt
julien-lecomte: I hesitated. I'll remove this on next push.

drybjed (approved):
Thanks! It seems that most of these are documentation fixes, so I'll backport the PR to the stable releases and fix any merge issues.

I wonder if codespell should be added to the CI test suite... Is it feasible, or are there too many false positives without a way to mark them properly?

julien-lecomte:
It could check but it's might be hard to auto-fix. There's a couple places with errors in licenses. I recall also one or two variables which I prefer not to touch.

PR #2236, opened 2022-11-26
Enable resources__files like in mount

julien-lecomte:
Like discussed in chat, this PR enables the user to create a file before mount actually tries to mount anything.

Sample usage:
```
      mount__enabled: True
      mount__group_devices:
        - name: '/share'
          src:  '//nas-backup/foobar'
          fstype: 'cifs'
          opts: [ "rw", "_netdev", "x-systemd.automount", "credentials=/etc/cifs.credentials" ]

      mount__group_files:
        - dest: "/etc/cifs.credentials"
          mode: "0600"
          content: |
            username=foobar
            password=secret1234
```

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -50,6 +50,15 @@ New DebOps roles
 General
 '''''''
 
+:ref:`debops.mount` role
+'''''''''''''''''''''''''''''
+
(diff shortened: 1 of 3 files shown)


drybjed (approved):
Looks good, thanks. I'll clean up the documentation a bit and make bacports to stable releases.

PR #2238, opened 2022-11-30
Fix ASCII password lookup character sets

gaudenz:
Multiple invocations of the password lookup used the term `ascii` instead of `ascii_letters` in the `chars` parameter. Because of this only the letters a,c,i,s were added to the character set instead of all lowercase and uppercase ASCII letters.

As all password lookups did at least also include all digits and the minimum password length used is 20 characters, this did not result in weak passwords.


julien-lecomte:
FYI, similar to https://github.com/debops/debops/pull/1821

drybjed (approved):
Thanks! This will definitely be backported to the stable releases. :-)

PR #2248, opened 2022-12-19
feat(owncloud): Add support for Nextcloud 24 and 25, drop 22; Drop ownCloud full auto upgrade support

ypid:


drybjed (approved):
Sounds good, but I won't be backporting these changes to stable releases. I need to check what's the current upgrade procedure in Nextcloud, I hope they have some documentation about that somewhere...

ypid:
Thanks for merging and fixing the docs build!

See https://docs.debops.org/en/latest/ansible/roles/owncloud/getting-started.html#upgrade

Some background info: There is https://github.com/nextcloud/server/wiki/Maintenance-and-Release-Schedule and https://docs.nextcloud.com/server/latest/admin_manual/maintenance/upgrade.html. The Ansible role is intended to be used with https://docs.nextcloud.com/server/latest/admin_manual/maintenance/update.html but it also assists the admin with https://docs.nextcloud.com/server/latest/admin_manual/maintenance/manual_upgrade.html. I sometimes have to manually upgrade in case the build-in upgrade does not work. The role can download, verify and unpack for you when you move the `nextcloud` dir away for a minute and then run the role. Auto setup will fail, then just move `data` and `config` from old `nextcloud` dir to new on and run 

I extended the role docs a bit in https://github.com/debops/debops/pull/2208.

PR #2249, opened 2022-12-19
Remove warn argument from command and shell module

gaudenz:
The `warn` argument to the command and shell modules was deprecated in Ansible 2.11 and removed in 2.14. This argument is no longer needed as no warnings are issued anymore if these modules are used in ways which could be replaced by an Ansible module. For cases where ansible-lint reports an error without the argument, appropriate `noqa` comments are added.

See https://github.com/ansible/ansible/pull/70504.

### DIFF
--- a/ansible/roles/apparmor/tasks/main.yml
+++ b/ansible/roles/apparmor/tasks/main.yml
@@ -150,9 +150,7 @@
   when: apparmor__fact_apparmor_enabled_and_loaded | bool and apparmor__register_tunables is changed
 
 - name: Unload AppArmor profiles
-  ansible.builtin.command: service apparmor teardown
-  args:
-    warn: False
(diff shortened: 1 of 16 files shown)


drybjed (approved):
Thanks for the fixes! I'll use this PR instead of the other one since this one has `ansible-lint` warnings covered up. Fixes will be backported to the stable release as well.

morremeyer:
@drybjed Thanks for taking care of it! What do we need to do to get these changes released? The last release has been a while ago.

drybjed:
Sorry for the delay. I'll try to create a new release in the coming days, after merging a few more PRs.

morremeyer:
Nothing to be sorry for, you're the one doing the work here!

Thank you for the update, looking forward to it!

PR #2251, opened 2023-01-01
feature(owncloud): Support to host on a subpath for security reasons (Nextcloud/ownCloud)

ypid:
Implements: #1233


drybjed (approved):
Merging as promised. :-) Although I think that this is a substantial change, so I won't backport this to the current stable releases.

ypid:
Thanks :) I would also not have backported this.

PR #2254, opened 2023-01-05
feat(netbox): Upgrade from NetBox v3.3.8 to v3.4.2

ypid:
No other changes are needed. I checked all the release notes and even gone though this in the NetBox git repo:

```Shell
git wd 'v3.3.8..v3.4.2' -- upgrade.sh netbox/netbox/configuration.example.py docs
```

to stay up-to-date.

I first thought the new search feature would use something like Elasticsearch but that was only my own bias. It uses a single database table and works great from my bit of testing.


drybjed (approved):
Nice, thanks!

PR #2257, opened 2023-01-08
fix(owncloud): Use OPcache values from Nextcloud docker image

ypid:
Admin overview has warned me on one instance that a buffer of 8 was filled up. Increase to 16 which is used by the Nextcloud docker image as well. This is only a background change/minor performance improvement. Nothing work mentioning in the changelog.


drybjed (approved):
Sounds good. Thanks for the quick update.

PR #2258, opened 2023-01-12
Fix reuse issue 'Duplicate field "Copyright" in paragraph'...

julien-lecomte:


drybjed (approved):
Thanks for the quick fix, will be backported to the stable releases as well.

PR #2259, opened 2023-01-12
Remove ansible-lint warning

julien-lecomte:


drybjed (approved):
Thanks! Since this is just the `ansible-lint` tag rename, I'll check if this can be backported to the stable release as well.

PR #2260, opened 2023-01-12
[dovecot] LDAP fixes + a bunch of new features

Alphix:


Alphix:
The first three patches are important (IMHO), the remaining patches are rather feature/performance related. I could split the PR in two, if that's preferable.

@imrejonk : I think you'll be interested in this PR....care to have a look?

Alphix:
Superseded by #2677

PR #2263, opened 2023-01-15
Add precommit and add codespell

julien-lecomte:
The following PR adds pre-commit and some actions with the aim of normalizing and upgrading debops.

I've done several commits in order to be able to remove any unwanted action of this PR.

Current actions done and committed: ensure that docstring comes before code, executables with shebang are executable, and the opposite for files without shebang, ensure toml is correct (license information moved to .reuse/dep5), ensure no extra EOL at EOF (a common practice), remove emacs/python "-*- coding: utf-8 -*-", prettify all json, use consistent EOLs.

Commented in precommit are 4 actions that would rewrite python: upgrade code to py36, and automatic linting


julien-lecomte:
Of note, a github action could be added, but I only know gitlab CI.

LICENSES/Apache-2.0.txt
```diff
@@ -7,17 +7,17 @@ AND DISTRIBUTION
 
    1. Definitions.
 
-      
+
```
ypid: Modifying LICENSE files does not help us. We are only the downstream of those files (see the `reuse` tool). If you want to change this, you would need to change it at the upstream/source.

src/debops/_data/defaults.toml
```diff
@@ -1,7 +1,3 @@
-{# Copyright (C) 2020-2021 Maciej Delmanowski <drybjed@gmail.com>
```
ypid, line 1: Yet another reason not to have another file format (TOML) when we already widely use YAML. Anyway, I suggest you use something like this:

```
#{# Copyright (C) 2020-2021 Maciej Delmanowski <drybjed@gmail.com>
```

instead because the reuse.software spec recommends to include the copyright in the file when possible.

.pre-commit-config.yaml
```diff
@@ -11,3 +11,5 @@ repos:
       - id: check-docstring-first
       - id: check-executables-have-shebangs
       - id: check-merge-conflict
+      - id: check-shebang-scripts-are-executable
```
ypid, line 18: I argue against `check-shebang-scripts-are-executable` or at least exclude `*.j2` files. They are mostly not valid scripts before being templated.

.pre-commit-config.yaml
```diff
@@ -0,0 +1,53 @@
...
+      - id: check-symlinks
+      - id: check-toml
+      # - id: check-yaml
+      - id: destroyed-symlinks
+      - id: debug-statements
+      - id: end-of-file-fixer
+      - id: fix-byte-order-marker
+      - id: fix-encoding-pragma
+        args:
+          - "--remove"
```
ypid, line 29: Can you add a reference/PEP that states that omitting the `coding` line is now best practice?

ypid (changes requested):
I appreciate the introduction of pre-commit and did a quick review of your PR.

.pre-commit-config.yaml
julien-lecomte: There's a nice SO answer here: https://stackoverflow.com/questions/14083111/should-i-use-encoding-declaration-in-python-3 .

I believe that the whole repository is UTF-8, and all our tools are UTF-8 by default. In the case of vim, the "-*- coding: latin-1 -*-" comment itself has no effect as it's not a correct vim line.

LICENSES/Apache-2.0.txt
julien-lecomte: I think we can add exclusions to the pre-commit configuration file, but is it worth it? This is only whitespace.

.pre-commit-config.yaml
julien-lecomte: I've added j2 exclusion.

LICENSES/Apache-2.0.txt
julien-lecomte: I've excluded LICENSES/.*

julien-lecomte:
A github action was added.

LICENSES/Apache-2.0.txt
ypid: reuse.software specifies that the LICENSE file should not be changed at all. It might even be checksums (not sure about that). Any changes make it more difficult to see if we changed the license or not. Also, as I said, making changes downstream is not the solution.

.pre-commit-config.yaml
ypid: Ok. I have no strong opinions about this so I am fine with that.

ypid (changes requested):
What has changed the `*.svg` files? Otherwise I have no further change requests.

julien-lecomte:
@ypid The svg were changed by the "fix end of files". They ended abruptly with ">" (3E) and now with ">\n" (3E 0A)

ypid:
I did a short research and found https://github.com/mermaid-js/mermaid/issues/400. GitHub also shows a warning if newline is missing so that is an argument to do it as you proposed. So fine for me.

ypid (dismissed):
I reviewed the pre-commit config and only peeked at the other changes. Looks good to me.

julien-lecomte:
Add codespell via precommit while retaining the possibility to run it directly from root of repo without pre-commit.

The github precommit action will automatically run codespell and bail out on error.

Files are ignored either by name, or by file contents (.codespell.exclude).

The keyserver word is whitelisted as a keyword, and referer is ignored to avoid renaming variables and mismatching the "official" http header typo.

drybjed:
Lots of files to check and it's after midnight, so I'll do the review tomorrow. I just wonder - is there some documentation about how to setup `pre-commit` locally? Hopefully in development docs.

ansible/roles/netbox/defaults/main.yml
```diff
@@ -1128,7 +1128,7 @@ netbox__nginx__dependent_servers:
         proxy_connect_timeout {{ netbox__app_timeout }};
         proxy_send_timeout {{ netbox__app_timeout }};
         proxy_read_timeout {{ netbox__app_timeout }};
-        add_header P3P 'CP="ALL DSP COR PSAa PSDa OUR NOR ONL UNI COM NAV"';
+        add_header P3P 'CP="ALL DSP COR PSAa PSDa OUR NOR ONLY UNI COM NAV"';
```
drybjed, line 1131: I'm not sure about this change. I did a brief search for Compact Policies and found [this website](http://etutorials.org/Mobile+devices/mobile+location+services/Appendix+D.+Platform+for+Privacy+Preferences+P3P/4+Compact+Policies/) where `ONL` is used as a category:
```
"PHY" | ; for <physical/>
"ONL" | ; for <online/>
"UNI" | ; for <uniqueID/>
"PUR" | ; for <purchase/>
"FIN" | ; for <financial/>
"COM" | ; for <computer/>
"NAV" | ; for <navigation/>
"INT" | ; for <interactive/>
"DEM" | ; for <demographic/>
"CNT" | ; for <content/>
"STA" | ; for <state/>
"POL" | ; for <political/>
"HEA" | ; for <health/>
"PRE" | ; for <preference/>
"LOC" | ; for <location/>
"GOV" | ; for <government/>
"OTC"   ; for <other-category/>
```
Is there a way to mark this as a false positive?

julien-lecomte: False positives (lines to ignore) can be added to the ./.codespell.exclude file at the root of the project.

julien-lecomte:
> Lots of files to check and it's after midnight, so I'll do the review tomorrow. I just wonder - is there some documentation about how to setup `pre-commit` locally? Hopefully in development docs.

I run precommit locally via `precommit run --all`.

ansible/roles/netbox/defaults/main.yml
drybjed: Thanks, I'll check it out before merging the PR.

ansible/roles/postgresql/defaults/main.yml
```diff
@@ -162,7 +162,7 @@ postgresql__user_clusters: []
 
 # .. envvar:: postgresql__roles [[[
 #
-# List of PostgresSQL roles, specified as YAML dicts. See
+# List of postgresql roles, specified as YAML dicts. See
```
drybjed, line 165: The `PostgreSQL` spelling should be added to the dictionary. :-) And yeah, I see the additional `s` there now, I wonder if `codespell` will trigger when that's fixed. Just adding an earmark to check this later.

ansible/roles/sshd/templates/etc/ssh/ldap_authorized_keys_bindpw.j2
```diff
@@ -2,4 +2,4 @@
  # Copyright (C) 2022 DebOps <https://debops.org/>
  # SPDX-License-Identifier: GPL-3.0-only
  #}
-{{- sshd__ldap_bindpw -}}
\ No newline at end of file
+{{- sshd__ldap_bindpw -}}
```
drybjed, line 5: New line in password file might be significant. I hope that the template will be correct here, might need to be tested in an LDAP environment to be sure.

src/debops/_data/defaults.toml
drybjed: The template file is used only internally in the script to generate defaults in the runtime configuration structure. The Jinja `{# #}` are not needed here, a mistake on my part during development. I'll fix that in a commit before merging this PR.

src/debops/cli.py
```diff
@@ -1,4 +1,3 @@
-# -*- coding: utf-8 -*-
```
drybjed, line 1: *slaps himself in the head*

This leaves an empty line in this file, and all others with this fix, doesn't it... Is there a check for that in `codespell` which can be enabled?

drybjed (approved):
I reviewed all files. Great work on integrating `pre-commit` and `codespell`, Julien! Now I need to check how to use them locally on Debian... :-)

That's a lot of changes in the project, most of that is on the level of documentation. I'll check if backporting this to `stable-3.0` branch makes sense, to propagate fixes to this stable release at least.

src/debops/cli.py
julien-lecomte: Not that I know of.

This isn't codespell, it's another hook from the precommit configuration file.

PR #2265, opened 2023-01-16
[ntp] Do not purge any ntp daemon if ntp support is not enabled

prahal:
Allow leaving the current target ntpd daemon untouched if ntpd daemon support is not enabled.


drybjed (approved):
Good call. I'm not sure why I missed this while creating the role. I'll backport the change to the stable releases as well.

PR #2266, opened 2023-01-16
[apt] set new debian-security generic mirror url for newer Debian

prahal:
Currently this is buster, bullseye and bookworm. That is all Debian debops supported releases except stretch.

-----------------

Note that stretch and Jessie are also available in http://deb.debian.org/debian-security/ (ie http://deb.debian.org/debian-security/dists/) so it may be we could set this URL for the whole of Debian.

We could also let the code as is until the old URL is decommissioned. As http://security.debian.org/dists/ shows buster/updates, bullseye-security and bookworm-security are also available with this old URL.


drybjed (approved):
Sounds good, I'll backport this to the stable DebOps releases as well.

PR #2278, opened 2023-02-02
Add pip/virtual env to resources

julien-lecomte:


ypid:
The resources role is already quite bloated. How about we draw a line here. I recently found `pipx`. So my proposal, how about a role called `debops.pipx` using https://docs.ansible.com/ansible/latest/collections/community/general/pipx_module.html that should also be able to solve your use case in a bit easier and cleaner why? If not, please state your use case :)

julien-lecomte:
The use case is creating a virtualenv that will later require a systemd file and launch a service proxied by apache.

`pipx` wasn't on my radar, but it's fine as a alternative solution.

scibi:
Long time ago I made a role to deploy Python apps inside virtualenv. You can take a look: https://github.com/scibi/ansible-python_deploy

drybjed:
With my user hat on, I think that the sane way to deploy applications outside of APT is to create an unprivileged UNIX account for a given application and run deployment commands directly in it, not going through `root` via `sudo`. Unfortunately, Ansible cannot work with partial `sudo` access, it requires unrestricted command privileges to do its job. To work around it, my solution is to configure `ssh` access for the deployment UNIX account and run Ansible against it. The deployment account doesn't need full `root` privileges, just access to various resources like databases, directories in the filesystem, etc. This way, various Ansible commands work just fine without raising privileges via `become`. This allows for much simplier roles and playbooks focused on unprivileged operation.

I started to think how to make this easy in DebOps a few years ago. Going at it via Ansible inventory configuration is probably the best way - each application can have its own "view" of the infrastructure, which means its own separate Ansible inventory with `ansible_user` variable pointing to an unprivileged UNIX account. I have the general layout of the project directory planned out, just need to spend some time with the `debops` codebase to implement it. After that, unprivileged roles and playbooks can be added in DebOps in a separate Ansible Collection.

As for this PR, I think it's a good stop gap for the moment. I'll merge it and backport to the stable releases.

PR #2279, opened 2023-02-06
Refresh SSH service configuration, improve support for socket activation

drybjed:


ypid:
Can you document why the role now opens the ports 3333 and 4444 by default? I assume to give alternative ports in case 22 is blocked for some reason? You don’t need to answer, you can write it into the `defaults` file directly to safe keystrokes ;-)

drybjed:
Umm, what? Hold on, looks like these were left there by mistake during testing. Checking...

drybjed:
Fixed by https://github.com/debops/debops/commit/28bb0731777d6f8ab1af2c568c2a2b7a8288178c, thanks for the heads up @ypid. :-)

PR #2281, opened 2023-02-07
fix(sshd): use "init" for parameter to uncomment when set

ypid:
Did I miss something about https://docs.debops.org/en/master/user-guide/universal-configuration.html#universal-configuration? I know that you use `comment` in other roles as well.


drybjed (approved):
Looks good. And the `init` state works correctly in cases were complex syntax is used, uncommenting options as needed, so let's go with it instead.

ypid:
You are welcome. I am done with testing of the redesigned sshd role. Nice job! I finally start to appreciate universal configuration. I updated my inventory presets:

* https://github.com/ypid/ypid-ansible-inventory/blob/master/group_vars/public_ypid_defaults_debops/public_ypid_defaults_debops_sshd.yml
* https://github.com/ypid/ypid-ansible-inventory/blob/master/group_vars/public_ypid_paranoid_debops/public_ypid_paranoid_debops_sshd.yml

Accordingly.

julien-lecomte:
@ypid There's an extra space line 32 of file #2 : 'DisableForwarding '

ypid:
Thanks @drybjed and @julien-lecomte, fixed (https://github.com/ypid/ypid-ansible-inventory/issues/1). It actually worked because DisableForwarding is not defined in DebOps. I was a bit tired yesterday evening :)

PR #2284, opened 2023-02-14
Handle change in extrepo Enabled/Disabled sources file

julien-lecomte:
Extrepo facts file did not detect a disabled repository as being disabled
due to a change in the extrepo file format.


drybjed (approved):
Thanks, works as expected. I'll backport this to stable releases as well.

PR #2288, opened 2023-03-02
[sshd] Fix typos in original configuration

Alphix:
Both configuration options are not present, so the typos don't matter by default. Still good to avoid potential footguns though.


drybjed (approved):
Thank you. I have a new keyboard, so maybe I'm just writing too fast sometimes.

Alphix:
> Thank you. I have a new keyboard, so maybe I'm just writing too fast sometimes.

Better too fast than too slow?

PR #2206, opened 2022-10-06
Support provisioning a remote psql db for icinga web

Cadair:
This seemed to do the trick for me.

PR #1466, opened 2020-08-14
[icinga_web] feat remote database installation

lewer:
This PR allows to configure icinga_web database if it not installed on the same host as icinga_web application.

PR #2295, opened 2023-03-10
[icinga] Add remote DB functionality and SSL support

Alphix:
This PR builds on the two separate PRs submitted by @lewer back in 2020 (#1443 and #1466),
but they've been substantially refresh. adapted to the current coding standards and I've also taken
the opportunity to add SSL support (seems like a good idea for remote DB access).

In addition, the x509 module in `icinga_web` is now prepared to support PostgreSQL once a new
upstream version is released.

Closes: #1443, #1466, #2206

### DIFF
--- a/ansible/roles/icinga_db/defaults/main.yml
+++ b/ansible/roles/icinga_db/defaults/main.yml
@@ -55,6 +55,130 @@ icinga_db__feature: '{{ "ido-pgsql"
                         else ("ido-mysql"
                               if (icinga_db__type == "mariadb")
                               else "unknown") }}'
+
+                                                                   # ]]]
+
+# .. envvar:: icinga_db__dbconfigs [[[
+#
+# Debconf questions and answers for the icinga2-ido-pgsql and icinga2-ido-mysql
+# packages, to install these packages without user interaction.
+icinga_db__dbconfigs:
+
+  - question: 'icinga2-{{ icinga_db__feature }}/dbconfig-install'
+    vtype: 'boolean'
+    value: 'false'
+
+  - question: 'icinga2-{{ icinga_db__feature }}/dbconfig-upgrade'
+    vtype: 'boolean'
+    value: '{{ icinga_db__host == "localhost" }}'
+
+  - question: 'icinga2-{{ icinga_db__feature }}/enable'
+    vtype: 'boolean'
+    value: 'true'
+
+  - question: 'icinga2-{{ icinga_db__feature }}/remote/host'
+    vtype: 'string'
+    value: '{{ icinga_db__host }}'
+
+  - question: 'icinga2-{{ icinga_db__feature }}/remote/port'
+    vtype: 'string'
+    value: '{{ icinga_db__port }}'
+
+  - question: 'icinga2-{{ icinga_db__feature }}/db/dbname'
+    vtype: 'string'
+    value: '{{ icinga_db__database }}'
+
+  - question: 'icinga2-{{ icinga_db__feature }}/db/app-user'
+    vtype: 'string'
+    value: '{{ icinga_db__user }}'
+
+  - question: 'icinga2-{{ icinga_db__feature }}/{{ "pgsql"
+                                                   if icinga_db__type == "postgresql"
+                                                   else ("mysql" if icinga_db__type == "mariadb"
+                                                                 else "")}}/app-pass'
+    vtype: 'password'
+    value: '{{ icinga_db__password }}'
+
+                                                                   # ]]]
+                                                                   # ]]]
+
+# Database configuration [[[
+# --------------------------
+# .. envvar:: icinga_db__host [[[
+#
+# The host where the Icinga 2 database will be installed.
+icinga_db__host: '{{ ansible_local.postgresql.server
+                     if (ansible_local.postgresql.server|d() and
+                         icinga_db__type == "postgresql")
+                     else (ansible_local.mariadb.server
+                           if (ansible_local.mariadb.server|d() and
+                               icinga_db__type == "mariadb")
+                           else "localhost") }}'
+
+                                                                   # ]]]
+# .. envvar:: icinga_db__port [[[
+#
+# The port used to connect to the Icinga 2 database.
+icinga_db__port: '{{ ansible_local.postgresql.port
+                     if (ansible_local.postgresql.port|d() and
+                         icinga_db__type == "postgresql")
+                     else (ansible_local.mariadb.port
+                           if (ansible_local.mariadb.server|d() and
+                               icinga_db__type == "mariadb")
+                           else "unknown") }}'
+
+                                                                   # ]]]
+# .. envvar:: icinga_db__user [[[
+#
+# Name of the database user for Icinga 2.
+icinga_db__user: 'icinga2'
+
+                                                                   # ]]]
+# .. envvar:: icinga_db__database [[[
+#
+# Name of the database for Icinga 2.
+icinga_db__database: 'icinga2'
+
+                                                                   # ]]]
+# .. envvar:: icinga_db__password_path [[[
+#
+# Path to database password file located on the Ansible Controller. See the
+# :ref:`debops.secret` role for more details.
+icinga_db__password_path: '{{ secret + "/" + icinga_db__type + "/" +
+                                        ansible_local[icinga_db__type].delegate_to }}{%
+                                         if icinga_db__type == "postgresql" %}/{{ ansible_local[icinga_db__type].port }}{% endif
+                                        %}{{ "/credentials/" + icinga_db__user +
+                                        "/password" }}'
+
+                                                                   # ]]]
+# .. envvar:: icinga_db__password [[[
+#
+# Password of the Icinga 2 database user.
+icinga_db__password: "{{ lookup('password', icinga_db__password_path
+                                   + ' length=48 chars=ascii_letters,digits,.-_') }}"
+
+                                                                   # ]]]
+# .. envvar:: icinga_db__init [[[
+#
+# Should the Icinga 2 database be initialized ? Per default, it should be
+# initialized only the first time this role runs.
+icinga_db__init: '{{ False if ansible_local.icinga_db.configured|d()
+                           else True }}'
+
+                                                                   # ]]]
+# .. envvar:: icinga_db__postgresql_schema [[[
+#
+# Path on the host where this role runs of the PostgreSQL schema. This schema
+# is used to populate the database when it is created.
+icinga_db__postgresql_schema: '/usr/share/icinga2-ido-pgsql/schema/pgsql.sql'
+
+                                                                   # ]]]
+# .. envvar:: icinga_db__mariadb_schema [[[
+#
+# Path on the host where this role runs of the MariaDB schema. This schema
+# is used to populate the database when it is created.
+icinga_db__mariadb_schema: '/usr/share/icinga2-ido-mysql/schema/mysql.sql'
+
                                                                    # ]]]
                                                                    # ]]]
 # APT packages [[[
@@ -79,3 +203,40 @@ icinga_db__base_packages:
 icinga_db__packages: []
                                                                    # ]]]
                                                                    # ]]]
+
+# Configuration for other Ansible roles [[[
+# -----------------------------------------
+
+# .. envvar:: icinga_db__postgresql__dependent_roles [[[
+#
+# Configuration of PostgreSQL roles for :ref:`debops.postgresql` Ansible role.
+icinga_db__postgresql__dependent_roles:
+  - name: '{{ icinga_db__user }}'
+    password: '{{ icinga_db__password }}'
+    db: '{{ icinga_db__database }}'
+    priv: ['ALL']
+
+                                                                   # ]]]
+# .. envvar:: icinga_db__postgresql__dependent_databases [[[
+#
+# Configuration of PostgreSQL databases for the :ref:`debops.postgresql` Ansible
+# role.
+icinga_db__postgresql__dependent_databases:
+  - name: '{{ icinga_db__database }}'
+    owner: '{{ icinga_db__user }}'
+
+                                                                   # ]]]
+# .. envvar:: icinga_db__mariadb__dependent_users [[[
+#
+# Configuration of MariaDB users for :ref:`debops.mariadb` Ansible role.
+icinga_db__mariadb__dependent_users:
+  - name: '{{ icinga_db__user }}'
+    password: '{{ icinga_db__password }}'
+    database: '{{ icinga_db__database }}'
+
+                                                                   # ]]]
+# .. envvar:: icinga_db__mariadb__dependent_databases [[[
+#
+# Configuration of MariaDB databases for :ref:`debops.mariadb` Ansible role.
+icinga_db__mariadb__dependent_databases:
+  - name: '{{ icinga_db__database }}'
--- a/ansible/roles/icinga_db/defaults/main.yml
+++ b/ansible/roles/icinga_db/defaults/main.yml
@@ -2,7 +2,9 @@
 # .. vim: foldmarker=[[[,]]]:foldmethod=marker
 
 # .. Copyright (C) 2018 Maciej Delmanowski <drybjed@gmail.com>
-# .. Copyright (C) 2018 DebOps <https://debops.org/>
+# .. Copyright (C) 2020 Gabriel Lewertowski <gabriel.lewertowski@trust-in-soft.com>
+# .. Copyright (C) 2023 David Härdeman <david@hardeman.nu>
+# .. Copyright (C) 2018-2023 DebOps <https://debops.org/>
 # .. SPDX-License-Identifier: GPL-3.0-only
 
 # .. _icinga_db__ref_defaults:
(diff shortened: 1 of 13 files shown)


Alphix:
Also, I should add that I've tested this using a bunch of VMs with:

- local pgsql
- remote pgsql
- local mariadb
- local mariadb + x509
- remote mariadb
- remote mariadb + x509

ansible/roles/icinga_db/templates/etc/ansible/facts.d/icinga_db.fact.j2
```diff
@@ -12,56 +12,24 @@ from json import loads, dumps
...
 output = loads('''{{ {"configured": False,
-                      "type": icinga_db__type}
+                      "type": icinga_db__type,
+                      "ido": icinga_db__database_map[icinga_db__type].ido,
+                      "db_name": icinga_db__database,
+                      "db_host": icinga_db__host,
+                      "db_port": icinga_db__port,
+                      "db_ssl": icinga_db__ssl,
+                      "db_user": icinga_db__user,
+                      "db_password": icinga_db__password}
```
drybjed: You included the database password directly in the fact script. I was able to run the `icinga_db.fact` from na unprivileged UNIX account on the same host, get the password and login to the Icinga database.

This shouldn't happen.

Existing setup relied on Debian's `dbconfig` infrastructure to setup the Icinga database. Apparently, `dbconfig-common` APT package should allow for this to work with remote databases as well. Perhaps DebOps needs to have a database configuration role to make distributed setup more manageable.

As for the current `icinga_db` setup - I would ussgest reading the connection details from the Icinga configuration instead of storing them directly in the fact script. This will be more accurate when they change, so there won't need to be a need to synchronize a changed password in icinga config and the fact script separately. This will also ensure that UNIX accounts which don't have access to the Icinga configuration files, cannot get the password through the fact script.

ansible/roles/icinga_db/templates/etc/ansible/facts.d/icinga_db.fact.j2
```diff
@@ -12,56 +12,24 @@ from json import loads, dumps
...
                      | to_nice_json }}''')
 
 db_config_files = [
-    '/etc/dbconfig-common/icinga2-ido-pgsql.conf',
-    '/etc/dbconfig-common/icinga2-ido-mysql.conf'
+    '/etc/icinga2/features-available/ido-{{ icinga_db__database_map[icinga_db__type].ido }}.conf'
```
drybjed: `pycodestyle` reports that line is too long here. You can write this as:
```python
db_config_files = [
    ('/etc/icinga2/features-available/'
     'ido-{{ icinga_db__database_map[icinga_db__type].ido }}.conf')
]
```
It will look weird in the generated script, but it will get the job done.

ansible/roles/icinga_db/defaults/main.yml
```diff
@@ -64,13 +230,7 @@ icinga_db__feature: '{{ "ido-pgsql"
...
 # The list of APT packages to install for Icinga 2 database support. The
 # database configuration will be performed by the ``dbconfig`` infrastructure.
-icinga_db__base_packages:
-
-  - '{{ "icinga2-ido-pgsql"
-        if (icinga_db__type == "postgresql")
-        else ("icinga2-ido-mysql"
-              if (icinga_db__type == "mariadb")
-              else []) }}'
+icinga_db__base_packages: [ '{{ "icinga2-ido-" + icinga_db__ido }}' ]
```
drybjed, line 233: There is a problem with installation of the Icinga 2 IDO packages - `dbconfig` replaces database authentication details (namely, password) for the database with their own generated randomly, and the installation breaks on the initial role execution. The second run of the role updates the database password for the one generated by DebOps roles, and the installation works fine again.

I think that defining `debconf` questions before installing the APT package should handle this problem. You can either disable `dbconfig` assistance and handle the database by Ansible, or define the user/password in the `debconf` database, so that `dbconfig` gets the correct values defined by DebOps. That goes both for PostgreSQL and MariaDB databases, of course.

drybjed (changes requested):
I tried out the changes on a single host, and after a few issues with `dbconfig` changing the password, I managed to spin up Icinga and get to its web interface.  There are still a few issues I see with this PR, check the comments on specific files.

The PostgreSQL changes are good, and I can merge them in a separate commit, which should modify just the `postgresql` role.

ansible/roles/icinga_db/defaults/main.yml
Alphix: > There is a problem with installation of the Icinga 2 IDO packages - `dbconfig` replaces database authentication details (namely, password) for the database with their own generated randomly, and the installation breaks on the initial role execution. The second run of the role updates the database password for the one generated by DebOps roles, and the installation works fine again.

Hm, really? I'm not seeing this. The role (with the patches from this PR) first installs the packages, then diverts the DB configuration files generated by `dbconfig` in the package installation scripts, and then installs the new (templated) configuration.

The order I used in my testing was roughly like this (assuming `qtest1` runs both the DB and icinga).

Either:
```
debops run service/postgresql_server -l qtest1
debops run service/postgresql
```

Or:
```
debops run service/mariadb_server -l qtest1
debops run service/mariadb
```

Then:
```
debops run service/icinga -l qtest1
debops run service/icinga_db -l qtest1
debops run service/icinga_web -l qtest1
# Run on all, including qtest1, so also qtest1 registers with the director
debops run service/icinga
```

And it's working fine for me. Which order did you do things when it broke on initial role execution?

ansible/roles/icinga_db/templates/etc/ansible/facts.d/icinga_db.fact.j2
Alphix: D'oh. I'll write a fact script that reads the connection details from the Icinga configuration files instead.

As for `dbconfig`, I think the current approach (diverting and ignoring the result of `dbconfig`) is the most `KISS`-compatible approach :)

ansible/roles/icinga_db/templates/etc/ansible/facts.d/icinga_db.fact.j2
Alphix: Ok, I've fixed that in the new facts script

ansible/roles/icinga_db/templates/etc/ansible/facts.d/icinga_db.fact.j2
Alphix: Ok, the new facts script will pull the connections details from the configuration files instead. That should fix this issue...

ansible/roles/icinga_db/defaults/main.yml
Alphix: I've written a new facts script, but this particular issue (the installation breaking on the initial role execution) I can't reproduce.

I just tore down all VMs and did a reinstall from scratch using the latest version of this PR, and I ended up with a working Icinga installation in the end.

drybjed: I'll test the changes and diversion ASAP. I'm doing this by creating a fresh LXC container and running commands:
```shell
debops run bootstrap -l <container> -u root
debops run common -l <container>
debops run service/postgresql_server -l <container>
debops run service/icinga service/icinga_db -l <container>
```
The specific steps of database authentication can be done like this, after setting up PostgreSQL server:
```shell
debops run service/icinga -l <container>
debops run service/icinga_db -l <container> -t role::postgresql
```
This should setup the database where authentication works based on the credentials stored in secrets, for example:
```shell
user@container:~$ psql -h localhost -U icinga2 -W icinga2
```
After that, when you apply the `service/icinga_db` role on the host, password authentication doesn't work until you run the playbook again, which re-runs the `service/postgresql` role in the icinga2 context, updates the database password again and sets up the database itself.

I tried this like 3 times with the same results, with PostgreSQL database on the same host as Icinga. I didn't test MariaDB support yet.

Alphix: I just tested with a local PostgreSQL installation (from scratch), and that also seemed to work....basically `bootstrap` on a couple of VMs and then:

```
debops run service/postgresql_server -l qtest1
debops run service/postgresql
debops run service/icinga -l qtest1
debops run service/icinga_db -l qtest1
debops run service/icinga_web -l qtest1
debops run service/icinga
```

Dunno if there's some interaction between `icinga_db` and `icinga_web`...I'll do another test run without `icinga_web` next...

Alphix: Still can't reproduce this...even after not running `icinga-web` at all:

```
debops run service/postgresql_server -l qtest1
debops run service/postgresql
debops run service/icinga -l qtest1
debops run service/icinga_db -l qtest1
ssh root@qtest1
root@qtest1:~# /etc/ansible/facts.d/icinga_db.fact 
{
    "configured": true,
    "db_host": "localhost",
    "db_name": "icinga2",
    "db_password": "blablabla",
    "db_port": "5432",
    "db_ssl": "disable",
    "db_ssl_ca": "/etc/pki/realms/domain/CA.crt",
    "db_user": "icinga2",
    "ido": "pgsql",
    "type": "postgresql"
}
root@qtest1:~# psql -h localhost -U icinga2 -W icinga2
Password: blablabla
psql (13.9 (Debian 13.9-0+deb11u1))
SSL connection (protocol: TLSv1.3, cipher: TLS_AES_256_GCM_SHA384, bits: 256, compression: off)
Type "help" for help.

icinga2=>
```

drybjed: That's fine, I'll try to find the exact issue on my side.

drybjed: I tried changes again, and still the installation process is broken. This is on Debian Bullseye. When I run the `service/icinga_db` role, the Ansible output is:
```
TASK [debops.debops.icinga_db : Assert that the DB type is valid] ********************************************************************************************************************************************************
ok: [icinga -> localhost] => changed=false 
  msg: All assertions passed
[DEPRECATION WARNING]: "listify_lookup_plugin_terms" does not use "dataloader" anymore, the ability to pass it in will be removed in future versions. This feature will be removed in version 2.18. Deprecation warnings 
can be disabled by setting deprecation_warnings=False in ansible.cfg.

TASK [debops.debops.icinga_db : Install required packages] ***************************************************************************************************************************************************************
changed: [icinga]

TASK [debops.debops.icinga_db : Create Icinga PostgreSQL tables] *********************************************************************************************************************************************************
fatal: [icinga]: FAILED! => changed=false 
  cmd: /usr/bin/psql --dbname=icinga2 --host=localhost --port=5432 --username=icinga2 --file=/usr/share/icinga2-ido-pgsql/schema/pgsql.sql < /usr/share/icinga2-ido-pgsql/schema/pgsql.sql
  msg: |-
    psql: error: FATAL:  password authentication failed for user "icinga2"
    FATAL:  password authentication failed for user "icinga2"
  rc: 2
  stdout: ''
  stdout_lines: <omitted>
```
The issue as I understand it is that you are diverting the Icinga2 configuration file that configures access to the database, but you are not dealing it with the `dbconfg` configuration file located in `/etc/dbconfig-common/icinga2-ido-pgsql.conf` which is generated during package installation (look into `/var/lib/dpkg/info/icinga2-ido-pgsql.postinst` script to see what's happening). This configuration file tells `dbconfig` how to configure the database in PostgreSQL for Icinga. There's no way to disable this during installation as far as I can see. We cannot divert this file because it's generated during package installation, based on the `debconf` database. The only way to deal with this is to use `debconf` to populate database credentials for Icinga2.

This is an unfortunate consequence of using Debian Icinga2 packages, which are integrated this way with the rest of the Debian ecosystem. If Icinga was installed from upstream, I'm sure that this could be simpler, but we're pulling it from Debian.

Well, there's one more way we can try. Apparently, installing `dbconfig-no-thanks` package should disable `dbconfig` from messing up the database entirely. I'm currently testing if this needs to be done as a separate APT install step, or both `icinga-ido-pgsql` and this package can be installed in one go. Still, `dbconfig` role needs to be created at some point to deal with these kinds of situations in the future, since this is a shared resource.

drybjed: Adding `dbconfig-no-thanks` to the list of APT packages to install in `icinga_db` role works like a charm. :-) I'll add it and merge the PR after checking CI tests locally, unless you do it first, @Alphix :-)

Alphix: > The issue as I understand it is that you are diverting the Icinga2 configuration file that configures access to the database, but you are not dealing it with the `dbconfg` configuration file located in `/etc/dbconfig-common/icinga2-ido-pgsql.conf` which is generated during package installation (look into `/var/lib/dpkg/info/icinga2-ido-pgsql.postinst` script to see what's happening). This configuration file tells `dbconfig` how to configure the database in PostgreSQL for Icinga. There's no way to disable this during installation as far as I can see. We cannot divert this file because it's generated during package installation, based on the `debconf` database. The only way to deal with this is to use `debconf` to populate database credentials for Icinga2.

Hmm, I don't think that's quite it...does your test LXR container have the `dbconfig-pgsql` package installed at the point where it fails?

drybjed: No, there was just `dbconfig-common` pulled as a dependency of `icinga2-dbo-pgsql`. If you have `dbconfig-pgsql` installed then it probably was added either via your Ansible inventory, or your container provisioner.

Alphix: > No, there was just `dbconfig-common` pulled as a dependency of `icinga2-dbo-pgsql`. If you have `dbconfig-pgsql` installed then it probably was added either via your Ansible inventory, or your container provisioner.

No, I also don't have it installed. I was just trying to guess why you see failures and I don't (without `dbconfig-no-thanks` installed).

As far as I can tell, `db-config` only generates a user/password but doesn't actually setup any database or user, which is why I'm not seeing any errors even without `dbconfig-no-thanks`.

For you, it seems that it does, which would explain the errors...

Basically:

1. `ansible/playbooks/service/icinga_db.yml` first calls the `debops.postgresql` role, which sets up a DB/user/pass and stores the password in `secret`
2. `ansible/roles/icinga_db/tasks/main.yml` installs the `icinga2-ido-pgsql` package
3. The package calls `/var/lib/dpkg/info/icinga2-ido-pgsql.postinst`, which calls `dbc_go` from `/usr/share/dbconfig-common/dpkg/postinst`
4. `ansible/roles/icinga_db/tasks/main.yml` creates the icinga tables in the postgresql db
5. `ansible/roles/icinga_db/tasks/main.yml` diverts `/etc/icinga2/features-available/ido-pgsql.conf` and generates its own version (this works because the file generated by `db-common` is marked as a config file)

So I think the difference must be that in step `2`, `db-common` actually alters the PostgreSQL configuration (i.e. changes the user/pass/etc) for you, while for me it only generates a password, but no actual changes on the DB server are performed.

That was why I guessed that `dbconfig-pgsql` presence/absence might explain the different experiences we had....but apparently not...

`dbconfig-no-thanks` feels pretty heavy-handed since it affects all packages using `db-config` (not such a big deal when the host is anyway Ansible-managed, but still).

I'll try to see if there's some other way to disable `dbconfig-common` on a per-package basis.

drybjed: @Cadair offered to test this PR on the IRC channel. Can you test it with and without `dbconfig-no-thanks` package to see if you can reproduce the issue with the database being inacessible during first installation? Let us know what you find.

Alphix: > @Cadair offered to test this PR on the IRC channel. Can you test it with and without `dbconfig-no-thanks` package to see if you can reproduce the issue with the database being inacessible during first installation? Let us know what you find.

Hah, I think I've found the cause of our different experience wrt this issue. I had `postgresql__server` set in my debug inventory to something like `qtest1.example.com`, which creates `/etc/postgresql-common/user_clusters` with a line like this:

```
# USER         GROUP          VERSION    CLUSTER              DATABASE
*              *              13         <host>:<port>        *
```

If I leave out `postgresql__server`, the same line will be:

```
# USER         GROUP          VERSION    CLUSTER              DATABASE
*              *              13         main                 *
```

With the latter line, passwordless access works for the `postgres` user, and `dbconfig` manages to create/update the DB/user/pass....with the first line, I think `dbconfig` gets confused enough that it fails to do any actual changes to the DB.

Still need to see if I can find a per-package way to disable `dbconfig`

Alphix: Creating `/etc/dbconfig-common/icinga2-ido-pgsql.conf` (*before* the `icinga2-ido-pgsql` package is installed, i.e. before the `icinga_db` role is executed) with the following content:

```
dbc_install='false'
dbc_upgrade='false'
dbc_remove='false'
```

Seems to fix this issue and only disable `dbconfig` for the `icinga2-ido-pgsql` package and not system-wide.

See also: [this issue](https://github.com/Icinga/deb-icinga2/issues/10)

drybjed: Good to know that you found the solution. Remember to add a comment in the file explaining what's going on in case someone logs in to the server and looks for explanation when something breaks.

Alphix: > Good to know that you found the solution. Remember to add a comment in the file explaining what's going on in case someone logs in to the server and looks for explanation when something breaks.

See https://github.com/debops/debops/pull/2318

PR #2296, opened 2023-03-10
[nginx] Add support for toplevel options to the default template

Alphix:
Some configurations require options to be defined at the toplevel, i.e. outside any "server" blocks. This adds support for such configurations to the default template (which is included by the other templates except custom).


Alphix:
I should add that I need this functionality for a different role that I'm working on (`bind`, where it can be used to create a reverse proxy for DNS-over-HTTPS).

drybjed (approved):

PR #2297, opened 2023-03-10
[bind] Add new role

Alphix:
This is a role I've been working on for the venerable `bind` DNS server.

It should be pretty much feature complete ATM, and supports things like:

- DNS key handling
- DNSSEC
- DNS-over-TLS/HTTPS
- Key rollovers
- etc

### DIFF
--- a/ansible/playbooks/layer/srv.yml
+++ b/ansible/playbooks/layer/srv.yml
@@ -162,6 +162,9 @@
 - name: Configure PowerDNS service
   import_playbook: '../service/pdns.yml'
 
+- name: Configure BIND DNS server
+  import_playbook: '../service/bind.yml'
+
 - name: Configure rspamd service
   import_playbook: '../service/rspamd.yml'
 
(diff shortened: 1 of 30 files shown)


drybjed (approved):
This is fantastic! Thanks for working on this role, and forgive me that it took me a while to get to it. I'm looking forward to seeing it in action.

I'll fix some `ansible-lint` spacing issues before merging, otherwise it looks like it's ready to go.

Alphix:
> This is fantastic! Thanks for working on this role, and forgive me that it took me a while to get to it. I'm looking forward to seeing it in action.
> 
> I'll fix some `ansible-lint` spacing issues before merging, otherwise it looks like it's ready to go.

Cool, thanks for reviewing the role...and I'm happy it's in. This was probably the most complicated (or 2nd most complicated, `kerberos` might be 1st, though not submitted yet) role that I've written so far.

Note that this makes PR #2209 even more relevant (since `bind` is AppArmor-enabled and the `/etc/pki` hierarchy doesn't work well with that...if one wishes to use TLS).

PR #2300, opened 2023-03-11
[sshd] Don't use "Include" on pre-Bullseye releases

### LINKED ISSUE #2293: [sshd] "Include" isn't available on openssh-server sshd_config file on Debian 10
Hello,

It seems `Include` isn't available on `sshd_config` of Debian 10 although Debian 11 doesn't complain about that.
Please check these articles:
https://stackoverflow.com/questions/43371489/ssh-config-line-1-bad-configuration-option-include
https://lists.debian.org/debian-user/2017/02/msg00096.html
https://www.openssh.com/txt/release-7.3

Tested with
```
ssh -V
OpenSSH_7.9p1 Debian-10+deb10u2, OpenSSL 1.1.1n  15 Mar 2022
```

Error message receive:
```
sudo sshd -t
/etc/ssh/sshd_config: line 12: Bad configuration option: Include
/etc/ssh/sshd_config: terminating, 1 bad configuration options
```

We should remove the line from the config
```
Include /etc/ssh/sshd_config.d/*.conf
```

Alphix:
The "Include" directive is only supported from Bullseye onwards.

Resolves: https://github.com/debops/debops/issues/2293

### DIFF
--- a/ansible/roles/sshd/defaults/main.yml
+++ b/ansible/roles/sshd/defaults/main.yml
@@ -247,6 +247,7 @@ sshd__original_configuration:
   - name: 'Include_sshd_config.d'
     option: 'Include'
     value: '/etc/ssh/sshd_config.d/*.conf'
+    state: '{{ "absent" if ansible_distribution_release in [ "stretch", "buster" ] else "present" }}'
 
   - name: 'Port'
     value: 22


drybjed (approved):
Thanks for the quick fix. :-)

PR #2301, opened 2023-03-11
[ansible] Use fully-qualified names for ipv4/ipv6/ipaddr/ipwrap

### LINKED ISSUE #2271: use ansible.utils.ipvX module instead of ansible.netcommon.ipvX
```
TASK [debops.debops.ferm : Generate firewall rules]
[DEPRECATION WARNING]: Use 'ansible.utils.ipv4' module instead. This feature will be removed from ansible.netcommon in a release after 2024-01-01. Deprecation warnings can be disabled by setting 
deprecation_warnings=False in ansible.cfg.
[DEPRECATION WARNING]: Use 'ansible.utils.ipv6' module instead. This feature will be removed from ansible.netcommon in a release after 2024-01-01. Deprecation warnings can be disabled by setting 
deprecation_warnings=False in ansible.cfg.
```

Alphix:
This fixes warnings from ansible such as:
[DEPRECATION WARNING]: Use 'ansible.utils.ipv4' module instead. This feature will be removed from ansible.netcommon in a release after 2024-01-01...

Resolves: https://github.com/debops/debops/issues/2271

### DIFF
--- a/ansible/playbooks/tools/6to4.yml
+++ b/ansible/playbooks/tools/6to4.yml
@@ -26,7 +26,7 @@
     debops_6to4_var_ipv4_interface: '{{ debops_6to4_ipv4_interface | default(ansible_default_ipv4.interface) }}'
 
     # IPv6 address converted from IPv4 public address
-    debops_6to4_var_ipv6_address: '{{ hostvars[inventory_hostname]["ansible_" + debops_6to4_var_ipv4_interface].ipv4.address | ipv4("6to4") }}'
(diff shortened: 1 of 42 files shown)


drybjed (approved):
Awesome work as always, thank you @Alphix! :-) I'll add `ansible.utils` collection to requirements to make sure it's installed by `ansible-galaxy` with the DebOps collection.

I'll backport these changes to the `stable-3.0` release, because otherwise it's broken on the current stable Ansible. This is only a filter name change, so there shouldn't be any issues with such a large diff.

PR #2305, opened 2023-03-20
Manage Ferm rules for OpenSearch

BerkhanBerkdemir:
This patch is inspired from debops.elasticsearch role.  The idea is that
if the current installation is not secured with SSL/TLS, then I would 
need a firewall to protect the endpoint.  Also, this patch is an initial
work to let user to dynamically choice HTTP and TCP transport ports.


drybjed (approved):
Nicely done, thanks! :-)

PR #2306, opened 2023-03-20
Upgrade actions checkout plugin

BerkhanBerkdemir:
`actions/checkout` plugin from version 2 was using Node.js 12.
According to GitHub, that should have been upgraded.  This patch work
upgrades all `actions/checkout` from 2 to 3.


drybjed (approved):
Thanks! I'll backport this to stable branches as well.

PR #2307, opened 2023-03-21
[proc_hidepid] Avoid issues with polkit(d)

Alphix:
polkitd expects to be able to dig around in /proc using an unprivileged UID/GID. In addition, polkitd is started as root/root and later drops privileges to polkit/polkit, meaning that the SupplementaryGroups= trick used for other systemd services will not work for polkitd.

The error messages that polkit emits when /proc/pid hiding is enabled are difficult to diagnose ("/proc/pid/blah" missing, though it's clearly "there") and the things that break are not (from a user PoV) obviously linked to polkit or pid hiding (for example: things which are no longer possible for the local user include reboots, connecting/disconnecting from a wireless network, etc)

### DIFF
--- a/ansible/roles/proc_hidepid/defaults/main.yml
+++ b/ansible/roles/proc_hidepid/defaults/main.yml
@@ -36,6 +36,15 @@ proc_hidepid__base_packages: [ 'libcap2-bin' ]
 #
 # List of additional APT packages to install with hidepid support.
 proc_hidepid__packages: []
+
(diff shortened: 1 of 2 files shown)


drybjed (approved):
I was considering if moving this check to the fact script could be better, but we probably don't need to check the package list every time Ansible gathers facts. Let's go with this solution then.

I'll backport this to the stable releases as well.

PR #2308, opened 2023-03-23
* [etc_aliases] condition for save dependent recipients on controller

hungpr0:
When using role with AWX, "Save etc_aliases dependent recipients on Ansible Controller" task create the file in AWX job container while it shouldn't if etc_aliases__dependent_recipients is []

### DIFF
--- a/ansible/roles/etc_aliases/tasks/main.yml
+++ b/ansible/roles/etc_aliases/tasks/main.yml
@@ -73,3 +73,4 @@
     mode: '0644'
   become: False
   delegate_to: 'localhost'
+  when: etc_aliases__dependent_recipients | length > 0


drybjed (approved):
Sounds good, thanks. I'll backport this to the stable release.

Also, how's DebOps on AWX, especially the 'secret/' directory? Are they stored in some sane place, or in a temporary container that gets removed after a job run? Any suggestions how to improve DebOps on AWX?

hungpr0:
> Sounds good, thanks. I'll backport this to the stable release.
> 
> Also, how's DebOps on AWX, especially the 'secret/' directory? Are they stored in some sane place, or in a temporary container that gets removed after a job run? Any suggestions how to improve DebOps on AWX?

By default on AWX operator, there isn't any persistent storage on the temp container. Of course we can mount custom volume (https://github.com/ansible/awx-operator#custom-volume-and-volume-mount-options) but that is what I tried to avoid. AWX executation environment container just need to update from source control repository.  Yes, secrets can be writable to pod but get removed after job run - pod lifetime.

I'm checking the setup of `debops.postfix` and some or another roles (`postconf`, `dovecot`) and found that we would need to store `secret` somewhere on persistent storage or ansible run isn't `idempotent`

```
TASK [debops.roles01.postfix : Save dependent configuration on Ansible Controller] ***
Tuesday 11 April 2023  06:03:13 +0000 (0:00:01.672)       0:05:48.796 ********* 
[WARNING]: Unable to find '/runner/secret/postfix/dependent_config/my_hostname/maincf.json' in expected paths (use -vvvvv to see paths)
--- before
+++ after: /home/runner/.ansible/tmp/ansible-local-20siwakygy/tmprw9g400d/maincf.json.j2
```

This is something by role design, not sure we can have workaround without ansible controller persistent storage for secret.

drybjed:
For secrets themselves - most instances are defined as role default variables, which can be redefined in the Ansible inventory to reference secrets directly instead of a `lookup("password")` call. In some cases, item lists provide a way to specify a `password` parameter or similar to pass the password indrectly instead of creating it randomly.

There's a lot, and I mean a LOT of instances of this in DebOps, without a simple way to handle this change wholesale. In the past I talked with some Ansible developers about finding a way to provide some kind of redirect mechanism so that lookups could be funneled to Hashicorp Vault, for example, but for now nothing came of it. It's a tradeoff between having to specify each password manually before deploying a service vs having a ready-made randomized password stored in the filesystem. I prefer the second approach, but I understand that with something like a Kubernetes or other containerized deployment it can be a hassle.

The tasks you noticed try to handle idempotency with dependent roles in a case where a service uses a single configuration file, like Postfix with its `main.cf` and `master.cf` configuration files. I had to make a choice between storing that data on the remote hosts vs Ansible Controller and I selected the second approach to avoid security issue with data retrieved from the remote host being incorporated into the configuration pipeline. This unfortunately requires some kind of data storage to store the dependent configuration and the `ansible/secret/` directory was an obvious choice being already implemented.

There are a few instances of this in various DebOps roles. I think that this can be dealt with by making these tasks conditional - users can then disable this mechanism when not desired and handle configuration from dependent roles by adding it via Ansible inventory. I'm not sure yet about the scope of this change (have a per-role variable, or add a variable to control this centrally in the `debops.secret` role), but I think that the latter might work better in this case. I'll keep that in mind and try to implement this at some point.

Thanks for sharing your insights. :)

PR #2311, opened 2023-03-27
Improvements in 'debops' subcommands

drybjed:


ypid:
@drybjed Out of curiosity, are you familiar with ansible-navigator? I just started playing with it but it looks interesting.

drybjed:
@ypid I looked at `ansible-navigator` briefly. The UI requires some getting used to. :) DebOps inventory seems to work fine, but the collection side is not recognized, I wonder how could that be improved.

Ansible has multiple utilities under its belt, probably all of which utilize `$ANSIBLE_CONFIG` to orient themselves in a particular environment. I wonder how DebOps should deal with this. Adding support for each command directly like it is currently with `ansible` and `ansible-playbook` probably is not ideal. I think that I can add `debops env` command, in the spirit of `env(1)` - that way when you run `debops env ansible-navigator` anywhere in the project directory, you get a proper Ansible environment. As for other commands, adding support for `ansible-galaxy collection` is probably benefical so that users can manage collections in the project directory easily. What do you think?

ypid:
@drybjed I still think we should integrate with existing tools as good as possible and limit the Python code in DebOps. Getting stuff like bell upstream might take more time but seems to be more sustainable. I will play with `ansible-navigator` more. It is a great step forward for checking out the inventory and for running playbooks and getting an overview of a playbook run.

drybjed:
> I still think we should integrate with existing tools as good as possible and limit the Python code in DebOps.

Problem with this is that many, many tools in Ansible ecosystem rely on environment variables to configure their behavior. For example, it's very hard to tell `ansible-lint` where it can find DebOps collection in the repository. Actually, I don't think that linting detects the collection at all, that's why there are weird issues in CI right now.

Problem with environment variables being sources of configuration is that you need to create them somehow in the environment. This is done either by sourcing variables into the current process environment (think Python's `source bin/activate` to enable a given virtualenv), or a command is started by a wrapper script that prepares execution environment for that command. From these two choices, I don't really like the first one - seems prone to human errors and mistakes. Providing a wrapper for a command seems to be much easier thing to understand, just like `docker` is a wrapper for CGroups, in a sense.

> Getting stuff like bell upstream might take more time but seems to be more sustainable.

I agree. `debops run --bell` might be a good example to show the Ansible team that this functionality might be worth implementing at some point. I wonder if it will catch on and will be used by our users.

PR #2316, opened 2023-04-04
[dovecot] use fqdn for template_src

hungpr0:
Use FQDN for template_src


drybjed (approved):
I wonder how it was missed, this change apparently was done two years ago. Weird. Also, thanks! :-)

PR #2319, opened 2023-04-04
[apparmor] Fix bugs and integrate with other roles

Alphix:
This PR fixes some bugs in the AppArmor role and adds integration for:

- dnsmasq
- apt_cacher_ng
- bind
- slapd

Especially `bind` and `slapd` have AppArmor support in Debian `stable` (not enabled by default for `slapd` IIRC), and break due to the custom `/etc/pki` hierarchy.

`dnsmasq` and `apt_cacher_ng` had AppArmor support via the `debops-contrib` role.

The only missing step is adding the role to the `common` playbook. I'm not 100% sure where in the playbook it should go.


Alphix:
Draft for now since I've only tested `dnsmasq` so far

Alphix:
Ok, now I've (lightly) tested the other roles as well.

- `slapd` using the AppArmor profile from Ubuntu, fails to start without this patchset, works with it
- `apt_cacher_ng` only needs AppArmor customisation if the profile dir is changed
- `bind` with PKI usage, fails to start without this patchset, works with it

Alphix:
@drybjed: I'll rebase the PR ASAP, meanwhile, could you have a look? I'd like to have this PR merged soon since `slapd` and `bind` are both AppArmor-enabled in Bookworm by default and might break (depending on configuration) out-of-the-box

BerkhanBerkdemir:
I didn't know that PR already exist in here for BIND and AppArmor. I had to find my own solution by patching the PyPI distribution. When would this PR be merged to the `master` branch, @drybjed?

PR #2330, opened 2023-04-18
[dhcp_probe] Add a Wants= on network-online.target

boucman:
According to https://systemd.io/NETWORK_ONLINE/ network-online.target is an "active target". This means that this target will not be started unless explicitely needed by another service

We already have an After=network-online.target but After= does not start the target, it only guarantees ordering

This means that if dhcp_probe is the only network-related service,network- online.target will not be started and dhcp_probe will be started too early and fail

See the first FAQ in the page documented above.

**NOTE** This PR is agains stabe-2.3 since this is what I have for testing, but it is probably needed for master too


boucman:
Ping ? 
This should be a quick review...

Issue #2232, opened 2022-11-17
PKI acme-tiny isn't converting the intermediate cert to PEM correctly


Cadair:
I have two realms both using acme-tiny, and the `acme/intermediate.pem` certificate is not converted to the ascii-armoured PEM format correctly and therefore the chain and other files in `public/` are also broken.

goetzk:
I believe #2198 should resolve this for you once merged.

ypid:
Looks like the same as what I reported after you in #2250. #2250 has a bit more details so I close this issue. Hope that is ok for you.

Issue #2250, opened 2022-12-24
Debian 12 (Bookwork) pki acme intermediate_root.pem is binary which makes nginx fail


ypid:
Just a heads up. I have not checked for the root cause yet.

## Symptom

`/etc/pki/realms/$realm/trusted.crt` contains a binary intermediate cert.

Nginx fails with:

```
nginx -t 
2022/12/24 14:06:20 [emerg] 30127#30127: SSL_CTX_load_verify_locations("/etc/pki/realms/$realm/trusted.crt") failed (SSL: error:05800088:x509 certificate routines::no certificate or crl found)
nginx: configuration file /etc/nginx/nginx.conf test failed
```

## Manual fix

```Shell
cd /etc/pki/realms/$realm/public
openssl x509 -in intermediate_root.pem > intermediate_root.pem1 && mv intermediate_root.pem1 intermediate_root.pem
```

## Proper fix

Probably do the binary to PEM conversion in `pki-realm`.

goetzk:
I've hit this a couple of times now too. I see it was reported in #2232 as well as here and the pull request.

Removing the binary component allows nginx to start though the certification chain is borked so its not a fix per se.

goetzk:
@ypid did #2331 resolve this for you?

thiras:
I confirm this bug. It even occurs at the default welcome server.

PR #2331, opened 2023-04-18
[pki] pki-realm specify input format to convert DER to PEM

bits:
When the input format has been determined to be other than PEM, `openssl -in` by itself fails to read the DER format file. Adding `-inform DER` enables openssl to read the DER file and output a PEM format file as `intermediate.pem.tmp`, allowing the script to continue by renaming it to `intermediate.pem`, preparing `root.pem`, and completing successfully.

Resolves the issue where Let's Encrypt ACME certificates were requested, verified, issued and retrieved, but stored in binary DER format as `intermediate.pem.tmp`, and where conversion to PEM failed, leaving the `/etc/pki/realm/example.org/acme` directory containing:

`account_key.pem`
`cert.pem`  ← *good cert*
`error.log`  ← *no errors*
`intermediate.pem.tmp`  ← *DER format file*
`openssl.conf`
`request.pem`

The certificate chain was left half-prepared, so the pki role deployed the domain CA-signed certificate instead.  With this patch, the ACME certificate chain can be converted into PEM format and the pki role can use and deploy the ACME-issued Let's Encrypt certificate.

### DIFF
--- a/ansible/roles/pki/files/usr/local/lib/pki/pki-realm
+++ b/ansible/roles/pki/files/usr/local/lib/pki/pki-realm
@@ -1072,7 +1072,7 @@ convert_der_to_pem () {
 
     if [ -n "${input_file}" ] && [ -r "${input_file}" ] ; then
         if ! openssl x509 -inform PEM -in "${input_file}" -noout 2>/dev/null ; then
-            openssl x509 -in "${input_file}" -outform PEM -out "${input_file}.tmp"
+            openssl x509 -inform DER -in "${input_file}" -outform PEM -out "${input_file}.tmp"
             mv "${input_file}.tmp" "${input_file}"
         fi
     fi


bits:
This refines #2198, and fixes #2232 and #2250

bits:
Separately, could this be the cause of ACME problems mentioned in commit 1b6e0a9?

drybjed:
Thanks for checking this and finding a fix. I'll try to test this in a short while, unless somebody else tries first - I have a few people on the IRC channel who encountered this issue and might test this fix soon.

lewer:
> Separately, could this be the cause of ACME problems mentioned in commit [1b6e0a9](https://github.com/debops/debops/commit/1b6e0a9b87838ce066b2ed513b9f9429e8abfdbf)?

Indeed, I was able to test the current PR together with https://github.com/debops/debops/commit/1b6e0a9b87838ce066b2ed513b9f9429e8abfdbf : everything is working fine; no more `intermediate.pem.tmp` remaining at the end, the realm directory is what it must be, with symlinks to `/etc/letsencrypt`. Thanks!

PR #2332, opened 2023-04-18
[swapfile] Use intended logic when checking whether to enable swapfiles

bits:
Previously, parentheses resulted in re-enabling swapfile on every run based on value of ansible_system_capabilities_enforced instead of intended logic


bits:
Closed, similar changes were committed in ae1457e4bbbe69c1af954aafcc22342067f98b44

PR #2343, opened 2023-05-01
[system_users] Add support for NSS-based lookups to getent_passwd.py3

Alphix:
The script currently only consults /etc/passwd, and errors out with rc = 2 if the local user cannot be found, which generates a rather unhelpful output from Ansible (connection closed, rc = 2, basically).

This patch adds support for using the libc getent binary as a fallback, which allows users to be found even if they are only defined in databases such as LDAP.


drybjed (approved):

PR #2349, opened 2023-05-16
[dnsmasq] allow empty dhcp_ipv6_mode

jankowa:
if dhcp_ipv6_mode is empty, starting dnsmasq fails
since there is a double separation comma.


drybjed (approved):
Thanks! I'll backport this change to the stable release.

PR #2350, opened 2023-05-16
Fix libvirtd kvm packages

jankowa:
In bullseye no packages at all will get installed.
Parentheses are evaluated wrong.


ypid:
@jankowa Something is wired with your branch. If you don‘t know how to resolve it or what went wrong I would probably `git reset --hard master` and `git cherry-pick HEAD~` (untested) something like that.

drybjed (changes requested):
Thanks for fixing the branching issue. Can you also update the jinja spacing warnings emitted by `ansible-lint` before I merge this?

drybjed (approved):
Looks good now, thanks. I'll backport this to the stable releases as well.

PR #2355, opened 2023-05-24
Fixes ansible shell with pipefail

### LINKED ISSUE #2354: few instances of pipefail without calling bash explicitely
ansible/roles/owncloud/tasks/tarball.yml explicitely set 'sh' as its shell for pipefail she'll script which fails with:
```
TASK [debops.debops.owncloud : Query for the full version string of the current release] ************************************************************************************
fatal: [argos]: FAILED! => {"changed": false, "cmd": "set -o nounset -o pipefail -o errexit && curl -s -m 900 https://download.nextcloud.com/server/releases/ | sed --silent 's/.*href=\"nextcloud-\\(24\\.0[^\"]\\+\\).zip.asc\".*/\\1/p' | sort --version-sort --reverse", "delta": "0:00:00.005296", "end": "2023-05-23 23:56:58.329173", "msg": "non-zero return code", "rc": 2, "start": "2023-05-23 23:56:58.323877", "stderr": "sh: 1: set: Illegal option -o pipefail", "stderr_lines": ["sh: 1: set: Illegal option -o pipefail"], "stdout": "", "stdout_lines": []}
```
this since commit 53e6ca4ae07562b36a6df73d0f065b50cecb33ab "[ci] Fix 'risky-shell-pipe' ansible-lint rule"

I am cooking a PR to fix the ansible shell calls with pipefail and not explicitly setting bash as executable.

Note that a few tasks files set the executable to '/bin/bash' while other do to 'bash'. Would you like them all to be set to 'bash'?

The Vagrantfile has pipefail inline shell script, though  I have been unable to find out if the vagrant shell provisioner is the default /bin/sh (ie dash on Debian per default) or hardcoded to bash. Requires testing on a box with dash as shell.

prahal:
pipefail requires bash.
default dash shell on Debian errors out when this option is set.

Fixes #2354.

### DIFF
--- a/ansible/roles/controller/handlers/main.yml
+++ b/ansible/roles/controller/handlers/main.yml
@@ -17,6 +17,8 @@
   ansible.builtin.shell: |
     set -o nounset -o pipefail -o errexit &&
     type debops-update > /dev/null 2>&1 && (echo 'debops-update' | batch > /dev/null 2>&1) || true
+  args:
+    executable: 'bash'
   register: controller__register_update_batch
   changed_when: controller__register_update_batch.changed | bool
   when: (not controller__update_method == 'sync' and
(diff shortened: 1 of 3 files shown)


drybjed (approved):
Good idea, thanks! I'll backport this to the stable releases.

PR #2357, opened 2023-05-24
[dovecot] fix ssl_dh when dovecot__ssl_dh_file is empty

hungpr0:
when `dovecot__ssl_dh_file` is empty, `ssl_dh = <` doens't allow dovecot to start. Error message `doveconf: Fatal: Error in configuration file /etc/dovecot/dovecot.conf line 53: ssl_dh: read(/etc/dovecot/) failed: Is a directory`. This fixes the issue


drybjed (approved):
Good catch, thanks. I'll add parentheses in the if statement to make it more explicit. This will be backported to the stable release.

PR #2359, opened 2023-06-01
Add support for configuring /etc/issue.net

goetzk:
/etc/issue.net is the network specific twin of /etc/issue so is likely to want modification at the same time.
For ease of administration it defaults to the same template as /etc/issue but can be overridden separately should the requirement call for it.


drybjed (changes requested):
The idea is good, but I don't like using the `/etc/issue` template by default for this. The two files have completely different formats and escape sequences are interpreted differently, it will be impossible to make both of them look good.

I would do this by adding an `etc/issue.net.j2` template with simple contents similar to the default one, but I would use the `debops.debops.template_src` lookup plugin to get the template. This way, users can add their own template using the overrides system in DebOps if they want to.

PR #2362, opened 2023-06-12
feat(netbox): Upgrade from NetBox v3.4.2 to v3.7.2

ypid:


ypid:
@drybjed I rebased.

drybjed (approved):
Thanks for the update. Since this is a release update, I'll see if this PR can be backported cleanly to the stable DebOps releases, and backport if it that works without issues.

PR #2363, opened 2023-06-13
RFC: Switch all Debian releases : bookworm is now stable

prahal:
Add trixie as the new testing.

---

I did not run any testsuite against this patchset on my side (ansible-lint ok)

I did not migrate ipxe to bookworm (I need to check if adding the new name is ok).

I did not add non-free-firmware in the dist-upgrade playbook and the apt role. Should we add it or leave it to the user? (and I did not check if and how to handle a non-free-firlmware in the pressed role)

I also did not add the vagrant box bookworm64 as the new default for Vagrantfile and lib/tests/jane.

bookworm switch from php7.4 to php8.2 is not taken care of. 

I removed stretch from reprepro.

### DIFF
--- a/ansible/playbooks/tools/dist-upgrade.yml
+++ b/ansible/playbooks/tools/dist-upgrade.yml
@@ -30,10 +30,10 @@
     # * ansible/roles/reprepro/defaults/main.yml
     # * ansible/playbooks/tools/dist-upgrade.yml
     dist_upgrade_version_map:
-      'jessie': 'stretch'
       'stretch': 'buster'
       'buster': 'bullseye'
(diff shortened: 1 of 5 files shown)


drybjed (approved):
The changes look good, thanks! I'll make sure that they are backported to the stable releases as well.

The `non-free-firmware` issue in the `apt` role will have to be coded carefully, since we are in a transition phase - older releases cannot use it yet. Some conditions are needed, I'll look into it.

I'll check the `ipxe` update as well, since I need to be able to install Bookworm easily. Preseed changes will definitely help with that, thanks!

PR #2369, opened 2023-06-16
[sshd] set ChallengeResponseAuthentication to False by default

lewer:
Currently, public key authentication is not enforced if KbdInteractiveAuthentication is set to  *yes*, even if PasswordAuthentication is set to *no*.

Setting KbdInteractiveAuthentication to *no* seems a good idea: this is the default value in /usr/share/openssh-server/sshd_config.

### DIFF
--- a/ansible/roles/sshd/defaults/main.yml
+++ b/ansible/roles/sshd/defaults/main.yml
@@ -360,6 +360,13 @@ sshd__original_configuration:
     value: True
     state: 'init'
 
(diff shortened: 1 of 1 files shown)


drybjed (approved):
I'm not sure why `ChallengeResponseAuthentication` isn't in the revamped configuration options, weird...

Thanks for the patch. In the older releases, `ChallengeResponseAuthentication` is hardcoded to `no` in the configuration file template, so they should be covered.

PR #2373, opened 2023-06-22
[ifupdown]: escape and unescape interface names in systemd unit name

### LINKED ISSUE #2371: ifupdown - interface names in unit names should be systemd escaped and unescaped
man systemd.unit tells in "STRING ESCAPING FOR INCLUSION IN UNIT NAMES": "The escaping algorithm operates as follows: given a string, any "/" character is replaced by "-", and all other characters which are not ASCII alphanumerics, ":", "_" or "."  are replaced by C-style "\x2d" escapes." and "This escaping is fully reversible, as long as it is known whether the escaped string was a path (the unescaping results are different for paths and non-path strings)."

The minus '-' in the interface name ought to be escaped by a C-style "\x2d" escape.
This bug would be in debops ifup-allow-boot.service generation of `iface@$i` with the vanilla interface name output from ifquery.
```
$ systemd-escape eth0-test
eth0\x2dtest
```

I believe %I is correct here.
https://github.com/debops/debops/pull/569 should be reverted and ifup-allow-boot.service fixed (I made a local version that replace "$i" in the systemctl calls  by `$escaped_iface="$(systemd-escape "$i")"`. Though I see no other calls to systemd-escape.

This will complicate the manual calling of these units as it will require typing `systemctl start 'iface@eth0\x2d-test'` instead of `systemctl start iface@eth0-test`.

prahal:
This fixes systemd unit names escaping. For one it let us escape the "hyphen" in the network interface name when we add them to systemd unit name in ifupdown iface@ services.
This restores unescaping of the systemd instance name to get the network interface name (the removal of this unescaping was a hack to avoid escaping the systemd unit instance name).

Closes #2371.

Note that the old hackish behavior could be fine too but I believe the ifupdown-reconfigure-interfaces script treats ifup@ and iface@ the same while one unescapes the instance name to get the systemd instance name to get the interface name to pass to ifup/ifquery/ifdown while the other does not. I believe (without a thorough check) that this script is tailored for iface@.service but will not play well for ifup@.service when the interface includes a "hyphen".

The plus of this PR is that with it the ifupdown-reconfigure-interfaces script plays well with "hyphen" in the network interface name for both ifup@.service and iface@.service. The down is that the service name will now C escape any non-alphanumeric characters rendering hand typing these instance service names and reading the output of systemctl difficult to read.

### DIFF
--- a/ansible/roles/ifupdown/files/script/ifupdown-reconfigure-interfaces
+++ b/ansible/roles/ifupdown/files/script/ifupdown-reconfigure-interfaces
@@ -159,17 +159,19 @@ interface_down () {
         local -a if_hotplug_interfaces
         local -a if_boot_interfaces
         local -a systemd_ifup_instances
+        local systemd_escaped_interface
         mapfile -t if_hotplug_interfaces < <(ifquery --list --allow=hotplug)
         mapfile -t if_boot_interfaces < <(ifquery --list --allow=boot)
-        mapfile -t systemd_ifup_instances < <(systemctl list-units --no-legend --state=active 'ifup@*.service' | awk '{print $1}' | sed -e 's/^ifup\@//' -e 's/\.service$//')
+        mapfile -t systemd_ifup_instances < <(systemctl list-units --no-legend --state=active 'ifup@*.service' | awk '{print $1}' | sed -e 's/^ifup\@//' -e 's/\.service$//' | xargs -0 systemd-escape -u)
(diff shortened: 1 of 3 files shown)


drybjed (approved):
Thank you for this patch. I think that the tradeoff (correctly working executable code vs harder to parse logs) is worth it, since if the code works as expected, logs won't be as needed to debug things.

I'll backport this to the stable releases.

PR #2374, opened 2023-06-22
[ifupdown]: order iface@ service after apparmor

prahal:
Sync with Debian ifup@.service. This has no functional effect as networking.service already adds an "After" apparmor.service. Still it does not hurt and reduce the difference with the Debian provided ifup@ service.


drybjed (approved):
Thanks for the update. I'll backport it to stable releases as well.

PR #2375, opened 2023-06-26
[sshd] set ChallengeResponseAuthentication fo False by default

lewer:
Following PR https://github.com/debops/debops/pull/2369

For openssh-server piror to version 8.7, `KbdInteractiveAuthentication  no` does not seem to be sufficient to really disable interactive authentication.

`KbdInteractiveAuthentication` and `ChallengeResponseAuthentication` can both be set to no to ensure that interactive authentication is disabled for all versions of openssh-server.


drybjed (approved):
Nicely done, thanks! :-)

Issue #2377, opened 2023-06-28
workaround to apt_key DEPRECATION warning - Bookworm


gardouille:
A third-party repository on Debian Bookworm can lead to a warning on an `apt update`:
```
W: http://example.com/debian/dists/bookworm/InRelease: Key is stored in legacy trusted.gpg keyring (/etc/apt/trusted.gpg), see the DEPRECATION section in apt-key(8) for details.
```

These variables previously worked correctly with ≤Bullseye:
``` yaml
apt__keys:
  - url: 'https://example.com/debian/example.com.asc'
    state: 'present'

apt__repositories:
  - filename: 'my-repo'
    repo: 'deb https://example.com/debian/ stable main'
    state: 'present'
```

Pending [deb822_repository](https://docs.ansible.com/ansible/latest/collections/ansible/builtin/deb822_repository_module.html) support in DebOps, as a first workaround (and not the cleanest), adding **keyring** parameter in `apt__*keys`:
  - Keyring file and repository filename should be the same.
  - Set **.gpg** extension to have a dearmored key.
``` yaml
apt__keys:
  - url: 'https://example.com/debian/example.com.asc'
    keyring: '/etc/apt/trusted.gpg.d/my-repo.gpg'
    state: 'present'

apt__repositories:
  - filename: 'my-repo'
```

apt-key(8) recommended to use **Signed-by** option (compatible DebOps 👍) with a keyrings directory such as /etc/apt/keyrings (not directly compatible with debops.apt 👎, at least i haven't find an easy solution).

See also [ansible issue #78063](https://github.com/ansible/ansible/issues/78063) for extra infos ;)

seven-beep:
IMHO this issue could be closed. apt__keys is now documented as deprecated and the use apt__repositories is documented for importing keys from multiple sources.

@gardouille do you mind closing the issue ?

gardouille:
Sure, thanks for the update :)

PR #2378, opened 2023-06-29
Deb822 repository bookworm

gardouille:
This PR should bring [deb822_repository](https://docs.ansible.com/ansible/latest/collections/ansible/builtin/deb822_repository_module.html) module support.

One thing missing, this module requires **python3-debian** (which is not installed by default by Debian or DebOps) but i don't know the best place/module to install it :)

This should also avoid apt_key deprecation warning (see #2377).

### DIFF
--- a/ansible/roles/apt/defaults/main.yml
+++ b/ansible/roles/apt/defaults/main.yml
@@ -219,6 +219,31 @@ apt__group_repositories: []
 apt__host_repositories: []
                                                                    # ]]]
                                                                    # ]]]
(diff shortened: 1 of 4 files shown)


drybjed (approved):
Sounds good! I'm currently working on refreshing the `apt` role, and I already have an idea to combine both legacy and deb822 repositories in one list, so I'll update these changes in the `master` branch. But I'll backport this to the stable branches so that other infrastructure users can still benefit from this.

PR #2387, opened 2023-07-14
no boot interface at boot - [ifupdown] Fix breakage introduced by hyphen support fixup.

prahal:
I missed a semicolon in the oneliner.

But it turns out that curly braces around a variable (parameter expansion) are also broken in a systemd service ExecStart.

systemd expand environment variable substition use with this same dollar syntax. And from the systemd documentation it seems that the fact that variable with a dollar prepended but without curly brace not being replaced by an empty string is a bug in systemd:
https://www.freedesktop.org/software/systemd/man/systemd.service.html#Command%20lines 'To pass a literal dollar sign, use "$$". Variables whose value is not known at expansion time are treated as empty strings.'

tested systemd version: 252.6-1

Reproducer:
ExecStart=/bin/sh -ec '\
  escaped_iface="br1"; \
  echo $escaped_iface ; \
  echo ${escaped_iface} ; \
  '
systemctl status on such a service will only echo once.

As this fix is critical to ifupdown users I only fix what is broken that is variable identifier surrounded by curly braces. I leave replacing all single dollar before a shell variable in systemd exec commands by a double dollars to a later lower piority push request. Also I believe there are other role that have this broken use of non double dollars shell variables in systemd unit. So doing them all at once might be better.

--
sorry last time I tested thoroughly the script.
I also tested the systemd unit changes but at least I made a last-minute change to add curly braces and I believe I forgot to test it anew.
Today I rebooted the boxes and this issue triggered.


prahal:
@drybjed ping for this ifupdown hyphen in the interface name fixup of the previous PR.

drybjed (approved):
Thanks for the bugfix. I tested the changes on a default Bookworm installation and the `ifup-allow-boot.service` unit starts correctly now. There was still an issue with the `iface@br0.service` not starting up correctly, but that can be fixed later.

I'll backport the changes to the stable releases as well.

prahal:
The workaround for br0 (I don't know if we could call that a clean fix) https://github.com/debops/debops/pull/2441
Maybe we should spend a bit more time investigating. I ask you for clarification about one of your comment on reddit  in https://github.com/debops/debops/pull/2441#issuecomment-1768600971. I don't remember the in and out, but maybe we should resume the discussion.
And maybe apply the workaround to let debops works with br0 in the meantime.

PR #2389, opened 2023-07-17
Fix disable apache vhost if enabled is not set

### LINKED ISSUE #2388: apache vhosts error if state: disabled and enabled not set
There seems to be an error in the apache role. If you set a vhost to 

```
  state: absent
```
but not set

```
  enabled: False
```
I get the **error:**

```
msg: src option requires state to be 'link' or 'hard'
```

It look there is one condition in `- name: Enable/disable Apache virtual hosts` missing:

```
--- a/ansible/roles/apache/tasks/main.yml
+++ b/ansible/roles/apache/tasks/main.yml
@@ -153,7 +153,7 @@
                                       | d(item.name
                                           if (item.name is string)
                                           else item.name[0] | d("default")) + ".conf")
-             if (item.enabled|d(True) | bool)
+             if (item.enabled|d(True) | bool and (item.state|d("present") != "absent"))
              else omit }}'
     force: '{{ item.force|d(ansible_check_mode) | bool }}'
     state: '{{ item.enabled|d(True) | bool | ternary("link", "absent")
```

jankowa:
fixes #2388

### DIFF
--- a/ansible/roles/apache/tasks/main.yml
+++ b/ansible/roles/apache/tasks/main.yml
@@ -142,7 +142,7 @@
                                       | d(item.name
                                           if (item.name is string)
                                           else item.name[0] | d("default")) + ".conf")
-             if (item.enabled | d(True) | bool)
+             if (item.enabled | d(True) | bool and (item.state|d("present") != "absent"))
              else omit }}'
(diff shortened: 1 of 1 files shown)


drybjed (approved):
Thanks! I'll backport this to the stable releases as well.

PR #2408, opened 2023-09-02
Prevent using broken pyyaml versions in docker_server virtualenv

hannahdiels:
- When deploying the docker_server role, I get the same error as https://github.com/yaml/pyyaml/issues/724. The fix is to install a different pyyaml version (that is also still compatible with docker-compose).
- Pip installing one by one won't do as e.g. even if I put pyyaml first in the list, it would probably install pyyaml>6 because it's not taking into account the version constraint of docker-compose (it needs pyyaml<6). I could include that constraint in the pyyaml item but that constraint may change in the future so this is more robust. Instead this installs all packages in one go so pip can take into account all constraints; afterwards it applies any non-present packages which I assume are only absent packages.

### DIFF
--- a/ansible/roles/docker_server/defaults/main.yml
+++ b/ansible/roles/docker_server/defaults/main.yml
@@ -183,6 +183,9 @@ docker_server__default_pip_packages:
     path: '/usr/local/bin/docker-compose'
     src:  '{{ docker_server__virtualenv + "/bin/docker-compose" }}'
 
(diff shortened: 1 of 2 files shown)


hannahdiels:
Probably no longer necessary in the latest debops as [the virtualenv has been replaced](https://github.com/debops/debops/blob/master/CHANGELOG.rst#refdebopsdocker_server-role), but I haven't tried whether the new thing works.

PR #2409, opened 2023-09-06
[debops.postfix]: Restart postfix if requested

### LINKED ISSUE #2223: [postfix] "inet_interfaces" is not applied because postfix requires "restart" not "reload"
I have applied this config to the postfix role:

```
postfix__maincf:

  - name: 'inet_interfaces'
    value: '127.0.0.1'
    section: 'base'
```

Applying this configuration triggers handler [Check postfix and reload](https://github.com/debops/debops/blob/master/ansible/roles/global_handlers/handlers/postfix.yml#L20-L22) but reload doesn't apply the new config. If I manually execute `systemctl restart postfix` in the machine the new setting is properly applied. 

I could workaround the problem triggering handler [Check postfix and restart](https://github.com/debops/debops/blob/master/ansible/roles/global_handlers/handlers/postfix.yml#L16-L18) [in this line](https://github.com/debops/debops/blob/master/ansible/roles/postfix/tasks/main.yml#L94) but I am not sure if this could have any other side effect.

If this change would be accepted I can send a PR

GerMalaz:
Define the optional attribute 'need_restart' (default: false). If present and true on any variable, restart postfix instead of reloading it.

Closes #2223.


GerMalaz:
In the example given:
```
postfix__maincf:

  - name: 'inet_interfaces'
    value: '127.0.0.1'
    section: 'base
    need_restart: true'
```

Alphix:
> Define the optional attribute 'need_restart' (default: false). If present and true on any variable, restart postfix instead of reloading it.

No documentation though?

GerMalaz:
Very good point indeed. 
Is this good/enough?

Alphix:
> Very good point indeed. Is this good/enough?

That's a question for @drybjed 😄

PR #2412, opened 2023-09-07
[debops.redis] parent folder needs to be writtable by the redis user

alinalexandru:
`CONFIG REWRITE` falls on newer version of redis if the parent folder is not writable by the redis user.

https://stackoverflow.com/questions/38214269/redis-err-rewriting-config-file-permission-denied-when-using-docker-data-volum


hungpr0:
This makes sense I think. Instance folders should be owned by `redis` user.

alinalexandru:
@drybjed Is there anything left to do for on this PR? Do you have a feedback?

PR #2415, opened 2023-09-11
[apt] Add support for non-free-firmware in Bookworm+

### LINKED ISSUE #2372: apt: non-free-firmware component introduced in bookworm
As discussed in https://github.com/debops/debops/pull/2363 the introduction of the non-free-firmware component in the apt role is not done yet.
I open this bug report as a reminder.

A question remains, that is should the `non-free-firmware` component be added to `apt__distribution_components_free` or `apt__distribution_components_nonfree`? Or in a new variable (since it is non-free but one might want to narrow the scope of  the non-free components to essential packages only).

Alphix:
Closes: #2372


drybjed (approved):
Thanks. I'll merge this and backport to the stable releases, since changes in the APT role are not backportable. I'll fix the linting issues as well.

PR #2417, opened 2023-09-11
[icinga_web] Update package versions and implement module removal

Alphix:
This was enough to make icinga_web work for me after updating to Bookworm. I also took the opportunity to add support for the removal of modules and to make the output from the role during an Ansible run bit saner.


Alphix:
Note, I just realised this needs a thorough rework...consider this a band-aid for now, I'll post a better version soon...

Alphix:
> Note, I just realised this needs a thorough rework...consider this a band-aid for now, I'll post a better version soon...

Updated PR with a more complete solution pushed

PR #2419, opened 2023-09-13
[slapd] Bookworm updates

Alphix:
This PR implements the changes that were necessary for me to migrate from Bullseye to Bookworm.

Mostly it's about missing schemas and renamed attributes.

The AppArmor integration PR is also relevant.


ansible/roles/slapd/defaults/main.yml
```diff
@@ -151,32 +155,26 @@ slapd__combined_schemas: '{{ slapd__default_schemas
...
+# List of APT packages that contain LDAP schemas to be loaded into the
+# directory by the server.
+slapd__schema_packages: []
 
                                                                    # ]]]
 # .. envvar:: slapd__packages [[[
 #
 # List of additional APT packages to install with OpenLDAP service.
 slapd__packages: []
+
```
drybjed: It's just a aesthetic nitpick - last variable in a section shouldn't have an empty line after it and before the fold markers. That way all variables have consistent spacing when you ignore fold markers.

ansible/roles/slapd/files/etc/ldap/schema/debops/ldapns.schema
```diff
@@ -0,0 +1,30 @@
+# SPDX-License-Identifier: GPL-2+
```
drybjed: DebOps uses the `reuse` Python package to check validity of all files with the https://reuse.software/ project. All files require standardized set of Copyright and License information. You can install the `reuse` Python package and run `reuse lint` to see which files are not correctly marked.

The `GPL-2+` string is probably not a correct one, it would probably be `GPL-2.0-or-later`.

It would be great to check the source repositories and find out the author(s) of these files and approximate ranges of when they maintained them.

docs/ansible/roles/slapd/examples/multi-master-replication.yml
```diff
@@ -69,7 +69,7 @@ slapd__cluster_tasks:
              tls_key="{{ slapd__tls_private_key }}"
              tls_cipher_suite="{{ slapd__tls_cipher_suite }}"
 
-      olcMirrorMode: 'TRUE'
+      olcMultiProvider: 'TRUE'
```
drybjed, line 72: Does this also work on older `slapd` releases, ie 2.4 from older Debian versions? Ah, I guess not.  Hmmm... To make things easy for migration, we could duplicate the sections and use conditional state with `exact` and `ignore` for the specific `slapd` version. The version should be available in the `slapd__version` variable. After 1-2 Debian releases, duplicate entries for old versions could be removed.

drybjed (changes requested):
Thanks for making the effort to keep the role updated, really appreciate this. :-) There are still some issues I'd like to see resolved before the merge. I can look into them later if you don't have time for now; I'll try to handle this PR in 1-2 weeks.

ansible/roles/slapd/files/etc/ldap/schema/debops/ldapns.schema
BerkhanBerkdemir: My guess is that `ldapns.schema` is copied directly from [fusiondirectory-schema] package. [the schema] is [listed] as GPL-2+ in the COPYRIGHT file for that Debian package.

  [fusiondirectory-schema]: https://packages.debian.org/buster/all/fusiondirectory-schema/filelist
  [the schema]: https://salsa.debian.org/debian/fusiondirectory/-/blob/master/debian/copyright?ref_type=heads#L1120
  [listed]: https://salsa.debian.org/debian/fusiondirectory/-/blob/master/debian/copyright?ref_type=heads#L1412-1415

drybjed: Yes, but the file still needs to reflect that in a way that `reuse` can parse. That's why we need to change the format from `GPL-2+` to `GPL-2.0-or-later`, and add Copyright information. The reason is, the file now comes from the DebOps repository itself, not from an outside package.

ansible/roles/slapd/defaults/main.yml
Alphix: Should be fixed in the rebase...I'm not sure I like the style though....not having an empty line after each variable increases the risk that the fold markers end up as part of the value of the variable....

docs/ansible/roles/slapd/examples/multi-master-replication.yml
Alphix: @drybjed I would tend to disagree....this is a documentation example...and the patch also adds a notice that the name has changed for older releases....adding complex conditionals just means we'll have to revisit this later to cleanup the conditionals (and we all know how likely that is....)

ansible/roles/slapd/files/etc/ldap/schema/debops/ldapns.schema
Alphix: Ok, I've done some archaeology, dug up some authors, and updated the SPDX headers accordingly...

PR #2421, opened 2023-09-19
[apt] Refresh of the 'debops.apt' role

drybjed:


hungpr0:
Hello @drybjed 

Just wanted to report, with this commit, we have following issues:
- `apt__combined_repositories` requires `item.name` or `'{{ apt__combined_repositories | debops.debops.parse_kv_items }}'` will be empty
```
apt__repositories:
  - repo: 'deb http://ab.com/debian {{ ansible_distribution_release }} main contrib'
    filename: 'main_repo'
'{{ apt__combined_repositories | debops.debops.parse_kv_items }}' returns empty
```

- `Install required packages` task should be present before `Configure custom APT keys` task. We must install `apt__base_packages` before adding apt_key because if `gnupg` isn't installed on host, we can't add apt key.

- If there is conflict (`E: Conflicting values set for option Trusted regarding source`) in 1 entry in `/etc/apt/sources.list` and `/etc/apt/sources.list.d/main_repo.list`, we are stopped at `Install required packages` task (Can't install any packages, error message `E: Conflicting values set for option Trusted regarding source`). The modification of `/etc/apt/sources.list` can never executed. For example:
```
#/etc/apt/sources.list
deb [trusted=yes] http://ab.com/debian bookworm main contrib

#/etc/apt/sources.list.d/main_repo.list
deb http://ab.com/debian bookworm main contrib
```

I think we need to re-order the tasks to workaround those issue.
- run first apt cache update
- install base packages
- add apt key
- modify /etc/sources.list
- modify /etc/sources.list.d/*
- run apt cache update again
- etc...

Thanks

PR #2427, opened 2023-10-02
[nginx] Remove systemd override for nginx.service

Alphix:
Since Debian package version 1.22.0-2 (Bookworm, Mantic), the nginx.service unit has been updated to make "After=network-online.target" the default, making this override unnecessary.


drybjed (approved):
Looks good, thanks! I'll backport this to the stable releases.

PR #2429, opened 2023-10-03
[sshd] sshd authorized_key fixes

### LINKED ISSUE #2384: /etc/ssh/ldap_authorized_keys_bindpw contains line break, authentication fails
Hi,

Auto generated passwords
https://github.com/debops/debops/blob/c6075e3a4351c5915cde3ec935fb1a856929e358/ansible/roles/sshd/defaults/main.yml#L1008-L1011
contain the line break in the very end and while that **should be ok** because jinja2 should take care of removing it via https://github.com/debops/debops/blob/master/ansible/roles/sshd/templates/etc/ssh/ldap_authorized_keys_bindpw.j2 - it fails. Resulting in _ldapsearch_ command saying that provided credentials are invalid (via https://github.com/debops/debops/blob/master/ansible/roles/sshd/templates/etc/ssh/authorized_keys_lookup.d/ldap.j2 )

Is it jinja2 bug?  Stripping the password from line breaks at ansible controller's side still re-introduces it when it's deployed on target VM.

Thanks

Alphix:
The first two commits are bugfixes to the "ldap" lookup, the third commit improves the default handling of the alternate lookup scripts.

Closes: #2384


njc-pdixon:
@Alphix 
Thanks for putting this pull request together. I've just spent longer than I care to admit trying to work out why `sshd` couldn't log in to the LDAP server after a Debops update!

Anyway, it's probably worth updating the `.pre-commit-config.yaml` file so that the `end-of-file-fixer` plugin excludes this file, since that's what introduced this bug (in commit f5e68fa8b59fa4fd1f9a8b91a4fc45579b8f6c70). Looks like it's probably blocking your CI, too.

This seems to do the trick on my machine:
```yaml
      - id: end-of-file-fixer
        exclude: ^ansible/roles/sshd/templates/etc/ssh/ldap_authorized_keys_bindpw.j2$
```

Alphix:
> @njc-pdixon

Thanks, FWIW, I've updated the PR to include your suggested fix...

PR #2434, opened 2023-10-10
[borgbackup] Add new role

### LINKED ISSUE #792: Create a 'debops.borgbackup' role

Alphix:
Starting from @ypid's draft in PR #835 and hacking for a week, I reworked the draft version of the role to properly integrate it with other DebOps roles, making the setup much simpler for end users in the process.

Some notable features:
- Updated (and simplified) docs
- Encrypted passphrases (systemd-creds)
- SPDX compliance
- Group-based host roles (server/client/controller)
- systemd (.timer/.service) and cron integration
- systemd sandboxing
- Modern compression/encryption choices
- Automatic configuration of users/keys/authorized_keys/etc
- Optional "append only" support

I've tested the various modes (client/server and client/server/controller), and it all works in my devel environment.

For now I've kept the PR as two patches (the draft I started from, rebased into one patch, and my changes as a separate patch). I could rebase this into a single patch, but I wanted to make the provenance clear.

Closes: #792
Closes: #835


Alphix:
@ypid: this should interest you 😁

Alphix:
Actually....I think this needs some minor rework. The current template for the `borgmatic` configuration (`ansible/roles/borgbackup/templates/etc/borgmatic/config.yaml.j2`) doesn't properly handle complicated YAML syntax (e.g. `checks:`, which expects a list of dicts)

PR #2435, opened 2023-10-11
[netbox] Add support for standby NetBox on read only database

k-304:
These changes add support for running NetBox on a PSQL cluster with a primary and read-only node (AKA secondary/standby).
Certain tasks must only run on the primary node, and must be skipped on the standby (read-only) node, others only on the standby node.

A new variable named `netbox__primary` of type `boolean` can be added to define primary/standby nodes.
Default is set to `True` for non-HA setups with only 1 NetBox instance.


ypid (approved):
Looks solid to me. Well done!

drybjed (approved):
Looks good, thanks! I'll try to backport these changes to stable releases.

PR #2437, opened 2023-10-15
[networkd] Improve networkd__configuration and networkd__units

Alphix:
The default configuration of systemd-networkd has already changed from Bullseye to Bookworm. It used to look like this in Bullseye:

  [Network]
  #SpeedMeter=no
  #SpeedMeterIntervalSec=10sec
  #ManageForeignRoutes=yes

  [DHCP]
  #DUIDType=vendor
  #DUIDRawData=

And now it looks like this in Bookworm:

  [Network]
  #SpeedMeter=no
  #SpeedMeterIntervalSec=10sec
  #ManageForeignRoutingPolicyRules=yes
  #ManageForeignRoutes=yes
  #RouteTable=

  [DHCPv4]
  #DUIDType=vendor
  #DUIDRawData=

  [DHCPv6]
  #DUIDType=vendor
  #DUIDRawData=

In order to simplify the configuration, and the templates, this patch removes the networkd__dhcp_* variables and lets the networkd__*_configuration variables define the whole content of /etc/systemd/networkd.conf.d/ansible.conf. This should be more future-proof and avoids a proliferation of configuration variables (one set of variables per section).

Furthermore, networkd__units has been changed to also support the typical universal configuration style (in addition to raw support, which has been kept).

Also, note that the behaviour of "x | bool" is not exactly intuitive:
    https://medium.com/opsops/wft-bool-filter-in-ansible-e7e2fd7a148f

With that in mind, I took the liberty of simplifying the shared macro logic a bit by removing the (IMHO confusing) usage of the bool filter.

The role has also been changed to use numbered unit files, which is what the man pages recommend, and is already used in some (but not all) of the examples provided in the role documentation.

Per-unit file permissions support is added (useful for e.g. WireGuard netdevs which will contain private keys).

Finally, the use by other roles is now supported via the usual __dependent__ variable.


drybjed (commented):
Very nice work! I'll probably remove my name from authors in the `macros.j2` file since it's entirely new. :-)

BTW, why not use the [community.general.to_ini](https://docs.ansible.com/ansible/latest/collections/community/general/to_ini_filter.html) filter? EIther it was unavailable at the time, or the formatting is not optimal for our purposes, would be my guess. I'll wait with merging for your answer, since you might want to change the code.

Alphix:
> Very nice work! I'll probably remove my name from authors in the `macros.j2` file since it's entirely new. :-)

I probably based it on some other `j2` file where you were the author...I'm not just randomly adding your name left and right :)
 
> BTW, why not use the [community.general.to_ini](https://docs.ansible.com/ansible/latest/collections/community/general/to_ini_filter.html) filter? EIther it was unavailable at the time, or the formatting is not optimal for our purposes, would be my guess. I'll wait with merging for your answer, since you might want to change the code.

I think it was a combination of not being aware that it exists and being paranoid about the output matching our expectations (for example, making sure that all boolean values end up in a systemd supported format). I'd suggest merging this as-is, and then the macro can be changed later if there's interest...

PR #2440, opened 2023-10-18
Fix pid max 32bits

prahal:
Only ship sysctl pid-max conf with pid-max set to 2^22 to 64bits archs.

First add the is_64bits fact to debops.core.


prahal:
ping?
I still own an armv7 armhf board that requires this fix (Odroid U2).

prahal:
Might be of interest to document this new ansible_local.core.is_64bits in the debops doc.

prahal:
Probaly this is an ansible missing feature, ie ansible missing flag that should be added in ansible not in debops.
So maybe it should not be documented here and pushed upstream. If it ever goes into ansible we could get rid of the Debops core flag workaround.

drybjed (approved):
Nice workaround, good idea with placing the check in the `core` role, so that it can be available elsewhere. The 32-bit or 64-bit check could definitely be in Ansible itself, since listing all 64-bit or 32-bit platforms supported by Debian could be messy and require updates in the future. For now, let's go with this solution instead.

Issue #2370, opened 2023-06-21
ifupdown : bookworm - bridge interfaces are down at boot


prahal:
The bridge interfaces are broken with debops iface@.service on bookworm.

systemd `252.6-1`
ifupdown `0.8.41`

Current iface@.service from debops ifupdown:

- on dhcp server:
```
systemd[1]: Started iface@br1.service - Bring network interface up for br1.
systemd[1]: Stopping iface@br1.service - Bring network interface up for br1...
ifdown[2085]: ifdown: waiting for lock on /run/network/ifstate.br1
sh[2133]: resolvconf: Error: Run lock held by another process for longer than 10 seconds
systemd[1]: iface@br1.service: Main process exited, code=exited, status=1/FAILURE
sh[2249]: Waiting for DAD... Done
sh[2305]: resolvconf: Error: Run lock held by another process for longer than 10 seconds
ifdown[2412]: resolvconf: Error: Run lock held by another process for longer than 10 seconds
ifdown[2473]: resolvconf: Error: Run lock held by another process for longer than 10 seconds
ifdown[2562]: Cannot find device "br1"
ifdown[2566]: Device "br1" does not exist.
ifdown[2570]: Cannot find device "br1"
systemd[1]: iface@br1.service: Failed with result 'exit-code'.
systemd[1]: Stopped iface@br1.service - Bring network interface up for br1.
```

- on dhcp client:
```
systemd[1]: Started iface@br0.service - Bring network interface up for br0.
systemd[1]: Stopping iface@br0.service - Bring network interface up for br0...
ifdown[877]: ifdown: waiting for lock on /run/network/ifstate.br0
dhclient[888]: Internet Systems Consortium DHCP Client 4.4.3-P1
sh[888]: Internet Systems Consortium DHCP Client 4.4.3-P1
sh[888]: Copyright 2004-2022 Internet Systems Consortium.
sh[888]: All rights reserved.
sh[888]: For info, please visit https://www.isc.org/software/dhcp/
dhclient[888]: Copyright 2004-2022 Internet Systems Consortium.
dhclient[888]: All rights reserved.
dhclient[888]: For info, please visit https://www.isc.org/software/dhcp/
dhclient[888]: 
dhclient[888]: Listening on LPF/br0/9a:1a:61:e9:a3:8b
sh[888]: Listening on LPF/br0/9a:1a:61:e9:a3:8b
sh[888]: Sending on   LPF/br0/9a:1a:61:e9:a3:8b
[...]

PR #2441, opened 2023-10-18
RCF: [ifupdown] Fix bridges down at boot with Debian bookworm systemd

prahal:
With Debian bookworm systemd 252.6-1, the bridge is stopped immediately after attempting its startup.
It could be that this newer systemd removed the unexpected behavior that allowed to start the service while its "BindsTo" device does not exists.

Workaround for #2370 .


prahal:
@drybjed Note that I cooked this workaround without a reply to my request in #2370 for you to explain me the idea you told about on reddit
https://www.reddit.com/r/debian/comments/7nw98g/comment/ds57s74/ "stick to ifupdown and [ifup@.service](mailto:ifup@.service) but configure virtual network interfaces as needed in udev beforehand. I'm not sure how this should be done properly, maybe with aid of .device units that define these devices on boot. But this involves more moving parts and I didn't really check if that solution is viable."
I don't think one can create an interface with udev without binding it to a physical interface event (add, change, remove). So to me udev cannot help to create this bridge device before running iface@ service.
We could ask the systemd projects devs how they envision this bridge device creation (or if they have other ideas to cope with this issue).

drybjed (approved):
This is probably a good workaround for current issues, I hope that you tested this. :)

The solution to the issue you mentioned is to create bridge devices in `udev` ourselves, yes. I'm not sure how `ifupdown` developers handle this in newer Debian releases, but in `networkd` it's just a few configuration options (see examples in the `networkd` role documentation). When the `ifupdown` role is refreshed this probably will be taken care of in the way `ifupdown` service does it. In simpler setups, switching to `networkd` is probably an easier way than to use `ifupdown`.

PR #2446, opened 2023-11-06
Flatten directory paths for docker-gen role

BerkhanBerkdemir:
This will resolve the crash that happens whenever docker-gen playbook runs.


drybjed (approved):
Sounds good. This probably worked way back when the role was written and at some point Ansible changed its behaviour. I'll backport this to older releases as well.

PR #2448, opened 2023-11-06
Remove GPG requirement and reconstruct j2 strings

BerkhanBerkdemir:
Miniflux project does not use GPG anymore.  With this, having GPG key in there results in errors.  Maintainer moved APT repository to another URL.

Other than that, Jinja templating wasn't working as expected in three different strings.  Now, they are working as expected in the dev env.

### DIFF
--- a/ansible/roles/miniflux/defaults/main.yml
+++ b/ansible/roles/miniflux/defaults/main.yml
@@ -1,7 +1,7 @@
 ---
-# Copyright (C) 2021-2022 Berkhan Berkdemir <berkhan@tekdanisman.com>
+# Copyright (C) 2021-2023 Berkhan Berkdemir <berkhan@tekdanisman.com>
 # Copyright (C) 2022      Maciej Delmanowski <drybjed@gmail.com>
-# Copyright (C) 2022      DebOps <https://debops.org/>
+# Copyright (C) 2023      DebOps <https://debops.org/>
 # SPDX-License-Identifier: GPL-3.0-only
 
 # .. _miniflux__ref_defaults:
@@ -132,14 +132,6 @@ miniflux__database_password: '{{ lookup("password", secret + "/postgresql/" +
 # installation definition can be found in the
 # :envvar:`miniflux__golang__dependent_packages` variable.
 
-# .. envvar:: miniflux__upstream_gpg_key [[[
-#
-# The fingerprint of the GPG key which is used to sign the Miniflux
-# releases. It will be used to verify the downloaded signature file as
-# well as the :command:`git` tags in the source repository.
-miniflux__upstream_gpg_key: '5A7A 89AC F055 24CA A0F3  6F9B AEB7 A164 0710 8DB5'
-
-                                                                   # ]]]
 # .. envvar:: miniflux__upstream_type [[[
 #
 # Specify the method which should be used to install Miniflux binary.  The
@@ -155,7 +147,7 @@ miniflux__upstream_type: 'apt'
 # .. envvar:: miniflux__upstream_version [[[
 #
 # Version of the Miniflux upstream.
-miniflux__upstream_version: '2.0.35'
+miniflux__upstream_version: '2.0.48'
 
                                                                    # ]]]
 # .. envvar:: miniflux__upstream_url_mirror [[[
@@ -177,7 +169,7 @@ miniflux__upstream_platform: 'linux-amd64'
 #
 # The URL of the upstream :command:`git` repository which contains
 # Miniflux source code.
-miniflux__upstream_git_repository: 'https://github.com/miniflux/miniflux'
+miniflux__upstream_git_repository: 'https://github.com/miniflux/v2'
 
                                                                    # ]]]
 # .. envvar:: miniflux__binary [[[
@@ -253,9 +245,7 @@ miniflux__combined_configuration: '{{ miniflux__default_configuration
 
 miniflux__keyring__dependent_apt_keys:
 
-  - id: '{{ miniflux__upstream_gpg_key }}'
-    url: 'https://apt.miniflux.app/KEY.gpg'
-    repo: 'deb https://apt.miniflux.app/ /'
+  - repo: 'deb [trusted=yes] https://repo.miniflux.app/apt/ /'
     state: '{{ "present"
                if (miniflux__upstream_type == "apt")
                else "absent" }}'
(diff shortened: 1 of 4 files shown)


ansible/roles/miniflux/defaults/main.yml
```diff
@@ -132,14 +132,6 @@ miniflux__database_password: '{{ lookup("password", secret + "/postgresql/" +
 # installation definition can be found in the
 # :envvar:`miniflux__golang__dependent_packages` variable.
 
-# .. envvar:: miniflux__upstream_gpg_key [[[
-#
-# The fingerprint of the GPG key which is used to sign the Miniflux
-# releases. It will be used to verify the downloaded signature file as
-# well as the :command:`git` tags in the source repository.
-miniflux__upstream_gpg_key: '5A7A 89AC F055 24CA A0F3  6F9B AEB7 A164 0710 8DB5'
```
BerkhanBerkdemir, line 140: Miniflux doesn't offer GPG key for their APT packages anymore. They are using `trusted=yes` attribute in the [APT configuration](https://miniflux.app/docs/debian.html).

ansible/roles/miniflux/defaults/main.yml
```diff
@@ -155,7 +147,7 @@ miniflux__upstream_type: 'apt'
 # .. envvar:: miniflux__upstream_version [[[
 #
 # Version of the Miniflux upstream.
-miniflux__upstream_version: '2.0.35'
+miniflux__upstream_version: '2.0.48'
```
BerkhanBerkdemir: It can be bumped to [2.2.1](https://miniflux.app/releases/2.2.1.html) which is what I use it

ansible/roles/miniflux/defaults/main.yml
```diff
@@ -177,7 +169,7 @@ miniflux__upstream_platform: 'linux-amd64'
 #
 # The URL of the upstream :command:`git` repository which contains
 # Miniflux source code.
-miniflux__upstream_git_repository: 'https://github.com/miniflux/miniflux'
+miniflux__upstream_git_repository: 'https://github.com/miniflux/v2'
```
BerkhanBerkdemir, line 171: I don't know why previous code includes `miniflux/miniflux`. It could be a mistake on my end, but current repository name is `v2`.

ansible/roles/miniflux/defaults/main.yml
```diff
@@ -253,9 +245,7 @@ miniflux__combined_configuration: '{{ miniflux__default_configuration
 
 miniflux__keyring__dependent_apt_keys:
 
-  - id: '{{ miniflux__upstream_gpg_key }}'
-    url: 'https://apt.miniflux.app/KEY.gpg'
-    repo: 'deb https://apt.miniflux.app/ /'
+  - repo: 'deb [trusted=yes] https://repo.miniflux.app/apt/ /'
```
BerkhanBerkdemir: As I also mentioned for the variable name, the APT repository URL changed. Here is the [documentation](https://miniflux.app/docs/debian.html) for that.

ansible/roles/miniflux/defaults/main.yml
```diff
@@ -321,21 +311,18 @@ miniflux__golang__dependent_packages:
...
-      - src: 'releases/{{ miniflux__upstream_platform }}/miniflux/miniflux/" + miniflux__upstream_version + "/miniflux_" + miniflux__upstream_version + "_linux_amd64'
+      - src: '{{ "releases/" + miniflux__upstream_platform + "/miniflux/v2/" + miniflux__upstream_version + "/miniflux_" + miniflux__upstream_version + "_" + miniflux__upstream_platform.replace("-", "_") }}'
         dest: 'miniflux'
         notify: [ 'Restart miniflux' ]
     git:
       - repo: '{{ miniflux__upstream_git_repository }}'
         version: '{{ miniflux__upstream_version }}'
         build_script: |
-          make clean linux-amd64
+          'make clean {{ miniflux__upstream_platform }}'
```
BerkhanBerkdemir, line 315: I wanted to use `miniflux__upstream_platform` variable to use the same variable in both ARMv7 (Raspberry Pi 2 Model B) and x86_64 architectures. I didn't try it on other architectures.

BerkhanBerkdemir (commented):
Hi @drybjed! I wanted to add my comments to this PR since it's been a year. This PR has been in the use since it is initially created.

ansible/roles/miniflux/defaults/main.yml
drybjed: Well, that's unfortunate. This breaks the chain of trust for that APT repository; the `miniflux` package could be replaced by anything over the wire and users wouldn't be able to verify the authenticity of the package.

In this case, the APT installation should be removed entirely from the role, so we're left with either URL download or local build based on the source code.

ansible/roles/miniflux/defaults/main.yml
drybjed: Sure, please update the version for the latest one.

ansible/roles/miniflux/defaults/main.yml
drybjed: Looks good.

ansible/roles/miniflux/defaults/main.yml
drybjed: Since w're dropping APT-based installation, you can remove the `keyring` role and its configuration from `miniflux` role dependencies.

ansible/roles/miniflux/defaults/main.yml
drybjed: Looks good. Make sure that you remove the `apt_packages` key here so that the `golang` role won't try to install `miniflux` using APT.

drybjed (changes requested):
Thanks for the update. I wonder if the `miniflux` team gave a reasoning for dropping the GPG-signed APT repository?

Please make changes based on the comments, then I'll be able to merge this PR.

BerkhanBerkdemir:
> I wonder if the `miniflux` team gave a reasoning for dropping the GPG-signed APT repository?

Looks like a good reason...

> I chose the lazy way at the moment, and the new repo does not have a GPG key. The repository URL must be prefixed with `[trusted=yes]`
>
> https://github.com/miniflux/v2/issues/1516#issuecomment-1200684889

BerkhanBerkdemir:
Debian is maintaining their own Miniflux APT package.  It is on Trixie right now. In the future, we might need to enable APT option.

https://packages.debian.org/trixie/miniflux

BerkhanBerkdemir:
Before you review, I also need to update the documentation about changes to APT.

PR #2451, opened 2023-11-12
workaround bag due to docker/network issue - https://bugs.debian.org/…

monah1744:
workaround bag due to docker/network issue

https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=1023656


drybjed (approved):
Thanks. I guess it's a good workaround for now. The `snmpd` role should be refreshed at some point, since newer Debian releases have their own `snmpd.service` files already.

I'll backport the change to stable DebOps releases.

PR #2460, opened 2023-11-30
Avoid an error if ferm rules are empty

julien-lecomte:
Quick fix to skip an error ("not defined") in the ferm role.


drybjed (approved):
OK, this fix is in two separate PRs now, so I'll merge it. This might cause the debugging of undefined variables in the firewall configuration to be harder in the future, but maybe not... I'll backport the fix to the stable releases as well.

PR #2461, opened 2023-12-04
[docker_gen] Fix loop to create required directories

yethee:
This PR fixes regression after converting `with_flattened` to `loop` (at #2141)


BerkhanBerkdemir:
I have been using #2446 for months.  What are the benefits for using `when` clause in here?

yethee:
Sorry, I don't found your PR when faced with this issue.

> What are the benefits for using `when` clause in here?

I didn't know that if set an empty list (`[]`) as the value of an element, it will be automatically skipped.

PR #2463, opened 2023-12-18
Add default value for ferm_parsed_rules

julien-lecomte:
Without this, the play can error out with a "Unable to look up a name or access an attribute in template string"

### DIFF
--- a/ansible/roles/ferm/tasks/main.yml
+++ b/ansible/roles/ferm/tasks/main.yml
@@ -122,7 +122,7 @@
     owner: 'root'
     group: 'adm'
     mode: '0644'
-  loop: '{{ ferm__parsed_rules | dict2items }}'
+  loop: '{{ ferm__parsed_rules | d({}) | dict2items }}'
   loop_control:
     label: '{{ item.key }}'
   register: ferm__register_rules_created


goetzk:
Ah, thanks for opening this PR - I started having this issue recently!

Its not obvious to me why this has started happening (the affected code hadn't changed in some time). I wonder if its a combination of commit dc938e19c24597a02501e29d6cf6f4c703f67d19 and some recent change to ansible.

Anyway, thanks for proposing this fix!

Edit: This affects me on version 3.0.5 of the collection using ansible-core 2.15.4 and python 3.11.2

drybjed:
I'd like to replicate the issue so that I can check what's the underlying problem. Can you give me details about the environment in which this happens? The error message you get from Ansible, for example. Did you set anything `ferm`-related in the Ansible inventory? Check all the files to be sure: `grep -r 'ferm__' ansible/iventory/*`.

julien-lecomte:
We're getting this issue with our inventory with every new VM we're trying to configure. My colleage got it yesterday, and wasn't aware of the issue, and posted this error message on our internal boards:

MicrosoftTeams-image

drybjed:
The messages point out to a variable which is undefined. Debugging this might be difficult via back and forth issue messages. Can we set up a TeamViewer session so that I can see the environment in question and I'm ablle to check the state of various variables? I can do it usually after 16:00 UTC on workdays. Contact me via Matrix/Element DebOps channel.

drybjed:
Fixed via https://github.com/debops/debops/pull/2460, closing.

PR #2475, opened 2024-02-16
Use template_src to overrive of nginx templates

l00ptr:
We use the debops template_src filter to give user the possibility to override default nginx templates.


drybjed (approved):
Looks good, thanks. I try to avoid using this template lookup and instead leverage Universal Configuration system for these things, but at this point in the `nginx` role lifecycle it's probably a better option to use `template_src`. I'll backport this to the stable releases as well.

PR #2476, opened 2024-02-25
Fix logic in `owncloud__do_autosetup`

Cadair:
I am pretty sure this is a logic bug (missing brackets)


l00ptr:
Indeed it looks like a missing parenthesis! nice catch :)

drybjed (approved):
Nice work, thanks. I'll backport this to the stable releases.

PR #2477, opened 2024-03-04
Disable systemd-resolved

### LINKED ISSUE #1778: unbound: Fails to start because systemd-resolved is already running
Running the role `debops.unbound` on an Ubuntu 20.04 fails during the handler:

```
[...]
RUNNING HANDLER [debops.debops.global_handlers : Check unbound configuration and reload] ****************************************************************************************************************************
changed: [dns-resolver-test1]

RUNNING HANDLER [debops.debops.global_handlers : Reload unbound] ****************************************************************************************************************************************************
fatal: [dns-resolver-test1]: FAILED! => {"changed": false, "msg": "Unable to start service unbound: Job for unbound.service failed because the control process exited with error code.\nSee \"systemctl status unbound.service\" and \"journalctl -xe\" for details.\n"}
```

And indeed, Unbound can't bind on ports UDP/TCP 53 because `systemd-resolve` is still running:

```bash
root@dns-resolver-test1:~# ss -natup |grep 53
udp     UNCONN   0        0          127.0.0.53%lo:53            0.0.0.0:*       users:(("systemd-resolve",pid=40179,fd=12))                                    
tcp     LISTEN   0        4096       127.0.0.53%lo:53            0.0.0.0:*       users:(("systemd-resolve",pid=40179,fd=13))                                    

root@dns-resolver-test1:~# journalctl -xe -u unbound
[...]
Jul 23 09:54:35 dns-resolver-test1 unbound[146845]: [1627026875] unbound[146845:0] error: can't bind socket: Address already in use for 0.0.0.0 port 53
Jul 23 09:54:35 dns-resolver-test1 unbound[146845]: [1627026875] unbound[146845:0] fatal error: could not open ports
Jul 23 09:54:35 dns-resolver-test1 systemd[1]: unbound.service: Main process exited, code=exited, status=1/FAILURE
[...]
```

Does I do something wrong? Or does the role should include some way to check port availability, and if they aren't, first stop and disable running process using these ports?

Adding tasks to stop and disable systemd-resolved in a `pre_tasks` section in the playbook is not a good solution because between the execution of these tasks and the role's, the host would fail DNS resolution, which could have dangerous consequences.

PS: Of course, manually stopping and disabling systemd-resolved works:

```bash
root@dns-resolver-test1:~# systemctl stop systemd-resolved.service 
root@dns-resolver-test1:~# systemctl disable systemd-resolved.service 
Removed /etc/systemd/system/dbus-org.freedesktop.resolve1.service.
Removed /etc/systemd/system/multi-user.target.wants/systemd-resolved.service.
root@dns-resolver-test1:~# systemctl start unbound
root@dns-resolver-test1:~# ss -natup |grep 53
udp     UNCONN   0        0                0.0.0.0:53            0.0.0.0:*       users:(("unbound",pid=147942,fd=3))                                            
tcp     LISTEN   0        256              0.0.0.0:53            0.0.0.0:*       users:(("unbound",pid=147942,fd=4))                                            
tcp     LISTEN   0        256            127.0.0.1:8953          0.0.0.0:*       users:(("unbound",pid=147942,fd=8))                                            
```

4ijlnu:
Trying to run both unbound and systemd-resolved on a system, there will be conflicts regarding port usage. Therefore, before installing and configuring unbound, make sure systemd-resolved is stopped and disabled. 

resolves #1778

### DIFF
--- a/ansible/roles/unbound/tasks/main.yml
+++ b/ansible/roles/unbound/tasks/main.yml
@@ -15,6 +15,12 @@
   ansible.builtin.set_fact:
     unbound__fact_ansible_connection: '{{ ansible_connection }}'
 
+- name: Make sure systemd-resolved is stopped and disabled
+  ansible.builtin.service:
+    name: systemd-resolved.service
+    state: stopped
+    enabled: false
+
 - name: Create Unbound configuration directory
   ansible.builtin.file:
     path: '/etc/unbound/unbound.conf.d'


drybjed:
Stopping `systemd-resolved` by the `unbound` role makes sense, but I'm afraid that doing it this early in the role execution might break name resolution in the remote host that is being managed. This can cause issues, like APT not being able to download packages later on when `unbound` is first installed. The `resolved` role should also account for this and not try to restart the service later on.

I'll investigate this issue and try to make a proper solution soon.

prahal:
Maybe we should have global handlers for both unbound and systemd-resolved that stops the other one when it starts (likely restarts are not affected as that imply we are already running the new daemon)

PR #2480, opened 2024-03-11
[onwcloud] Add missing password for postgresql

l00ptr:
When using PostgreSQL as a backend for owncloud / nextcloud the password for authentification against the database is missing. Without that password the occ command to install fails. So we simply add that password.


drybjed (approved):
Thanks! I'll backport this to the stable releases as well.

PR #2481, opened 2024-03-14
[dovecot] Add ability to list users using doveadm

href:
Before this change, dovecot could only check if users existed, but it could not iterate over them. That is, the following would not work

    doveadm user '*'

Nor any command that takes `-A` (all-users), like so:

    doveadm purge -A

With this change, a new default query is added that works like the existing default queries. Instead of searching for a user, it lists them however.


drybjed (approved):
Looks good, thanks. I'll improve the code styling a bit. :) And backport this to the stable releases.

PR #2482, opened 2024-03-15
[grub] setting grub user is broken

pob0111:
The script `/etc/grub.d/01_users` generated from the corresponding template `etc/grub/01_users.j2` is using an unsafe tail command with an absolute number to print out a portion of itself.
Now there has been added additional comments to the head of the template some time ago. The result is that the output of the grub script contains some garbage.
As a result, calling `update-grub` is outputing error messages, e.g.
```
Generating grub configuration file ...
/etc/grub.d/01_users: 10: -exec: not found
/etc/grub.d/01_users: 13: password_pbkdf2: not found
Found linux image: /boot/vmlinuz-6.1.0-21-amd64
Found initrd image: /boot/initrd.img-6.1.0-21-amd64
Warning: os-prober will not be executed to detect other bootable partitions.
Systems on them will not be added to the GRUB boot configuration.
Check GRUB_DISABLE_OS_PROBER documentation entry.
Adding boot menu entry for UEFI Firmware Settings ...
done
```
This leads to an empty user section in `/boot/grub/grub.cfg`:
```
...
### BEGIN /etc/grub.d/01_users ###
### END /etc/grub.d/01_users ###
...
```
This patch fixes this.


pob0111:
I've updated my comment and subject because I realized that it's actually a bug and that the grub user is not set anymore.

drybjed (approved):
Thanks. After looking at the `exec` method of tailing part of the script out... what? I guess it works but it's so convoluted... I hope that this change will be easier to understand in the future.

PR #2483, opened 2024-03-18
[icinga] fix typo in handler condition

lewer:
There's a typo in the `changed_when` condition (see variable defined just a line before).


drybjed (approved):
Thanks! No idea why this doesn't cause Ansible to error when the task is executed... I'll backport this to the stable releases as well.

PR #2485, opened 2024-03-27
Support follow parameter for authorized_keys

gaudenz:
Support the `follow` parameter for the authorized_keys module where it makes sense. This allows having the `authorized_keys` file symlinked. The default behaviour remains unchanged.


drybjed (approved):
Nice work, bonus points for documentation updates. :-) I'll backport these to the stable releases as well.

PR #2488, opened 2024-04-10
[netbox] Improve superuser account creation

k-304:
Creating a superuser is not idempotent. Since it is based on the `migration` output, it may be run when performing NetBox updates (at this time, the user does exist already). The task will fail and stop the update.

This change adds a check if the superuser is present or not. Depending on this (rather than depending on the migration task output), a user will be created or the task skipped.

### DIFF
--- a/ansible/roles/netbox/tasks/main.yml
+++ b/ansible/roles/netbox/tasks/main.yml
@@ -220,6 +220,24 @@
   register: netbox__register_migration
   changed_when: netbox__register_migration.changed | bool
 
+- name: Check if Django superuser account exists
+  # In case of an upgrade (with DB migrations), the superuser will
+  # be existent but the createsuperuser command will run anyway (and fails).
+  # So we check if it exist beforehand and only run the create command if needed.
+  environment:
+    DJANGO_SUPERUSER_PASSWORD: '{{ netbox__superuser_password }}'
+  django_manage:
+    command: shell -c 'import sys; from django.contrib.auth.models import User; print(User.objects.filter(username="{{ netbox__superuser_name }}").count())'
+    app_path: '{{ netbox__git_checkout + "/netbox" }}'
+    virtualenv: '{{ netbox__virtualenv }}'
+  become: True
+  become_user: '{{ netbox__user }}'
+  when: netbox__primary|bool
+  no_log: '{{ debops__no_log | d(True) }}'
+  register: netbox__check_superuser
+  ignore_errors: true
+  changed_when: false
+
 - name: Create Django superuser account
   community.general.django_manage:
     command: 'createsuperuser --noinput --username={{ netbox__superuser_name }} --email={{ netbox__superuser_email }}'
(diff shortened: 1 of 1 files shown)


l00ptr:
your proposition look good to me !

ansible/roles/netbox/tasks/main.yml
```diff
@@ -220,6 +220,24 @@
...
+  django_manage:
+    command: shell -c 'import sys; from django.contrib.auth.models import User; print(User.objects.filter(username="{{ netbox__superuser_name }}").count())'
+    app_path: '{{ netbox__git_checkout + "/netbox" }}'
+    virtualenv: '{{ netbox__virtualenv }}'
+  become: True
+  become_user: '{{ netbox__user }}'
+  when: netbox__primary|bool
+  no_log: '{{ debops__no_log | d(True) }}'
+  register: netbox__check_superuser
+  ignore_errors: true
```
drybjed, line 238: I try not to add `ignore_errors: True` in DebOps code. Any error should be accounted for and handled appropriately.

Did you encounter any errors while testing this task? Does it work when the superuser account doesn't exist yet?

ansible/roles/netbox/tasks/main.yml
```diff
@@ -220,6 +220,24 @@
...
+    DJANGO_SUPERUSER_PASSWORD: '{{ netbox__superuser_password }}'
+  django_manage:
+    command: shell -c 'import sys; from django.contrib.auth.models import User; print(User.objects.filter(username="{{ netbox__superuser_name }}").count())'
+    app_path: '{{ netbox__git_checkout + "/netbox" }}'
+    virtualenv: '{{ netbox__virtualenv }}'
+  become: True
+  become_user: '{{ netbox__user }}'
+  when: netbox__primary|bool
+  no_log: '{{ debops__no_log | d(True) }}'
+  register: netbox__check_superuser
```
drybjed, line 237: Please rename this variable to `netbox__register_check_superuser` for consistency with the rest of the code.

ansible/roles/netbox/tasks/main.yml
k-304: I actually don't recall the reason on why this is required. But I do recall having some trouble with the jango part, as it was not idempotent at all.
Do you want me to check this again and verify if it is required or not?

drybjed: Yes please, if you can. It could be an issue in your environment, but then if we can replicate it, we could try and find the cause.

drybjed: A few changes were merged recently in the `netbox` role, you might want to check them out.

k-304: @drybjed The following commit (https://github.com/debops/debops/commit/a6896f579f83cf2d14cc1d12a1600f41e57fedfc) is actually doing the same but different.  From looking at the code only, I'd say this should resolve the initial issue I tried fixing in this PR.

We can close this one.

PR #2489, opened 2024-04-17
[gitlab] set /etc/gitlab/ssl mode to '0755'

lewer:
The current mode is '0775', but when `gitlab-ctl reconfigure` runs, it sets the mode back to `0755`, which makes the role not idempotent.


drybjed (approved):
Thanks, it makes sense. I don't remember why 0775 here specifcally, but I guess if GitLab still works with the 0750 mask, it shouldn't be an issue. I'll backport this to stable releases as well.

PR #2497, opened 2024-05-15
[gitlab_runner] Fix compatible with GitLab 17.0 or higher

yethee:
Since GitLab 17.0 API server returns `410 Gone` status for the `/api/v4/runners` endpoint. Because the registration token has been [deprecated](https://docs.gitlab.com/ee/update/deprecations.html#registration-tokens-and-server-side-runner-arguments-in-post-apiv4runners-endpoint) and will be removed on GitLab 18.0.

The PR changes the runner registration using new API - [create a runner linked to a user](https://docs.gitlab.com/ee/api/users.html#create-a-runner-linked-to-a-user).
According to GitLab documentation, this should be compatible with GitLab 15.10 or higher.

The new API requires a personal access token, so the `gitlab_runner__api_token` parameter becomes mandatory.

**Second commit**

Removed the task to check GitLab API availability. This guard is useless in cases when the endpoint is overridden in the instance parameters (`api_url`).

### DIFF
--- a/ansible/roles/gitlab_runner/defaults/main.yml
+++ b/ansible/roles/gitlab_runner/defaults/main.yml
@@ -174,7 +174,7 @@ gitlab_runner__api_url: 'https://{{ gitlab_runner__api_fqdn }}/'
 # The personal GitLab API access token used to register or remove the runners in GitLab.
 # You can generate an access token in GitLab "User Settings", "Access Tokens"
 # configuration page.
-gitlab_runner__api_token: ''
+gitlab_runner__api_token: '{{ lookup("env", "GITLAB_API_TOKEN") }}'
 
                                                                    # ]]]
 # .. envvar:: gitlab_runner__runner_type [[[
--- a/ansible/roles/gitlab_runner/defaults/main.yml
+++ b/ansible/roles/gitlab_runner/defaults/main.yml
@@ -171,24 +171,38 @@ gitlab_runner__api_url: 'https://{{ gitlab_runner__api_fqdn }}/'
                                                                    # ]]]
 # .. envvar:: gitlab_runner__api_token [[[
 #
-# The personal GitLab API access token used for API operations, for example
-# removal of existing Runners. This is not a Runner registration token. You can
-# generate an access token in GitLab "User Settings", "Access Tokens"
+# The personal GitLab API access token used to register or remove the runners in GitLab.
+# You can generate an access token in GitLab "User Settings", "Access Tokens"
 # configuration page.
 gitlab_runner__api_token: ''
+
+                                                                   # ]]]
+# .. envvar:: gitlab_runner__runner_type [[[
+#
+# The scope of the runner.
+# It can be one of the following values: ``instance_type``,
+# ``group_type`` or ``project_type``
+gitlab_runner__runner_type: 'instance_type'
+
+                                                                   # ]]]
+# .. envvar:: gitlab_runner__group_id [[[
+#
+# The ID of the group that the runner is created in.
+# Required if ``gitlab_runner__runner_type`` is ``group_type``.
+gitlab_runner__group_id: ~
+
+                                                                   # ]]]
+# .. envvar:: gitlab_runner__project_id [[[
+#
+# The ID of the project that the runner is created in.
+# Required if ``gitlab_runner__runner_type`` is ``project_type``.
+gitlab_runner__project_id: ~
 
                                                                    # ]]]
 # .. envvar:: gitlab_runner__executor [[[
 #
 # Default "executor" used by the runners.
 gitlab_runner__executor: 'shell'
-
-                                                                   # ]]]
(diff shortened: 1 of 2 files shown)


ansible/roles/gitlab_runner/defaults/main.yml
```diff
@@ -171,24 +171,38 @@ gitlab_runner__api_url: 'https://{{ gitlab_runner__api_fqdn }}/'
                                                                    # ]]]
 # .. envvar:: gitlab_runner__api_token [[[
 #
-# The personal GitLab API access token used for API operations, for example
-# removal of existing Runners. This is not a Runner registration token. You can
-# generate an access token in GitLab "User Settings", "Access Tokens"
+# The personal GitLab API access token used to register or remove the runners in GitLab.
+# You can generate an access token in GitLab "User Settings", "Access Tokens"
```
drybjed, line 175: Are the changes compatible with older GitLab versions, or just 17.x and higher?

ansible/roles/gitlab_runner/defaults/main.yml
```diff
@@ -171,24 +171,38 @@ gitlab_runner__api_url: 'https://{{ gitlab_runner__api_fqdn }}/'
...
+# Required if ``gitlab_runner__runner_type`` is ``group_type``.
+gitlab_runner__group_id: ~
 
                                                                    # ]]]
-# .. envvar:: gitlab_runner__token [[[
+# .. envvar:: gitlab_runner__project_id [[[
 #
-# The GitLab CI registration token used to register the runners in GitLab.
-# See :ref:`gitlab_runner__token` for more details.
-gitlab_runner__token: '{{ lookup("env", "GITLAB_RUNNER_TOKEN") }}'
```
drybjed, line 191: It would be useful to have this kind of lookup in `gitlab_runner__api_token` variable. Since runners are per-user, the token can be set in the `.env` file of the DebOps project directory, and then at runtime the variable will be available in the environment.

ansible/roles/gitlab_runner/tasks/main.yml
```diff
@@ -78,32 +78,36 @@
...
+                     + (gitlab_runner__shell_tags
+                        if (item.executor == "shell") else [])
+                     + gitlab_runner__combined_tags)
+                     | unique | join(",") }}'
+      run_untagged: '{{ item.run_untagged | d(gitlab_runner__run_untagged) }}'
+      paused: '{{ item.paused | d(omit) }}'
+      locked: '{{ item.locked | d(omit) }}'
+      access_level: '{{ item.access_level | d(omit) }}'
+      maximum_timeout: '{{ item.maximum_timeout | d(omit) }}'
+      maintenance_note: '{{ item.maintenance_note | d(omit) }}'
```
drybjed, line 103: Are we sure that the `ansible.builtin.url` Ansible module does a correct thing here and removes the parameters with the `omit` value from the request, or is the current value of `omit` passed along the request? If it is, are there any issues in GitLab when this happens?

drybjed (commented):
Thanks for the PR! There are some questions I'd like to get the answers first. Also, if the code is compatible with GitLab 17.0+ only, I might not backport it to the current stable releases.

ansible/roles/gitlab_runner/tasks/main.yml
yethee: I'm using this version to manage the gitlab-runner 17.0+ and it works without issues.

Quick check, I've modify [uri.py](https://github.com/ansible/ansible/blob/v2.16.7/lib/ansible/modules/uri.py) to this:
```patch
+   module.fail_json(msg="Encoded body: %s" % to_native(body))
+
    # Make the request
    start = utcnow()
    r, info = uri(module, url, dest, body, body_format, method,
                  dict_headers, socket_timeout, ca_path, unredirected_headers,
                  decompress, ciphers, use_netrc)
```
Result:
```
msg: 'Encoded body: runner_type=group_type&group_id=887490&description=gl-runner5&tag_list=managed-by-debops&run_untagged=True'
```

ansible/roles/gitlab_runner/defaults/main.yml
yethee: Should be compatible with GitLab >= 15.11.

> In GitLab 15.11 and later, you can use the [POST /user/runners REST API](https://docs.gitlab.com/ee/api/users.html#create-a-runner-linked-to-a-user) to create a runner as an authenticated user.

From: https://docs.gitlab.com/ee/ci/runners/new_creation_workflow.html#creating-runners-programmatically

ansible/roles/gitlab_runner/defaults/main.yml
yethee: Fixed

drybjed (approved):
Thanks for the explanation. :) I'll backport the changes to the stable releases.

PR #2499, opened 2024-06-03
[pam_access] fix issue with unknown variable

pob0111:
When renaming the variable `access_conf` to `pam_access__var_access_conf` one occurrence of it was not renamed.


drybjed (approved):
Thanks! I wonder why Ansible didn't altert about this during testing... I'll backport the fix to stable releases.

PR #2500, opened 2024-06-10
[ntp] Fix ntp__daemon_enabled ansible system capabilities check

urscale:
Since `ansible_system_capabilities_enforced` tends to be a string ("True" / "False") rather than a boolean, it needs to be cast to `bool` in order to successfully set `ntp__daemon_enabled` in environments where `ansible_system_capabilities_enforced` is "False"


drybjed (approved):
Thanks! This should ensure that NTP support is handled correctly.

PR #2501, opened 2024-06-18
Fix ifup-allow-boot.service failing to start

href:
Without this fix, the interface name seems to be empty:

    ifup-allow-boot[786]: Interface floating is down, bringing it up
    sh[787]: Failed to start iface@.service: Unit name iface@.service is missing the instance name.
    sh[787]: See system logs and 'systemctl status iface@.service' for details.
    systemd[1]: ifup-allow-boot.service: Main process exited, code=exited, status=1/FAILURE
    systemd[1]: ifup-allow-boot.service: Failed with result 'exit-code'.
    systemd[1]: Failed to start Bring up non-hotpluggable interfaces on boot.

At least from my tests I'm not sure if this ever worked. In our case, it fails on Ubuntu 20.04 with an interface named `floating`, and IIUC, at least the semicolon is absolutely vital, or `escaped_iface` won't be correct.


prahal:
The proper fix is properly escaping the way systemd except it by double dollars as in https://github.com/debops/debops/pull/2387,  is waiting for review, but should do the same the way systemd expects it.
Please close this less correct duplicate PR.

drybjed:
This issue should be fixed by https://github.com/debops/debops/pull/2387, closing.

PR #2503, opened 2024-06-26
[nginx] Remove /index.html from default try_files

Erethon:
I might be missing something as to why /index.html and /index.htm are tried by default, but this seems wrong to me for most cases. It results in no 404s to ever be returned if any of these two files were present in the root. This patch removes them from the default list of files to try.


drybjed (approved):
After thinking it over, agreed - presence of `/index.html` might cause issues with subdirectories where we might expect missing files to return 404. However, I would still like to keep easy access to "/index.html" file on the URI, so adding `$uri/index.html` should do what's expected. I'll add the change and make backports to stable releases.

Issue #2504, opened 2024-07-01
Some roles not included in ansible galaxy rollout


computerlyrik:
During integration from *contrib, some roles were disabled in the make script deploying the collection to galaxy.

I want to use the apparmor role. It's been moved here but it's still disabled.

Refs 
- https://github.com/debops/debops/commit/a13b76766ee403928eec0b61800e537434b893bf
- https://github.com/debops/debops/pull/2319

drybjed:
The issue was fixed in https://github.com/debops/debops/pull/2505, the `apparmor` role will be available in the next DebOps 3.1.x release on Ansible Galaxy. Sorry for the issue.

PR #2505, opened 2024-07-01
Update make-collection

computerlyrik:
Enabling build of apparmor role

solves  #2504


drybjed (approved):
Thanks! That's on me for adding the `apparmor` role officialy to the project but not including it in the Galaxy collection. I'll try to make a new release soon so that it's included on Galaxy properly.

PR #2506, opened 2024-07-02
fix typo in icinga handler

Cadair:


drybjed (approved):
Thanks for catching that. I guess the handler wasn't triggered when I tested this code. I wonder if `ansible-lint` would be able to catch this typo...

PR #2509, opened 2024-07-20
feat(proc_hidepid): ensure hidepid is disabled when libvirtd is used 

ypid:
The current detection does not work when running the site playbook only once because by the time the `proc_hidepid` role runs, polkit is not yet installed. This is fixed by checking for membership in the `debops_service_libvirtd` Ansible host group. Also check Ansible local facts for good measure.


drybjed (approved):
Adter reading up on issues with Polkit and hidepid=2, it seems that hiding `/proc` information can cause more issues in various places than it solves. The proposed solution in regards to `libvirtd` is find, but I think that the `proc_hidepid` role should be removed from the common playbook and made optional. Then it can be enabled easily via an inventory group in instances where it might be useful, for example jump/bastion hosts.

The removal from the common playbook will be done at a later date after making sure nothing breaks.

PR #2515, opened 2024-08-29
enable override template in sshd role

borrown:
Adjust the src field into the "Generate /etc/pam.d/sshd configuration" task of the role sshd in order to allow to be override by the user


drybjed (approved):

Issue #2410, opened 2023-09-06
[debops] Installation instructions need to be updated


Alphix:
The setup.py file comments:
https://github.com/debops/debops/blob/5abfaafd6ae5dc2c6677d3f30a083c048f113bca/setup.py#L10
indicate that for development, you should run `pip3 install --user --editable .`

Similar instructions are provided [here](https://docs.debops.org/en/master/introduction/install.html).

Since `bookworm`, PEP668 has been implemented, meaning that the above instructions will just give this result:

```
    error: externally-managed-environment
  
    × This environment is externally managed
    ╰─> To install Python packages system-wide, try apt install
        python3-xyz, where xyz is the package you are trying to
        install.
  
        If you wish to install a non-Debian-packaged Python package,
        create a virtual environment using python3 -m venv path/to/venv.
        Then use path/to/venv/bin/python and path/to/venv/bin/pip. Make
        sure you have python3-full installed.
  
        If you wish to install a non-Debian packaged Python application,
        it may be easiest to use pipx install xyz, which will manage a
        virtual environment for you. Make sure you have pipx installed.
  
        See /usr/share/doc/python3.11/README.venv for more information.
  
    note: If you believe this is a mistake, please contact your Python installation or OS distribution provider. You can override this, at the risk of breaking your Python installation or OS, by passing --break-system-packages.
    hint: See PEP 668 for the detailed specification.
```

I know I can use a `venv` (or `pipx`, or `--break-system-packages`), but the instructions should be updated. Not sure which approach you prefer @drybjed?

drybjed:
I checked out `pipx` with Ansible and DebOps. Wow, I like Python virtualenvs now! :-D I'll probably add `pipx` package to the `python` role by default due to this. Thanks!

drybjed:
Fixed by #2517

PR #2517, opened 2024-08-30
Update installation documentation with pipx instead of pip

seven-beep:
As pointed by the issue #2410, the documentation is not applicable for Debian bookworm which implemented PEP668.

This commit provides simple examples to use pipx in place of pip to still satisfy the first method of installation :

"As a Python package in a Python environment of a given UNIX account.
  Different UNIX accounts can use their own versions of DebOps."


drybjed (approved):
Looks great, thanks! I'll backport the changes to the stable releases as well.

PR #2518, opened 2024-09-02
Add separate option enabling to set basedir variable

he0s:
This is necessary in a case, when someone needs to have a custom value for this variable to be able to define custom reprepro repository structure for a repository.

It follows the same approach that is defined for the "outdir" variable, that can be set separately, if a defuault generated value can't be used.


drybjed (approved):
Sounds good. I'll add the documentation and backport the change to the stable release.

he0s:
Wow, thank you! I didn't expect you will prepare the documentation yourself. I appreciate that! :)

PR #2524, opened 2024-09-19
chore(bitcoind): remove role because it is unmaintained

ypid:
I have no interesting in supporting/maintaining bitcoind anymore. Rather than carrying this role over and fixing linting issues, I propose we remove it.


drybjed (approved):
Sounds good. I never used the role, not sure if it was used elsewhere. The role still will be available in current stable releases and if anybody is interested, it can be brought back.

PR #2525, opened 2024-09-19
chore(volkszaehler): remove role, switch to Grafana and InfluxDB

ypid:


drybjed (approved):
Looks good, thanks. I won't backport this PR to existing stable releases, but the role will be removed on the next stable release (v3.3).

PR #2534, opened 2024-10-10
Fixed bug in dist-upgrade.yml

FrankyFire:
In cases where dist_upgrade_register_upgrade is undefined or unchanged, the 'Check if /etc/services' will register a dict without any of the stat return parameters. This leads to 'Assemble /etc/services' to fail with '"dict object" has no attribute "stat"'.

Adding another check in the when statement to see if the previous task was skipped will cause this error to not appear again as it does skip there already.


drybjed (approved):
Nice workaround, thanks!

Issue #2442, opened 2023-10-23
Secret is undefined


Sagebati:
When running the postgresql_server, postgresql roles, or the site playbook, on clean hosts, I get: 

```
fatal: [prep1]: FAILED! => 
  msg: |-
    The task includes an option with an undefined variable. The error was: {{ lookup('password', secret + '/credentials/' + inventory_hostname + '/postgresql/default/' + postgresql_server__user + '/password length=' + postgresql_server__password_length + ' chars=' + postgresql_server__password_characters) }}: {{ (secret__root + "/" + ((secret__levels + "/") if secret__levels else "") + secret__name) | realpath }}: 'secret__root' is undefined. 'secret__root' is undefined. {{ (secret__root + "/" + ((secret__levels + "/") if secret__levels else "") + secret__name) | realpath }}: 'secret__root' is undefined. 'secret__root' is undefined. {{ lookup('password', secret + '/credentials/' + inventory_hostname + '/postgresql/default/' + postgresql_server__user + '/password length=' + postgresql_server__password_length + ' chars=' + postgresql_server__password_characters) }}: {{ (secret__root + "/" + ((secret__levels + "/") if secret__levels else "") + secret__name) | realpath }}: 'secret__root' is undefined. 'secret__root' is undefined. {{ (secret__root + "/" + ((secret__levels + "/") if secret__levels else "") + secret__name) | realpath }}: 'secret__root' is undefined. 'secret__root' is undefined
  
    The error appears to be in '/home/sam/src/infra-debops/ansible/collections/ansible_collections/debops/debops/roles/postgresql_server/tasks/secure_installation.yml': line 6, column 3, but may
    be elsewhere in the file depending on the exact syntax problem.
  
    The offending line appears to be:
  
  
    - name: Update default admin password
      ^ here
```
Seem like the same problem as #2234, the variables of the secret playbook are not defined for whatever reason. 
Also tested this: https://github.com/debops/debops/issues/2234#issuecomment-1333722884 and got the same behavior.

gardouille:
Hi \o

With Ansible 8.0.0 :
```
ansible [core 2.15.5]
  config file = /home/gardouille/.ansible.cfg
  configured module search path = ['/home/gardouille/.ansible/plugins/modules', '/usr/share/ansible/plugins/modules']
  ansible python module location = /home/gardouille/src/debops3-venv/lib/python3.11/site-packages/ansible
[...]

PR #2539, opened 2024-10-11
[ci] Don't pin versions of specific packages in CI

drybjed:
Continuous Integration should use newest versions of packages to check if the code is compatible.


requirements-integration.txt
```diff
@@ -1,28 +1,29 @@
 # Copyright (C) 2023 Berkhan Berkdemir <berkhan@tekdanisman.com>
+# Copyright (C) 2024 Maciej Delmanowski <drybjed@gmail.com>
 # Copyright (C) 2024 DebOps <https://debops.org/>
 # SPDX-License-Identifier: GPL-3.0-or-later
 
 # List of Python packages used by GitHub CI pipeline for continuous integration
 
-ansible==8.0.0a1
```
BerkhanBerkdemir, line 7: This was done because in the past we thought Ansible broke the API. See https://github.com/debops/debops/issues/2442#issuecomment-1787160157

PR #2540, opened 2024-10-12
[docker_gen] Fix create required directories

yethee:
This PR fixes regression of refactoring when `with_flattened` was replaced with `loop`.

Related to #2141

### DIFF
--- a/ansible/roles/docker_gen/tasks/main.yml
+++ b/ansible/roles/docker_gen/tasks/main.yml
@@ -15,10 +15,13 @@
     group: 'root'
     mode: '0755'
   loop:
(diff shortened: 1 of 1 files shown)


BerkhanBerkdemir:
Hi @yethee!

Thank you for fixing the bug that has been there forever. I already had #2446 which has been waiting there for a review.

https://github.com/debops/debops/pull/2446/files#diff-dea0796f831cc0725856369ec9fbb51559fa72e95026da22cf4e1fc60bcdccdbR22

I wanted to keep this just in case if the variable is not defined.

BerkhanBerkdemir:
Looks like you also had #2461.

yethee:
I'm sorry for another duplicate.

Looks like I need to fix this locally in my project before #2446 is merged.

PR #2542, opened 2024-10-18
[icinga] update upstream gpg key id

lewer:
See https://icinga.com/blog/2024/08/26/icinga-package-repository-key-rotation-2024/

The GPG key was changed on September 30, 2024. With the old one, I cannot run `apt update` because of 

> Failed to update apt cache: W:GPG error: https://packages.icinga.com/debian icinga-bookworm InRelease: The following signatures couldn''t be verified because the public key is not available: NO_PUBKEY CC116F55AA7F2382


drybjed (approved):
Looks good, thanks. I'll backport this to stable releases as well.

PR #2546, opened 2024-10-29
feat(apt): add option to disable package purging

jon4hz:
Hi,

This PR adds an option to disable apt purging.
Currently, the role automatically purges packages and the only way to prevent it is to set `apt__enabled: false` which is a bit suboptimal if you want the role to execute an `apt update`.


prahal:
I don't know the use case for this feature.
Though after reading the PR, the task it modifies is about purging apt packages, so I believe the purge flag to apt should be kept to true. To disable the purge of the package, I believe you should use the variable to skip apt purge task altogether. At least it would then be logged as skipped instead of succeeding but not purging anything.

Something like
` when: apt__enabled | bool &&  apt__purge_allowed | bool`

The apt__autoremove_allowed looks like it should be split in another PR as unrelated.

jon4hz:
The thing is that purge and autoremove do two slightly different things, and depending on your setup you might want one or the other. 
If you set `purge: true` and `autoremove: true`, ansible does the equivalent of `apt purge` and if you only set `autoremove: true`, it does the same as `apt autoremove`.

And completely skipping the task seems like a bad idea to me, because then you don't have the possibility anymore to remove certain packages. 

> I don't know the use case for this feature.

The use case is that I want more control over how debops removes packages on my systems. `apt purge` isn't always what people want when removing packages.

jon4hz:
@prahal did that make things clearer?

prahal:
The task is named "name: Purge APT packages if requested". You should not make it autoremove without purging in my opinion. You should write another tasks to handle autoremove without purging if you want that feature.
I strongly believe that if you don't want the task "Purge APT packages if requested" to purge packages if requested, you should disable it based on a flag. Not silently making it not purge based on a global flag. Purging is the purpose of this task.

If you want to only autoremove, you could disable this purge task and execute another autoremove only task. Mangling all features in one task is making it impossible to debug afterwards.

jon4hz:
fair enough, I've updated the PR to only add an option to disable the purge task.

jon4hz:
any chance to get a review for this?

drybjed (approved):
The changes look good, thank you. I'll backport them to the stable releases.

PR #2547, opened 2024-10-30
[opensearch] Minor fixes

haffael:
This fixes two minor issues with `openserach`.

1. The role wants to stop the service `opensearch.service` without before checking if it exists, causing a crash at first install.
**Fixed** by populating service facts beforehand and placing a condition for the stop command.

2. The permissions for directory `/etc/opensearch` are missing the executable bit. Thus, opensearch fails to start with error `/usr/local/share/opensearch/bin/opensearch-env: line 106: cd: /etc/opensearch: Permission denied`.
**Fixed** by changing the permissions from `640` to `750`.

### DIFF
--- a/ansible/roles/opensearch/tasks/main.yml
+++ b/ansible/roles/opensearch/tasks/main.yml
@@ -39,10 +39,14 @@
       become_user: '{{ opensearch__user }}'
       changed_when: False
 
+    - name: Populate service facts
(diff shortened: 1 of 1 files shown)


drybjed (approved):
Thanks for the bugfixes, well done. I'll backport this to the stable releases.

PR #2548, opened 2024-11-03
fix(netbox): match upstream for SESSION_FILE_PATH (`None`, not `False`)

ypid:
Setting `SESSION_FILE_PATH` to `False` has worked for many years but it is not good to derive from upstream config example. We now either set it to `None` or to a string.

Cc: @k-304 as you recently touched those lines #2435 and thus I dug deeper. But the derivation from upstream (`False` and not `None`) was there before you. With your change, it was just changed to "" (empty string). Both `False` and empty string are not ideal I guess (see `settings.py`).

Ref: https://github.com/netbox-community/netbox/blob/27d15615b3590000901984bddfa7a815cd1343d8/netbox/netbox/configuration_example.py#L222
Ref: https://github.com/netbox-community/netbox/blob/27d15615b3590000901984bddfa7a815cd1343d8/netbox/netbox/settings.py#L348-L349

### DIFF
--- a/ansible/roles/netbox/defaults/main.yml
+++ b/ansible/roles/netbox/defaults/main.yml
@@ -828,7 +828,7 @@ netbox__config_metrics_enabled: False
 # can be useful for enabling authentication on a standby instance with
 # read-only database access.) Note that the user as which NetBox runs must have
 # read and write permissions to this path.
-netbox__config_session_file_path: '{{ "" if netbox__primary | bool else netbox__data + "/sessions" }}'
+netbox__config_session_file_path: '{{ "None" if (netbox__primary | bool) else (netbox__data + "/sessions") }}'
 
                                                                    # ]]]
 # .. envvar:: netbox__config_media_root [[[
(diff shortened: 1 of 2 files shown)


k-304:
> Setting `SESSION_FILE_PATH` to `False` has worked for many years but it is not good to derive from upstream config example. We now either set it to `None` or to a string.
> 
> Cc: @k-304 as you recently touched those lines #2435 and thus I dug deeper. But the derivation from upstream (`False` and not `None`) was there before you. With your change, it was just changed to "" (empty string). Both `False` and empty string are not ideal I guess (see `settings.py`).
> 
> Ref: https://github.com/netbox-community/netbox/blob/27d15615b3590000901984bddfa7a815cd1343d8/netbox/netbox/configuration_example.py#L222 Ref: https://github.com/netbox-community/netbox/blob/27d15615b3590000901984bddfa7a815cd1343d8/netbox/netbox/settings.py#L348-L349

Using `None` here is fine if you ask me. This would work as well and is probably better. Just ensure its `None`, not `'None'` (without quotes).

PR #2550, opened 2024-11-04
Remove disutils leftovers for python 3.12 support

prahal:
I switch to the looseversion package and import the strtobool code per PEP-632.
The latter as str2bool does not work for our use case. That is, str2bool does not raise an exception when the string to convert is invalid but returns "False" which breaks "debops project refresh" for all .debops.cfg fields that are not valid boolean strings.

I have been able to run debops on a controller with python 3.12.7 (Debian Testing/Unstable)

### DIFF
--- a/ansible/plugins/lookup/file_src.py
+++ b/ansible/plugins/lookup/file_src.py
@@ -58,7 +58,7 @@ try:
 except ImportError:
     LookupBase = object
 
-from distutils.version import LooseVersion
+from looseversion import LooseVersion
 from ansible import __version__ as __ansible_version__
 
 
--- a/ansible/plugins/lookup/task_src.py
+++ b/ansible/plugins/lookup/task_src.py
@@ -58,7 +58,7 @@ try:
 except ImportError:
     LookupBase = object
 
-from distutils.version import LooseVersion
+from looseversion import LooseVersion
 from ansible import __version__ as __ansible_version__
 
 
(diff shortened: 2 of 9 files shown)


prahal:
The strtobool file is missing a license. disutils.util it was extracted from is MIT https://github.com/pypa/distutils/blob/main/LICENSE, https://github.com/pypa/distutils/blob/main/distutils/util.py.
I don't the policy with code that PEP-632 asks to import into our own with regard to relicensing. Seems obvious to me that not all project would be MIT too. DO you know if they grant us the right to license it under the GPL in debops? Or it is considered obvious code that is not bound by the license?

Edit: turns out disutils was from python before its removal from it https://github.com/python/cpython/blob/3.10/Lib/distutils/util.py,  so the code is PSF which is GPL compatible, ie we could have both code in the same project. But I don't know if we can ship the strtobool code in the same file as the other GPL code, so we might have to sort out if PEP-632 meint we are allowed to reuse distutils code under our own license (as they don't ship this code anywhere anymore and they tell us to copy it to our code).
For now, I will add a PSF license header.

prahal:
I went deeper and took the liberty to include strtobool into src/debops/util.py as at the very least we are not a derivative works of strtobool. I still don't know if the python project allows us to relicense the strtobool code (ie it tells us to copy it into every project that use strtobool and do not provide alternatives, I think they don't expect all these projects to be licensed under the PSF). Maybe it simply requires us to ask them if we can.

prahal:
Obsolete by https://github.com/debops/debops/pull/2577

PR #2555, opened 2024-11-06
Update to nextcloud 33

prahal:
Nextcloud 26 requires PHP 8 (recommands php 8.2 bookworm, php 8.0 is deprecated but still working)  so this PR depends on https://github.com/debops/debops/pull/2416 to be merged before.
Even though this PR does not set the version to 26 on its own so could be applied as is, it advertises support for 26 which itself requires PHP 8.

Nextcloud 32 still supports php 8.1, but it is deprecated. https://docs.nextcloud.com/server/stable/admin_manual/release_notes/upgrade_to_32.html
32 support php 8.4 added by https://github.com/debops/debops/pull/2621 but recommends php 8.3 (but Trixie is php 8.4). Nextcloud 31 is minimum version for Trixie (to support php 8.4).

For nextcloud 33 no changes to debops required. Note that 33 minimum php version is 8.2 (deprecated though, so bookworm support will be soon deprecated).

### DIFF
--- a/ansible/roles/owncloud/defaults/main.yml
+++ b/ansible/roles/owncloud/defaults/main.yml
@@ -320,6 +320,7 @@ owncloud__variant_name_map:
 # Upcoming:
 #
 # * Nextcloud ``25.0`` (Implemented based on documentation changes but untested).
+# * Nextcloud ``26.0`` (Implemented based on documentation changes, no changes from ``25.0``).
 #
 # Unsupported:
 #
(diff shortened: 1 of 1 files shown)


prahal:
https://github.com/debops/debops/pull/2416 is now enough for Nextcloud 33, as it was updated to Trixie support. I closed https://github.com/debops/debops/pull/2621 as a duplicate of https://github.com/debops/debops/pull/2416

PR #2557, opened 2024-11-07
[nginx] Add nginx_events_extra_options for additional directives in events context inside nginx.conf file. Current template nginx.conf.j2 doesn't allow to have additional directlves

hungpr0:
[nginx] Add `nginx_events_extra_options` for additional directives in `events` context inside `nginx.conf` file. Current template `nginx.conf.j2` doesn't allow to have additional directlves. There are only 2 default `worker_connections` and `multi_accept`


drybjed (approved):
Sounds good, thanks. I'll backport this to stable releases.

PR #2561, opened 2024-11-27
Allow resolvconf update during early boot stage

bajo:
On Debian 11 (bullseye) we saw that problem, that the static update was not set at boot time. Restarting the service afterwards showed the expected behavior. 

It seems, that the service is in state `activating` as the `ExecStartPost` script did not finish yet. Thus we are checking from within this script if the script itself is already done (check for `active`), creating a dependency on itself.
This PR adds `activating` to the list of states, from which it can trigger the update of the static resolvconf file. 

Hope this helps.


drybjed (approved):
Nice solution, thank you. I'll backport this to the stable releases as well.

PR #2567, opened 2024-12-16
[rabbitmq_server] Use `to_json` filter for parameter values

dstatyvka:
Make sure that `community.rabbitmq.rabbitmq_parameter` gets a proper JSON as the parameter's value


drybjed (approved):
I think that syntax should work... Only one task needs this? Weird, but OK. Let's merge it, and in the future I will have to add some tests to verify in CI if the role works correctly in cases like this.

Issue #2512, opened 2024-08-05
distutils was removed from Python 3.12 stdlib


sio:
[distutils](https://docs.python.org/3.11/library/distutils.html) was deprecated a while ago and was recently removed from Python 3.12

Debops uses this module in [a number of places](https://github.com/search?q=repo%3Adebops%2Fdebops%20distutils&type=code), which makes us incompatible with current and future Python versions. 

distutils features have been [mostly replaced](https://docs.python.org/3.11/whatsnew/3.10.html#distutils-deprecated) by other Python modules, so it should be possible to migrate without resorting to vendoring the old stdlib version.

BerkhanBerkdemir:
Thanks @sio for the heads up. I am also in the process of working on a test environment with Trixie machines which come with [Python 3.12](https://packages.debian.org/trixie/python3).

@Alphix's #2160 is one of those improvements to use.

fry69:
Current workaround for installing via `pipx`: Run this afterwards ->
```shell
pipx runpip debops install setuptools
```

drybjed:
Thnks, I merged the above PR. There are still some instances to be replaced, so I'm leaving this issue open.

amedee:
I am using `debops.apt_install` (debops version 3.2.4) as a separate role and I am confronted with a `distutils` error: 
```
TASK [debops.debops.apt_install : Pre hooks] ***********************************
task path: /home/runner/.ansible/collections/ansible_collections/debops/debops/roles/apt_install/tasks/main.yml:19
An exception occurred during task execution. To see the full traceback, use -vvv. The error was: ModuleNotFoundError: No module named 'distutils'
fatal: [box.vangasse.eu]: FAILED! => 
  msg: 'Unexpected failure during module execution: No module named ''distutils'''
  stdout: ''
```

I haven't found a workaround so I have removed my dependency on debops.apt_install and I hand-rolled my own role to install packages.

I have found an open PR that seems to fix it but introduces a requirement in Python 3.8.
https://github.com/debops/debops/pull/2577
What work still needs to be done on that PR? I don't know any Python but maybe I can help?

PR #2577, opened 2025-02-27
Compatibility with Python 3.12 / 3.13

mbehrle:


mbehrle:
Probably only part of all necessary fixes, s.a. https://github.com/search?q=repo%3Adebops%2Fdebops%20distutils&type=code

mbehrle:
S.a. https://github.com/debops/debops/issues/2512

mbehrle:
Another MR which uses rather package looseversion instead of packaging: https://github.com/debops/debops/pull/2550

I don't know which one to prefer, but leaving this one open for the time being.

prahal:
I believe using packaging instead of looseversion is fine as this is only used to parse `__ansible_version__` which from https://release-monitoring.org/project/165410/ is always PEP-440 compliant. And it avoids adding a dependency on pypi looseversion.

Your push request adds the python 3.12 and 3.13 to the list of supported ones in setup.py (and also 3.10 and 3.11) than mine https://github.com/debops/debops/pull/2550.

Also, my strtobool implementation uses a MAP. I don't remind why, maybe because I took it from this reimplementation https://github.com/networktocode/netutils/issues/280. That is the cpython implementation https://github.com/python/cpython/blob/3.10/Lib/distutils/util.py#L308 and pypa https://github.com/pypa/distutils/blob/main/distutils/util.py#L337 ones are like yours and don't use MAP.

So I will close my PR.

Thanks for your work.

mbehrle:
> I believe using packaging instead of looseversion is fine as this is only used to parse `__ansible_version__` which from https://release-monitoring.org/project/165410/ is always PEP-440 compliant. And it avoids adding a dependency on pypi looseversion.

Well, it adds a dependency on packaging instead. Which is only available for Python >= 3.8 (nevertheless already [EOL](https://devguide.python.org/versions/)).

> Your push request adds the python 3.12 and 3.13 to the list of supported ones in setup.py (and also 3.10 and 3.11) than mine #2550.

I updated this MR accordingly, adapting also the then supported Python versions.

> So I will close my PR.

Since I don't know the scope of desired Python versions for debops perhaps yours is more adequate...

> Thanks for your work.

And yours as well.

prahal:
> I updated this MR accordingly, adapting also the then supported Python versions.
> 
> > So I will close my PR.
> 
> Since I don't know the scope of desired Python versions for debops perhaps yours is more adequate...

I believe the desired Python version depends on the oldest Debian releases referenced in debops playbooks.

python2 is likely not supported anymore so that would be at least >= 3.4 with the previous python_requires.

v3.1.0 tells "
- Support for end-of-life Debian and Ubuntu releases has been removed from
  Ansible roles included in the DebOps project. The releases dropped are:
  "Debian Wheezy", "Debian Jessie", "Ubuntu Precise Pangolin". The support is
  still available in stable DebOps releases up to v3.0.x if needed.
"

That is debops only references  wheezy and jessie in the Vagrantfile and apt roles. So we might assume that the lowest supported Debian release is likely Stretch for Debian.

The oldest supported release by Debian is bullseye (python 3.9.2) but the Debian LTS goes back to jessie (and even buster is python 3.7.3)
https://www.debian.org/releases/index.html
From https://wiki.debian.org/LTS/Extended Jessie LTS supports ends 30th of June 2025.
So Debian Stretch seems a fair bet.
Which is AFAIK python 3.5.

So you might want to reintroduce a form of above python3.4 like >=3.5 in python_requires.

For Ubuntu, I did not check.

So yes, requiring python 3.8 with packaging might be a no go.
Would you mind reworking your patch set to switch to pypi looseversion? Or maybe wait for a confirmation that indeed Stretch support is required, and then do the switch?
Your PR is more complete than mine.

mbehrle:
> So yes, requiring python 3.8 with packaging might be a no go.
Would you mind reworking your patch set to switch to pypi looseversion? Or maybe wait for a confirmation that indeed Stretch support is required, and then do the switch?

Yes, waiting for confirmation which way to go.

amedee:
Can I help with this PR?
It appears to solve an issue I have.
Although I don't know any Python, but then I'll just learn it.

prahal:
@amedee we are waiting for confirmation whether Debian stretch support is required or not. That is this PR depends on packaging which requires python 3.8, for stretch we should use looseversion from pipy instead.
Maybe we should use looseversion until we are confident stretch support is removed, either way.

But no commit has been done to the repository for a long time, I don't know when this PR even if fixed would be merged.

amedee:
@prahal what needs to be done to get that confirmation and who is to give that confirmation?

Don't worry about getting the PR merged, I'll take care of that. I am very good at communication with maintainers of projects and convincing them to make releases that fix bugs.

drybjed (approved):
Thanks for working on this PR and sorry for coming back here so late. I'm currently working on the next DebOps release and this really helps. Chages seem reasonable, so I'll merge this PR first and then look at the other related PRs to see if there are any modifications that are needed.

PR #2580, opened 2025-03-04
[libvirt] Update package default variables for bookworm

seven-beep:
The default variables of the libvirt roles are preventing its playbook to execute on Bookworm, since there is no "virt-top" package here.  Also "virt-goodies" seems only available in Buster so trying to install it should also result in errors.


drybjed (approved):
Looks good, thanks. I'll backport this to stable releases.

PR #2582, opened 2025-03-16
[libvirtd] Fix nsswitch

seven-beep:
Libvirt nss method will not be reached if there is a return action before them (eg: resolve).  This simple re-order ensures that the libvirt nss methods will be used.


muelli:
LGTM

drybjed (approved):
Sounds good to me, thanks. I'll backport this to the stable releases as well.

PR #2583, opened 2025-03-22
[nginx] Add support http2 directive

yethee:
The patch provides the ability to use the new `http2` directive for nginx version `1.25.1` or higher, instead of the corresponding parameter for `listen` directive. To avoid warnings in logs:

```
nginx: [warn] the "listen ... http2" directive is deprecated, use the "http2" directive instead
```

From nginx [changelog](https://nginx.org/en/CHANGES), for `1.25.1` version:

> Feature: the "http2" directive, which enables HTTP/2 on a per-server basis; the "http2" parameter of the "listen" directive is now deprecated.


drybjed (approved):

PR #2594, opened 2025-05-16
Only create root password secret when necessary

seven-beep:
When root_account__password_update was specifically disabled, a secret password
was still created (but not updated on the remote), leading to confusion.

Moreover, deactivating unnecessary secrets write is an improvement towards
compatibility with systems that does not support easily the secret method (eg:
AWX on containers).


drybjed (approved):
Sounds good, but I'll check the if condition and update it because I think that a `then` might be required here. I'll backport the change to the stable release as well.

Issue #2595, opened 2025-05-16
ubuntu sources have moved to /etc/apt/sources.list.d/ubuntu.sources


seven-beep:
Hello,

debops.depobs.apt fails to handle this change, leading to duplicated sources and warnings on updates.

On #2454 this was acknowledged for some cloud-init images, and the change was reverted upstream. Now it is active for standard server ubuntu 24.0 installations and is aknowledged by others as well (https://github.com/ossec/ossec-hids/issues/2140). 

I have not found an official announcement for this change.

PR #2596, opened 2025-05-16
Template upstream Ubuntu deb822 sources when present

seven-beep:
As stated on issue #2595, Ubuntu may ship deb822 sources, this commit detects
it, divert the file and template it accordingly to usual debops variables.


muelli:
LGTM! Thanks!

PR #2602, opened 2025-06-16
[pki] Remove ACME contacts feature

oxmie:
## Description

As _Let's encrypt_ ended it's _expiration notification_ (see https://letsencrypt.org/2025/01/22/ending-expiration-emails/)   the `contact` settings does not work anymore (see also https://github.com/diafygi/acme-tiny/issues/299 ).

DebOps uses `acme-tiny` to generate new or renewal _Let's encrypt_ certificates. Currently there will be an error thrown and no new certificate is generated.

```
root@host:/etc/pki/realms# cat <DOMAIN>/acme/error.log 
Parsing account key...
Parsing CSR...
Found domains: <DOMAIN>
Getting directory...
Directory found!
Registering account...
Registered!
Traceback (most recent call last):
  File "/usr/bin/acme-tiny", line 33, in <module>
    sys.exit(load_entry_point('acme-tiny==4.1.0', 'console_scripts', 'acme-tiny')())
  File "/usr/lib/python3/dist-packages/acme_tiny.py", line 194, in main
    signed_crt = get_crt(args.account_key, args.csr, args.acme_dir, log=LOGGER, CA=args.ca, disable_check=args.disable_check, directory_url=args.directory_url, contact=args.contact)
  File "/usr/lib/python3/dist-packages/acme_tiny.py", line 116, in get_crt
    log.info("Updated contact details:\n{0}".format("\n".join(account['contact'])))
KeyError: 'contact'
```

## Solution
**1. Remove `acme_contacts`-feature**
   * see this pull request
 
**2. Remove `config['acme_contacts']` from `.../config/realm.conf`**
```
find /etc/pki/realms -type f -path "*/config/realm.conf" -exec sed -i '/acme_contacts/d' {} \;
```
**3. Delete `error.log` containing `KeyError: 'contact'` to allow regeneration of new certificates.**
```
find /etc/pki/realms -type f -path "*/acme/error.log" -exec rm -i {} \;
```
**4. (Optional) Run certificate generation manually**
```sh
cd /etc/pki/realms/;
for realm in *;
do
  echo -n "$realm ... ";
  /usr/local/lib/pki/pki-realm run -n $realm;
  echo "DONE";
done
```

### DIFF
--- a/ansible/roles/pki/defaults/main.yml
+++ b/ansible/roles/pki/defaults/main.yml
@@ -222,14 +222,6 @@ pki_acme_ca_api_map:
   'le-staging-v2': 'https://acme-staging-v02.api.letsencrypt.org/directory'
 
                                                                    # ]]]
-# .. envvar:: pki_acme_contacts [[[
-#
-# List of (mailto:) URLs that the ACME server can use to contact you for issues
-# related to your account. For example, the server may wish to notify you about
-# server-initiated revocation or certificate expiration.
-pki_acme_contacts: [ 'mailto:{{ ansible_local.core.admin_public_email[0] | d("root@" + ansible_domain) }}' ]
-
-                                                                   # ]]]
 # .. envvar:: pki_acme_challenge_dir [[[
 #
 # Directory where the ACME client should store responses to ACME CA challenges.
--- a/ansible/roles/pki/files/usr/local/lib/pki/pki-realm
+++ b/ansible/roles/pki/files/usr/local/lib/pki/pki-realm
@@ -181,8 +181,6 @@ initialize_environment () {
     )
     config["acme_ca_api"]="${acme_ca_api_map[${config['acme_ca']}]}"
 
-    config["acme_contacts"]=""
-
     config["acme_expiration_days"]="30"
 
     # shellcheck disable=SC2154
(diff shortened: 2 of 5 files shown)


muelli:
LGTM! Thanks for your contribution!

prahal:
Why don't point 2 and 4 get done via ansible rules ?
Point 3 is less obvious are there might be other errors in these acme error.log that we might want the admin to see. But they could be moved (could adding an extension to the error.log filename also fix this file from preventing acme-tiny from working?).

alinalexandru:
meantime, an easy hack is to set
`pki_acme_contacts: []`

TwiggyWan:
@drybjed  fixed it already in https://github.com/debops/debops/commit/23c6f2e15b2ac6644b55021aeab1ad80d30f81a8,

fix is a bit different as it relies on installing an up-to-date acme-tiny version instead of removing --contacts.

 can we close this?

drybjed (approved):
I think that removing the support from the PKI role is a good idea, less code to maintain. I'll fix any conflicts and backport the changes to the stable releases since this touches an external service.

drybjed:
No, I was wrong with this decision - Let's Encrypt might not support `contact`, but other ACME registrars might still use it. So the code stays for now. Since upgraded `acme-tiny` fix works, I'll close this PR.

PR #2609, opened 2025-09-03
[ifupdown] Invoke systemctl instead of systemd

href:
The systemd binary is not meant to be invoked directly[1] and is no longer in the PATH in Trixie.

[1]: https://www.freedesktop.org/software/systemd/man/latest/systemd.html


ypid (approved):

drybjed:
This was fixed in a different PR. Closing.

PR #2610, opened 2025-09-04
Update cache if Deb822 source added

Peter33333:
If I add a extra deb822 source, I will need to do `apt update` for sure, 
So the `cache_valid_time` should be default'0', just like other change of source.


drybjed (approved):

PR #2614, opened 2025-10-09
Fixes for Debian 13 installation, preparation for next release

drybjed:


drybjed:
The current DebOps codebase works with Ansible v2.18, and since the CI Pipeline uses Ansible v2.19, there are errors. I'm merging these changes to make new DebOps release that works with Debian 13, work on compatibility with Ansible v2.19 will be done on the next release.

PR #2617, opened 2025-10-26
[postgresql_server] implement autopostgresql systemd timer support

prahal:
Systemd timer support was added in autopostgresql 2.3-1

-------------
autopostgresqlbackup equal and above deb 2.3-1 includes systemd timer/service. I disable the universal timer and ship a versioned timer/service.
I also add an early exit to the daily cron job if the autopostgresqlbackup shipped has the systemd timer.

### DIFF
--- a/ansible/roles/postgresql_server/tasks/manage_autopostgresqlbackup.yml
+++ b/ansible/roles/postgresql_server/tasks/manage_autopostgresqlbackup.yml
@@ -3,6 +3,26 @@
 # Copyright (C) 2015-2017 DebOps <https://debops.org/>
 # SPDX-License-Identifier: GPL-3.0-only
 
+- name: Get autopostgresqlbackup package version
+  environment:
+    LC_MESSAGES: 'C'
+  ansible.builtin.command: dpkg-query -W -f='${Version}\n' autopostgresqlbackup
+  register: postgresql_server__register_autopostgresqlbackup_version
+  changed_when: False
+  failed_when: False
+  check_mode: False
+
+- name: Check if autopostgresqlbackup package ships systemd timer
+  environment:
+    LC_MESSAGES: 'C'
+  ansible.builtin.shell: |
+       HAS_TIMER='{{ postgresql_server__register_autopostgresqlbackup_version | d("0.0") | regex_replace("[^0-9.]", "") is version_compare("2.3", ">=") | bool | int }}'
+       if [ "$HAS_TIMER" -eq 1 ]; then true; else false; fi
+  register: postgresql_server__register_autopostgresqlbackup_has_timer
+  changed_when: False
+  failed_when: False
+  check_mode: False
+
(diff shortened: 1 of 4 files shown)


prahal:
I added a commit with `# shellcheck disable=all` to fix the issue with jinja templating breaking spellcheck,
which depends on shellcheck 0.8+ (bookworm has 0.9 and bullseye 0.7)

ansible/roles/postgresql_server/templates/etc/cron.daily/autopostgresqlbackup.j2
```diff
@@ -6,9 +6,19 @@
...
+
 VERSION="{{ item.version | d(postgresql_server__version) }}"
 NAME="{{ item.name }}"
 
+{% if postgresql_server__register_autopostgresqlbackup_has_timer.rc | d(1) == 0 %}
+if [ -d /run/systemd/system ]; then
+    # Skip in favour of systemd timer.
+    exit 0
+fi
+{% endif %}
```
drybjed, line 20: The Bash script templates cannot include Jinja expressions directly, because they are verified using `shellcheck`. You should define a separate variable in which you can set a value based on a Jinja condition, then in the Bash code check the value and act accordingly.

ansible/roles/postgresql_server/tasks/manage_autopostgresqlbackup.yml
```diff
@@ -49,3 +69,51 @@
...
+    src: 'etc/systemd/system/autopostgresqlbackup.timer.j2'
+    dest: '/etc/systemd/system/autopostgresqlbackup-{{ (item.version | d(postgresql_server__version))
+                                                   | replace(".", "_") }}-{{ item.name }}.timer'
+    owner: 'root'
+    group: 'root'
+    mode: '0644'
+  when: postgresql_server__register_autopostgresqlbackup_has_timer.rc | d(1) == 0
+  loop: '{{ q("flattened", postgresql_server__clusters) }}'
+  notify:
+    - 'Reload systemd daemon'
```
drybjed, line 108: The `systemd` services and timers here could be instanced, so that there are complementary services to the PostgreSQL instances on the host.

ansible/roles/postgresql_server/templates/etc/systemd/system/autopostgresqlbackup.service.j2
```diff
@@ -0,0 +1,17 @@
+{# Copyright (C) 2014-2025 Maciej Delmanowski <drybjed@gmail.com>
```
drybjed, line 1: Please use your own copyright in the PR, don't impersonate me. :) You could also set the year as 2026 for the new files, so we know when they were added to the project.

drybjed (changes requested):
Nice approach to this issue, but there are some things I'd like to see improved.

PR #2621, opened 2025-10-30
Add php trixie

prahal:
This includes #2416 as it depends on it


Alphix:
@prahal I'd already added Trixie to my php branch, but I forgot to push it....I've updated #2416 now, so I think this PR can be closed as a dupe :)

prahal:
@Alphix indeed now a dup. Closing. Thanks

PR #2623, opened 2025-11-14
Fixed lxc_ssh.py deprecated selectors dependency

FrankyFire:
Selectors were deprecated in ansible-core 2.17.0.
See: https://github.com/ansible-collections/community.docker/pull/871


drybjed (approved):
Thanks, I'll backport this to the stable releases. The `community.docker` collection is already a requirement, so that's covered.

PR #2628, opened 2025-12-03
[postgresql_server] Correct the template to also use postgresql_server__version

mbehrle:
Introduced by "Change default syslog ident" and should be handled like other version lookups to avoid
'AnsibleUndefinedVariable: ''dict object'' has no attribute ''version'''


drybjed (approved):
This was deliberate to ensure that the `item.version` is defined in a given PostgreSQL cluster configuration, but... OK, I'll merge this. I'm not sure if other places in the role still don't wrap that default variable though. The PostgreSQL server role will have to be refreshed at some point, so this could be designed a bit better.

I'll backport the change to the stable releases as well, for completeness.

PR #2629, opened 2025-12-03
[logrotate] Add support for systemd timer

seven-beep:
Since Debian Bookworm, logrotate is not implemented via cron but with a systemd
timer.  On previous systems and Ubuntu, the crontab and the systemd timer are
both installed.

While is it not harmful to have multiple schedulers calling logrotate, it is
unnecessary, and it also pull cron as a dependency on systems where it may not
be installed.

This commit shift from cron to a systemd timer starting at Debian>11 and
Ubuntu>24.

### DIFF
--- a/ansible/playbooks/service/logrotate.yml
+++ b/ansible/playbooks/service/logrotate.yml
@@ -17,6 +17,7 @@
 
     - role: cron
       tags: [ 'role::cron', 'skip::cron' ]
+      when: logrotate__scheduler == 'cron'
 
     - role: logrotate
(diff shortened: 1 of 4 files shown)


drybjed (commented):
This is a very interesting solution to make a host `cron`-less, which I plan to do with the `common` playbook. But for that to work we still need to install the `systemd-sysv` Debian package, so that installing `logrotate` will avoid pulling the `cron` dependency. I'm not sure where to put that yet. I'll come back to this PR in the near future though, nice job.

PR #2632, opened 2025-12-26
fix(postgresql_server): missing default version in postgresql.conf.j2

seven-beep:
Hello,

With not version provided in `postgresql_server__cluster_main` (which is the
default), the template fails to render postgresql.conf.

Kind regards.


drybjed:
This is already fixed in other PRs, closing.

PR #2633, opened 2025-12-26
fix(postgresql_server): keyring with no postgresql_server__upstream

seven-beep:
When `postgresql_server__upstream` is false, we should not install the upstream
postgresql apt keys.

Notice that the keyring role seems broken on Trixie as apt-key is deprecated, so this commit actually help to fix the playbook when `postgresql_server__upstream` is false, however without addressing the root cause.


drybjed (commented):
The `keyring` role should be fixed now. The dependent configuration in `postgresql_server` role should skip installing the APT keys already, but I'll check for it to make sure.

Issue #2625, opened 2025-11-14
slapd: Support for Debian 13 Trixie and Ubuntu 24.04


mathieumd:
Package [fusiondirectory-schema](https://packages.debian.org/bullseye/fusiondirectory-schema) is no more available in Debian Trixie.

Role `slapd` therefore fails at task `Install APT packages with rfc2307bis LDAP schema` (in [prepare_rfc2307bis.yml](https://github.com/debops/debops/blob/master/ansible/roles/slapd/tasks/prepare_rfc2307bis.yml)) which is [enabled by default](https://github.com/debops/debops/blob/master/ansible/roles/slapd/defaults/main.yml#L26).

mathieumd:
Looks like `rfc2307bis.schema` is still available in [gosa-schema](https://packages.debian.org/trixie/all/gosa-schema/filelist).

mathieumd:
Same for `fusiondirectory-plugin-sudo-schema`, installed through [slapd__schema_packages](https://github.com/debops/debops/blob/master/ansible/roles/slapd/defaults/main.yml#L170), whose `sudo.schema` is still available in [gosa-plugins-sudo-schema](https://packages.debian.org/trixie/all/gosa-plugins-sudo-schema/filelist).

But unfortunately it does not include `sudoRunAsUser` and `sudoRunAsGroup` attributes, so role should be updated:

```diff
--- a/.collections/ansible_collections/debops/debops/roles/slapd/defaults/main.yml
+++ b/.collections/ansible_collections/debops/debops/roles/slapd/defaults/main.yml
@@ -1155,8 +1155,6 @@ slapd__structure_tasks:
       cn: '%admins'
       description: 'Grant privileged access to UNIX accounts in the "admins" UNIX group'
       sudoUser: '%admins'
-      sudoRunAsUser: 'ALL'
-      sudoRunAsGroup: 'ALL'
       sudoHost: 'ALL'
       sudoCommand: 'ALL'
       sudoOption:
```

mathieumd:
Helper `fusiondirectory-insert-schema` (used in [ldap-load-schema](https://github.com/debops/debops/blob/master/ansible/roles/slapd/files/script/ldap-load-schema#L54)) is no more available too. Replacing by `schema2ldif` looks to be working:

```yaml
slapd__schema_packages:
  # Replace fusiondirectory-plugin-sudo-schema
  - 'gosa-plugins-sudo-schema'
[...]

PR #2635, opened 2025-12-30
slapd: Support for Debian 13 Trixie

tblein:
Adaptation of the slapd role to take into accounts the changes introduced in Trixie:

- inclusion of the FusionDirectory schema in the role, since the Debian packages are no more available in Trixie
- compilation of slapd against OpenSSL instead of GnuTLS
- memberOf should not be set since it is a dynamic value
- Since OpenLDAP 2.6 `olcLastBindPrecision` is required to be configured on its backend entry instead and the overlay

Thanks to @mathieumd for the initial trouble shouting of the role for Trixie. Some test and adaptation might be necessary for Ubuntu 24.04 (https://github.com/debops/debops/issues/2625)

### DIFF
--- a/ansible/roles/slapd/defaults/main.yml
+++ b/ansible/roles/slapd/defaults/main.yml
@@ -427,10 +427,13 @@ slapd__tls_private_key: '{{ slapd__pki_path + "/" + slapd__pki_realm + "/" + sla
                                                                    # ]]]
 # .. envvar:: slapd__tls_cipher_suite [[[
 #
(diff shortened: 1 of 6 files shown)


Alphix:
@tblein see also https://github.com/debops/debops/pull/2419

tblein:
> @tblein see also #2419

I missed it, thanks a lot.
Indeed, a lot of my points were already included in your update for Bookworm.
I will modify my PR to include only the updated required for Trixie based on your PR (certainly close and create a new one). I will let it as it is until the refactoring is done.

Issue #2624, opened 2025-11-14
[bootstrap] CallbackModule' object has no attribute 'disabled'


FrankyFire:
Hi!
I just made a fresh install of debops according to documentation using pipx with a global install.
Afterwards I initialized a new project with 
`debops project init`
I added the ansible controller itself to the inventory file and trying to bootstrap.

```
$: sudo debops run bootstrap -e 'ansible_user=user' -k
Executing Ansible playbooks:
/opt/pipx/venvs/debops/lib/python3.11/site-packages/debops/_data/ansible/collections/ansible_collections/debops/debops/playbooks/bootstrap.yml
[WARNING]: Deprecation warnings can be disabled by setting `deprecation_warnings=False` in ansible.cfg.
[DEPRECATION WARNING]: DEFAULT_MANAGED_STR option. Reason: The `ansible_managed` variable can be set just like any other variable, or a different variable can be used.
Alternatives: Set the `ansible_managed` variable, or use any custom variable in templates. This feature will be removed from ansible-core version 2.23.

SSH password:
[WARNING]: Skipping plugin (/opt/pipx/venvs/debops/lib/python3.11/site-packages/debops/_data/ansible/collections/ansible_collections/debops/debops/plugins/connection/lxc_ssh.py), cannot load: cannot import name 'selectors' from 'ansible.compat' (/opt/pipx/venvs/ansible/lib/python3.11/site-packages/ansible/compat/__init__.py)
[DEPRECATION WARNING]: community.general.yaml has been deprecated. The plugin has been superseded by the option `result_format=yaml` in callback plugin ansible.builtin.default from ansible-core 2.13 onwards. This feature will be removed from collection 'community.general' version 12.0.0.
[DEPRECATION WARNING]: The 'debops.debops.profile_tasks' callback plugin implements the following deprecated method(s): playbook_on_stats, playbook_on_task_start. This feature will be removed from the callback plugin API in ansible-core version 2.23. Implement the `v2_*` equivalent callback method(s) instead.
[ERROR]: Unexpected Exception, this is probably a bug: 'CallbackModule' object has no attribute 'disabled'

Traceback (most recent call last):
  File "/opt/pipx/venvs/ansible/lib/python3.11/site-packages/ansible/cli/__init__.py", line 660, in cli_executor
    exit_code = cli.run()
                ^^^^^^^^^
  File "/opt/pipx/venvs/ansible/lib/python3.11/site-packages/ansible/cli/playbook.py", line 153, in run
[...]

PR #2636, opened 2025-12-30
Hotfix callbacks

ekin-oener-provsn:
fixing #2624 and updating callback methods to v2


drybjed:
I decided to remove the `profile_tasks.py` callback plugin from the project; users should use the upstream version, `ansible.posix.profile_tasks` plugin.

PR #2637, opened 2025-12-31
pki: adds the fullchain to the acme directory

tblein:
Add the fullchain in the acme pki directory so it can be symlinked and used for OCSP stapling verification.
Since OCSP is activated by default in nginx, its absence preven nginx start.


drybjed:
The `fullchain.pem` file actually includes both the end-entity X.509 certificate and its CA chain. I want to avoid setting it as `root.pem` for OCSP sptapling because that would defeat the purpose of the client providing the end-entity certificate and intermediates for the server to validate. Also, adding `acme/root.pem` this way would make the `pki-realm` script duplicate the chain by combining `acme/intermediate.pem` and `acme/root.pem` into `public/intermediate_root.pem` which is meant for cases like OCSP stapling.

I'll try to find a good way to handle this in the future, perhaps by symlinking the known Let's Encrypt certificates from system CA store.

PR #2639, opened 2026-01-02
[pki] Make sure changes to realm configuration are applied

TwiggyWan:
command.args.creates causes task "Initialize PKI realms" to be skipped if realm.conf already exists.

This means if user changes a pki_ variable (for instance pki_acme_contacts) for an existing realm, change would not be applied until user removes config/realm.conf to force task re-execution.

Solution: always re-execute "Initialize PKI realms" and rely on pki-realm for idempotency.

### DIFF
--- a/ansible/roles/pki/tasks/main.yml
+++ b/ansible/roles/pki/tasks/main.yml
@@ -185,8 +185,7 @@
     --dhparam "{{ (item.dhparam | d(pki_dhparam)) | bool | lower }}"
     --dhparam-file "{{ item.dhparam_file | d(pki_dhparam_file) }}"
     --selfsigned-sign-days "{{ item.selfsigned_sign_days | d('365') }}"
-  args:
-    creates: '{{ pki_root + "/realms/" + item.name + "/config/realm.conf" }}'
+
   loop: '{{ q("flattened", pki_realms
                            + pki_group_realms
                            + pki_host_realms


drybjed:
This change would cause the task and the whole `pki` role to not be idempotent. Any changes in the PKI realm configuration that affect the key would require regeneration of already created keys; the same goes with X.509 certificates. Unfortunately at this point, the role is not designed to handle these cases correctly and the suggested method to apply any change is to remove the entire PKI realm on a given host(s) and re-create them from scratch.

There might be cases of changing realm options without affecting the certificates/keys as you mention. Maybe the realm configuration file could be re-generated similarly to the environment file, could solve the problem you encountered. I'll try to look into it in the future.

TwiggyWan:
> This change would cause the task and the whole `pki` role to not be idempotent. Any changes in the PKI realm configuration that affect the key would require regeneration of already created keys; the same goes with X.509 certificates. Unfortunately at this point, the role is not designed to handle these cases correctly and the suggested method to apply any change is to remove the entire PKI realm on a given host(s) and re-create them from scratch.
> 
> There might be cases of changing realm options without affecting the certificates/keys as you mention. Maybe the realm configuration file could be re-generated similarly to the environment file, could solve the problem you encountered. I'll try to look into it in the future.

Hello @drybjed , I can see several options that I could contribute to, which one would you rather have?

Option 1: Improve the documentation; as of today, I believe nowhere is it stated how to behave when one wants to change the realm configuration. We could add a small page here: 
image
In addition to that, we could document in https://docs.debops.org/en/stable-3.3/ansible/roles/pki/defaults/main.html#configuration-of-pki-realms (not sure its the right place) that any change needs the whole realm to be re-created

Option 2: Split the realm configuration file into two parts depending on whether a change needs the realm to be recreated, then hook that to  `pki-realm` 's `enter_realm()`

If none of them suit you and you eventually have a better idea feel free to @ me

drybjed:
Documenting this is a good idea, sure. You picked a good section to put this in.

As for splitting the config file - currently the `config/realm.conf` file is generated by the `pki-realm` script itself. hence all those options in the command task. Perhaps an easier solution for now would be creation of a `config/realm.local.conf` file by the role after the PKI realm is created, and sourcing it after the default config is sourced - that way any variable definitions in the local config file will override those defined by `pki-realm`. The options could be passed per-realm using a new `config` parameter, which can be a YAML dictionary with `key=value` options. That way if you want to change something, the role re-generates the file and on the next execution of `pki-realm` the new values should be picked up. Want to try implementing that?

TwiggyWan:
I can try to do both documentation and splitting the config file. Is it fine if I re-use this pull request?

drybjed:
Sure, just remember to restore the removed lines.

Issue #2641, opened 2026-01-29
make debops.users append configurable


TwiggyWan:
Hi, in debops.users: https://github.com/debops/debops/blob/049a8638f55ac8689c67fa46a148e4f3e5dc29b2/ansible/roles/users/tasks/main.yml#L111

```
- name: Manage additional UNIX groups for UNIX accounts
  ansible.builtin.user:
    groups:       [...]
    append:       '{{ item.append | d(True) }}'
```

it is not possible to set a global setting for the 'append' value.

as a result, if I want append:false to be the default behavior, I must add it to each user block.
It is very easy to forget a  user, given that they can be spread in any host_vars or group_vars directory.

This can lead to drift between what is written in the configuration, and what is really on the hosts.

What I would like instead is to make that configurable so that I could set a default value in group_vars/all.

I did not find a way to do that through ansible.cfg, hence this issue.

TwiggyWan:
hello @drybjed , since you merged  https://github.com/debops/debops/pull/2733, I think I can close this ticket as resolved.

PR #2642, opened 2026-01-29
[users] Allow user to configure default group append behavior.

TwiggyWan:
Because users configuration can span many group_vars/ and hosts_vars/, current role forces one to set 'append' for each user if one wants to always set 'append' to 'false'.

This is not practical with users configuration spread in many group_vars/ and hosts/vars/.

The solution is to give the user the ability to set a default value through a global default variable.
It is set to 'True' by default, therefore will not alter existing inventories.

See #2641


TwiggyWan:
please advise on variable name.

drybjed (approved):
Good idea, but I wanted to have a more namespaced variable name, `users__group_*` already exists and could be confusing. I also added corresponding variable in the `system_users` role for completeness. The code will be merged in a separate PR.

Issue #2620, opened 2025-10-29
[sshd] Wrong max version detected


remi31000:
When using *sshd__kex_algorithms_map* with OpenSSH 10.0, Debops sshd role select the wrong version : 

```
sshd__kex_algorithms_map:

  '10.0': [ 'mlkem768x25519-sha256', 'sntrup761x25519-sha512',
           'sntrup761x25519-sha512@openssh.com' ]

  '9.0': [ 'sntrup761x25519-sha512', 'sntrup761x25519-sha512@openssh.com' ]

  # Source: https://wiki.mozilla.org/Security/Guidelines/OpenSSH
  '6.5': [ 'curve25519-sha256@libssh.org', 'ecdh-sha2-nistp521',
           'ecdh-sha2-nistp384', 'ecdh-sha2-nistp256',
           'diffie-hellman-group-exchange-sha256' ]

  # Source: https://xivilization.net/~marek/blog/2015/01/12/secure-secure-shell-on-debian-wheezy/
  '6.0': [ 'diffie-hellman-group-exchange-sha256' ]
```

Debug : 

```
   - name: 'KexAlgorithms'
    comment: 'List of allowed key exchange algorithms'
    raw: |
      {% set sshd__tpl_kex_algorithms_max_version = sshd__kex_algorithms_map.keys() | select('version_compare', sshd__version, '<=') | max %}
      {% set sshd__tpl_kex_algorithms = sshd__kex_algorithms_map[sshd__tpl_kex_algorithms_max_version] %}
      {% set sshd__tpl_kex_algorithms = (sshd__tpl_kex_algorithms + sshd__kex_algorithms_additional) | unique %}
      {% if sshd__tpl_kex_algorithms and sshd__paranoid | bool %}
      {{ 'KexAlgorithms {}'.format(([sshd__tpl_kex_algorithms | first] + sshd__kex_algorithms_additional) | unique | join(",")) }}
      {% elif sshd__tpl_kex_algorithms %}
      # SSH Version : {{ sshd__version }}
      # Configured sshd__kex_algorithms_map : {{ sshd__kex_algorithms_map.keys() | select('version_compare', sshd__version, '<=') }}
      # Using sshd__kex_algorithms_map : {{ sshd__kex_algorithms_map.keys() | select('version_compare', sshd__version, '<=') | max }}
[...]

PR #2644, opened 2026-02-05
Correctly detect max sshd version from kex_algorithms_map

bits:
When sshd__kex_algorithms_map versions list contains 9.0 and 10.0, `|max` returns 9.0 instead of 10.0 because of string sorting.

To get the largest version number reliably, cast each version number string to float, take the max, and re-cast to string: `… | map("float") | max | string`

This fix is applied to the three blocks for `Ciphers`, `KexAlgorithms`, and `MACs`


bits:
Fixes #2620

drybjed (approved):

Issue #2601, opened 2025-06-10
ansible 12 / ansible-core 2.19 support


zerwes:
As debian trixie comes with ansible 12 / ansible-core 2.19, supporting this version would be great.

https://docs.ansible.com/ansible/devel/porting_guides/porting_guide_12.html

There are several breaking changes, like:
 * broken conditionals (annoying, but easy to fix)
 * template trust model
 * jinja2 native templating mode
 * ...

We use a small subset of the debops roles, but currently they will not run with ansible 12

kosli:
is there an update regards the debian trixie support? It was [released on August 9th 2025](https://www.debian.org/releases/trixie/),

jon4hz:
According to the [release and maintenance](https://docs.ansible.com/projects/ansible/latest/reference_appendices/release_and_maintenance.html) page from ansible, `2.18.` will be EOL in May 2026.

It would be nice if there was some comment from the maintainers if they plan to support `2.19` any time soon.

scibi:
I prepared some PR: #2653, #2654, #2655, #2656, #2657, #2658, #2659, #2660.

debuglevel:
For anyone who is stuck: This
```
  - name: https://github.com/debops/debops.git
    type: git
    version: stable-3.3
```
might help (as this includes already lots of fixes) in the meantime, until `3.3.1` is released.

Also, some roles (like `etckeeper` in `3.3.0`) fail only on their first run, but succeed on subsequent runs (as `Unstage and remove ignored files from Git index` is then skipped).

PR #2645, opened 2026-02-09
Fix changed_when syntax return sometimes None which is no a boolean

Tr4sK:
Regex_search return None when not found which trigger warning

```
regex_search('.+ set to manually installed\.')
```

```
[DEPRECATION WARNING]: Conditional result at location
/home/tr4sk/scripts/tiersbox/srv/provisioning/venv/lib/python3.14/site-packages/debops/_data
/ansible/collections/ansible_collections/debops/debops/roles/python/tasks/main_raw.yml 33:17
 was of type 'NoneType'. Conditional results should only be True or False. The result was
interpreted as False. This feature will be removed in version 2.19. Deprecation warnings can
 be disabled by setting deprecation_warnings=False in ansible.cfg.```

### DIFF
--- a/ansible/roles/python/tasks/main_raw.yml
+++ b/ansible/roles/python/tasks/main_raw.yml
@@ -30,9 +30,9 @@
     LANG=C apt-get --no-install-recommends -yq install {{ python__core_packages | join(" ") }}
   register: python__register_raw
   when: python__enabled | bool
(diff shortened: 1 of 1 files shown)


Tr4sK:
Bump : https://github.com/debops/debops/issues/2601

PR #2654, opened 2026-03-04
fix(python): use explicit boolean in when conditional to avoid AnsibleUnicode deprecation

scibi:
Replace `python__raw_etc_hosts | d()` with a condition that evaluates to True/False: `(python__raw_etc_hosts | default('')) | length > 0`. Fixes deprecation warning about conditional results being `AnsibleUnicode` instead of `bool` (Ansible 2.19+).


scibi:
Superseded by @drybjed's commit 7ec2fcb

PR #2656, opened 2026-03-05
fix(system roles): ensure when conditionals return bool (Ansible 2.19)

scibi:

Affected roles: apt_install, debconf, grub, locales, sysctl, sysfs, mount, resources, postfix, sysnews


drybjed:
The pending fixes here were fixed by other pull requests using `is truthy` and `is falsy` tests. Closing.

PR #2657, opened 2026-03-05
fix(networking roles): ensure when conditionals return bool (Ansible 2.19)

scibi:

Affected roles: avahi, dnsmasq, nfs, nfs_server, unbound, tcpwrappers, iscsi, tinc


drybjed:
The proposed fixes from this PR were fixed via other PRs using `is truthy` and `is falsy` tests. Closing.

PR #2658, opened 2026-03-05
fix(web/apps/languages roles): ensure when conditionals return bool (Ansible 2.19)

scibi:

Affected roles: nginx, php, gunicorn, owncloud, foodsoft, ruby, golang, java, kodi


drybjed:
All issues being fixed in this PR were fixed using `is truthy` and `is falsy` tests instead. Closing.

PR #2659, opened 2026-03-05
fix(database/messaging roles): ensure when conditionals return bool (Ansible 2.19)

scibi:

Affected roles: postgresql_server, mariadb_server, redis_server, redis_sentinel, rabbitmq_server, elasticsearch, kibana


drybjed:
Issues being fixed in this PR were fixed using `is truthy` and `is falsy` tests in another PR. Closing.

PR #2666, opened 2026-03-21
slapd: Support for Debian 13 Trixie

tblein:
Based on the improvement already set in #2419. Therefore, only specificity of Trixie:
- compilation of slapd against OpenSSL instead of GnuTLS
- Since OpenLDAP 2.6 olcLastBindPrecision is required to be configured on its backend entry instead and the overlay

Thanks to @mathieumd for the initial trouble shouting of the role for Trixie. Almost everything was done in Bookworm by @Alphix.

Some test and adaptation might be necessary for Ubuntu 24.04 (https://github.com/debops/debops/issues/2625) .
Superseed #2635 .


Alphix:
Perfect, I was planning to upgrade my OpenLDAP server soon (might have to wait for the Easter holidays) and these two simple patches on top of what I already did for Bookworm makes the Trixie upgrade so much easier to review. Thank you :)

drybjed (approved):

PR #2671, opened 2026-03-28
Improve shellcheck jinja disabling

prahal:
Embed shellcheck disabling into jinja templated shell scripts instead of blacklisting them one by one in the loop.

---
Edit : removed shellcheck disabling for autopostgresqlbackup as this commit is already in PR #2617 

This might be a no go, as this requires shellcheck >= 0.8 included only starting from bookworm.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -65,6 +65,10 @@ General
   a certificate request. This fixes potential issues with the
   :command:`certbot` command setting wrong certificate name and signature.
 
+- In the :command:`pki-realm` script, ensure that certain :command:`certbot`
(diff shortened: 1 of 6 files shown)


prahal:
This might be a no go, as this requires shellcheck >= 0.8 included only starting from bookworm.

drybjed:
The fixed in the `pki-realm` script are now in the project (https://github.com/debops/debops/commit/d3e2b2eacba5e289ba081f1905e7a44575caa3a0).

The Jinja in Bash scripts should be wrapped in variables. The scripts being ignored in the test script makes the issue more prominent, so I would leave that as-is for now.

PR #2672, opened 2026-03-28
Draft: fix(nsswitch): allow systemd to shadow and gshadow to support dynamic users

prahal:
Gnome gdm v49 requires nss systemd for passwd, group, shadow and gshadow.

I believe a check ofr libnss-systemd to be installed is missing in this PR. But I cannot fix this for now. So marking as draft.


drybjed:
Fixed by https://github.com/debops/debops/pull/2692, closing

PR #2687, opened 2026-04-24
rabbitmq_server: required/opt-in feature flags and automatic cluster formation

scibi:
## Summary
This PR teaches `debops.rabbitmq_server` to work cleanly with RabbitMQ 4.x
and to form a cluster on its own, without changing anything for existing
users who do not opt in. It bundles four related commits on top of the
existing rolling-restart work already in `master`:
1. **Skip "required" feature flags idempotently** – stop failing the play
   when an inventory references a flag that has been promoted to
   `required` status (e.g. `detailed_queues_endpoint` on RabbitMQ 4.x).
2. **Opt-in feature flags** – new `opt_in: True` key on
   `rabbitmq_server__*_feature_flags` entries, enabling flags that the
   broker refuses to toggle without `--opt-in` (e.g. `khepri_db` on
   RabbitMQ 4.0/4.1).
3. **Automatic cluster formation (opt-in)** – non-seed nodes `reset` and
   `join_cluster` against a deterministic seed, guarded by a sanity check
   that never resets an already-clustered node. Disabled by default;
   operators enable it explicitly per inventory group.
No behavioural changes for users on RabbitMQ 3.x, users who keep the
previous defaults for feature flags, or users who bootstrap their
RabbitMQ cluster manually.
## Motivation
- On a fresh RabbitMQ 4.x install the role fails with
  `feature flag "detailed_queues_endpoint" is not available` because that
  flag is now `required` and no longer listed by
  `rabbitmqctl list_feature_flags`, while the
  `community.rabbitmq.rabbitmq_feature_flag` module still tries to enable
  it. Keeping such entries in the inventory is legitimate (they are used
  when upgrading from 3.13) and the role should tolerate them after the
  upgrade.
- `khepri_db` on 4.0/4.1 cannot be enabled via the Ansible module at all
  (the broker requires `--opt-in`), so operators had to drop down to
  shell commands and manage idempotency themselves.
- Even with `cluster_nodes` configured, `peer_discovery_classic_config`
  only acts on a completely fresh Erlang node, so the role used to leave
  existing single-node installs permanently unclustered.
### 1. Skip required feature flags (`a9f24ff7d`)
- New pre-task `Get list of toggleable RabbitMQ feature flags` runs
  `rabbitmqctl -q list_feature_flags name state` once.
- `Manage RabbitMQ feature flags` now loops only over names that are
  actually reported by `rabbitmqctl`, so flags that were promoted to
  `required` (and are therefore always on) are silently skipped.
- Typos in still-toggleable flag names continue to fail inside the
  module, preserving the previous safety net.
### 2. Opt-in feature flags (`e8fa6726f` + `1bbba879b`)
- New `opt_in` key on `rabbitmq_server__*_feature_flags` entries.
- New task `Manage RabbitMQ feature flags (opt-in)` uses
  `rabbitmqctl -q enable_feature_flag --opt-in <name>` for those entries.
- Idempotency is derived from the feature-flag list captured above,
  because `rabbitmqctl enable_feature_flag` prints nothing to either
  stdout or stderr for a no-op run on RabbitMQ 4.x. The task therefore
  skips entirely when the flag is already `enabled`, instead of relying
  on the unreliable `'already enabled' not in stdout` match that used to
  report `changed` on every run.
- The standard `community.rabbitmq.rabbitmq_feature_flag` task excludes
  `opt_in: True` entries so the two paths do not fight each other.
Example inventory:
```yaml
rabbitmq_server__feature_flags:
  - name: 'khepri_db'
    opt_in: True
```
### 3. Automatic cluster formation, opt-in (`e8fa6726f` + `23d6070cd`)
- New variables in `defaults/main.yml`:
  - `rabbitmq_server__cluster_autojoin` (default **`False`**, opt-in)
  - `rabbitmq_server__cluster_hosts` (default:
    `groups["debops_service_rabbitmq_server"] | sort`)
  - `rabbitmq_server__cluster_seed_node` (default: first entry of the
    list)
- New `tasks/cluster_autojoin.yml` imported after handlers have flushed.
  On every host it runs `rabbitmqctl cluster_status` and:
  - no-ops if the seed is already in `running_nodes`;
  - otherwise asserts `disk_nodes == [self]` and performs
    `stop_app` → `reset` → `join_cluster rabbit@<seed>` → `start_app`.
- Combined with `serial: 1` from the rolling-restart work, both
  invocation modes work out of the box once auto-join is enabled:
  - `debops run service/rabbitmq_server` against the whole group –
    nodes are configured sequentially and auto-join the seed.
  - `debops run service/rabbitmq_server --limit hostN` – still works,
    provided the seed is reachable.
- Auto-join is **disabled by default** so that upgrading the role on
  existing deployments never alters cluster membership on its own. In
  particular, several independent single-node RabbitMQ instances that
  share the `debops_service_rabbitmq_server` inventory group stay
  independent. Operators enable auto-join explicitly:
  ```yaml
  # group_vars/debops_service_rabbitmq_server/rabbitmq_server.yml
  rabbitmq_server__cluster_autojoin: True
  ```
- For inventories hosting more than one independent cluster under the
  umbrella `debops_service_rabbitmq_server` group, narrow
  `rabbitmq_server__cluster_hosts` in the per-sub-group `group_vars` so
  the seed election and the join step do not mix clusters together. A
  full inventory example is in
  `docs/ansible/roles/rabbitmq_server/defaults-detailed.rst`.
## Documentation
- `CHANGELOG.rst` – three new entries under "debops master -
  unreleased" (Added section); auto-join entry notes the `False` default
  and the opt-in rationale.
- `docs/ansible/roles/rabbitmq_server/defaults-detailed.rst` – documents
  the `opt_in` parameter, the "required flag" skip behaviour, the new
  cluster auto-join variables, a `.. warning::` around the default of
  `rabbitmq_server__cluster_hosts`, and a full "Running multiple
  independent clusters under one inventory" example with two dedicated
  inventory sub-groups.
- `docs/ansible/roles/rabbitmq_server/getting-started.rst` – new
  "Automatic cluster formation" section in opt-in tone (minimal enable
  snippet + safety-assertion description) plus a "Multiple independent
  clusters in one environment" paragraph that cross-references the
  defaults reference.
## Compatibility
- **No default behaviour change.** `rabbitmq_server__cluster_autojoin`
  defaults to `False`, so upgrading the role is a no-op with respect to
  cluster membership. Operators who want the new automation **must opt
  in** explicitly, per inventory group. Deployments that run several
  independent single-node RabbitMQ instances under one
  `debops_service_rabbitmq_server` group are therefore safe to upgrade
  without any inventory changes.
- The `reset + join_cluster` sequence, even once enabled, is guarded by
  an `assert` that `disk_nodes == [self]`, so an already-clustered node
  is never reset – it fails loudly instead.
- RabbitMQ 3.x: no functional change. All added tasks are gated on
  feature-flag list contents or on `rabbitmq_server__cluster_autojoin`.
- RabbitMQ 4.x: the role now applies cleanly on fresh installs,
  including fleet-wide `peer_discovery_classic_config` deployments and
  post-upgrade inventories that still carry `detailed_queues_endpoint`.
## Testing
Verified on a 3-node RabbitMQ 4.x cluster on Debian Trixie:
- Fresh install (all 3 hosts together) with
  `rabbitmq_server__cluster_autojoin: True` – cluster forms
  automatically, `rabbitmqctl cluster_status` reports all three
  `running_nodes`.
- Bootstrap one node at a time (`--limit host1`, then `host2`, `host3`)
  – each subsequent run joins the running seed.
- Re-run on a healthy cluster – all feature-flag tasks report
  `ok`/`skipped`, auto-join task reports `ok` (no-op), idempotency
  preserved across repeated runs.
- Inventory with `detailed_queues_endpoint` and `khepri_db` (opt-in) –
  both flags reported as `enabled` via
  `rabbitmqctl list_feature_flags`, task reports `skipped` on second
  run (fixes previous "always changed" behaviour).
- Default upgrade path with `rabbitmq_server__cluster_autojoin`
  unset – the role runs to completion without touching cluster
  membership on existing single-node deployments.
- Two independent clusters in one inventory
  (`rabbitmq_cluster_a` / `rabbitmq_cluster_b` sub-groups of
  `debops_service_rabbitmq_server`, each with a narrowed
  `rabbitmq_server__cluster_hosts`) – each cluster forms on its own
  and stays independent; seed election in group A never reaches hosts
  of group B.
## Commits
- `a9f24ff7d` rabbitmq_server: skip required feature flags idempotently
- `e8fa6726f` rabbitmq_server: automatic cluster formation + opt-in feature flags
- `1bbba879b` rabbitmq_server: make opt-in feature flag task idempotent
- `23d6070cd` rabbitmq_server: default cluster_autojoin to False (opt-in)

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -113,6 +113,16 @@ General
   command options and their arguments are separated with a spaca. This fixes an
   issue with ACME DNS-01 challenge not being processed correctly.
 
+:ref:`debops.rabbitmq_server` role
+''''''''''''''''''''''''''''''''''
+
+- The ``Manage RabbitMQ feature flags`` task no longer fails when the
+  inventory references feature flags that have become ``required`` in the
+  running RabbitMQ version. Required flags are auto-enabled and no longer
+  appear in ``rabbitmqctl list_feature_flags`` output; they are now detected
+  and skipped. Fixes a hard failure observed with RabbitMQ 4.x on Debian
+  Trixie where ``detailed_queues_endpoint`` was required.
+
 
 `debops v3.3.0`_ - 2026-03-13
 -----------------------------
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -41,6 +41,23 @@ Added
   --default-queue-type`` for the vhost. Useful on RabbitMQ 3.13+/4.x where
   ``quorum`` queues can be made the default without resorting to policies.
 
+- The role can now form a RabbitMQ cluster automatically. Non-seed nodes
+  run ``rabbitmqctl reset`` and ``rabbitmqctl join_cluster`` against the
+  seed node (first host of the ``debops_service_rabbitmq_server`` group,
+  sortable override via ``rabbitmq_server__cluster_hosts``) once RabbitMQ
+  has been configured and restarted. The join is guarded by a sanity
+  check that the current node's ``disk_nodes`` contains only itself, so
+  an existing cluster member is never reset. Controlled by
+  ``rabbitmq_server__cluster_autojoin`` (default ``True``); flip to
+  ``False`` for the classic manual workflow.
+
+- The ``rabbitmq_server__*_feature_flags`` list variables now accept an
+  ``opt_in: True`` key per entry. When set, the role enables the feature
+  flag via ``rabbitmqctl -q enable_feature_flag --opt-in`` instead of the
+  standard ``community.rabbitmq.rabbitmq_feature_flag`` module. This lets
+  operators switch on flags that require explicit opt-in, such as
+  ``khepri_db`` on RabbitMQ 4.0/4.1.
+
 Changed
 ~~~~~~~
 
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -43,13 +43,19 @@ Added
 
 - The role can now form a RabbitMQ cluster automatically. Non-seed nodes
   run ``rabbitmqctl reset`` and ``rabbitmqctl join_cluster`` against the
-  seed node (first host of the ``debops_service_rabbitmq_server`` group,
-  sortable override via ``rabbitmq_server__cluster_hosts``) once RabbitMQ
-  has been configured and restarted. The join is guarded by a sanity
-  check that the current node's ``disk_nodes`` contains only itself, so
-  an existing cluster member is never reset. Controlled by
-  ``rabbitmq_server__cluster_autojoin`` (default ``True``); flip to
-  ``False`` for the classic manual workflow.
+  seed node (first host of ``rabbitmq_server__cluster_hosts``, which
+  defaults to the ``debops_service_rabbitmq_server`` inventory group
+  sorted alphabetically) once RabbitMQ has been configured and restarted.
+  The join is guarded by a sanity check that the current node's
+  ``disk_nodes`` contains only itself, so an existing cluster member is
+  never reset. Controlled by ``rabbitmq_server__cluster_autojoin``
+  (default ``False``, opt-in) so that upgrading the role on existing
+  deployments never alters cluster membership on its own; in particular,
+  several independent single-node RabbitMQ instances that happen to share
+  the ``debops_service_rabbitmq_server`` group stay independent until
+  auto-join is explicitly enabled on the groups that should form a
+  cluster. See the role documentation for a multi-cluster inventory
+  layout example.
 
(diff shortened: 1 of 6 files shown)


drybjed (approved):
The changes look good, I assume that the code was tested with partial Ansible runs using `--limit`, so the role should work fine in such case. There are small cosmetic changes I'll add in separate commits.

The failing CI tests were fixed in a separate PR so should work fine now; the issue was with the `sudo` role which this PR does not touch.

PR #2692, opened 2026-05-02
[nsswitch] Update to not break libnss-systemd

Alphix:
First, systemd needs to be included for "shadow" and "gshadow" as well in order to handle users created via the /usr/lib/sysusers.d/ mechanism (since systemd 251.3-2). systemd (258.1-2) will automatically add these groups (to fix bug #1118640), but debops.nsswitch will undo that fix, which breaks any package using the /usr/lib/sysusers.d/ mechanism to create users (e.g. gdm3).

Second, since glibc 2.33 (Bookworm onwards), nsswitch.conf is automatically reloaded if the file is changed, so update the tasks and docs to remove an unnecessary restart.

### DIFF
--- a/ansible/roles/nsswitch/defaults/main.yml
+++ b/ansible/roles/nsswitch/defaults/main.yml
@@ -2,7 +2,7 @@
 # .. vim: foldmarker=[[[,]]]:foldmethod=marker
 
 # .. Copyright (C) 2017 Maciej Delmanowski <drybjed@gmail.com>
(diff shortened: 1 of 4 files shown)


Alphix:
See also #2672 (but I'd consider this PR to be a bit more thorough)

drybjed (approved):
Sounds good, thanks. Restarting `systemd-logind` was always problematic since it broke existing user session, so the removal should be fine, although it makes more cumbersome on older systems. But let's go with it, time to update.

PR #2693, opened 2026-05-07
[PKI] Extend ACME DNS support to other providers

bhwright:
Added support for ACME DNS-01 for other providers; ACME HTTP-01 already passes the "--server" flag.  Additional providers must be added to pki_acme_ca_api_map variable.  

Documentation, including a warnings that "staging" may result in a **valid** certificate, has been updated.


drybjed (approved):
Changes look good. I'll fix the CI issues before merging.

PR #2701, opened 2026-05-26
Add docker_compose_service role for multi-container Compose projects

scibi:
## Summary

Adds a new `debops.docker_compose_service` role that manages Docker Compose
projects (multi-container applications) on DebOps-managed hosts, with optional
integration with `debops.nginx` for automatic reverse-proxy setup. Wires the
new playbook into `layer/app.yml` so that `debops run site` deploys Compose
projects out of the box.

This complements the existing `debops.docker_service` role which manages
**individual Docker containers** (single-container apps such as Grafana,
VictoriaMetrics, Bugsink, etc.). The new role is for **multi-container apps**
distributed as `docker-compose.yml` files (Immich, Paperless-ngx, etc.) that
rely on Compose features such as `depends_on` with health checks, named
networks, and `extends` for hardware-acceleration overlays.

## Motivation

Some self-hosted applications ship a `docker-compose.yml` as the primary
deployment format and rely on Compose semantics that aren't expressible
through `community.docker.docker_container` alone:

- **`depends_on` with health checks** — Compose waits for a healthcheck to
  pass before starting dependent containers (e.g. ML service waiting for
  PostgreSQL with pgvector).
- **`extends` overlays** — common pattern for opt-in hardware acceleration
  (CUDA, ROCm, OpenVINO) where the upstream project ships a base
  `docker-compose.yml` plus additional `hwaccel.*.yml` files.
- **Named networks and service discovery** — internal service-to-service
  hostnames (`http://tika:9998`, `postgres://db:5432`) without manual
  network plumbing.

Re-implementing these on top of `docker_service` would mean either rewriting
the upstream Compose file as a flat list of `docker_container` calls (losing
upstream maintenance) or shelling out to `docker compose` from a custom
playbook (losing DebOps idioms — secret integration, nginx dependent vars,
inventory groups).

## Design

The role mirrors the structure of `debops.docker_service` so the two are
interchangeable for users who know one of them:

| Aspect | `docker_service` | `docker_compose_service` |
|---|---|---|
| Inventory group | `debops_service_docker_service` | `debops_service_docker_compose_service` |
| Service-list var | `docker_service__host_services` | `docker_compose_service__host_services` |
| Per-service `nginx:` block | yes | yes |
| Secret integration | yes | yes |
| Underlying module | `community.docker.docker_container` | `community.docker.docker_compose_v2` + `docker_compose_v2_pull` |
| Service definition | inline `image:`, `volumes:`, `env:` | `compose_src:` template + `env:` -> `.env`, plus `compose_files:`, `config_files:`, `config_dir:` |

The role:

1. Renders a `docker-compose.yml` per service (from a Jinja template under
   `<inv>/../docker_compose_service/by-host/<host>/<service>/docker-compose.yml`
   or from inline `compose_content`).
2. Generates a `.env` file from `item.env` (or from a separate template).
3. Deploys auxiliary files: `compose_files` (extra YAML overlays for
   `extends`), `config_files` (single-file deploys), `config_dir` (whole
   subtrees with rsync semantics for app config).
4. Pulls images (`docker compose pull`) and brings the project up
   (`docker compose up -d` via `docker_compose_v2`).
5. Tracks which service definitions changed since last run and forces
   `recreate=always` on those projects only — others get `recreate=auto`.

The `pre_tasks` build the `docker_compose_service__nginx__dependent_servers`
and `..._upstreams` lists by rendering two Jinja templates against the
combined service list, which are then fed into the `nginx` role via standard
DebOps dependent-vars conventions. So when a service entry contains:

```yaml
nginx:
  enabled: true
  fqdn: 'app.example.com'
  port: '8000'
```

…the role automatically configures an upstream `app-upstream` pointing at
`127.0.0.1:8000` plus a server block on `app.example.com` that proxies to it.
SSL termination uses the existing `debops.pki` infrastructure.

### Per-host template tree

Following the established DebOps convention where each role that consumes
user-supplied per-host content owns a top-level directory next to the
inventory (`debops.resources` -> `resources/`, `debops.nixos` -> `nixos/`,
`debops.secret` -> `secret/`), this role uses a private tree:

```
<inventory>/../docker_compose_service/
  by-host/
    <inventory_hostname>/
      <service-name>/
        docker-compose.yml
        <other files...>
```

The default value of `docker_compose_service__templates_path` points there.
This **must not** be changed to point at `<inv>/../resources/templates/`
because the `debops.resources` role iterates that tree via
`with_community.general.filetree` and deploys every file it finds verbatim
to `/<relative-path>` on the host. Hosts that are members of both
`debops_service_docker_compose_service` and `debops_service_resources`
would get their Compose templates rendered through the standard `template`
module and copied to `/` -- silent corruption of the host filesystem.

With the role-private tree the two roles never overlap regardless of
inventory group membership.

Inventory parameters that reference templates (`compose_src`,
`compose_files.src`, `config_files.src`, `config_dir.src`) are written
relative to `by-host/<inventory_hostname>/`, e.g.:

```yaml
docker_compose_service__host_services:
  - name: 'immich'
    compose_src: 'immich/docker-compose.yml'    # not 'docker_compose_service/immich/...'
    compose_files:
      - src: 'immich/hwaccel.ml.yml'
```

## What this PR adds

- **Role**: `ansible/roles/docker_compose_service/`
  - `defaults/main.yml`, `meta/main.yml`, `COPYRIGHT`
  - `tasks/main.yml` (project dir + data dirs + compose template +
    `.env` generation + auxiliary files + `compose pull` + `compose up`)
  - `tasks/build_nginx_vars.yml` (pre-task building nginx dependent vars)
  - `tasks/manage_config_dir.yml` (rsync-style config tree management)
  - `templates/lookup/docker_compose_service__nginx_servers.j2`
  - `templates/lookup/docker_compose_service__nginx_upstreams.j2`
  - `templates/etc/ansible/facts.d/docker_compose_service.fact.j2`
- **Playbook**: `ansible/playbooks/service/docker_compose_service.yml`
  - Imports `secret`, `keyring`, `apt_preferences`, `ferm`, `python`,
    `nginx`, then the role itself — same shape as `service/docker_service.yml`
  - Wraps nginx-dependent roles in `when: ..._dependent_servers | length > 0`
    so hosts that don't expose any HTTP services skip the nginx machinery
- **Site integration**: `ansible/playbooks/layer/app.yml`
  - Adds the import after `service/debconf.yml`, immediately after
    `service/docker_service.yml` so that hosts running both mechanisms get
    `docker_service` (and its postgresql dependent_roles) configured first
- **Documentation**: `docs/ansible/roles/docker_compose_service/`
  - `getting-started.rst` — prerequisites, comparison vs `docker_service`,
    nginx integration, minimal example
  - `defaults-detailed.rst` — every default variable annotated
  - `guides.rst` — full Immich deployment walkthrough
  - `index.rst`, `man_index.rst`, `man_synopsis.rst`, `man_description.rst`
- **Bug fix**: `1404840a8` — replaces single-quoted Jinja in conditionals
  with `is defined and X is truthy` form, required for Ansible v2.19
  strictness around bare-string conditions.

## Dependencies

This PR depends on
#2699 
which adds the `docker_service` import to `layer/app.yml`. The merge commit
in this PR's history (`939ae068b`) is what carries that change forward;
once the dependency PR is merged, this branch can be rebased and the merge
commit will drop out cleanly.

## Testing

Tested in a homelab DebOps deployment running Debian 13 (trixie) on
unprivileged Proxmox LXC containers:

- **Immich** (multi-container: server + ML + postgres + redis) deployed via
  `compose_src` template, brought up cleanly via
  `debops run site -l immich.example.com`. Verified `docker compose ps`
  shows all containers healthy, nginx vhost on `immich.example.com` proxies
  correctly with SSL.
- **Paperless-ngx** (multi-container: webserver + Gotenberg + Tika) deployed
  with host-managed PostgreSQL/Redis via UNIX sockets bind-mounted into the
  containers. Verified that `docker_compose_service__nginx__dependent_servers`
  is populated correctly during `pre_tasks` and that the resulting nginx
  config matches what `docker_service` would produce for an equivalent
  single-container app.
- **Idempotency**: re-running the playbook with no changes results in
  `changed=0` for the role; modifying a single env var causes only the
  affected project to be `recreate`d, others keep `recreate=auto`.
- **State transitions**: tested `state: stopped` (containers stopped, project
  files retained) and `state: absent` (project removed entirely). Both work
  as expected.

## Compatibility

- Ansible 2.16+ (tested on 2.19; `is truthy` conditional form requires 2.16+)
- `community.docker` collection 3.4.0+ (for `docker_compose_v2` module)
- Docker Engine with Compose plugin v2 (`docker compose ...`, not the
  legacy `docker-compose` Python script)

## Checklist

- [x] Role follows DebOps role conventions (defaults/meta/tasks structure,
      `<role>__` variable namespace, `[ 'role::<name>', 'skip::<name>' ]`
      tags, `become: True` on the playbook level)
- [x] Playbook follows the same structure as `service/docker_service.yml`
      (pre_tasks build nginx dependent vars, roles list applies them
      conditionally)
- [x] Documentation under `docs/ansible/roles/docker_compose_service/`
      (getting-started, defaults-detailed, guides, man pages)
- [x] All new files include SPDX-License-Identifier and Copyright headers
- [x] Role index updated where applicable
- [ ] CI green (will be verified after push)


ansible/roles/docker_compose_service/templates/etc/ansible/facts.d/docker_compose_service.fact.j2
```diff
@@ -0,0 +1,53 @@
...
+    'projects': {},
+    'firewall': {
+{% for svc in docker_compose_service__combined_services | d([]) %}
+{%   if svc.state | d('present') not in ['absent', 'ignore']
+        and svc.published_ports | d([]) | selectattr('allow', 'defined') | list | length > 0 %}
+        '{{ svc.name }}': {{ svc.published_ports
+                              | selectattr('allow', 'defined')
+                              | list | to_json }},
+{%   endif %}
+{% endfor %}
```
drybjed: This code should be in a heredoc-like variable that can contain the dictionary for the `firewall` key. This way it can be properly linted by pyCodeStyle and treated as a Python script by other tools.

Actually, what this key is used for? I don't see `ansible_local.docker_compose_service.firewall` used anywhere.

drybjed (changes requested):
Very comprehensive role. Hopefully in near future we can add some CI tests on GitHub Actions for it to use it as a benchmark. There's slight problem with the fact script, I'll merge the PR when that's fixed. The current CI test failures are due to bugs in the `sudo` role which have been subsequently fixed.

PR #2702, opened 2026-05-26
Add cloudflared role for managing Cloudflare Tunnel instances

scibi:
## Summary

Adds a new `debops.cloudflared` role that installs and configures the
`cloudflared` daemon (Cloudflare Tunnel client) on Debian-family hosts, with
support for multiple concurrent tunnel instances managed via a single systemd
template unit. Wires the new playbook into `layer/agent.yml` so that
`debops run site` configures Cloudflare Tunnels alongside other outbound
agents (Filebeat, Metricbeat, Telegraf, Zabbix Agent).

`cloudflared` maintains outbound-only encrypted connections from the host to
the Cloudflare edge, eliminating the need for inbound firewall ports or a
public IP on the origin server. This makes it a natural building block for
Zero Trust homelab and self-hosted deployments where the public-facing
attack surface should be minimised.

## Motivation

Cloudflare Tunnel deployments in DebOps environments today rely on either
(a) manual `apt-get install cloudflared` plus hand-rolled systemd units
and `config.yml` files, or (b) running `cloudflared` inside a container,
which adds an unnecessary network namespace between the daemon and the
host services it proxies to.

A dedicated DebOps role offers:

- **Multi-tunnel support out of the box** — a single host can run several
  named tunnel instances (`cloudflared@web.service`,
  `cloudflared@admin.service`, …) sharing one binary and one APT repo, with
  per-tunnel credentials, configuration, and lifecycle.
- **Token mode and local mode** — both Cloudflare-managed (remote) and
  Ansible-managed (local) tunnel configurations are first-class, with
  validation at playbook time so misconfigured tunnels fail fast.
- **DebOps secret integration** — credentials live under
  `secret/cloudflared/tunnels/<name>/{token,credentials.json}` on the
  Ansible Controller, never in cleartext in inventory.
- **Hardened systemd unit** — sandboxing, capability dropping, and a
  `TimeoutStopSec` greater than the in-flight grace period so the kernel
  doesn't `SIGKILL` the daemon mid-drain.
- **Idempotent restart-on-change semantics** — only restart a tunnel when
  *its* credentials/config/env file changed, not on every playbook run.

### Systemd template unit

A single `cloudflared@.service` template unit drives all tunnel instances.
Per-tunnel state lives in two files under `/etc/cloudflared`:

- `<name>.yml` — `config.yml` for local-mode tunnels, or a stub for token
  mode containing only `token-file` references.
- `<name>.env` — `EnvironmentFile` for token mode (holds `TUNNEL_TOKEN`),
  or absent in local mode.

This means adding a new tunnel is a one-line YAML addition; removing one is
`state: 'absent'` with full cleanup of unit/config/credentials.

### Tunnel modes

| Mode | Where ingress lives | Required secret | Use case |
|---|---|---|---|
| `token` (default) | Cloudflare Zero Trust dashboard or IaC (Terraform/OpenTofu Cloudflare provider) | `secret/cloudflared/tunnels/<name>/token` | Cloudflare Access integration, dashboard-driven workflows |
| `local` | Inline `ingress:` list in inventory, rendered to `<name>.yml` | `secret/cloudflared/tunnels/<name>/credentials.json` | GitOps for tunnel routing, no Cloudflare-side IaC |

Both modes use the standard DebOps `secret` mechanism. The `main_env`
pre-task computes the list of required `secret/` directories per host and
passes it to the `secret` role via `secret_directories`.

### Validated tunnel definitions

Each tunnel passes through an `assert` in `manage_tunnel.yml` before any
files are deployed:

```yaml
- name: Validate tunnel '{{ tunnel.name | d("<unnamed>") }}' definition
  ansible.builtin.assert:
    that:
      - tunnel.name is defined and tunnel.name | length > 0
      - tunnel.mode | d('token') in ['token', 'local']
      - tunnel.state | d('present') in ['present', 'absent']
```

…so a typo in `mode:` or a missing `name:` fails the play immediately with
a clear message, rather than silently deploying broken units.

### Restart semantics

The role registers separate change variables for credentials, config, and
env files, and only restarts a tunnel when one of *its own* files changed:

```yaml
- name: Restart tunnel '{{ tunnel.name }}' if configuration changed
  ansible.builtin.systemd:
    name: 'cloudflared@{{ tunnel.name }}.service'
    state: 'restarted'
  when: (cloudflared__register_credentials is changed) or
        (cloudflared__register_config is changed) or
        (cloudflared__register_env is changed)
```

Changes to a *different* tunnel on the same host don't disturb running
tunnels — important when a single host hosts multiple production tunnels
and you only want to roll one of them.

### Hardened systemd unit

`cloudflared` in tunnel-run mode only needs outbound TCP/UDP to the
Cloudflare edge and loopback access for the metrics endpoint, so the
hardened defaults are safe. The unit applies `ProtectSystem=strict`,
`PrivateTmp`, capability dropping, namespace isolation, and a
`TimeoutStopSec` of `45s` (greater than the default `grace_period: '30s'`)
to allow graceful connection draining without `SIGKILL`. Extra
restrictions (e.g. `IPAddressAllow=`) can be appended via
`cloudflared__systemd_hardening_extra` without forking the role.

### Hook points

`tasks/cloudflared/{pre_main,post_main}.yml` are empty placeholders that
project-level overrides can populate via the standard
`debops.debops.task_src` lookup. Useful for hooks like "wait for tunnel
to be `/ready`" or "register Prometheus scrape target after deploy".

## What this PR adds

- **Role**: `ansible/roles/cloudflared/`
  - `defaults/main.yml` (315 lines, fully documented)
  - `meta/main.yml`, dependencies on `secret`, `keyring`, `ansible_plugins`,
    `global_handlers`
  - `tasks/main.yml` — package install, user/group/dirs, systemd template
    unit, loop over `cloudflared__combined_tunnels`
  - `tasks/main_env.yml` — pre-task computing
    `cloudflared__secret_directories` for the `secret` role
  - `tasks/manage_tunnel.yml` — per-tunnel deploy/remove block with
    validation and selective restart
  - `tasks/cloudflared/{pre_main,post_main}.yml` — hook placeholders
  - `templates/etc/cloudflared/config.yml.j2` — supports both modes,
    catch-all rule, ingress with origin-request options
  - `templates/etc/cloudflared/tunnel.env.j2` — token-mode env file
  - `templates/etc/systemd/system/cloudflared@.service.j2` — hardened
    systemd template unit
  - `templates/etc/ansible/facts.d/cloudflared.fact.j2` — local facts
    exposing installed version + tunnel inventory
  - `templates/lookup/cloudflared_env_secret_directories.j2`
- **Global handler**: `roles/global_handlers/handlers/cloudflared.yml`
  with `Reload service manager` and `Restart cloudflared tunnels` handlers,
  imported from `global_handlers/handlers/main.yml`.
- **Playbook**: `ansible/playbooks/service/cloudflared.yml`
  - Imports `cloudflared/main_env` as `pre_task` to populate
    `cloudflared__secret_directories`
  - `roles:` order: `secret` (creates secret dirs) → `keyring` (Cloudflare
    APT key) → `cloudflared` (main role)
- **Site integration**: `ansible/playbooks/layer/agent.yml`
  - Adds the import at the top of the layer (before Filebeat) so the
    tunnel comes up before observability agents start emitting traffic
    that may be relayed through services exposed via the tunnel.
- **Documentation**: `docs/ansible/roles/cloudflared/`
  - `getting-started.rst` — what cloudflared is, prerequisites, two tunnel
    modes, secret management, minimal example, available tags
  - `defaults-detailed.rst` — every default variable annotated
  - `guide-docker-integration.rst` — using cloudflared to expose Docker
    Compose services on the same host
  - `guide-nginx-integration.rst` — using cloudflared in front of an
    nginx reverse proxy with origin SSL
- **Role index**: `docs/ansible/role-index.rst` updated.
- **Bug fixes / refinements** in commit `8c1a62558`:
  - **APT key rotation**: Cloudflare changed their signing key from
    `cloudflare-main.gpg` (fingerprint `A5FC51C5...`) to
    `cloudflare-public-v2.gpg` (fingerprint `CC94B39C77AE7342A68B89628A682D308D4E5E73`),
    and the repository now uses the universal `any` distribution component
    instead of per-codename pools.
  - **Hook file placeholders**: empty
    `tasks/cloudflared/{pre_main,post_main}.yml` files so the
    `include_tasks` lookups in `tasks/main.yml` don't fail when no
    project-level override is provided.
  - **Spurious notify removed**: changing APT preferences doesn't affect
    running tunnel instances, so the `Restart cloudflared tunnels`
    handler notification was dropped from that task.

## Dependencies

Independent — depends only on `master`. No coupling with any other PR
in flight.

## Testing

Tested in a homelab DebOps deployment running Debian 13 (trixie) on
unprivileged Proxmox LXC containers, behind a Cloudflare Zero Trust
account with multiple tunnels:

- **Token-mode tunnels** for several services (Paperless-ngx, Bugsink,
  Immich) provisioned via OpenTofu using the
  `cloudflare_zero_trust_tunnel_cloudflared` resource. The token is
  written to `secret/cloudflared/tunnels/<name>/token` by
  `local_sensitive_file` and consumed by this role on the next
  `debops run site`.
- **Multi-tunnel hosts** — one LXC running two named tunnels
  (`cloudflared@public.service` + `cloudflared@admin.service`) for
  different ingress audiences. Verified that modifying ingress on one
  tunnel via Cloudflare dashboard doesn't disturb the other (no restart
  on `debops run`).
- **Idempotency** — re-running the playbook with no changes results in
  `changed=0` for the role; `cloudflared.fact` reports the running
  version unchanged. Modifying a tunnel's `config.yml` via the inventory
  triggers exactly one restart of that tunnel.
- **State transitions** — tested `state: 'absent'` on a tunnel: systemd
  unit stopped + disabled, config/env/credentials files removed,
  `journalctl -u cloudflared@<name>.service` shows clean shutdown.
- **Hardening** — verified via `systemd-analyze security cloudflared@<n>`
  that the unit has `Exposure level: 1.x SAFE` rating with the default
  hardening block enabled.
- **Restart timing** — verified that with `grace_period: '30s'` and
  `TimeoutStopSec: '45s'`, in-flight HTTP requests complete during a
  config-triggered restart instead of being killed mid-stream.

## Compatibility

- Debian 12 (bookworm), Debian 13 (trixie), Ubuntu 22.04+ — all use the
  same Cloudflare APT repo with the `any` distribution component.
- `cloudflared` ≥ 2024.x (any version that supports tunnel tokens, which
  is the path most Cloudflare Zero Trust workflows now use).
- systemd ≥ 247 (for the hardening directives used in the unit template;
  older systemd versions silently ignore unknown directives, but the
  hardening will be partial).

## Checklist

- [x] Role follows DebOps role conventions (defaults/meta/tasks structure,
      `cloudflared__` variable namespace, `[ 'role::cloudflared',
      'skip::cloudflared' ]` tags, `become: True` on the playbook level)
- [x] Playbook follows the standard `service/<name>.yml` structure
      (pre_task computes dependent vars, then `secret` → `keyring` →
      main role)
- [x] Documentation under `docs/ansible/roles/cloudflared/` (getting-
      started, defaults-detailed, two integration guides, all listed in
      `role-index.rst`)
- [x] All new files include SPDX-License-Identifier and Copyright headers
- [x] `tasks/cloudflared/{pre_main,post_main}.yml` hook placeholders
      provided so default-config users don't need to create empty files
- [x] APT signing key rotated to current Cloudflare key
      (`CC94B39C77AE7342A68B89628A682D308D4E5E73`)
- [x] Hardened systemd unit with `TimeoutStopSec` > `grace_period` so the
      kernel doesn't `SIGKILL` mid-drain
- [ ] CI green (will be verified after push)


ansible/roles/global_handlers/handlers/cloudflared.yml
```diff
@@ -0,0 +1,14 @@
...
+
+- name: Restart cloudflared tunnels
+  ansible.builtin.systemd:
+    name: 'cloudflared@{{ item.name }}.service'
+    state: 'restarted'
+  loop: '{{ cloudflared__combined_tunnels }}'
+  loop_control:
+    label: '{{ item.name }}'
+  when: cloudflared__enabled | d(False) | bool and
+        item.state | d('present') == 'present'
```
drybjed: Using role-scoped variables here will break the entire `global_handlers` role when the `cloudflared` role is not included in a playbook. Move this handler into the `cloudflared` role to avoid that.

ansible/roles/cloudflared/defaults/main.yml
```diff
@@ -0,0 +1,315 @@
...
+
+# Role state and deployment [[[
+# -----------------------------
+
+# .. envvar:: cloudflared__enabled [[[
+#
+# Enable or disable the ``cloudflared`` role.
+cloudflared__enabled: True
+
+                                                                   # ]]]
```
drybjed: The role is not part of the `common.yml` playbook, so adding this variable is probably superfluous.

ansible/roles/cloudflared/defaults/main.yml
```diff
@@ -0,0 +1,315 @@
...
+# The time to wait between restarts of cloudflared service instances.
+cloudflared__systemd_restart_sec: '5'
+
+                                                                   # ]]]
+# .. envvar:: cloudflared__systemd_after [[[
+#
+# The ``After=`` directive in cloudflared systemd service units.
+cloudflared__systemd_after: 'network-online.target'
+                                                                   # ]]]
+                                                                   # ]]]
```
drybjed, line 262: These options could probably be set directly in the `systemd` unit files deployed by this role, especially the `cloudflared__systemd_after` variable. The options can be overridden via the `systemd` role if needed using the systemd-level override files, and most likely will be more useful per-instance, for example to order different tunnels after different services are started.

ansible/roles/cloudflared/defaults/main.yml
```diff
@@ -0,0 +1,315 @@
...
+# The default log level for cloudflared. Possible values:
+# ``debug``, ``info``, ``warn``, ``error``, ``fatal``.
+cloudflared__loglevel: 'info'
+
+                                                                   # ]]]
+# .. envvar:: cloudflared__metrics [[[
+#
+# Address and port to expose Prometheus metrics. Set to empty string to
+# disable. Example: ``127.0.0.1:2000``.
+cloudflared__metrics: ''
```
drybjed: The default could be enabled by default on localhost instead of being a proposal via comment - less work during deployment. A nice QoL would be adding an `etc_services` entry for this port when enabled. Also, is there an upstream default for Prometheus metrics used by `cloudflared`? Port 2000 seems to be already reserved for Cisco SCCP service, I would pick a different one.

drybjed (changes requested):
Looks like a solid role. Looking forward for some changed requested elsewhere, then we can merge it. The CI test failures are due to an issue with the `sudo` role, already fixed by a different PR.

Issue #2705, opened 2026-06-03
[ifupdown] `systemd` command not found (Debian Trixie)


oxmie:
## Description

During the `ifupdown` role run, DebOps tries to evaluate the *systemd`-version with the (deprecated) `systemd` command:
https://github.com/debops/debops/blob/6ed3c3047f63af66772fe73f0b118ff76c50e6ce/ansible/roles/ifupdown/tasks/ifup_systemd.yml#L8

(see https://github.com/debops/debops/blob/master/ansible/roles/ifupdown/tasks/ifup_systemd.yml#L7 )

In *Debian Trixie* the command `/bin/systemd` is not anymore provided (see https://packages.debian.org/trixie/amd64/systemd/filelist ) (compare with *Debian Bookworm* https://packages.debian.org/bookworm/amd64/systemd/filelist ).

## Solution 
As the command `/usr/bin/systemctl` provides the same output as the previous ``/bin/systemd` command, the `ifupdown` role can easy use `systemctl` at this point.

**Debian Bookworm** (example system)
```sh
# cat /etc/os-release 
PRETTY_NAME="Debian GNU/Linux 12 (bookworm)"
[...]

# systemd --version
systemd 252 (252.38-1~deb12u1)
+PAM +AUDIT +SELINUX +APPARMOR +IMA +SMACK +SECCOMP +GCRYPT -GNUTLS +OPENSSL +ACL +BLKID +CURL +ELFUTILS +FIDO2 +IDN2 -IDN +IPTC +KMOD +LIBCRYPTSETUP +LIBFDISK +PCRE2 -PWQUALITY +P11KIT +QRENCODE +TPM2 +BZIP2 +LZ4 +XZ +ZLIB +ZSTD -BPF_FRAMEWORK -XKBCOMMON +UTMP +SYSVINIT default-hierarchy=unified

# systemctl --version
systemd 252 (252.38-1~deb12u1)
[...]

PR #2706, opened 2026-06-03
Change `systemd` command to `systemctl` (Solves #2705)

oxmie:
## Description

This PR solves the issue described in https://github.com/debops/debops/issues/2705.


drybjed (approved):
Good idea, thanks. I'll backport this to the stable branches as well.

PR #2708, opened 2026-06-04
chore(sudo): ubuntu 26 use sudo-rs

seven-beep:
As ubuntu 26 use `sudo-rs` now, we should avoid to install the old `sudo`
alongside.

Unfortunately `sudo-rs` does not support all `sudo` configurations and using
'!requiretty' would not pass the validation of the sudoers files.

Note that I did not test sudo with ldap, so further attention may be required.

https://ubuntu.com/server/docs/reference/other-tools/sudo-rs/


drybjed (approved):
Good idea, thanks. But I prefer brackets to explicitly control the code, I'll add them in the variable.

PR #2709, opened 2026-06-08
Extract multiple intermerdiate certificates from the result of acme-tiny

### LINKED ISSUE #2707: certificate chain is missing the second intermediate with Generation Y letsencrypt certificates with acme-tiny
Since last month, letsencrypt started to roll out Generation Y Certificates. [1]
These certificates have two intermediates, YR1 and YR.
The /usr/local/lib/pki/pki-realm script supports only one intermediate

This leads to the following:
* Clients are fetching the missing certificate in the chain themselves 
* Since there was an issue with the first version of the YR Certificate, some clients will have an YR certificate which is revoked [2]

## References

1. https://letsencrypt.org/2025/11/24/gen-y-hierarchy
2. https://community.letsencrypt.org/t/today-re-issued-certificates-from-issuer-yr2-have-a-revoked-cert-in-chain/247612

pfy:
Fixes https://github.com/debops/debops/issues/2707
Lets encrypt uses multiple intermediate certificates now. The pki script has to support that or it will break some browsers / clients

### DIFF
--- a/ansible/roles/pki/files/usr/local/lib/pki/pki-realm
+++ b/ansible/roles/pki/files/usr/local/lib/pki/pki-realm
@@ -1119,56 +1119,25 @@ request_acme_tiny_certificate () {
             fi
 
             if [ -r acme/cert.pem.tmp ] && [ -s acme/cert.pem.tmp ] ; then
-
                 # Let's Encrypt provides certificates with attached
                 # intermediates but we don't want that, PKI realms handle the
                 # intermediates themselves. Extract the first certificate from
                 # the generated file and swap it in, so the rest of the script
                 # works.
-                openssl x509 -in acme/cert.pem.tmp -out acme/first-cert.pem.tmp
-                mv acme/first-cert.pem.tmp acme/cert.pem.tmp
+
+                csplit -s -z -f acme/temp_cert_ acme/cert.pem.tmp '/-----BEGIN CERTIFICATE-----/' '{*}'
+                if [ ! -r acme/temp_cert_00 ] ; then
+                  echo "no certificates found in acme/certm.pem.tmp"
+                  exit 1
+                fi
 
                 if [ -r acme/cert.pem ] ; then
                     mv acme/cert.pem acme/cert.pem.old
                 fi
-                mv acme/cert.pem.tmp acme/cert.pem
-
-                local acme_intermediate_uri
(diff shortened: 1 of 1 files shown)


Gomez:
Thanks, worked for me!

oxmie:
@pfy thx for you patch.

I hit this issues as well, but I wasn't sure yet how to fix it probably.

Your fix works (for me as well) if the cert provider will return it's complete chain. For me and all the others using _Let's Encrypt_ it's solved with your PR :tada: 

The previous code has downloaded the intermediates via the issuer URI from the signed certs (only one intermediate was possible). Your code changes will now break this feature, downloading the intermediates via the issuer URI. 

I do not have any idea, if there exists a specification that a cert provider MUST provide the full chain. If yes , than I would prefer your simple code solution, otherwise as written before your code will break an existing feature.

oxmie (approved):
This PR works and will solve #2707 with the side effect, that the function that downloads the intermediate via the issuer URI is removed (as comment in https://github.com/debops/debops/pull/2709#issuecomment-4669004980 )

pfy:
> @pfy thx for you patch.
> 
> I hit this issues as well, but I wasn't sure yet how to fix it probably.
> 
> Your fix works (for me as well) if the cert provider will return it's complete chain. For me and all the others using _Let's Encrypt_ it's solved with your PR 🎉
> 
> The previous code has downloaded the intermediates via the issuer URI from the signed certs (only one intermediate was possible). Your code changes will now break this feature, downloading the intermediates via the issuer URI.
> 
> I do not have any idea, if there exists a specification that a cert provider MUST provide the full chain. If yes , than I would prefer your simple code solution, otherwise as written before your code will break an existing feature.

The server SHOULD answer with the full chain. See https://www.rfc-editor.org/info/rfc8555/#section-9.1. 

acme-tiny EXPECTS the server to answer with the fullchain. See https://github.com/diafygi/acme-tiny/blob/main/README.md#step-4-get-a-signed-certificate

The code fetching the intermediates is from 2016, https://github.com/debops/debops/commit/abc15d421ef00562da4803b32e63f40ebcc1839f and replaced a static way for fetching the intermediates from before. 

I don't see any discussion why it was done as it was done and not just splitting the certificate chain from the server, which would have worked back then as well.


I see the risk of breaking other solutions and would be happy to make the patch in a form which supports the old form for downloading the intermediates as well. But i don't know where i could test that, i'm not aware of any acme complaint CA which does NOT send any intermediates in the certificate request.

drybjed (approved):
Thanks for the patch. From the other commenters I'm assuming that the code works as expected and shouldn't affect `certbot` installations since it's inside the `request_acme_tiny_certificate`. I'll backport this change to the stable releases as well.

The old code indeed was written in 2016 when Let's Encrypt sent only one intermediate CA certificate, so it's time for a change.

PR #2712, opened 2026-06-15
add an rsyslog__shell variable and change the default shell to /usr/s…

arjanoosting:
When rolling out the rsyslog role it changes the shell of the already present syslog user from /usr/sbin/nologin to /bin/false. I would rather keep it the /usr/sbin/nologin shell which is the default shell according to the adduser manpage:

> Unless  a  shell  is explicitly set with the --shell option, the new system user will have the shell set to /usr/sbin/nologin.

The rsyslog.postinst doesn't specify the shell thus it is created with the /usr/sbin/nologin shell when the package is installed

```
case "$1" in
    configure)
        adduser --system --group --no-create-home --quiet syslog || true
```

To make the shell configurable I have added the `rsyslog__shell` variable with the default /usr/sbin/nologin but this could also be /bin/false if you prefer that.

### DIFF
--- a/ansible/roles/rsyslog/defaults/main.yml
+++ b/ansible/roles/rsyslog/defaults/main.yml
@@ -124,6 +124,12 @@ rsyslog__home: '{{ "/home/syslog"
                    if (ansible_distribution in ["Ubuntu"])
                    else "/var/log" }}'
 
+ # .. envvar:: rsyslog__shell [[[
+#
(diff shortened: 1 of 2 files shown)


drybjed:
The `rsyslog__shell` variable has been added in the https://github.com/debops/debops/commit/483311bd1f165cc1206c0fba976ce0d1696b3a29 commit, I'm closing this one as ducplicate.

PR #2720, opened 2026-06-23
Rspamd: DKIM: Add NetBox method to rspamd-dkim-update (and optional Healthchecks ping)

mathieumd:


cogito-mens (commented):
## Design concerns

### `requests` as hard module-level dependency

`import requests` is at module scope. Every user of the script — even those with `method: log` or `method: email` — must install `python3-requests` or the script crashes on import. Consider guarding behind a `try/except ImportError` with a clear error message, or deferring to an import inside `do_netbox`.

### Stricter `get_keys` for all methods

The DNS file read/parse loop was moved outside `if action == "add":` so that `rr_data` is available for both add and delete (needed for NetBox value matching). This means DNS file parsing — and potential `die()` on malformed files — now also runs on delete actions for all methods. Previously a malformed file on delete was silently skipped; now it aborts the entire run.

### `strip_dkim_spaces` / value comparison fragility

The four `replace()` calls handle the common chunked DKIM TXT format (`"..." "..." "..."`), but if NetBox normalizes the value differently on read-back (whitespace, quoting, wrapping), the existing-record match in `do_netbox` may silently fail to find the record, creating a duplicate.

ansible/roles/rspamd/files/usr/local/sbin/rspamd-dkim-update
```diff
@@ -152,6 +161,17 @@ def read_config_file(path):
                 not config["nsupdate_gsstsig_princ"]):
             die("nsupdate gsstsig principal not configured")
 
+    elif config["method"] == "netbox":
+        if not config["netbox_url"]:
+            die("netbox_url not configured")
+
+        if not config["netbox_token_file"]:
+            die("netbox_token_file not configured")
+            if not os.access(config["netbox_token_file"]):
```
cogito-mens: **Unreachable code.** The `os.access` guard is inside `if not config["netbox_token_file"]:` which already calls `die()` — this branch will never execute. Should be moved outside:

```python
if not config["netbox_token_file"]:
    die("netbox_token_file not configured")
if not os.access(config["netbox_token_file"], os.R_OK):
    die("...is not readable")
```

ansible/roles/rspamd/files/usr/local/sbin/rspamd-dkim-update
```diff
@@ -284,6 +312,170 @@ def do_nsupdate(config, keys):
         die("Failed to call {}".format(main_cmd))
 
 
+def netbox_api(config, method, path, data=None):
+    base = config["netbox_url"].rstrip("/")
+    url = base + path
+    headers = {
+        "Authorization": "{}".format(open(config["netbox_token_file"]).read()),
```
cogito-mens: **Authorization header format** — NetBox DRF's `TokenAuthentication` expects `Authorization: Token <hex-token>`, not just the raw token. Without the `Token ` prefix, all API calls will return 401/403. Also, the file should be opened with a `with` block and the token should be `.strip()`'d to drop the trailing newline:

```python
with open(config["netbox_token_file"]) as f:
    token = f.read().strip()
headers = {
    "Authorization": "Token " + token,
    ...
}
```

ansible/roles/rspamd/files/usr/local/sbin/rspamd-dkim-update
```diff
@@ -321,6 +513,11 @@ def do_email(config, keys):
         # smtp.set_debuglevel(1)
         smtp.sendmail(config["email_from"], config["email_to"], msg)
 
+    hc_list = []
+    for key in keys:
+        hc_list += key["selector"] + "." + key["domain"]
```
cogito-mens: **`hc_list += string`** — `+=` on a list with a string iterates over characters. So `"sel.example.com"` becomes `['s','e','l',...]` and the eventual `" ".join(hc_list)` produces garbage like `"s e l . e x a m p l e"`. Should be:

```python
hc_list.append(key["selector"] + "." + key["domain"])
```

drybjed (commented):
The issue with the `rsyslog` role in the common playbook has been fixed. If you rebase the PR on the `master` branch, all tests should pass fine now.

mathieumd:
> ### `requests` as hard module-level dependency

Will replace it by:
```py
try:
    import requests
except ImportError:
    print("Please install Python module 'requests' if you need NetBox or Healthchecks.")
```

> ### Stricter `get_keys` for all methods
> Previously a malformed file on delete was silently skipped; now it aborts the entire run.

Not sure how to actually improve that.

> ### `strip_dkim_spaces` / value comparison fragility
> the existing-record match in `do_netbox` may silently fail to find the record, creating a duplicate.

I apply the same function to both data to be recorded into and to data retrieved from NetBox. According to [`_split_text_value`](https://github.com/sys4/netbox-plugin-dns/blob/main/netbox_dns/validators/dns_value.py#L72) used by NetBox DNS, I think it should behave coherently.

ansible/roles/rspamd/files/usr/local/sbin/rspamd-dkim-update
mathieumd: Will fix it.

ansible/roles/rspamd/files/usr/local/sbin/rspamd-dkim-update
mathieumd: Since NetBox API v2, it expects both `Token` or `Bearer`. I will add a comment about what should contain this file.

ansible/roles/rspamd/files/usr/local/sbin/rspamd-dkim-update
mathieumd: Thanks.

PR #2721, opened 2026-06-23
feat(lookup_plugins): add global config dir fallback for Galaxy installs

cogito-mens:
The `debops.debops.file_src`, `debops.debops.template_src`, and `debops.debops.task_src` lookup plugins now fall back to reading `override_paths` from the global DebOps configuration directories when the `debops` Python module is unavailable — needed when the collection is installed via Ansible Galaxy. Relative paths in those config files are resolved against the current working directory.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -70,6 +70,14 @@ General
   installations should update their Ansible Collections to get the updated
   Community collection.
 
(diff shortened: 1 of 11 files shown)


drybjed (approved):
The changes were tested by an user interested in the feature and work as expected.

PR #2730, opened 2026-07-31
docker_service: parity with docker_compose_service (data_dirs, hooks, ferm)

scibi:
This brings `docker_service` to parity with its sibling `docker_compose_service` role in three areas that came up while planning a single-container deployment: explicit `data_dirs` for host directory management (owner/group/mode) alongside a hardened version of the existing `volumes`-based directory auto-creation, which now skips paths already covered by `data_dirs` or `config_files` and skips paths that exist but are not directories, so it no longer tries to turn a bind-mounted file like `/etc/localtime` into a directory; `pre_main`/`post_main` custom hooks resolved through the same `debops.debops.task_src` mechanism `docker_compose_service` already uses, letting a project inject tasks before or after the role runs without patching it; and a `published_ports` key that renders per-service firewall rules into the `DOCKER-USER` iptables chain, which is where Docker-published ports actually get filtered in the `FORWARD` path (a plain `INPUT` rule from ferm's `host_rules` silently never matches them).

All three additions are opt-in and keep the `docker_service` and `docker_compose_service` schemas aligned (same `data_dirs` shape, same hook mechanism, same `published_ports`/ferm default names) so inventory snippets stay portable between the two roles. Verified with `--check --diff` against three hosts already running `docker_service` in production, with no reported changes on any of them.


anzil:
Thanks for contributing `docker_service`.

Would it be possible to add the parameter `devices` from `community.docker.docker_container` in task *Manage Docker containers*?

Instead of adding required parameters from this module explicitly could one allow/pass any parameter, e.g. through a dictionary and key-value lookup?

PR #2660, opened 2026-03-05
fix(monitoring/containers/misc roles): ensure when conditionals return bool (Ansible 2.19)

scibi:
Replace | d() truthiness checks with | d() | length > 0 to produce proper boolean results in when conditions, avoiding deprecation warnings that will become errors in Ansible 2.19.

Affected roles: telegraf, metricbeat, filebeat, mosquitto, snmpd, lxc, gitlab_runner, reprepro, rsnapshot, ipxe, libvirtd, snapshot_snapper, minio, icinga_web, nixos, controller, yadm

PR #2736, opened 2026-08-13
Fix non-boolean conditionals for Ansible 2.19 (~all roles)

tblein:
This PR makes `when:` and `changed_when:` conditions in the roles of every playbook layer compatible with Ansible 2.19, which enables Jinja2 native types by default so a condition that evaluates to a bare string or list no longer passes the boolean check and fails at runtime. The bare `| d()` calls used in boolean contexts are replaced with `is defined` and `truthy`/`falsy` tests, following the same approach @drybjed started in commit 7ec2fcb. 

I am aware of the alternative approach proposed by @scibi in the debops PRs #2656–#2660, which uses the `| d() | length > 0` idiom. I deliberately stayed with the `is defined`/`is truthy` form: unlike the length filter, it also handles scalar booleans and numbers without raising. Furthermore, it keeps the fix consistent with the style already established in the project. I also chose to submit this as a single PR covering all layers rather than splitting it per role group.

### DIFF
--- a/ansible/roles/unbound/tasks/main.yml
+++ b/ansible/roles/unbound/tasks/main.yml
@@ -28,14 +28,14 @@
     path: '/etc/unbound/unbound.conf.d/ansible-server.conf'
     state: 'absent'
   notify: [ 'Check unbound configuration and reload' ]
-  when: not unbound__combined_server | d()
+  when: unbound__combined_server is falsy
 
 - name: Remove Unbound remote control configuration
   ansible.builtin.file:
     path: '/etc/unbound/unbound.conf.d/ansible-remote-control.conf'
     state: 'absent'
   notify: [ 'Check unbound configuration and reload' ]
-  when: not unbound__combined_remote_control | d()
+  when: unbound__combined_remote_control is falsy
 
 - name: Generate Unbound server configuration
   ansible.builtin.template:
@@ -45,7 +45,7 @@
     group: 'root'
     mode: '0644'
   notify: [ 'Check unbound configuration and reload' ]
-  when: unbound__combined_server | d()
+  when: unbound__combined_server is truthy
 
 - name: Generate Unbound remote control configuration
   ansible.builtin.template:
@@ -55,7 +55,7 @@
     group: 'root'
     mode: '0644'
   notify: [ 'Check unbound configuration and reload' ]
-  when: unbound__combined_remote_control | d()
+  when: unbound__combined_remote_control is truthy
 
 - name: Remove DNS zones if requested
   ansible.builtin.file:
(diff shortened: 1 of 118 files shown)


drybjed:
Thanks for the PR, I actually prefer this solution due to the reasons you mentioned. I'll try to review and merge the PR as soon as possible.

ansible/roles/unbound/tasks/main.yml
```diff
@@ -28,14 +28,14 @@
...
-  when: not unbound__combined_server | d()
+  when: unbound__combined_server is falsy
 
 - name: Remove Unbound remote control configuration
   ansible.builtin.file:
     path: '/etc/unbound/unbound.conf.d/ansible-remote-control.conf'
     state: 'absent'
   notify: [ 'Check unbound configuration and reload' ]
-  when: not unbound__combined_remote_control | d()
+  when: unbound__combined_remote_control is falsy
```
drybjed: I'm not sure why the codespell check fails here. This will probably happen again if `falsy` is used. Could you try changing it to `ansible.builtin.falsy`, to see if that will also trigger `codespell`?

Alternatively, I think that `not unbound__combined_remote_control is truthy` will probably work as expected.

drybjed (changes requested):
Great work, I hope that it can be merged soon. There's an issue with `codespell` tripping on `falsy` in the Unbound role, it should be resolved before merging.

ansible/roles/unbound/tasks/main.yml
tblein: A shame: I thought `falsy` was quite elegant and mirrors nicely `truthy`.
I used the not ... is trusthy approach that looks nicer than using the long `ansible.builtin.falsy`.

`codespell` seems to pass this time.

drybjed: This might be a known false positive, maybe we can ignore it. Try this: create `.codespellrc` file in the root of the repo and add:

```ini
[codespell]
ignore-words-list = falsy
```

Revert back the unbound change. Did codespell fail for you locally? Check then, and if the fix works you can amend the last PR commit and change it to creation of the config file. If it doesn't work locally, commit the changes in a new commit and push, let's see it in CI.

tblein: Locally, codespell do not report any error.
```bash
$ codespell --version
2.4.3
```

There is already a codespell exclude file at the root of the project: `.codespell.exclude`
Do I add `falsy` there or to a new `.codespellrc`?

drybjed: Asd the falsy to that file, I think it's used but let's see in CI.

tblein: I finally added it to the `ignore-word-list` included in the `pyproject.toml` as some others.
`.codespell.exclude` file seems to include usernames or software names

drybjed: Great, all checks passed. I'll do final check and merge tomorrow.

drybjed (approved):
All tests passed, merging. I'll backport the changes to stable releases as well.

PR #2737, opened 2026-08-13
[authorized_keys] Fix ansible breaking due to undefined loop label

LoniasGR:
When running debops with ansible-core > =14, the playbook fails on step `Ensure the required groups exist`, if an account is defined with `home: True`. For example:

```
  - name: "loniasgr"
    sshkeys:
      ["{{ lookup('file', '{{ playbook_dir }}/files/ssh_keys/loniasgr')}}"]
    state: present
    accounts:
      - name: "loniasgr"
        home: True
```
would fail, because group is not defined in the loop label, even though it should skip. This PR fixes this issue in the two following steps:
- Ensure the required groups exist
- Enforce state of authorized_keys user files

by adding a default value on the inciminating values.

Edit: I had a previous PR, but for some reason it got closed automatically when I tried to rebase on the current master.

### DIFF
--- a/ansible/roles/authorized_keys/tasks/main.yml
+++ b/ansible/roles/authorized_keys/tasks/main.yml
@@ -32,7 +32,7 @@
     state: 'present'
   loop: '{{ lookup("template", "lookup/authorized_keys__identities.j2") | from_yaml }}'
   loop_control:
-    label: '{{ {"identity": item.identity, "group": item.group} }}'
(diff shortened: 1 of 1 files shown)


LoniasGR:
This not fixed by the changes in #2736, since the `when` check happens after the `loop_control` field is evaluated. 

Another option would be to remove the group from the loop label, but I feel that setting a  default may be more sensible.

Issue #2739, opened 2026-08-18
apt_preferences: pins are never created since '| bool' was added to the Create task when (debops >= 3.2.6)


matlads:
## Summary

Since **debops 3.2.6** (commit `2a401af50`, cherry-picked into master as `7ec2fcb`, "Fix conditional expressions for Ansible v2.19"), the **`debops.apt_preferences` role can no longer create (or update) any Apt preferences/pin files**. The `when:` condition on the *Create APT preferences* task was changed from ending with `item.pin | d())` to `item.pin | d())) | bool`, and the `| bool` filter coerces the *string* pin expression to `false`, so the task is silently skipped.

This affects **both ansible-core 2.17 and 2.19** (and is still present on `master` at `6d6485b`).

## Behaviour

The role's *Create APT preferences* task has this `when:` (current master / 3.2.6+):

```yaml
when: ((((item.state | d('present')) != 'absent' and (not ((item.delete is defined and item.delete is truthy) | bool))) and
       (item.package | d() or item.packages | d())) and
      (item.raw | d() or
       item.version | d() or
       (item.pin is undefined and ...) or
        item.pin | d())) | bool
```

Jinja's `and`/`or` return the last-evaluated operand, not a boolean. For a normal pin entry like:

```yaml
apt_preferences__group_list:
  - package: 'python3-certbot-dns-desec'
    pin: 'release n=plucky'
    priority: 900
```

the whole expression evaluates to the **string** `'release n=plucky'`, and `'release n=plucky' | bool` is `false` because the `bool` filter only recognises `yes/on/1/true`. Same for `raw:`, `version:`, and plain `package:`+`priority:` entries. Result: the task always skips:

```
skipping: [host] => (item={'package': 'python3-certbot-dns-desec', 'pin': 'release n=plucky', 'priority': 900})
```

## Impact

- New pin files are never written to `/etc/apt/preferences.d/`.
- Existing pin files (created before 3.2.6) are **not updated** — edits to pin content/templates silently do nothing.
- This is a silent failure with no error or changed-detection (files simply end up missing/stale). There is no way to work around it via role variables; only by downgrading to debops 3.2.5 or editing the role.

## Repro

Run `debops.apt_preferences` with any `apt_preferences__group_list` entry that uses `pin:`, `raw:`, `version:` or `package:` under ansible-core 2.17 or 2.19 (roles from debops 3.2.6+, 3.3.0, or master):

```yaml
- hosts: testhost
  roles:
[...]

PR #2741, opened 2026-08-18
Make sure everything is a proper boolean in the when clause to fix pin file creating

arjanoosting:
Today i needed to adjust some pinning at our site and encountered bug #2739 
I have created an fix for the creation of the pinning files. I did not verify all the other modules with|bool changes mentioned in #2739 

Kind regards,

Arjan


drybjed (approved):
Looks good, thanks. I'll backport the changes to the stable releases as well. This role will have to be redesigned soon, it was made more than 12 years ago and a lot has changed...

PR #2743, opened 2026-08-19
Fixed various bugs in apparmor.fact.j2 fact script

4utomate:
Apparmor fact script did contains some bugs resulting in errors after running apparmor role and later fact gathering. Namely there's a missing colon, the output dict uses append, cmd_exists leaves room for errors with empty PATH variable.


drybjed (approved):
Thanks for the fixes. I'll backport them to the stable releases as well.

PR #2746, opened 2026-08-23
pki: silence successful certbot output in weekly pki-realm jobs

scibi:
## Summary

The weekly `pki-realm schedule` path queues `batch`/`at` jobs that always run `certbot certonly` for DNS-01 and manual ACME flows. Even when a certificate is not yet due for renewal, certbot and verbose `ln -v` write to stdout, so `at` mails root on every host with messages like “Certificate not yet due for renewal” and symlink listings.

This change runs certbot with `--quiet` and creates the realm symlinks with `ln -sfT` (without `-v`) in those two ACME paths. Successful runs and “not yet due” cases stay silent; failures still emit stderr and therefore still produce mail. The `--expand` behaviour is unchanged, so Ansible-driven SAN updates continue to work.

## Test plan

- Deploy the updated `pki-realm` script via the `debops.pki` role.
- Manually run `pki-realm run -n <realm>` on a host whose certificate is not due and confirm no stdout from certbot/`ln`.
- Confirm a deliberate certbot failure still produces stderr (and thus mail from `at` if that is how the job is scheduled).

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -261,6 +261,12 @@ General
 :ref:`debops.pki` role
 ''''''''''''''''''''''
 
+- The :command:`pki-realm` script now runs :command:`certbot` with the
+  ``--quiet`` option for DNS-01 and manual ACME flows, and creates realm
+  symlinks without :command:`ln` ``-v``. Periodic ``batch``/``at`` jobs no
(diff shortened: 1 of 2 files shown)


drybjed (approved):
Thanks for testing with a real ACME environment; I'll make the backports to the stable releases as well.

PR #2751, opened 2026-08-26
[docker_service] Add device-related options to container management task

anzil:
Implementation of device options as asked for in https://github.com/debops/debops/pull/2730#issuecomment-5368194553. @scibi


drybjed (approved):
Looks good, thanks.

Issue #2549, opened 2024-11-04
slow startup: takes 2 minutes loading global_handlers/handlers/*.yml - local setup issue?


prahal:
Starting a debops run with extra verbose a flag (-vvvv) I see the startup takes two minutes doing "statically imported" of  global_handlers. The issue is probably in ansible (o my local setup see further) not debops but maybe debops could work around it. And the fact is I don't know debops and ansible enough to sort out if it is a debops or ansible issue.

Each global handler is loaded 555 times and there are 79 of them. Maybe the global_handler could be concatenated before being imported, so only 555 imports are done?

Or maybe my setup is not supported. 
If there is a simpler way to achieve a setup with local roles not in the debops repository and at the same time being able to test changes I make to upstream debops playbooks and roles before sending them upstream, I would be glad to migrate to it.

I have a debops project folder with its own local playbooks and roles appended to a git clone of the debops repository playbooks and roles (while still using a virtualenv pip3 debops install).
debops-project
- ansible/collections/ansible_collections/debops a symlink to <debops git clone>
- playbooks/ my local roles ad hoc playbooks including a site.yml a a folder hierarchy like the debops playbooks one
- playbooks/roles my local roles

$HOME/.ansible with collections, galaxy.yml, playbooks, plugins and roles as symlink to the debops git clone ad hoc folder
```
ls -l ~/.ansible/
total 20
drwxr-xr-x 1 prahal prahal   0  4 nov.  22:04 collections
drwx------ 1 prahal prahal   0  4 nov.  21:57 cp
lrwxrwxrwx 1 prahal prahal  76  4 nov.  02:36 debops-contrib-playbooks -> /home/prahal/Projects/Admin/ansible/debops/ansible/debops-contrib-playbooks/
drwx------ 1 prahal prahal  16  3 mars   2022 galaxy_cache
-rw------- 1 prahal prahal   0  3 mars   2022 galaxy_token
lrwxrwxrwx 1 prahal prahal  61  4 nov.  02:36 galaxy.yml -> /home/prahal/Projects/Admin/ansible/debops/ansible/galaxy.yml
lrwxrwxrwx 1 prahal prahal  60  4 nov.  02:36 playbooks -> /home/prahal/Projects/Admin/ansible/debops/ansible/playbooks
lrwxrwxrwx 1 prahal prahal  58  4 nov.  02:36 plugins -> /home/prahal/Projects/Admin/ansible/debops/ansible/plugins
[...]

PR #2758, opened 2026-09-17
swapfile: add task hooks

TwiggyWan:
Follow pattern explained at : https://docs.debops.org/en/stable-3.3/user-guide/debops-for-ansible.html#file-template-and-task-hooks

and already implemented in other roles (for instance sysctl)


ansible/roles/swapfile/tasks/main.yml
```diff
@@ -4,6 +4,18 @@
...
+  ansible.builtin.import_role:
+    name: 'ansible_plugins'
+
+- name: Import DebOps global handlers
+  ansible.builtin.import_role:
+    name: 'global_handlers'
+
+- name: Pre hooks
+  ansible.builtin.include_tasks: '{{ lookup("debops.debops.task_src", "swapfile/pre_main.yml") }}'
+  when: sysctl__enabled | bool
```
drybjed: This variable should be `swapfile__enabled`.

ansible/roles/swapfile/tasks/main.yml
```diff
@@ -91,3 +103,7 @@
   ansible.builtin.file:
     path: '{{ swapfile__sysctl_file | d("/etc/sysctl.d/30-debops.swapfile.conf") }}'
     state: 'absent'
+
+- name: Post hooks
+  ansible.builtin.include_tasks: '{{ lookup("debops.debops.task_src", "swapfile/post_main.yml") }}'
+  when: sysctl__enabled | bool
```
drybjed: This variable should be `swapfile__enabled`.

ansible/roles/swapfile/tasks/main.yml
```diff
@@ -4,6 +4,18 @@
 # Copyright (C) 2015-2016 DebOps <https://debops.org/>
 # SPDX-License-Identifier: GPL-3.0-only
 
+- name: Import custom Ansible plugins
+  ansible.builtin.import_role:
+    name: 'ansible_plugins'
+
+- name: Import DebOps global handlers
+  ansible.builtin.import_role:
+    name: 'global_handlers'
```
drybjed: These imports don't seem to be needed for anything at the moment, unless you want to use them in some tasks located in `pre_tasks/post_tasks`, can you explain the use case?

drybjed (changes requested):
Pre and post tasks could be added, but I'm not sure about the import tasks. Looking forward for the fixed variable names. :)

ansible/roles/swapfile/tasks/main.yml
TwiggyWan: oops, sorry! Fixed.

ansible/roles/swapfile/tasks/main.yml
```diff
@@ -4,6 +4,18 @@
...
+
+- name: Pre hooks
+  ansible.builtin.include_tasks: '{{ lookup("debops.debops.task_src", "swapfile/pre_main.yml") }}'
```
drybjed, line 12: Sorry, one more thing. For this lookup to work without any overrides present, the role needs `tasks/swapfile/pre_main.yml` and `tasks/swapfile/post_main.yml` files present, with correct Copyright and SPDX headers. You can get examples from other roles, for example `resources` role.

ansible/roles/swapfile/tasks/main.yml
```diff
@@ -4,6 +4,18 @@
...
+  when: swapfile__enabled | bool
```
drybjed: And you can actually remove the `when:` lines. The `swapfile` role is not included in the common playbook, and does not have `swapfile__enabled` variable, nothing to disable. Sorry for the confusion.

ansible/roles/swapfile/tasks/main.yml
TwiggyWan: For global_handlers, I did not read the [documentation](https://docs.debops.org/en/latest/ansible/roles/global_handlers/getting-started.html) thoroughly before submitting this, and just thought all roles ought to import it for standardization. I agree with your statement and will remove that.

For ansible_plugins, although I did not need it for my exact use case (as I just wanted to add debugging hooks to print `ansible_memtotal_mb`), as a new developer, I wasn't aware some roles would not have these features.

Reading  : 
https://docs.debops.org/en/stable-3.3/user-guide/debops-for-ansible.html
https://docs.debops.org/en/latest/user-guide/universal-configuration.html

I thought all roles were supposed to support  universal configuration, and that the filters and lookup facilities provided by ansible_plugins were required. Therefore I thought it would make sense to follow a pattern of always adding this import. What do you think?

ansible/roles/swapfile/tasks/main.yml
TwiggyWan: ok, done.

ansible/roles/swapfile/tasks/main.yml
TwiggyWan: I added the mentioned files :

```
$ ls roles/swapfile/tasks/swapfile/
post_main.yml  pre_main.yml
```

ansible/roles/swapfile/tasks/main.yml
drybjed: The `global_handlers` role causes lengthy slowdowns in some environments (https://github.com/debops/debops/issues/2549) so I plan to move most of the handlers from it back to their respective Ansible roles. The `global_handlers` role might be useful for very common handlers like fact reloads and `systemd daemon-reload`, but it might be removed from most of the roles otherwise.

The `ansible_plugins` role is an old mechanism from before Ansible Collections to provide roles with access to custom plugins. Since DebOps playbooks include the `debops.debops` collection, and custom plugins are supposed to be referenced via their FQCN names, the `ansible_plugins` role might be removed at some point to ensure that everything goes through Ansible collection mechanisms.

Roles should support Universal Configuration, but not all roles do it just yet, and it requires a bit more changes to the role, like use of the `parse_kv_items/config` filter plugin and possible fixes in templates or Jinja code in tasks. As for use of UC in the `swapfile` role - since it's so simple, I don't think that's really needed here. You can specify swap files directly per-host or per-group, I don't think any merged lists would improve things much. But if you think that might be useful in the role, let me know.

ansible/roles/swapfile/tasks/swapfile/post_main.yml
```diff
@@ -0,0 +1,4 @@
+---
+# Copyright (C) 2016-2019 Maciej Delmanowski <drybjed@gmail.com>
```
drybjed, line 2: Please set your own copyright here and in the other file, don't impersonate me. :)

ansible/roles/swapfile/tasks/main.yml
TwiggyWan: Thanks for the explanation! I agree with you, and removed `ansible_plugins` .

> Roles should support Universal Configuration, but not all roles do it just yet, and it requires a bit more changes to the role, like use of the parse_kv_items/config filter plugin and possible fixes in templates or Jinja code in tasks. 

is this part of https://github.com/debops/debops/issues/1461 ? I tried searching open issues for "Universal Configuration" but had trouble to find which roles need work.

Or is there a role you would recommend looking at for the correct way of supporting Universal Configuration without ansible_plugins` ?

ansible/roles/swapfile/tasks/swapfile/post_main.yml
TwiggyWan: fixed.

ansible/roles/swapfile/tasks/main.yml
drybjed: Part of #1461 - this is more a thread about updating the filter plugins to be a bit more consistent and easier to reason about. There are some issues with the current design, for example with how the weight parameter works. I have an improved version of the plugin somewhere, need to dig it up and check what roles need to be updated. I'll try to merge the updates soon, right now I'm going through old PRs to get the project running somewhat faster.

Currently there are no roles that use UC without `ansible_plugins` role also imported, it's just the current convention. If you omit it and reference the filter using its FQCN, `debops.debops.parse_kv_config` or `debops.debops.parse_kv_items`, Ansible should do the correct thing and look it up in the `debops.debops` collection.

Here's an example playbook:
```yaml
---

- name: Test filter reference
  hosts: 'all'

  vars:
    parameters:

      - name: 'option1'
        value: 'value1'

      - name: 'option2'
        value: 'value2'

  tasks:

    - name: "List vars"
      debug:
        msg: '{{ parameters | debops.debops.parse_kv_config }}'
```

And expected output:
```json
{
    "msg": [
        {
            "id": 0,
            "name": "option1",
            "real_weight": 0,
            "section": "unknown",
            "separator": false,
            "state": "present",
            "value": "value1",
            "weight": 0
        },
        {
            "id": 10,
            "name": "option2",
            "real_weight": 10,
            "section": "unknown",
            "separator": false,
            "state": "present",
            "value": "value2",
            "weight": 0
        }
    ]
}
```

drybjed (approved):
That should work nicely now, thanks. I'll backport this to the stable releases as well.

